PR #48, opened 2014-05-28
Experimental support for LPC17xx devices

richardeoin:
These are now somewhat out of date devices, and they don't seem to implement ADI_v5 in the same way as newer devices (like the lpc11xx). Full debug support for these devices seems to require the native firmware to be compiled with the -O0 flag - with probably indicates some kind of timing issue. However without this they are still identified and code can be loaded.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -26,6 +26,7 @@ SRC =			\
 	jtagtap.c	\
 	lmi.c		\
 	lpc11xx.c	\
+	lpc17xx.c	\
 	lpc43xx.c	\
 	main.c		\
 	nrf51.c		\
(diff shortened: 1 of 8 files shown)


gsmcmullin:
I haven't merged this because it's yet another fork of the lpc* code which is a maintenance problem.

I've refactored the upstream on my new_flash branch (subject to change/rebase):
https://github.com/gsmcmullin/blackmagic/tree/new_flash

When I've had a chance to test on real hardware I'll revisit support for these lpc17xx devices.

esden:
@richardeoin It would be nice to see LPC17xx support. As I do not have the eval boards for this I can not fix your patch. Would you be able to make the changes @gsmcmullin requested and update this patch? :)

richardeoin:
@esden It would indeed be nice. However I'd start over using [lpc15xx.c](src/target/lpc15xx.c) as a template, and then debug. I may do this if I have some spare time!

Eval boards are still available if you hunt around, e.g. [digikey](https://www.digikey.com/products/en?mpart=OM13085UL&v=568).

esden:
As this PR is quite out of date now and it needs improvements to be merged into the codebase I will close it for now.

PR #51, opened 2014-06-11
Support erasing + flashing the nRF51 User Information Configuration Registers (UICRs)

richardeoin:


giobauermeister:
Hi @richardeoin!

Does this change also supports writing to UICR registers? If I understang correctly it only erases UICR.

I'm looking for a way to write to address 0x10001080, 0x10001084 and 0x10001088

PR #64, opened 2015-01-10
STM32L05x target support.

beewoolie:
o Program flash write and erase.
o Options modification and status.
o Stub support for faster programming.
o There are two versions of the prog_write function.  There is one
  that runs through the debug interface and another that uses stubs.
  Stubs are faster, but the MCU may not run a stub if it is in a
  crashed state.
o Option commands added.
o Command to disable stubs as well.
o Stubs generated from C++ code and converted to arrays of half-words.
o Removing #defines and favoring enum.
o Replacing references to NVM registers is code with constant values.
o Copyrights.
o Segment writes to EEPROM region supported for executable data.
o Monitor command to write to EEPROM.

### DIFF
--- a/flashstub/README
+++ b/flashstub/README
@@ -1,5 +1,13 @@
-These are the assembler routines for executing a flash write
-on the supported targets.  They are kept here for reference, but
-are not used, as the compiled binary code is included in the 
-target drivers.
+Flash Stubs
+===========
 
+For most of the targets, these are assembler routines for executing
+a flash write on the supported targets.  They are kept here for
+reference, but are not used, as the compiled binary code is included
+in the target drivers.
+
+For the STM32l0x, the stubs are written in C++ and emitted as arrays
+of half-words for inclusion in the target driver.  The use of a higher
+level language allows more detailed code and for easy revisions.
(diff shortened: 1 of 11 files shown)


beewoolie:
This is a complete implementation for the STM32L0xx targets.  It would probably work for the L1's as well, but I don't have hardware on which I could test it.  If you would prefer to have a merged driver, let me know and I'll buy an L151 dev kit and make the driver support both.  I know, for example, that the erase block sizes are different, but that should not be a significant hurdle.

zyp:
I wish I saw this before I did my own take (#65), unfortunately I didn't notice your pull request until now.

As your implementation is more complete than mine, yours should be favored. For consistency's sake I think it should also be extended to cover L1, which I could assist with testing if you'd like.

beewoolie:
I don't have an L1 ATM.  I'll make a change that *should* support L1 and post that to my repo where you can pull it to try.

beewoolie:
I'm going to close this pull request for two reasons.  I've found a couple of errors that don't seem to break the functionality of the functions, but they are incorrect.  Second, I'm adding the stm32L1x features and these are turning out to require more than simply changing the IDCODEs.

PR #69, opened 2015-01-17
Fixed run command on SAM D20 old revision B silicon.

richardeoin:
This appears to be related to Atmel Errata Reference 12015 (see § 35.4.1 DSU in the datasheet) but only applies to revision B silicon - not revision C. While this is an old silicion revision (latest is revision E as of 17 Jan 2015) I found one on a development board.

### DIFF
--- a/src/samd20.c
+++ b/src/samd20.c
@@ -198,6 +198,51 @@ static const char samd20_xml_memory_map[] = "<?xml version=\"1.0\"?>"
 #define CORTEXM_DHCSR_C_HALT		(1 << 1)
 #define CORTEXM_DHCSR_C_DEBUGEN		(1 << 0)
 
+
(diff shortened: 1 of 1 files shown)


gsmcmullin:
I prefer not to repeat the Cortex-M code in this driver.  Please check my samd20-rev-fix branch (commit a3ab9b24) and confirm this works for you.

richardeoin:
I didn't refactor cortexm to avoid breaking other drivers, but this is definitely the better solution!  I've checked your [samd20-rev-fix](https://github.com/blacksphere/blackmagic/tree/samd20-revb-fix) branch and it works as expected.

richardeoin:
Thanks for pulling this @gsmcmullin . I'm also working on SAM D10/11/21 support, expect a new pull request in the next few days

PR #65, opened 2015-01-12
Add STM32L0 support.

zyp:
This is mostly a copy of the existing STM32L1 code.
Changes consists of ID code location, flash base and sector size.

PR #75, opened 2015-01-24
STM32Lx target support.

beewoolie:
Target support for stm32l0's and stm32l1's including option bytes and
data EEPROM.  This module will superceed the previous stm32l1 driver.

o Program flash write and erase.
o Options modification and interpretive status.
o Stubs for program flash writes and erases.  Stubs are modestly
  faster than non-stub version.   The stm32l0 will not execute stubs
  when the MCU has crashed.  A monitor option may be used to force
  non-stub flash writes.
o Stubs generated from C++ code and converted to arrays of half-words.
o Writes to data EEPROM supoprted when loading segments.
o EEPROM data monitor command to write words.
o Stubs supported on stm32l1.


beewoolie:
This is a reworked version of the first stm32l0xx effort.   It would replace the existing stm32l1 if it were merged, but the removal of the stm32l1.c file is not part of the changeset.

gsmcmullin:
@zyp Please confirm this works for you and supersedes your PR #65 and I'll merge this and close yours.

zyp:
I can do that, but right now I'm travelling and won't have access to hardware to test with until I'm home in about two weeks.

beewoolie:
Zyp, have you had a chance to test the Lx patch?

zyp:
Sorry, I forgot about it. I'll try to get it done some time this week.

src/stm32l0.c
```diff
@@ -0,0 +1,1076 @@
...
+       while an operation is in progress.
+
+   o Performance.  The throughput for writing is not high.  We suspect
+     it may be possible to improve throughput significantly by increasing
+     the MCU clock.  The code, as is, offers a simplicity without
+     undue knowledge of the inner workings of the MCUs.  Increasing
+     clock frequencies would require substantial knowledge of the
+     clock tree.
+
+  0x1ff80000: 0x00aa 0xff55 OK
```
gsmcmullin: This stuff is cryptic and should probably go.

src/stm32l0.c
```diff
@@ -0,0 +1,1076 @@
...
+p *(int*)0x40023c10 = 0x8c9daebf
+p *(int*)0x40023c10 = 0x13141516
+
+
+p *(int*)0x40023c14 = 0xfbead9c8
+p *(int*)0x40023c14 = 0x24252627
+
+*/
+
+#define CONFIG_STM32L1          /* Include support for STM32L1 */
```
gsmcmullin: Is there a reason we would want this support conditional?

src/stm32l0.c
```diff
@@ -0,0 +1,1076 @@
...
+      inhibit_stubs = 0;
+    if (!strncasecmp (argv[1], "disable", cb))
+      inhibit_stubs = 1;
+  }
+  gdb_outf ("stubs: %sabled\n", inhibit_stubs ? "dis" : "en");
+
+  return true;
+}
+
+#if 0
```
gsmcmullin: These dead functions should be removed.

src/stm32l0.c
```diff
@@ -0,0 +1,1076 @@
...
+
+  if (argc == 4) {
+    uint32_t addr = strtoul (argv[2], NULL, 0);
+    uint32_t val  = strtoul (argv[3], NULL, 0);
+
+    if (   addr <  STM32Lx_NVM_EEPROM_PHYS
+        || addr >= STM32Lx_NVM_EEPROM_PHYS + stm32lx_nvm_eeprom_size (t))
+      goto usage;
+
+#if 0
```
gsmcmullin: This dead code should be removed.

src/stm32l0.c
beewoolie: NP.  These were working notes while I developed the options code.

src/stm32l0.c
beewoolie: That depends on you.  If you want to replace the existing L1 support, then I'll make it permanent.  If you don't I'd prefer to leave the conditional in and work toward coalescing all of the Lx support in one place.

src/stm32l0.c
beewoolie: I can remove it if you prefer.  The hitch with this function is that it doesn't work because the reset of the emulator code gets confused by the target behavior when we perform the OBL_LAUNCH.  I would hope that we could leave this in a work toward solving the option byte loader issue.  However, if you prefer I'll remove it and leave it in my independent tree.

src/stm32l0.c
beewoolie: I'll see if I can fix this instead of removing it.  I think the catch is that the eeprom write code presently doesn't permit byte writes.  IIRC, the MCU does so this is really a shortcoming of my implementation.

gsmcmullin:
The generated *.stub files shouldn't be in git, they should be generated at build time.
Also please use the Linux kernel coding style.
Is there a reason to keep stubbed and unstubbed flash routines?  What is the advantage?

I haven't finished my review, so there may be more comments when I get time to look over it more.

gsmcmullin:
ok to test

beewoolie:
On 3/6/15 10:22 AM, Gareth McMullin wrote:
>
> The generated *.stub files shouldn't be in git, they should be 
> generated at build time.
>

IMHO, it would be unwise to generate stubs at build time.  We need to 
make sure the stub code is correct and changes to the toolchain could 
affect this.  It's very difficult to guarantee that the compiler will 
produce code that works properly in the context for which we are 
building, no matter how carefully we code the routines.

So, I think it is wiser to let the stubs be static, just as they are for 
the other targets.  It is certainly convenient that we can rebuild them 
from source should we want to change their behavior. However, there 
isn't any reason to believe that a recompile at build time is going to 
make them significantly better.

The other reason to not build them is that the conversion depends on a 
Perl script and I wouldn't want to add that kind of dependency to the 
build unless there is a significant justification.

> Also please use the Linux kernel coding style.
>

Can you give me an example of where you want the coding style changed?


> Is there a reason to keep stubbed and unstubbed flash routines? What 
> is the advantage?
>

That depends.  The stubbed versions are faster, though not by much. I 
would prefer to use them all of the time.  However, there is a 
circumstance where the stubs cannot be used.  This is described in the 
first source file comment.  I believe that this is a change to the MCU 
core becase the F parts doesn't seem to have the same behavior.

> I haven't finished my review, so there may be more comments when I get 
> time to look over it more.
>

Thanks.

src/stm32l0.c
gsmcmullin: Understood.  I think this driver is more complete and should replace the old driver.  I can make that change after merging.

src/stm32l0.c
gsmcmullin: I'd prefer to leave it out and merge it in when it's working.  Code that isn't built has a higher tendency to rot.

src/stm32l0.c
```diff
@@ -0,0 +1,1076 @@
...
+  uint32_t sr;
+  do {
+    sr = adiv5_ap_mem_read (ap, STM32Lx_NVM_SR(nvm));
+  } while (sr & STM32Lx_NVM_SR_BSY);
+
+  return !(sr & STM32Lx_NVM_SR_ERR_M);
+}
+
+static bool stm32lx_cmd_stubs (target* t,
+                               int argc, char** argv)
```
gsmcmullin: Warning on unused parameter t should be suppressed by casting to void.

src/stm32l0.c
```diff
@@ -0,0 +1,1076 @@
...
+
+  /* We can't handle unaligned destination or non-word writes. */
+  /* *** FIXME: we should handle misaligned writes by padding with
+     *** zeros.  Probably, the only real time we'd see something
+     *** misaligned would be on a write to a final half-word.  Perhaps
+     *** this could be handled with the stub?  In fact, aligning the
+     *** start is going to be mandatory.  We will change the code to
+     *** cope with a trailing half-word. */
+  if ((destination & 3) || (size & 3))
+    return -1;
```
gsmcmullin: Although rare, cases do exist where the section bounds are unaligned.  See how this is handled in the STM32F1 driver.

gsmcmullin:
On Fri, Mar 6, 2015 at 10:38 AM, Marc Singer wrote:
> IMHO, it would be unwise to generate stubs at build time. We need to
> make sure the stub code is correct and changes to the toolchain could
> affect this. It's very difficult to guarantee that the compiler will
> produce code that works properly in the context for which we are
> building, no matter how carefully we code the routines.

I mostly agree with this, which is why my stubs are tiny assembler routines.
I acknowledge that yours are more complicated.

> So, I think it is wiser to let the stubs be static, just as they are for
> the other targets. It is certainly convenient that we can rebuild them
> from source should we want to change their behavior. However, there
> isn't any reason to believe that a recompile at build time is going to
> make them significantly better.

That seems reasonable.  I am just opposed to having hundred lines of generated
junk added to the repo.  Could we just include the actual instruction
codes of the
compiled stubs and not the whole commented out source file?

> The other reason to not build them is that the conversion depends on a
> Perl script and I wouldn't want to add that kind of dependency to the
> build unless there is a significant justification.

Following on above, the compiled stub can be dumped to binary with objdump,
and converted into includable form with hexdump, this saves us the perl script
and dependence on perl to regenerate the stubs.

> > Also please use the Linux kernel coding style.
>
> Can you give me an example of where you want the coding style changed?

Use tabs for indentation.
In function calls, don't put a space between the function name and
opening bracket.

> > Is there a reason to keep stubbed and unstubbed flash routines? What
> > is the advantage?
>
> That depends. The stubbed versions are faster, though not by much. I
> would prefer to use them all of the time. However, there is a
> circumstance where the stubs cannot be used. This is described in the
> first source file comment. I believe that this is a change to the MCU
> core becase the F parts doesn't seem to have the same behavior.

Ok, understood.  I still don't think that the monitor command to disable the
stubs is useful.

src/stm32l0.c
zyp: Keep in mind that STM32L flash reads 0 when erased while STM32F flash reads 1. The STM32F1 driver pads with ones, while this will have to pad with zeros. You can also look to my old STM32L1 code for an example.

beewoolie:
On 3/7/15 7:21 PM, Gareth McMullin wrote:
> On Fri, Mar 6, 2015 at 10:38 AM, Marc Singer wrote:
> > IMHO, it would be unwise to generate stubs at build time. We need to
> > make sure the stub code is correct and changes to the toolchain could
> > affect this. It's very difficult to guarantee that the compiler will
> > produce code that works properly in the context for which we are
> > building, no matter how carefully we code the routines.
>
> I mostly agree with this, which is why my stubs are tiny assembler 
> routines.
> I acknowledge that yours are more complicated.
>
> > So, I think it is wiser to let the stubs be static, just as they are for
> > the other targets. It is certainly convenient that we can rebuild them
> > from source should we want to change their behavior. However, there
> > isn't any reason to believe that a recompile at build time is going to
> > make them significantly better.
>
> That seems reasonable. I am just opposed to having hundred lines of 
> generated
> junk added to the repo. Could we just include the actual instruction
> codes of the
> compiled stubs and not the whole commented out source file?

Certainly.

I leave those things in for the sake of readability.  One of my primary 
concerns is obfuscation.  It's really easy for a complex program to 
become impenetrable.  Leaving the comments in the stub makes it easier 
to audit should someone be interested in doing so.

The simple alternative is to have two ways to generate the stub. I'll 
add some comments to the generation routine so that it is clear what you 
would have to do.

If we can generate stubs without Perl do you want to make it part of the 
build?
>
> > The other reason to not build them is that the conversion depends on a
> > Perl script and I wouldn't want to add that kind of dependency to the
> > build unless there is a significant justification.
>
> Following on above, the compiled stub can be dumped to binary with 
> objdump,
> and converted into includable form with hexdump, this saves us the 
> perl script
> and dependence on perl to regenerate the stubs.
Hmm.  I've done this as well in the past.  I don't recall why I didn't 
do so this time.  I'll revisit this step.

>
> > > Also please use the Linux kernel coding style.
> >
> > Can you give me an example of where you want the coding style changed?
>
> Use tabs for indentation.
> In function calls, don't put a space between the function name and
> opening bracket.

OK.  Can do.

FWIW, I tend not to use tabs for indents because programmers like to 
change the tab width which leads to sometimes unreadable source.

>
> > > Is there a reason to keep stubbed and unstubbed flash routines? What
> > > is the advantage?
> >
> > That depends. The stubbed versions are faster, though not by much. I
> > would prefer to use them all of the time. However, there is a
> > circumstance where the stubs cannot be used. This is described in the
> > first source file comment. I believe that this is a change to the MCU
> > core becase the F parts doesn't seem to have the same behavior.
>
> Ok, understood. I still don't think that the monitor command to 
> disable the
> stubs is useful.

The problem is that there doesn't seem to be a way to detect this 
state.  I've tried to do so, but it doesn't seem to be reliable. The 
disable command is the only way to guarantee it.

If you have an idea about how to detect the lock-up state of the MCU I'd 
certainly prefer to make this automatic.

src/stm32l0.c
beewoolie: On 3/7/15 7:06 PM, Gareth McMullin wrote:
>
> In src/stm32l0.c 
> <https://github.com/blacksphere/blackmagic/pull/75#discussion_r26001878>:
>
> > +                                           const uint8_t* source,
> > +                                           int size)
> > +{
> > +  struct stm32lx_nvm_stub_info info;
> > +  const uint32_t nvm = stm32lx_nvm_phys (target);
> > +  const size_t page_size = stm32lx_nvm_prog_page_size (target);
> > +
> > +  /* We can't handle unaligned destination or non-word writes. */
> > +  /* *** FIXME: we should handle misaligned writes by padding with
> > +     *** zeros.  Probably, the only real time we'd see something
> > +     *** misaligned would be on a write to a final half-word.  Perhaps
> > +     *** this could be handled with the stub?  In fact, aligning the
> > +     *** start is going to be mandatory.  We will change the code to
> > +     *** cope with a trailing half-word. */
> > +  if ((destination & 3) || (size & 3))
> > +    return -1;
>
> Although rare, cases do exist where the section bounds are unaligned. 
> See how this is handled in the STM32F1 driver.
>
>


OK.  I'll look at the fix there and make the necessary change.

src/stm32l0.c
beewoolie: On 3/7/15 6:08 PM, Gareth McMullin wrote:
>
> In src/stm32l0.c 
> <https://github.com/blacksphere/blackmagic/pull/75#discussion_r26001592>:
>
> > +    adiv5_ap_mem_write (ap, address, value);
> > +  else if (cb == 2)
> > +    adiv5_ap_mem_write_halfword (ap, address, value);
> > +  else
> > +    return false;
> > +
> > +  uint32_t sr;
> > +  do {
> > +    sr = adiv5_ap_mem_read (ap, STM32Lx_NVM_SR(nvm));
> > +  } while (sr & STM32Lx_NVM_SR_BSY);
> > +
> > +  return !(sr & STM32Lx_NVM_SR_ERR_M);
> > +}
> > +
> > +static bool stm32lx_cmd_stubs (target* t,
> > +                               int argc, char** argv)
>
> Warning on unused parameter t should be suppressed by casting to void.
>
>
Done.

src/stm32l0.c
beewoolie: On 3/7/15 6:02 PM, Gareth McMullin wrote:
>
> In src/stm32l0.c 
> <https://github.com/blacksphere/blackmagic/pull/75#discussion_r26001569>:
>
> > +  if (argc == 1) {
> > +    gdb_out ("usage: mon stubs [enable/disable]\n");
> > +  }
> > +  else if (argc == 2) {
> > +    size_t cb = strlen (argv[1]);
> > +    if (!strncasecmp (argv[1], "enable", cb))
> > +      inhibit_stubs = 0;
> > +    if (!strncasecmp (argv[1], "disable", cb))
> > +      inhibit_stubs = 1;
> > +  }
> > +  gdb_outf ("stubs: %sabled\n", inhibit_stubs ? "dis" : "en");
> > +
> > +  return true;
> > +}
> > +
> > +#if 0
>
> I'd prefer to leave it out and merge it in when it's working. Code 
> that isn't built has a higher tendency to rot.
>
>

OK.  I'll add a comment about the fact that this feature is missing with 
an explanation of the reason.

src/stm32l0.c
beewoolie: On 3/7/15 6:00 PM, Gareth McMullin wrote:
>
> In src/stm32l0.c 
> <https://github.com/blacksphere/blackmagic/pull/75#discussion_r26001556>:
>
> > +p *(int*)0x1ff80000 = 0xff5500aa
> > +
> > +  l1 program
> > +p *(int*)0x40023c04 = 1
> > +p *(int*)0x40023c0c = 0x89abcdef
> > +p *(int*)0x40023c0c = 0x02030405
> > +p *(int*)0x40023c10 = 0x8c9daebf
> > +p *(int*)0x40023c10 = 0x13141516
> > +
> > +
> > +p *(int*)0x40023c14 = 0xfbead9c8
> > +p *(int*)0x40023c14 = 0x24252627
> > +
> > +*/
> > +
> > +#define CONFIG_STM32L1          /* Include support for STM32L1 */
>
> Understood. I think this driver is more complete and should replace 
> the old driver. I can make that change after merging.
>
>

OK.  IMHO, it's better for you to make that kind of change.  I like to 
have an easy way to go back if something unexpected happens when we 
roll-out the change.  IOW, we deprecate the old driver but leave it in 
for a time before removing it.

gsmcmullin:
On Sun, Mar 8, 2015 at 11:41 AM, Marc Singer wrote:

> The simple alternative is to have two ways to generate the stub. I'll
> add some comments to the generation routine so that it is clear what you
> would have to do.
>
> If we can generate stubs without Perl do you want to make it part of the
> build?
>

This is still a work in progress, be see what I've done with the STM32
driver here:
https://github.com/gsmcmullin/blackmagic/tree/flashstubs

I'm doing some work around generalising the calls to stubs and passing
parameters in registers.  I'll also rewrite the stubs in C, and write some
notes on how it all works to make it easier for newcomers.

src/stm32l0.c
beewoolie: On 3/7/15 7:33 PM, zyp wrote:
>
> In src/stm32l0.c 
> <https://github.com/blacksphere/blackmagic/pull/75#discussion_r26001973>:
>
> > +                                           const uint8_t* source,
> > +                                           int size)
> > +{
> > +  struct stm32lx_nvm_stub_info info;
> > +  const uint32_t nvm = stm32lx_nvm_phys (target);
> > +  const size_t page_size = stm32lx_nvm_prog_page_size (target);
> > +
> > +  /* We can't handle unaligned destination or non-word writes. */
> > +  /* *** FIXME: we should handle misaligned writes by padding with
> > +     *** zeros.  Probably, the only real time we'd see something
> > +     *** misaligned would be on a write to a final half-word.  Perhaps
> > +     *** this could be handled with the stub?  In fact, aligning the
> > +     *** start is going to be mandatory.  We will change the code to
> > +     *** cope with a trailing half-word. */
> > +  if ((destination & 3) || (size & 3))
> > +    return -1;
>
> Keep in mind that STM32L flash reads 0 when erased while STM32F flash 
> reads 1. The STM32F1 driver pads with ones, while this will have to 
> pad with zeros. You can also look to my old STM32L1 code for an example.
>

I know!  It's kinda strange that they changed this very long-standing 
convention.

beewoolie:
On 3/8/15 11:41 AM, Marc Singer wrote:
>
>>
>> > The other reason to not build them is that the conversion depends on a
>> > Perl script and I wouldn't want to add that kind of dependency to the
>> > build unless there is a significant justification.
>>
>> Following on above, the compiled stub can be dumped to binary with 
>> objdump,
>> and converted into includable form with hexdump, this saves us the 
>> perl script
>> and dependence on perl to regenerate the stubs.
> Hmm.  I've done this as well in the past.  I don't recall why I didn't 
> do so this time.  I'll revisit this step.

Objcopy (not objdump) will dump to binary.  However, it doesn't generate 
the correct byte stream.  I changed the parse to use sed and removed the 
source code.  The stub is just the instruction stream.

beewoolie:
On 3/8/15 12:01 PM, Gareth McMullin wrote:
> On Sun, Mar 8, 2015 at 11:41 AM, Marc Singer wrote:
>
> > The simple alternative is to have two ways to generate the stub. I'll
> > add some comments to the generation routine so that it is clear what you
> > would have to do.
> >
> > If we can generate stubs without Perl do you want to make it part of the
> > build?
> >
>
> This is still a work in progress, be see what I've done with the STM32
> driver here:
> https://github.com/gsmcmullin/blackmagic/tree/flashstubs
>
> I'm doing some work around generalising the calls to stubs and passing
> parameters in registers. I'll also rewrite the stubs in C, and write some
> notes on how it all works to make it easier for newcomers.
>
> —
>

I considered switching to a register calling convention as well.  I 
didn't want to blaze that trail because I wasn't sure why you hadn't 
done so from the beginning.  Also, I didn't see a way to selectively 
update registers.  The only call I saw was one that sets all of the 
registers.

There were three changes I would have liked to make:

1) Pass the base address of the information structure in r0.  It seems 
unnecessary to pass all of the args in registers since this can make the 
linkage tricky if the args don't fit in the first four registers.  Also, 
I don't think our performance is limited by memory bandwidth.  The flash 
operations take millseconds in many cases. Reading a structure from 
memory vs registers isn't going to move the needle.

2) Only write the stub when necessary.  This would require that we have 
a way to verify that the stub is present.  My hunch has been that some 
of the performance is lost in writing the stub every time.  Perhaps this 
is state that the emulator could keep for us. IOW, it lets us know if 
there is a chance that our flash stub was clobbered since the last time 
we were called.

3) Coalesce writes to flash.  Currently, we only write ~900 bytes at a 
time.  We would get far better performance if we could write more full 
pages.  This needs changes to the upper layers and may even necessitate 
some IO overlap with USB to eliminate delays.

BTW, I found that objcopy doesn't do what you might expect in some 
cases.  I couldn't use it to convert my stub to an array as it left off 
some of the instructions.  I favored objcopy and sed, both of which seem 
to be readily available and have greater fidelity.

gsmcmullin:
See #82 for changes to the flash stub mechanism.  We can discuss further there.

beewoolie:
I pushed to the branch to accommodate the changes we've discussed.

I added code to handle the misalignment of the destination.  It will execute for any write to program memory.  However, I don't know how to test this since I have never seen this occur.  Generally, I don't like to push code I cannot test which is why I never write this before.  I've read it a couple of times to convince myself it is correct...which is no substitute for testing.

So, if you know of a way to test this code path I'm all ears.

gsmcmullin:
Thanks for all your effort.  I'm going to merge this now.
I think there are still some improvements we can make around the stubs.  I've started PR #82 in that direction, so please check that out and comment.

PR #77, opened 2015-01-30
platform: native: no TXS workaround on mini

machinaut:
blacksphere_mini board (hwversion = 1) doesn't need the TXS0108 workaround.


gsmcmullin:
Why is this an issue?  This pin is not connected on the mini.

PR #78, opened 2015-03-02
Exceptions minor cleanup and stlink

esden:
This pull request contains minor improvements, and ports the changes to the stlink probe platform.

I have tested it on the stm32f4-discovery board, using the [stm32f4 timer example](https://github.com/libopencm3/libopencm3-examples/tree/master/examples/stm32/f4/stm32f4-discovery/timer), that uses __WFI in the main loop.

It is able to swdp_scan and attach to the target, this was not working all the time before. Now it seems to be quite reliable. I also tested the behavior when stuck in __WFI (disabled the interrupt in the example). It correctly points out that the swdp_scan failed due to timeout and that it might be due to WFI.

So far so good, will continue using this firmware and see how things go. :)

### DIFF
--- a/src/platforms/native/platform.h
+++ b/src/platforms/native/platform.h
@@ -191,6 +191,8 @@ int platform_init(void);
 void morse(const char *msg, char repeat);
 const char *platform_target_voltage(void);
 int platform_hwversion(void);
+void platform_set_timeout(uint32_t ms);
+bool platform_timeout_expired(void);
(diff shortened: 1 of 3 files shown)


gsmcmullin:
I'm not going to merge this.  The exceptions branch here is gone now, moved to gsmcmulllin/blackmagic, and rebased.  These changes are included.

PR #81, opened 2015-03-05
SAM4S: Erase flash in 8K instead of 16K chunks.

dlaw:
This pull request (1) changes SAM4S flash erases to use 8 page chunks instead of 16 page chunks, and (2) corrects a bug in the generation of the FARG field for SAM4S flash erases.

(1) This issue is due to an error in the SAM4S datasheet. Table 20-4 states that 16 page erases may be used on all sectors, but buried in section 8.1.3.1 is a note that only 8 page erases are guaranteed to work in sectors 0 and 1.

(2) FARG[15:2] is not supposed to be shifted left by two bits. This bug was masked by the fact that an invalid 16 page erase on sector 0 also erased sectors 1-3.

I've been testing this change for more than a month now, so I am confident that there are no more SAM4S flash erase issues lurking within.

Cheers,
Dave

### DIFF
--- a/src/sam3x.c
+++ b/src/sam3x.c
@@ -291,26 +291,27 @@ static int sam3x_flash_erase(struct target_s *target, uint32_t addr, int len)
 	uint32_t base = sam3x_flash_base(target, addr, &offset);
 
 	/* The SAM4S is the only supported device with a page erase command.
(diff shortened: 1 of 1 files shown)


gsmcmullin:
ok to test

PR #83, opened 2015-03-17
samd: Remove low level ADIv5 calls an favour of target_mem_write.

gsmcmullin:


gsmcmullin:
Thanks @richardeoin for testing.

PR #86, opened 2015-03-31
sam3x: add SAM3U support

florolf:
This patch adds support for the U subfamily of the SAM3 controllers.


gsmcmullin:
ok to test

florolf:
I forgot to mention that I was only able to test this on a ATSAM3U4E. But as this is a device with two flash banks, all smaller SAM3U chips should be working as well.

gsmcmullin:
Thanks.  I'm merging as is.  I have work in progress to change how these flash drivers are implemented.  I will ask you to retest when I've reworked the driver.

PR #91, opened 2015-06-23
Fix usage of USB_VBUS

schodet:
USB_VBUS is not an alternate function, it is an additionnal function which is
always enabled.

If configured as an alternate function, it will draw current from VBUS.


gsmcmullin:
ok to test

gsmcmullin:
Can anyone using the f4discovery port confirm this?  @UweBonnes

esden:
Tested and I can confirm that this patch drops the current usage on the OTG_FS port from 30mA to less than my current measurement can indicate (<10mA). Also tested that I can connect to the f4discovery Black Magic Probe firmware and able to scan for a target device.

PR #92, opened 2015-06-29
Fixes #90 flashing of STM32 m0 chips

### LINKED ISSUE #90: Current master fails to flash stm32f0discovery (stm32f051)
it works on v1.6-rc0

i will test this on a native probe, but i assume it will be the same result.

    No. Att Driver
     1      STM32F0
    0xfffffffe in ?? ()
    (gdb) load
    Loading section .isr_vector, size 0xc4 lma 0x8000000
    Error writing data to flash

maknoll:
These are 3 commits from UweBonnes wich i shamelessly sending in cause he did not yet do so.
They are fixing flashing of M0 ST's, because the stubs aren't compliled for ARMv6.
I tested the changes and they work on the M0, but should be tested on a STM32F1 target before merging because the code affects them as well.


gsmcmullin:
ok to test

gsmcmullin:
This is working for me on the F103, so it seems the change is good.

PR #93, opened 2015-07-12
[Bugfix] stm32f1_probe would always return true, breaking support for…

richardeoin:
… all other targets

The intention of `if (t->driver)` conditional was to test if any of the cases in the preceeding switch/case were met. However t->driver was previously set to point to a default value in cortexm.c:200 and therefore this would always evaluate to true. I've replaced the broken if statement with a duplicate of the switch/case run above.

It looks like this was introduced in 09544bc7101981f2688e0a678138dc4580ea3f76 (PR #92) but 492d6c9cf8f735eea5478c99aee43608549218d7 maybe contributes to the confusion in this instance.

### DIFF
--- a/src/stm32f1.c
+++ b/src/stm32f1.c
@@ -162,7 +162,12 @@ bool stm32f1_probe(target *t)
 		block_size = 0x800;
 		break;
 	}
-	if (t->driver) {
+	switch(t->idcode) {
+	case 0x444:  /* STM32F03 RM0091 Rev.7 */
+	case 0x445:  /* STM32F04 RM0091 Rev.7 */
+	case 0x440:  /* STM32F05 RM0091 Rev.7 */
+	case 0x448:  /* STM32F07 RM0091 Rev.7 */
+	case 0x442:  /* STM32F09 RM0091 Rev.7 */
 		flash_size = (target_mem_read32(t, FLASHSIZE_F0) & 0xffff) *0x400;
 		gdb_outf("flash size %d block_size %d\n", flash_size, block_size);
 		target_add_ram(t, 0x20000000, 0x5000);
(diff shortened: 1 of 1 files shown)


UweBonnes:
Hello,

Thanks for the hint. I didn't see that t->driver was initialized. I think the better solution is to add a default case to the switch that returns false:

        case 0x442:  /* STM32F09 RM0091 Rev.7 */
                t->driver = "STM32F09";
                block_size = 0x800;
                break;
+       default: return false;
        }
        if (t->driver) {

gsmcmullin:
ok to test

gsmcmullin:
I'm not really a fan of that switch block that fills in the driver string.  The driver string, in my opinion, was just meant to indicate this is the stm32f1 driver, not the particular device.  I'd rather just put a generic driver string in the new switch block added by this PR, and remove the first one altogether.

UweBonnes:
>>>>> "Gareth" == Gareth McMullin <notifications@github.com> writes:

    Gareth> I'm not really a fan of that switch block that fills in the
    Gareth> driver string. The driver string, in my opinion, was just meant
    Gareth> to indicate this is the stm32f1 driver, not the particular
    Gareth> device. I'd rather just put a generic driver string in the new
    Gareth> switch block added by this PR, and remove the first one
    Gareth> altogether.

I think some user feedback is very usefull. If you work with several boards,
it is easy to get confused. The feedback must not be the driver string, but
some gdb printout should be done.

-- 
Uwe Bonnes                bon@elektron.ikp.physik.tu-darmstadt.de

Institut fuer Kernphysik  Schlossgartenstrasse 9  64289 Darmstadt
--------- Tel. 06151 162516 -------- Fax. 06151 164321 ----------

richardeoin:
Bump - Please accept either #93 or #94 to un-break master.

PR #95, opened 2015-07-31
Add 0x0072 as a valid & supported idcode for NRF51

bvanheu:
Hi,

The NRF51 chip I'm using has the idcode 0x0072. I've tested the Blackmagic debugger with the chip on my board and everything is working fine.

Thanks


gsmcmullin:
ok to test

PR #96, opened 2015-08-17
Add support for STM32F2xx family

schodet:
The F405 rev A uses a wrong CPUID, this patch use the core identifier to make
the distinction.

Tested with:
 - F205
 - F405 rev Z
 - F429

Unfortunately, I do not have any F405 rev A.

### DIFF
--- a/src/stm32f4.c
+++ b/src/stm32f4.c
@@ -52,6 +52,7 @@ static int stm32f4_flash_write(struct target_flash *f,
                                uint32_t dest, const void *src, size_t len);
 
 static const char stm32f4_driver_str[] = "STM32F4xx";
+static const char stm32f2_driver_str[] = "STM32F2xx";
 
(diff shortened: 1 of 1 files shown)


gsmcmullin:
ok to test

LorenzMeier:
@gsmcmullin How is this looking? We use the F4 and would need HW support.

gsmcmullin:
I don't really see the point of this PR.  The STM32F2 has been supported by this driver since the beginning of time.  The driver string indicated the stm32f4, but this is the stm32f4 driver.

@LorenzMeier The F4 works perfectly.  What support do you need?

schodet:
Don't you find it disturbing to display F4 if the connected device is an F2?

If the driver can handle both devices, it should report something like F2/F4.

There is at least one difference between F4 and F2, there is no CCM memory on F2, so the current code in master branch will report a wrong memory map.

PR #98, opened 2015-08-22
Add STM32L4 handling. No option write handling yet.

UweBonnes:
Tested with a blackmagik-stlink connected to Nucleo-L476 and L4-Discovery.


gsmcmullin:
ok to test

PR #101, opened 2015-09-30
TI TM4C/LM4F cdc-acm fixes

umetronics:
Hi,

These changes fix USB enumeration and communication over the CDC serial port on the TI Launchpad platform. With them applied, connecting to the probe and scan using both JTAG and SWD works. I'm using a LM4F120 Launchpad but the TM4C123 shouldn't be any different in this regard.

Attaching to the target fails, however. I've traced it down to the snprintf call in map_ram() in target.c. In fact, any call to snprintf with a format argument seems to crash the MCU. Google suggests it might be a stack size issue, but I'm not sure how to rectify that.

### DIFF
--- a/src/platforms/common/cdcacm.c
+++ b/src/platforms/common/cdcacm.c
@@ -498,7 +498,7 @@ static void cdcacm_set_config(usbd_device *dev, uint16_t wValue)
 	configured = wValue;
 
 	/* GDB interface */
-#ifdef STM32F4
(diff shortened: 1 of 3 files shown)


gsmcmullin:
ok to test

umetronics:
Defining snprintf to sniprintf fixed the crashing.

gsmcmullin:
The sniprintf should come in for other newlib platforms too.  I'll push that in another commit.

PR #106, opened 2015-11-14
Support for Silicon Labs EFM32/EZR32

richardeoin:
Tested with EZR32LG230


gsmcmullin:
ok to test

lbonn:
It looks like the EUI64's top 24 bits are not really constant across all the family members (despite what the EZR32 RM says).

I got `0x24ee17` on a Gecko (and not  `0xb000b57`).

richardeoin:
Hi @lbonn, you're quite right. For the [EZR32LG](http://www.silabs.com/Support%20Documents/TechnicalDocs/EZR32LG-RM.pdf) and [EZR32WG](http://www.silabs.com/Support%20Documents/TechnicalDocs/EZR32WG-RM.pdf) families the top 24 bits are constant `0x000b57` (IEEE-MA-L for Silicon Labs). However for the [EFM32G](http://www.silabs.com/Support%20Documents/TechnicalDocs/EFM32G-RM.pdf) those bits are just part of the Unique ID for that particular piece of silicon and could be anything.

I've just fixed this with PR #107. We can use the SW-DP IDCODE register for identification rather than this one. While you're waiting for it be merged you could just comment out that if statement.

You might find further problems on a Gecko as I've only tested with the EZR32LG family myself.

Thanks @lbonn!

PR #107, opened 2015-12-01
[efm32] Use SW-DP IDCODE register for identification

richardeoin:
Instead of the Device ID register which isn't constant for EFM32G parts.

Fixes issue with #106 found by @lbonn

### DIFF
--- a/src/efm32.c
+++ b/src/efm32.c
@@ -251,12 +251,18 @@ static void efm32_add_flash(target *t, uint32_t addr, size_t length,
 char variant_string[40];
 bool efm32_probe(target *t)
 {
-	/* Read the extended unique identifier */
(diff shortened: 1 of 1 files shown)


lbonn:
Now the Gecko is correctly detected :) and `load` seems to work correctly.

I'll keep you in touch if I find other issues, I'll probably do some more tests in the upcoming days.

Actually we had our own support for efm32 (last version here: https://github.com/hikob/blackmagic/tree/jaguar-2.2) that we haven't cleaned for an upstream merge.

Also, I noticed that the efm32 flashstub is not listed in the Makefile's target list (https://github.com/richardeoin/blackmagic/blob/544e1080a36143c286e89d31ea7dc6fb34bf6831/flashstub/Makefile#L14).
Is there a specific reason for this?

PR #108, opened 2015-12-14
Some various fixes

lbonn:
These fixes several issues that we found during tests with our hardware:

* crc32: speed up the generic crc32 routine which caused us some trouble by buffering larger chunks of memory. I don't know if the STM32's routine needs similar modification.
* flashstub: mostly cosmetic adjustments in Makefile
* efm32: we had some trouble when doing loads and breaks with srst switching enabled
* gdb_main.c: fixes a buffer overflow, observed when issuing `dump` commands

In my sense, the last one is the most critical.

Please let me know if you'd prefer these to be submitted in their own pull request.


gsmcmullin:
ok to test

PR #114, opened 2016-02-11
Add nRF52 support.

markrages:
Lightly tested, appears to work for erasing, loading code, and basic debugging.


gsmcmullin:
ok to test

PR #115, opened 2016-02-15
[native] Check for hardware version by pull-up and -down. Print HW Ve…

esden:
…rsion in mon command.

Until now the native hardware was pulling PB5-7 down and checking if
they were asserted high. BMPMV2b is pulling the pins down instead of
high. The hardware version routine now determines the hardware version
based on the fact if a pin is asserted at all. This means that if a pin
is left floating, the version number bit will be 0, and if the pin is
asserted either high or low the bit will be set to 1. While we were
already at it the "monitor version" command in GDB will now also print
the hardware version number.

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -108,7 +108,7 @@ int command_process(target *t, char *cmd)
 
 bool cmd_version(void)
 {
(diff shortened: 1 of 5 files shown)


gsmcmullin:
add to whitelist

PR #116, opened 2016-02-15
[native] Make the native bootloader bit more flashy.

esden:
This change has also a practical reason. When flashing and testing the
hardware this change makes it easier to make sure all the LEDs work. Now
when the DFU bootloader is idle it is scanning the LEDs making it easy
to see if one of them has an issue.

In addition to that, the bootloader now indicates when there is data
being flashed using the DFU interface. In cases when one has more than
one device connected and accidently starts flashing a wrong device this
is very useful feature to have.


gsmcmullin:
ok to test

PR #119, opened 2016-03-03
STM32L07x

beewoolie:
Added STM32L07x's to the supported MCUs.

o Some regularizing of the target detection was done between L1's and L0's.
o New command for MCU info added to aid in determining which MCU we're using.

### DIFF
--- a/src/nrf51.c
+++ b/src/nrf51.c
@@ -119,6 +119,7 @@ bool nrf51_probe(target *t)
 	case 0x002F: /* nRF51822 (rev 1) CEAA B0 */
 	case 0x0031: /* nRF51422 (rev 1) CEAA A0A */
 	case 0x003C: /* nRF51822 (rev 2) QFAA G0 */
+	case 0x0057: /* nRF51422 (rev 2) QFAA G2 */
 	case 0x0040: /* nRF51822 (rev 2) CEAA CA0 */
 	case 0x0044: /* nRF51822 (rev 2) QFAA GC0 */
 	case 0x0047: /* nRF51822 (rev 2) CEAA DA0 */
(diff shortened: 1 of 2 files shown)


beewoolie:
Looks like there may be a catch.  The patch allows the L07x's to write to flash, but something goes wrong when the file being written approaches 64KiB.  I'll invetigate and update the patch accordingly.

UweBonnes:
41fb2a310475263d6d855862adfb26867c7a3f1c adds more L0 devices. Writing above 64 k is still a problem. See "STM32L0: Timeout when erasing large chunks of flash" on the mailing list.

esden:
@UweBonnes is the large chunk flash erase the known GDB timeout problem?

schodet:
Hello,

I also see the 64 Kb problem, actually, the problem is with the erase command, there is something wrong if it is done in one go.

I can make it work if I split the flash definition in 64 Kb parts.

schodet:
More precision on the > 64 Kb problem:

If programming a 70 Kb binary for example, GDB first erase, then write. When writing the first block at address 0x8010000, the MCU signals a "not erase" error. Actually the erase done in one go did not erase this block.

When programming a small binary at 0x8010000 directly, there is no problem.

I tried to put a loop in `stm32lx_nvm_prog_erase` around the whole function content, to program 64 Kb per 64 Kb, but it does not solve the problem.

If I declare separated blocks to GDB, I suppose it issues several erase commands and everything works.

You can see #246 for the code I use.

gsmcmullin:
What version of GDB is this problem being observed?  There is a known bug in GDB 7.12 where it times out on the flash erase packet.  See #181

schodet:
It's GDB 7.7.1, but there is no timeout, the problem is with the erase which do not erase.

Actually, I think we are supposed to wait after each erase operation, the current code erase every page in one go.

The commit 2216587 (in master) workarounds the problem by defining 3 different flash sections.

esden:
As this issue has been fixed in master I am closing this pull request.

If there is still an issue please feel free to reopen this pull request. :)

PR #120, opened 2016-03-15
added support for KL02 family

lyorak:
Kinetis for KL25 is extended with KL02 support. The family members are rather similar in terms of flash API and device ID location, so I generalized the KL25 device support APIs. I have tested the change on a MKL02Z32VFG4 device. Works from command line and from the Kinetis Design Studio as well. Please kindly review the changes.


gsmcmullin:
ok to test

PR #126, opened 2016-04-18
Cortex-A support

gsmcmullin:
Working on Xilinx Zynq 7000 (Cortex-A9 dual core)


jacobmcnamee:
- Probably not Cortex-A specific, but releasing SRST when on `mon connect_srst disable` would be useful
- Disconnect / kill seem to leave target suspended in debug state

PR #137, opened 2016-06-22
Presetting the connect_srst setting to false.

esden:
If we do not explicitly set it to something it might end up with some
random ram value.

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -68,7 +68,7 @@ const struct command_s cmd_list[] = {
 	{NULL, NULL, NULL}
 };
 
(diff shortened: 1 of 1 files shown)


gsmcmullin:
This should be in the `.bss` section and set to zero by the C runtime.  If this really is random we have a bigger problem.

esden:
Right, you are correct. From the coding style point of things, static analyzers will usually throw a little tantrum if you use a non initialized variable, this is why I prefer when variables are explicitly set, at least in some init function. My commit comment is definitely false regarding the random RAM value. :)

PR #139, opened 2016-06-23
Use Product ID and Componend ID tables instead of defines.

esden:
This patch adds tables that allow us to decode the Product ID and
Component ID registers. Based on those tables we can decide what probe
function to use. Also this makes it possible for us to understand what
ADIv5 components the target is offering to us, even if we do not
need/use them.

The tables include a comment describing how the register bits need to be
interpreted.

### DIFF
--- a/src/adiv5.c
+++ b/src/adiv5.c
@@ -32,16 +32,181 @@
 #define DO_RESET_SEQ 0
 #endif
 
-/* ROM table CIDR values */
-#define CIDR_ROM_TABLE  0xb105100d
-#define CIDR_GENERIC_IP 0xb105e00d
-#define CIDR_DEBUG      0xb105900d
+/* All this should probably be defined in a dedicated ADIV5 header, so that they
+ * are consistently named and accessible when needed in the codebase.
+ */
 
-#define PIDR_REV_MASK 0x0FFF00000ULL
-#define PIDR_ARMv7M   0x4000BB000ULL
-#define PIDR_ARMv6M   0x4000BB008ULL
-#define PIDR_ARMv7MF  0x4000BB00CULL
-#define PIDR_ARMv7A   0x4000BBC09ULL
+/* ROM table CIDR values */
+#define CIDR0_OFFSET    0xFF0 /* DBGCID0 */
+#define CIDR1_OFFSET    0xFF4 /* DBGCID1 */
+#define CIDR2_OFFSET    0xFF8 /* DBGCID2 */
+#define CIDR3_OFFSET    0xFFC /* DBGCID3 */
+
+/* Component class ID register can be broken down into the following logical
+ * interpretation of the 32bit value consisting of the least significant bytes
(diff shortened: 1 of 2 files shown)


src/adiv5.c
```diff
@@ -32,16 +32,181 @@
...
-#define PIDR_REV_MASK 0x0FFF00000ULL
-#define PIDR_ARMv7M   0x4000BB000ULL
-#define PIDR_ARMv6M   0x4000BB008ULL
-#define PIDR_ARMv7MF  0x4000BB00CULL
-#define PIDR_ARMv7A   0x4000BBC09ULL
+#ifdef PLATFORM_HAS_DEBUG
+/* The reserved ones only have an R in them, to save a bit of space. */
+static const char const *cidc_debug_strings[] =
+{
+	"Generic verification component",           /* 0x0 */
```
gsmcmullin: I'd prefer designated initialisers here.  https://gcc.gnu.org/onlinedocs/gcc/Designated-Inits.html
````
    [cidc_gvc] = "Generic verification component",
    [cidc_romtab] = "ROM Table",
    ...
````
Reserved can be left out, `printf` can handle a `NULL` should this happen.

File: src/adiv5.c [hunk 2]
```diff
@@ -86,18 +251,37 @@ static void adiv5_component_probe(ADIv5_AP_t *ap, uint32_t addr)
...
 
-	switch (cidr) {
-	case CIDR_ROM_TABLE: /* This is a ROM table, probe recursively */
+	/* CIDR preamble sanity check */
+	if ((cidr & ~CID_CLASS_MASK) != CID_PREAMBLE) {
+		DEBUG("0x%X: 0x%X <- does not match preamble (0x%X)\n", addr, cidr, CID_PREAMBLE);
+		return;
+	}
+
+  /* Extract Component ID class nibble */
```
gsmcmullin: What happened to the indentation here?

src/adiv5.c
```diff
@@ -86,18 +251,37 @@ static void adiv5_component_probe(ADIv5_AP_t *ap, uint32_t addr)
 	addr &= ~3;
 	uint64_t pidr = 0;
 	uint32_t cidr = 0;
+
+	/* Assemble logical Product ID register value. */
 	for (int i = 0; i < 4; i++) {
-		uint32_t x = adiv5_mem_read32(ap, addr + 0xfe0 + 4*i);
+		uint32_t x = adiv5_mem_read32(ap, addr + PIDR0_OFFSET + 4*i);
 		pidr |= (x & 0xff) << (i * 8);
+		//DEBUG("pidr %d 0x%08X\n", i, x);
```
gsmcmullin: Please remove the commented out lines.

src/adiv5.c
esden: the remaining strings should be in there as they are usually reserved, I wanted to prevent the debug to output just an empty string when the component class is one of them.

PR #157, opened 2016-10-12
Add signed driver so you can install it on windows 10 and a script fo…

clovett:
See makecert.cmd for info on how to sign the windows driver yourself.  This change makes the Black Magic drivers appear in Device Manager without yellow warning.  Titled: 

libusb-win32 devices
    Black Magic Firmware Upgrade
    Black Magic Trace Capture
Ports (COM & LPT)
    Black Magic GDB Server (COM6)
    Black Magic UART Port (COM7)


rogerclarkmelbourne:
Interesting...

FYI. Another way to do this is to use libwdi , see 

https://github.com/pbatard/libwdi

to compile a custom driver installer cli exe

gsmcmullin:
Thank you!  This is great!  I'll try it out when I can find a Windows box, and merge if it works for me.

gsmcmullin:
This is not working for me.

Working on Windows 10 Enterprise VM on Virtualbox from https://developer.microsoft.com/en-us/windows/downloads/virtual-machines

When I first connect, no prompt for drivers appears, and the COM ports magically appear:
screenshot from 2016-10-21 20-35-05
Microsoft has figured out they can use the device class information to associate drivers!

These ports actually work, they just appear in counter-intuitive order:
screenshot from 2016-10-21 20-47-22

If I select 'Update Driver Software' from the Device Manager, even after pointing it at this BMP diver, I get a message 'The best driver software for your device is already installed'.

If it try on the firmware upgrade to trace interfaces I get this this complaint:
screenshot from 2016-10-21 20-58-03

rogerclarkmelbourne:
Gareth
Like you. I dont have any real W10 hardware, but if I get time today (Sunday), I'll try installing it on a laptop, and see if I get the same errors.

rogerclarkmelbourne:
Just tried this on a real hardware, and like Gareth, Windows 10 automatically recognises the Serial devices, but doesn't recognise the firmware update device.

Using the signed driver I was able to get W10 to recognise the firmware update device, and also to change the names of the 2 existing serial devices.

I then uninstalled these drivers and tried to install the ones from this repo, but for the firmware upgrade it didnt seem to have any effect, i.e no driver was installed and the device remained unrecognised.
With the Serial com devices, I got a message saying the best driver was already installed

So overall this PR seems to be a slight improvement for W10 users, but perhaps it needs to be in a W10 sub folder as the existing drivers work fine on W7 (and probably W8 and W8.1), and on W10 only the firmware upgrade is non functional, which for most people would not be a big problem as I expect that 99% of people just use the com ports and never update the firmware (I don't even know how to ;-))

rpavlik:
The current drivers probably don't work on windows 8. I'd actually be surprised if they worked on windows 7 x64, they almost certainly wouldn't on win 8 x64 (as somebody who has participated in shipping hardware with a CDC interface and as a result ended up getting a code signing certificate to make it simpler)

FWIW, some time ago I submitted improvements to libwdi that allows it (and Zadig, the GUI tool based on it) to also generate custom usbser signed drivers (really just signed inf files), so it could replace the Windows default ones (really just changing the name on Win10)

esden:
As this driver does not work for everyone I will close this pull request.

You can use the [Zadig](http://zadig.akeo.ie/) driver to use the BMP on Windows.

PR #163, opened 2016-10-22
First attempt at publishing Travis builds to Amazon S3.

gsmcmullin:


gsmcmullin:
Oh well,  Travis ignores this on pull request builds, so there's now way to tell if it's going to work...

Issue #169, opened 2016-11-15
EFM32HP (happy gecko) support


linuxchris12:
The Happy Gecko is not in the targets file yet. Needs a

`#define EFM32_DI_PART_FAMILY_HAPPY_GECKO	77`

and

`		case EFM32_DI_PART_FAMILY_HAPPY_GECKO:
			sprintf(variant_string,
				"EFM32 Happy Gecko);
			flash_page_size = 1024;
			break;`

in efm32.c

gsmcmullin:
Applied the changes in #170.

PR #170, opened 2016-11-15
EFM32HP (happy gecko) support

gsmcmullin:
For #169 

@linuxchris12 please confirm this is correct.


linuxchris12:
Yes, tested it again. And its recognizing the EFM32HG310 as well as the EFM32HG322

Has problems printing the serial number (unique number) for the EFM32HG310 and some other issues still, I will fix some more when I have time.

gsmcmullin:
I'm going to merge this then.  The SIGLOST will be a result of a bad cable or other signal integrity problem, and is very unlikely to be the fault of the firmware.  I'll accept a pull request to fix the serial number or whatever other problems you have.

PR #171, opened 2016-11-21
Don't enable usb uart on v2 when debugging

gsmcmullin:


esden:
Tested. Works for me. ™️ 😸

PR #174, opened 2016-12-02
Target Driver for Atmetl AT91SAM4L target

ChuckM:
These are the changes to add the SAM4L target to the BMP code base. There are two intertwingled things here, one is the SAM4L driver and the other is an addition to the internal Target interface which adds an extended-reset OPTION and function. Documentation is:

##Extended Reset

This allows a target to do some additional house keeping after reset has been asserted but before returning control the target. In this driver the function re-enables the SMAP peripheral if it had become disabled, and resets the SMAP.HCR (hold core reset) bit if it has become set.

### DIFF
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -96,9 +96,6 @@ int gdb_main_loop(struct target_controller *tc, bool in_syscall)
 	int size;
 	bool single_step = false;
 
-	/*
-	DEBUG("Entering GDB protocol main loop\n");
-	*/
 	/* GDB protocol main loop */
 	while(1) {
 		SET_IDLE_STATE(1);
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -96,7 +96,9 @@ int gdb_main_loop(struct target_controller *tc, bool in_syscall)
 	int size;
 	bool single_step = false;
 
-	DEBUG("Entring GDB protocol main loop\n");
+	/*
+	DEBUG("Entering GDB protocol main loop\n");
+	*/
 	/* GDB protocol main loop */
 	while(1) {
 		SET_IDLE_STATE(1);
--- a/src/platforms/common/swdptap.c
+++ b/src/platforms/common/swdptap.c
@@ -36,7 +36,7 @@ static void swdptap_turnaround(uint8_t dir)
 	if(dir == olddir) return;
 	olddir = dir;
 
-#if 0
+#ifdef DEBUG_SWD_BITS
 	DEBUG("%s", dir ? "\n-> ":"\n<- ");
 #endif
 
@@ -58,7 +58,7 @@ bool swdptap_bit_in(void)
 	gpio_set(SWCLK_PORT, SWCLK_PIN);
 	gpio_clear(SWCLK_PORT, SWCLK_PIN);
 
-#if 0
+#ifdef DEBUG_SWD_BITS
 	DEBUG("%d", ret?1:0);
 #endif
 
@@ -67,7 +67,7 @@ bool swdptap_bit_in(void)
 
 void swdptap_bit_out(bool val)
 {
-#if 0
+#ifdef DEBUG_SWD_BITS
 	DEBUG("%d", val);
 #endif
 
--- a/src/platforms/common/swdptap.c
+++ b/src/platforms/common/swdptap.c
@@ -36,7 +36,9 @@ static void swdptap_turnaround(uint8_t dir)
 	if(dir == olddir) return;
 	olddir = dir;
 
+#if 0
 	DEBUG("%s", dir ? "\n-> ":"\n<- ");
+#endif
 
 	if(dir)
 		SWDIO_MODE_FLOAT();
@@ -56,14 +58,18 @@ bool swdptap_bit_in(void)
 	gpio_set(SWCLK_PORT, SWCLK_PIN);
 	gpio_clear(SWCLK_PORT, SWCLK_PIN);
 
+#if 0
 	DEBUG("%d", ret?1:0);
+#endif
 
 	return ret != 0;
 }
 
 void swdptap_bit_out(bool val)
 {
+#if 0
 	DEBUG("%d", val);
+#endif
 
 	swdptap_turnaround(0);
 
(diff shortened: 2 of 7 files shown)


src/gdb_main.c
```diff
@@ -96,7 +96,9 @@ int gdb_main_loop(struct target_controller *tc, bool in_syscall)
 	int size;
 	bool single_step = false;
 
-	DEBUG("Entring GDB protocol main loop\n");
```
gsmcmullin: We can just remove this.  Commented out code isn't useful

src/target/cortexm.c
```diff
@@ -262,6 +262,7 @@ bool cortexm_probe(ADIv5_AP_t *ap)
 	PROBE(lpc15xx_probe);
 	PROBE(lpc43xx_probe);
 	PROBE(sam3x_probe);
+    PROBE(sam4l_probe);
```
gsmcmullin: Whitespace

src/target/cortexm.c
```diff
@@ -442,6 +443,11 @@ static void cortexm_reset(target *t)
 	target_mem_write32(t, CORTEXM_AIRCR,
 	                   CORTEXM_AIRCR_VECTKEY | CORTEXM_AIRCR_SYSRESETREQ);
 
+	/* If target needs to do something extra (see Atmel SAM4L for example) */
```
gsmcmullin: I don't think we need an extra option:
```
if (t->extended_reset)
        t->extended_reset(t);
```

src/target/sam4l.c
```diff
@@ -0,0 +1,410 @@
...
+#define FLASHCALW_FGPFRHI			(FLASHCALW_BASE + 0x14)
+/* Flash General Purpose Registers (low) */
+#define FLASHCALW_FGPFRLO			(FLASHCALW_BASE + 0x18)
+
+static void sam4l_extended_reset(target *t);
+static int sam4l_flash_erase(struct target_flash *f, target_addr addr, size_t len);
+static int sam4l_flash_write_buf(struct target_flash *f, target_addr dest,
+                             const void *src, size_t len);
+
+const struct command_s sam4l_cmds[] = {
```
gsmcmullin: Remove this.  It's unused.

src/target/sam4l.c
```diff
@@ -0,0 +1,410 @@
...
+/*
+ * Populate a target_flash struct with the necessary function pointers
+ * and constants to describe our flash.
+ */
+static void sam4l_add_flash(target *t, uint32_t addr, size_t length)
+{
+	struct target_flash *f = calloc(1, sizeof(struct target_flash));
+	f->start = addr;
+	f->length = length;
+	f->blocksize = 512;
```
gsmcmullin: Use the named constant SAM4_PAGE_SIZE

src/target/sam4l.c
```diff
@@ -0,0 +1,410 @@
...
+	f->erase = sam4l_flash_erase;
+	f->write = target_flash_write_buffered;
+	f->done = target_flash_done_buffered;
+	f->write_buf = sam4l_flash_write_buf;
+	f->buf_size = 512;;
```
gsmcmullin: Use the named constant SAM4_PAGE_SIZE

src/target/sam4l.c
```diff
@@ -0,0 +1,410 @@
...
+	512 * 1024,		/* 10: 512K */
+	0,				/* 11: reserved */
+	1024 * 1024,	/* 12: 1024K (1M) */
+	0,				/* 13: reserved */
+	2048 * 1024,	/* 14: 2048K (2M) */
+	0				/* 15: reserved */
+};
+
+
+#define SAM4L_PAGE_SIZE 512
```
gsmcmullin: Remove.  You're using SAM4_PAGE_SIZE in the rest of the code

src/target/sam4l.c
```diff
@@ -0,0 +1,410 @@
...
+		}
+		return true;
+	}
+	return false;
+}
+
+/*
+ * We've been reset, make sure we take the core out of reset
+ */
+void
```
gsmcmullin: should be static

src/target/sam4l.c
```diff
@@ -0,0 +1,410 @@
...
+sam4l_flash_write_buf(struct target_flash *f, target_addr addr, const void *src, size_t len)
+{
+	target *t = f->t;
+	uint32_t *src_data = (uint32_t *)src;
+	uint32_t ndx;
+	uint16_t page;
+	
+	DEBUG("\nSAM4L: sam4l_flash_write_buf: addr = 0x%08lx, len %d\n", (long unsigned int) addr, (int) len);
+	/* This will fail with unaligned writes, the write_buf version */
+	while (len) {
```
gsmcmullin: This loop isn't needed. The buffered write will always be called with exactly one sector.

src/target/sam4l.c
```diff
@@ -0,0 +1,410 @@
...
+static int
+sam4l_flash_erase(struct target_flash *f, target_addr addr, size_t len)
+{
+	target *t = f->t;
+	uint16_t page;
+
+	DEBUG("SAM4L: flash erase address 0x%08x for %d bytes\n", 
+		(unsigned int) addr, (unsigned int) len);
+	/*
+	 *  NB: if addr isn't aligned to a page boundary, or length
```
gsmcmullin: You're safe here if you put the right values in the memory map.  GDB will do the right thing.  The comment is fine though.

src/target/sam4l.c
```diff
@@ -0,0 +1,410 @@
+/*
+ * vim: set noexpandtab ts=4:
```
gsmcmullin: Remove this

src/platforms/common/swdptap.c
```diff
@@ -36,7 +36,9 @@ static void swdptap_turnaround(uint8_t dir)
 	if(dir == olddir) return;
 	olddir = dir;
 
+#if 0
```
gsmcmullin: Maybe use a `#ifdef` with a named macro `DEBUG_SWD_BITS` or similar.  If someone wants one of these, they will probably want them all.

src/target/sam4l.c
gsmcmullin: Do that in a separate PR when this one is merged if you want to.  It's optional.

PR #195, opened 2017-03-27
Kinetis stuff

gsmcmullin:
- Add support for KL03Z and K22F devices.
- Force security byte to unsecured value by default.
- Allow security byte to be programmed with `mon unsafe enable` command
- Provide a fake recovery target to access mass erase over MDM-AP to recover secured devices.

### DIFF
--- a/src/target/adiv5.c
+++ b/src/target/adiv5.c
@@ -364,15 +364,6 @@ ADIv5_AP_t *adiv5_new_ap(ADIv5_DP_t *dp, uint8_t apsel)
 	if(!tmpap.idr) /* IDR Invalid - Should we not continue here? */
 		return NULL;
 
-	/* Check for ARM Mem-AP */
-	uint16_t mfg = (tmpap.idr >> 17) & 0x3ff;
-	uint8_t cls = (tmpap.idr >> 13) & 0xf;
-	uint8_t type = tmpap.idr & 0xf;
-	if (mfg != 0x23B) /* Ditch if not ARM */
-		return NULL;
-	if ((cls != 8) || (type == 0)) /* Ditch if not Mem-AP */
-		return NULL;
-
 	/* It's valid to so create a heap copy */
 	ap = malloc(sizeof(*ap));
 	memcpy(ap, &tmpap, sizeof(*ap));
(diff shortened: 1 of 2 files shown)


src/target/adiv5.c
```diff
@@ -281,6 +286,10 @@ static void adiv5_component_probe(ADIv5_AP_t *ap, uint32_t addr)
 	if (cid_class == cidc_romtab) { /* ROM table, probe recursively */
 		for (int i = 0; i < 256; i++) {
 			uint32_t entry = adiv5_mem_read32(ap, addr + i*4);
+			if (adiv5_dp_error(ap->dp)) {
+				DEBUG("Fault reading ROM table entry\n");
```
gsmcmullin: I don't really think these should be reported as errors.  Failure to read the ROM table is not significantly different from successfully reading the ROM table and not seeing anything we recognise. Both cases just indicate that this isn't a usable target.

In the current source tree this is only ever called from a GDB monitor command, so it's technically possible to send output with `O` GDB packets.  It is, however, conceivable that this target code could be used in another context where this is not available, so I want to avoid putting GDB dependencies into the target code.

`tc_printf` is the clean function for output in monitor commands from the target code, but that can't be used here as there is no target or attached target controller in this context.

src/target/adiv5.c
```diff
@@ -438,6 +438,9 @@ void adiv5_dp_init(ADIv5_DP_t *dp)
 		if (ap == NULL)
 			continue;
 
+		extern void kinetis_mdm_probe(ADIv5_AP_t *);
```
gsmcmullin: I agree having this two way dependency isn't very nice, but I'm leaving it as it is for now.  We should clean this up and have drivers register themselves with their parents in future, but that is a significant refactor that is out of scope of this PR.

src/target/adiv5.c
esden: Ok that makes sense. I am just advocating for having more verbose and easier to support user bug reports that come in. :)

src/target/adiv5.c
esden: Ok sounds good too. :)

PR #198, opened 2017-03-27
Support for STM32F103Cx Bluepill minimal board

mubes:
Full support for Bluepill including SWO/SWV. No changes are required in the main tree.

### DIFF
--- a/src/platforms/bluepill/Connector
+++ b/src/platforms/bluepill/Connector
@@ -1,0 +1,26 @@
+The connector used for debug is the 0.1" header to the West side of the board with the USB
+connector pointing North. You can get a pinout from;
+
+http://reblag.dk/wordpress/wp-content/uploads/2016/07/The-Generic-STM32F103-Pinout-Diagram.pdf
+
+My pinning and colours are;
+
+Pin number    Name        Colour   Purpose
+from bottom
+   1          3v3	  White    3v3 Feed
+   2          Gnd      !* Black    Ground Connection
+   3          5v          Brown    5V USB Derrived Feed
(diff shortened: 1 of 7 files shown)


tink3r:
Nice! Been tinkering with a blue pill these days and have one running using stlink platform. I will test your version next week!

BobRyan530:
Thanks! Works for me.

pragtich:
Hi,

Was trying this today, but keep ending up with oversized results. Even the 0c26377 commit ends up too large:
`
/usr/local/bin/arm-none-eabi-ld: blackmagic section '.data' will not fit in region 'rom'
/usr/local/bin/arm-none-eabi-ld: region 'rom' overflowed by 12 bytes
`

Any tips on this?

mubes:
Hi,

I pulled the build into line with the BMP 1.6 release, but unfortunately 
that has pushed it over the 64K of memory that's 'officially' on that 
part. You have two choices....either roll back to the version just 
before the integration of the 1.6 changes or change the linker config 
file in ./src/platforms/stm32/stlink.ld to up the LENGTH from 64K to 
128K....quite frankly, you're better of doing the latter because this 
problem will only become worse over time and you're already using a part 
beyond it's spec (the Bluepill chip doesn't even _officially_ have USB).

If you are burning the version onto the board using the DFU interface 
then you will need to include the 'force' option because the DFU loader 
doesn't think it's got memory beyond the 64K.

Regards

DAVE

pragtich:
Thanks,
That makes sense.

Joris

slurdge:
Hi, 

Thank for the work. I tried 6cd7e7d and it worked fine. I had to flash it externally though.

Issue #185, opened 2017-01-30
Unable to flash blackmagic firmware to Launchpad TM4C123GXL ICDI


hazelnusse:
I have the following setup:
- BMPM2 running the blackmagic firmware (sha: 7ba2987a83bab1adb1db9b7f167e2454c7cb9b44)
- EK-TM4C123GXL, power switch set to DEBUG
- Tag Connect cable: TC2050-IDC-NL
- GNU ARM Embedded Toolchain, version 6-2016-q4-major
- Gentoo Linux on a Lenovo T450s

I first connect the BMP2 to my laptop via the USB cable, here is what I see from dmesg:
[97703.515430] usb 2-3: New USB device found, idVendor=1d50, idProduct=6018
[97703.515433] usb 2-3: New USB device strings: Mfr=1, Product=2, SerialNumber=3
[97703.515435] usb 2-3: Product: Black Magic Probe
[97703.515436] usb 2-3: Manufacturer: Black Sphere Technologies
[97703.515437] usb 2-3: SerialNumber: D5DABEF4
[97703.516025] cdc_acm 2-3:1.0: ttyACM0: USB ACM device
[97703.516387] cdc_acm 2-3:1.2: ttyACM1: USB ACM device

So far so good.

Next, I connect the EK-TM4C123GXL to my laptop via the debug USB, here is what I see from dmesg:
[97807.165383] usb 2-1: new full-speed USB device number 66 using xhci_hcd
[97807.334651] usb 2-1: New USB device found, idVendor=1cbe, idProduct=00fd
[97807.334654] usb 2-1: New USB device strings: Mfr=1, Product=2, SerialNumber=3
[97807.334656] usb 2-1: Product: In-Circuit Debug Interface
[97807.334657] usb 2-1: Manufacturer: Texas Instruments
[97807.334658] usb 2-1: SerialNumber: 0E210604
[97807.335134] cdc_acm 2-1:1.0: ttyACM2: USB ACM device

And the LEDs are blinking on the EK-TM4C123GXL.

Next, I connect the Tag Connect to the BMP2 hardware via the 10pin connector. Next, I connect the other end to the Tag Connect JTAG connection on the EK-TM4C123GXL.

Then, I build the blackmagic firmware for the launchpad-icdi by running:
```
$ make PROBE_HOST=launchpad-icdi
```
This results in these two files:
```
ls -l src/blackmagic*
[...]

PR #202, opened 2017-04-12
lmi: Inhibit SRST on Tiva and add some fault checks.

gsmcmullin:
Fix for #185.  On this target, asserting SRST resets debug state causing flash programming to fail.

### DIFF
--- a/src/target/lmi.c
+++ b/src/target/lmi.c
@@ -84,6 +84,7 @@ bool lmi_probe(target *t)
 		t->driver = lmi_driver_str;
 		target_add_ram(t, 0x20000000, 0x10000);
 		lmi_add_flash(t, 0x80000);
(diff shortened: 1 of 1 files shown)


src/target/lmi.c
```diff
@@ -84,6 +84,7 @@ bool lmi_probe(target *t)
 		t->driver = lmi_driver_str;
 		target_add_ram(t, 0x20000000, 0x10000);
 		lmi_add_flash(t, 0x80000);
+		t->target_options |= CORTEXM_TOPT_INHIBIT_SRST;
```
esden: I think this line needs a comment saying why we inhibit this. Something like "The Tiva platform is resetting the debug subsystem together with the system reset preventing us from using SRST" or similar.

PR #203, opened 2017-04-12
Add LPC11xx command to read out unique ID from target.

dlaw:
This pull request adds the "readuid" command, to read out the unique ID from an LPC11xx/LPC8xx target. I've tested it with a couple of LPC824s and verified that the UID printed in GDB matches the UID that I read out in my own code.

The function `lpc_iap_call()` did not support IAP commands with a return value, so I had to alter it slightly. I've attached the relevant page of the user manual for reference. The maximum number of `uint32`s returned by any IAP command is 4.

iap_format

### DIFF
--- a/src/target/lpc11xx.c
+++ b/src/target/lpc11xx.c
@@ -35,6 +35,26 @@
 #define LPC11XX_DEVICE_ID  0x400483F4
 #define LPC8XX_DEVICE_ID   0x400483F8
 
+static bool lpc11xx_read_uid(target *t, int argc, const char *argv[])
+{
+	(void)argc;
+	(void)argv;
+	struct lpc_flash *f = (struct lpc_flash *)t->flash;
+	uint8_t uid[16];
+	if (lpc_iap_call(f, uid, IAP_CMD_READUID))
+		return false;
+	tc_printf(t, "UID: 0x");
+	for (uint32_t i = 0; i < sizeof(uid); ++i)
+		tc_printf(t, "%02x", uid[i]);
+	tc_printf(t, "\n");
+	return true;
+}
+
+const struct command_s lpc11xx_cmd_list[] = {
+	{"readuid", lpc11xx_read_uid, "Read out the 16-byte UID."},
+	{NULL, NULL, NULL}
+};
+
 void lpc11xx_add_flash(target *t, uint32_t addr, size_t len, size_t erasesize)
 {
 	struct lpc_flash *lf = lpc_add_flash(t, addr, len);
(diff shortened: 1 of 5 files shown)


dlaw:
Hi folks,

Could you let me know if you are interested in merging this PR?

I would be happy to make additional changes if you feel it is not suitable in its current state.

Cheers,
David

dlaw:
Dear Black Magic Probe maintainers,

Is there any timetable for evaluating this pull request? I remain available to make any changes or perform any additional tests that are requested. I just rebased my change onto the current head of master.

This functionality has become quite important to my work (I log the UID of every chip that is programmed by my Black Magic Probe setup).

Thank you for your consideration.

Sincerely,
David Lawrence

UweBonnes:
Gareth told about "September" for a review of the the patch and issue backlog...

gsmcmullin:
The code looks ok. I'm happy to merge.  The only problem with this is that making the IAP call may be problematic if the target is in an unknown state.  In the cases where these calls are made for programming flash, the target has just been reset, so we know what we're dealing with.  The monitor command could be given at any time.

I suggest at least save the memory used and restore after, and globally disable interrupts in the target while making the call.  Is there no way to read the unique ID my memory-mapped IO?

UweBonnes:
>  Is there no way to read the unique ID my memory-mapped IO?
Probably no documented way...

dlaw:
Thanks for the feedback, and for all your work on the Black Magic Probe. I will make the suggested modifications (to disable interrupts and restore the memory and registers which get clobbered). It may take me a few weeks to get it all polished and tested, because work is pretty busy right now.

I poked around and unfortunately the location of the unique ID in the boot ROM varies between different versions of the boot ROM, so the only safe way to read it is to use the API call. It is technically possible to get the UID using MMIO, by reading in the boot ROM and running it on the BMP in an emulator... but that seems like a lot more work than restoring state properly. :-)

dlaw:
After a *very* long delay (my apologies), we had a chance to revisit this pull request. Changes:
1. Rebased on master.
2. Added support for LPC15xx in addition to LPC8xx/LPC11xx.
3. Updated LPC IAP function call to save and restore registers and target RAM.
4. Tested on LPC824 and LPC1549.

Do you have any thoughts on the best way to disable interrupts during an IAP call? The DHCSR (debug halting control and status register) only permits interrupts to be disabled during single stepping. We could set FAULTMASK before the call and restore its value afterwards, but this still runs the risk of a non-maskable interrupt executing during an IAP call. And I am a little nervous about tinkering with so much of the target state and expecting it to be totally transparent to the target program.

Cheers,
David

UweBonnes:
Sorry, didn't notice the rebase and rework. It seem all issue that Gareth had are resolved now.

PR #205, opened 2017-04-15
stlink: Announce 124 kiB flash size in DFU to commplement #204.

UweBonnes:
STLinkV2-1 has F103CB on board! F103C8 can use upper flash with hopefully
acceptable error rate.

### DIFF
--- a/src/platforms/f4discovery/platform.h
+++ b/src/platforms/f4discovery/platform.h
@@ -33,9 +33,7 @@
 
 #define PLATFORM_HAS_TRACESWO
 #define BOARD_IDENT       "Black Magic Probe (F4Discovery), (Firmware " FIRMWARE_VERSION ")"
-#define BOARD_IDENT_DFU   "Black Magic (Upgrade) for F4Discovery, (Firmware " FIRMWARE_VERSION ")"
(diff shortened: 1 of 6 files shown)


UweBonnes:
Another approach would be to announce the amount of flash found in the flashsize register and leave it up to the user to force dfu to write not flash not announced...

gsmcmullin:
Please rebase this as it conflicts with your changes in #206 (merged).  I like your suggestion of reporting the flash size in the fuse registers and force the user to override if they choose to.

UweBonnes:
The pull request should be rebased now and the DFU String is taken from the  FLASH_SIZE register minus the bootloader area.

If you  look at the old calculation of max_address, the code did not what was intended...

UweBonnes:
The DFU update string now reflects device read or bootloader write protection, indication 0 available pages.

UweBonnes:
Up to now, dfu-utils at https://sourceforge.net/p/dfu-util/dfu-util/ci/master/tree/ does not erase flask pages above the l,imit given in the dfu-string. https://sourceforge.net/p/dfu-util/tickets/44/ helps in that situation.

Issue #199, opened 2017-03-30
No range-checking on gdb packets. (too-large packets crash probe)


markrages:
In gdb_main.c, in handling 'm' and 'M' at least, there is no check on the gdb-supplied length arguments, allowing writing past the end of the 1024-byte packet buffer.  This results in a crash, requiring a power-cycle to make the probe functional again.

This doesn't happen with gdb, which respects the PacketSize from qSupported, but it can happen with gdb.py or other user scripting.

(I have a patched gdb.py with many changes, including qSupported, if anyone is interested.)

gsmcmullin:
@markrages Please close this issue if #208 resolves this for you, otherwise let me know what is still not working.  Please also submit a PR with fixes for the python stuff.

PR #208, opened 2017-04-18
Add range checking on mem access packets.

gsmcmullin:
Fix for #199


gsmcmullin:
Well this makes it fail, which is better than crashing, but there isn't much we can do if the other side of the link breaks the rules.

PR #209, opened 2017-04-18
F4 discovery

UweBonnes:
Hello,

this patch cleans up the f4 discovery platform. Only the internal bootloader is used for dfu upload. For better handling with the dfu-utils, BMP generates the same serial number as the built-in bootloader. A single call like 
 dfu-util -a 0 -s 0x08000000:leave -D src/blackmagic.bin
now jumps to the bootloader, reflashes and starts the new user program as expected . For me, master (v288) did not do so.

For bootloader entry, a magic constant placed at _ebbs is now used. Bootloader entry is initialied by  a system reset. So bootloader is now called early in initialization, minimizing chances of interrupts or other settings getting into the way.

The concept of a magic constant in RAM is also usefull when using a custom bootloader as for native/stlink/swlink. However _ebss is not the right place in that situation, as _ebss is different for bootloader and user program. I have used _flash, with the flash definition in Makefile.inc decreased by 8 bytes. Again this allows bootloader entry by issuing a system reset.


UweBonnes:
Second to last sentence should read:
I have used _stack,  with the  _stack definition in Makefile.inc decreased by 8 bytes

UweBonnes:
Some slight changes needed. I will annonce another pull request.

PR #210, opened 2017-04-18
F4 discovery

UweBonnes:
Hello,

this patch cleans up the f4 discovery platform. Only the internal bootloader is used for dfu upload. For better handling with the dfu-utils, BMP generates the same serial number as the built-in bootloader. A single call like
dfu-util -a 0 -s 0x08000000:leave -D src/blackmagic.bin
now jumps to the bootloader, reflashes and starts the new user program as expected . For me, master (v288) did not do so.

For bootloader entry, a magic constant placed at _ebbs is now used. Bootloader entry is initialized by a system reset. So bootloader is now called early in initialization, minimizing chances of interrupts or other settings getting into the way.

The concept of a magic constant in RAM is also usefull when using a custom bootloader as for native/stlink/swlink. However _ebss is not the right place in that situation, as _ebss is different for bootloader and user program. I have used _stack, with the _stack definition in Makefile.inc decreased by 8 bytes. Again this allows bootloader entry by issuing a system reset.


src/include/general.h
```diff
@@ -42,5 +42,8 @@
 #undef MIN
 #define MIN(x, y)  (((x) < (y)) ? (x) : (y))
 
+#define BOOTMAGIC0 0xb007da7a
```
gsmcmullin: Please move these to src/platforms/f4discovery/platform.h

src/platforms/f4discovery/platform.c
```diff
@@ -34,25 +34,41 @@
...
 void platform_init(void)
 {
+	volatile uint32_t *magic = (uint32_t *) &_ebss;
 	/* Check the USER button*/
-	rcc_peripheral_enable_clock(&RCC_AHB1ENR, RCC_AHB1ENR_IOPAEN);
-	if(gpio_get(GPIOA, GPIO0)) {
-		platform_request_boot();
+	rcc_periph_clock_enable(RCC_GPIOA);
+	if (gpio_get(GPIOA, GPIO0) ||
+		((magic[0] == BOOTMAGIC0) && (magic[1] == BOOTMAGIC1))) {
```
gsmcmullin: Please fix the indentation here so it lines up with the opening bracket of the if condition.

UweBonnes:
I force pushed the wanted changes.

PR #215, opened 2017-04-22
Add 96Boards Carbon as a probe host

daniel-thompson:
The carbon contains two SoCs, an STM32 (host) and an nRF51 (BLE). The
STM32 implements the probe and allows the board to reprogram its own
radio firmware!

### DIFF
--- a/src/platforms/96b_carbon/Makefile.inc
+++ b/src/platforms/96b_carbon/Makefile.inc
@@ -1,0 +1,28 @@
+CROSS_COMPILE ?= arm-none-eabi-
+CC = $(CROSS_COMPILE)gcc
+OBJCOPY = $(CROSS_COMPILE)objcopy
+
+CFLAGS += -Istm32/include -mcpu=cortex-m4 -mthumb \
+	-mfloat-abi=hard -mfpu=fpv4-sp-d16 \
+	-DSTM32F4 -D_96B_CARBON -I../libopencm3/include \
+	-Iplatforms/stm32
+
+LDFLAGS = -lopencm3_stm32f4 -Wl,--defsym,_stack=0x20006000 \
+	-Wl,-T,platforms/stm32/96b_carbon.ld -nostartfiles -lc -lnosys \
+	-Wl,-Map=mapfile -mthumb -mcpu=cortex-m4 -Wl,-gc-sections \
+	-mfloat-abi=hard -mfpu=fpv4-sp-d16 \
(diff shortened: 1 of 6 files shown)


daniel-thompson:
Any new on this? I can see a few changes have landed on the mainline since I posted. It looks like the merge remains clean but if you think a rebase and re-test is needed then just let me know...

daniel-thompson:
@pfalcon reported a typo in the Carbon's Readme.md so whilst fixing that I have done the rebase anyway (there were no conflicts).

Any more changes needed?

pfalcon:
+1 for merging this

Mani-Sadhasivam:
Any reason for not merging this? I've tested this port on 96Boards Carbon to flash and debug STM32F411xE based custom board. It would be good to have this support in mainline as the usecase will more likely to increase.

pfalcon:
In regard to recently added "$3 Board Support" label. Well, the target board of this patch, 96Boards Carbon https://www.96boards.org/product/carbon/ , isn't a "$3 board". It's not as expensive as an original BMP too, but it's nowhere near being a real SWD programmer - it's a board intended to explore 6LoWPAN connectivity using Zephyr RTOS http://zephyrproject.org/ . The only realistic usecase for running BMP firmware on it is to program one onboard MCU with another (as mentioned in the description, the board has 2 - one "application MCU", and another is "BLE controller").

Hopefully this might help to assess a level of conflict of interest for maintainers' spending effort to merge this patch. (Otherwise, we're just happy to allow user to self-program the BLE part, and love to merge things upstream ;-) ).

dragonmux:
Sorry this PR has dragged on! Thank you for the contribution.

Once the v1.8 release cycle has been completed, we'll try and review this in short order and see about merging it. This shouldn't have dragged on as long as it has.

dragonmux (approved):
LGTM

esden (approved):
I do have to ask if the copyright in the file headers correct? I assume the files mostly just a copy from one of the other host boards with minor modifications? If so that is fine. But in general it would be good to keep some record who added the code.

That said, there are two reasons why I am ok with merging this PR none the less:
1. The host board support will be consolidated with the upcoming build system update. So the duplicated files will essentially be deleted and only the differences kept as a config file.
2. This patch is sitting for so long that I am not expecting the original authors to come back to make copyright header fixes. 

Keeping all that in mind. I will approve this PR.

PR #216, opened 2017-04-23
Support for NXP MK20

tommz9:
I added a (basic) flash writing support for [NXP MK20](http://www.nxp.com/products/microcontrollers-and-processors/arm-processors/kinetis-cortex-m-mcus/k-series-performance-m4/k2x-usb/kinetis-k20-120-mhz-full-speed-usb-mixed-signal-integration-microcontrollers-mcus-based-on-arm-cortex-m4-core:K20_120).

- Successfully identifies the MCU.
- It can write to flash, verify sections and do debugging (run, breakpoints, ...).
- Tested with MK20FX512VLQ12 (172kB of program and data written to flash).

I can see two problems so far:

- Programming the flash is very slow. Solution: Use _Program section_ and _Program Partition_ commands instead of _Program Phrase_ command.
- functions for KL and K series are very similar. Solution: refactor.

### DIFF
--- a/src/target/kinetis.c
+++ b/src/target/kinetis.c
@@ -59,6 +59,8 @@
 #define FTFA_CMD_ERASE_ALL         0x44
 #define FTFA_CMD_BACKDOOR_ACCESS   0x45
 
+#define FTFE_CMD_PROGRAM_PHRASE    0x07
+
 #define KL_GEN_PAGESIZE 0x400
 
 static bool kinetis_cmd_unsafe(target *t, int argc, char *argv[]);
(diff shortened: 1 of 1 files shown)


src/target/kinetis.c
```diff
@@ -152,6 +163,28 @@ bool kinetis_probe(target *t)
 		kl_gen_add_flash(t, 0, 0x40000, 0x800);
 		kl_gen_add_flash(t, 0x40000, 0x40000, 0x800);
 		break;
+	case 0x0: /* Chips that do not support [FAMILYID SUBFAMID SERIESID] fields. */
+
+		/* MK20 has 0x1 in FAMID (bits 4-6) */
+		if(((sdid >> 4) & 0x7) == 0x1){
+			t->driver = "MK20";
+			const uint32_t sram_l_start = 0x1C000000;
```
gsmcmullin: Why not just put the base and length as call parameters.  They're no less magic this way.
The length here is also one too low.

src/target/kinetis.c
```diff
@@ -253,6 +286,58 @@ static int kl_gen_flash_done(struct target_flash *f)
...
+                           uint32_t addr, size_t length, size_t erasesize){
+
+	struct target_flash *f = calloc(1, sizeof(*f));
+
+	f->start = addr;
+	f->length = length;
+	f->blocksize = erasesize;
+	f->erase = mk_gen_flash_erase;
+	f->write = mk_gen_flash_write;
+	f->done = kl_gen_flash_done;
```
gsmcmullin: Is this going to work?  `kl_gen_flash_done` may use `FTFA_CMD_PROGRAM_LONGWORD` to write the security byte.

cosmikwolf:
Curious if you guys have thought about this again since this pull request was closed a year ago? I'd love MK20 support so I can get this working with my Teensy 3.1. I am trying to get black magic probe to connect as a debugging tool.

PR #226, opened 2017-05-07
RFC: i.MX 7 support

agners:
With this modifications I can successfully scan an i.MX 7Solo or Dual SoC:

    (gdb) monitor jtag_scan
    Target voltage: 3.3V
    Available Targets:
    No. Att Driver
     1      ARM Cortex-A
     2      i.MX 7Solo/Dual ARM Cortex-M4

Attaching the M4 core works too, even when the core is in reset (the targets attach function takes it out of reset). Attaching the A7 does not work currently.

### DIFF
--- a/src/target/imx7.c
+++ b/src/target/imx7.c
@@ -1,0 +1,86 @@
+/*
+ * This file is part of the Black Magic Debug project.
+ *
+ * Copyright (C) 2011  Black Sphere Technologies Ltd.
+ * Written by Gareth McMullin <gareth@blacksphere.co.nz>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*
+ * This file implements NXP i.MX 7Solo/Dual detection and reset.
+ *
+ * Reference:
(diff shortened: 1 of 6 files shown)


src/target/imx7.c
```diff
@@ -0,0 +1,124 @@
...
+ * - MX7DRM, Rev. 0.1, 08/2016
+ * - SoC detection from U-Boot v2016.11
+ */
+
+#include "general.h"
+#include "target.h"
+#include "target_internal.h"
+#include "cortexm.h"
+
+#define IMX7_AHP_AP_IDR		0x64770001
```
gsmcmullin: This only identifies an ARM AHB.  It doesn't contain anything i.MX7 specific.

src/target/imx7.c
```diff
@@ -0,0 +1,124 @@
...
+{
+	uint32_t val;
+	switch(ap->idr) {
+	case IMX7_AHP_AP_IDR:
+		break;
+	default:
+		return;
+	}
+
+	val = adiv5_mem_read32(ap, SRC_A7RCR1);
```
gsmcmullin: I'm concerned about probing this on any AHB.  If this address is not implemented on some other SoC it will stall the AHB bus there.

src/target/imx7.c
```diff
@@ -0,0 +1,124 @@
...
+	if (!(val & SRC_A7RCR1_A7_CORE1_ENABLE)) {
+		DEBUG("i.MX 7 Secondary A7 core not enabled, installing ROM filter.\n");
+		ap->dp->rom_table_filter = imx7_rom_table_filter;
+	}
+}
+
+static bool imx7_rom_table_filter(uint32_t addr)
```
gsmcmullin: I think we should take an `apsel` argument here too.  Address space need not be common across all busses.
The code below could then only mask out these addresses on the debug APB, not on the M4, for example.

UweBonnes:
Could you have a look at at the debug port if it is DPv2? If it is V2, what is ADIV5_DP_IDCODE? Perhaps when knowing the TARGETID can help with the problem?

UweBonnes:
Posting the output of "blackmagic -t" (bmp-hosted)  here  would be a good reference for further work.

esden:
According to @agners in #227 this project is now orphaned. Thus I am closing this PR.

PR #231, opened 2017-05-20
Stlink V2 - generic clone "Baite"

andrewleech:
I'm testing a stlink v2 clone branded "Baite" with blackmagic firmware. This appears to be a relatively common clone with multiple references / images of it when searching on google.
It was originally purchased from adafruit I believe.

Its layout is basically half way between the stlink V1 and V2, but unfortunately I was unable to find a method of detecting it automatically. 
As such I've added a compile time setting to allow overriding hw version detection and adding support for this programmers differences.

The basic differences of this hardware amount to: 
* using V2 led gpio (but with only a single led) 
* using V2 srst pin
* not needing MCO on PORTA8 with HSE like V2
* doesn't have correct wiring to be detected as V2

For more details: http://www.stm32duino.com/viewtopic.php?t=1357&start=10#p17455

This changeset can be used as such:
`make PROBE_HOST=stlink HWVERSION=Baite`
where `HWVERSION={V1, V2, Baite, detect}`  with default `detect` matching original functionality.

### DIFF
--- a/src/platforms/stlink/Makefile.inc
+++ b/src/platforms/stlink/Makefile.inc
@@ -2,10 +2,12 @@ CROSS_COMPILE ?= arm-none-eabi-
 CC = $(CROSS_COMPILE)gcc
 OBJCOPY = $(CROSS_COMPILE)objcopy
 
+HWVERSION ?= detect
+
 OPT_FLAGS = -Os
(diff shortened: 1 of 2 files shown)


UweBonnes:
Baite board are now supported by the stlink platform.

PR #239, opened 2017-06-07
Support for asynchronous SWV

mubes:
Addition of Async SWV capability to native BMP.  This version assumes access to B7 and does not impact on the behaviour of the existing serial port.


src/platforms/stm32/traceswoasync.c
```diff
@@ -0,0 +1,136 @@
...
+
+{
+    static volatile char inBufDrain;
+
+    /* If we are already in this routine then we don't need to come in again */
+    if (__atomic_test_and_set (&inBufDrain, __ATOMIC_RELAXED))
+        return;
+
+    /* Attempt to write everything we buffered */
+  if ((w!=r) && (usbd_ep_write_packet(dev, ep, &trace_rx_buf[r*FULL_SWO_PACKET], FULL_SWO_PACKET)))
```
gsmcmullin: I'm confused about these packet sizes.  I'll need to go back and check the packet memory allocations, but I was sure everything fitted correctly before.  Now we only allocate for 48 byte packets in the call to `usbd_ep_setup`, but here potentially write up to 64 bytes.

src/platforms/common/cdcacm.c
```diff
@@ -188,14 +188,14 @@ static const struct usb_endpoint_descriptor uart_data_endp[] = {{
 	.bDescriptorType = USB_DT_ENDPOINT,
 	.bEndpointAddress = 0x03,
 	.bmAttributes = USB_ENDPOINT_ATTR_BULK,
-	.wMaxPacketSize = CDCACM_PACKET_SIZE,
+	.wMaxPacketSize = 48, //CDCACM_PACKET_SIZE,
```
gsmcmullin: Magic numbers are bad, and misleading comments are worse.  If this needs to change, perhaps we just change the value of `CDCACM_PACKET_SIZE`?

src/command.c
```diff
@@ -277,12 +277,21 @@ static bool cmd_target_power(target *t, int argc, const char **argv)
...
 
 #ifdef PLATFORM_HAS_TRACESWO
-static bool cmd_traceswo(void)
+static bool cmd_traceswo(target *t, int argc, const char **argv)
 {
-	extern char serial_no[9];
-	traceswo_init();
-	gdb_outf("%s:%02X:%02X\n", serial_no, 5, 0x85);
-	return true;
+  extern char serial_no[9];
```
gsmcmullin: Indentation should be tabs

src/command.c
```diff
@@ -277,12 +277,21 @@ static bool cmd_target_power(target *t, int argc, const char **argv)
...
+  uint32_t speed=0;
+  (void)t;
+
+  if (argc>1)
+    speed=atoi(argv[1]);
+
+  traceswo_init(speed);
+  if (!speed)
+    {
+      gdb_outf("%s:%02X:%02X\n", serial_no, 5, 0x85);
```
gsmcmullin: This information was output so that a GDB python extension could identify the USB endpoint for trace.  I don't see why it should be suppressed if speed is specified.

src/platforms/stm32/traceswoasync.c
gsmcmullin: The USB hardware also has pingpong buffers.  This may help performance if USB is a bottleneck.  I never finished adding support for this in libopencm3, it's possible that someone else has finished this by now.

mubes: Performance testing shows that the pingpong isn't a major cycle hog at the maximum speeds we can run the interface at (about 5% of the CPU), and I'd prefer to keep the changes as small as possible, so I propose to leave this one as-is.

src/platforms/common/cdcacm.c
mubes: It does indeed need to change, which is why I deliberately left this grot in so it would be seen.  If we change the size of CDCACM_PACKET_SIZE it will affect all the CDCs, including the gdb one which _does_ deal with fast data and probably benefits from the large size buffer. I worked out that the overall buffer was 32 octets overcommitted, and indeed the second 32 bytes of the trace buffer were corrupt. a new definition for the second serial port size is probably needed, assuming your assessment of USB allocation is the same as mine.

src/platforms/stm32/traceswoasync.c
mubes: There is room for this packet with the updated buffer allocations...see the comments below.

src/command.c
mubes: True enough. Reverted in next version.

src/command.c
mubes: Fixed.

UweBonnes:
Closing: Implemented in the meantime.

PR #241, opened 2017-06-09
Use vector_table symbol for assignment of SCB_VTOR.

gsmcmullin:


esden:
Just as a sidenote, you could also include `cm3/vector.h` and then do `extern vector_table_t vector_table` not that it really matters as you are only interested in the pointer.

PR #245, opened 2017-06-15
stm32f4: add support for STM32F4[67]9

schodet:


schodet:
Rebased after changes on master.

PR #246, opened 2017-06-15
stm32l0: add support for large STM32L0

schodet:
Up to 6 Ko EEPROM, 20 Ko SRAM and read flash size from flash size register.

I have actually done this before seeing #119, I suppose there is a better way than defining several flash banks (which does not map to hardware bank), but it works.

I also tried to add a loop in `stm32lx_nvm_prog_erase`, but it is not sufficient to solve the > 64 Kb problem.

Tested with a STM32L072.


schodet:
Rebased on current master.

This commit is less useful after 2216587b, what do you think?

esden:
Thank you @schodet for the submission. Yes you are correct the newly reported size should incorporate the larger devices now. This is the solution we will go with for now. If there are any more issues please feel free to reopen this pull request. :)

PR #247, opened 2017-06-16
Allow programming STM32F4 when using a low voltage

schodet:
This adds:
- programming using byte access (thanks @GordonHJSmith),
- allow selection of access size using `monitor psize (x8|x32)`.

It would be great to also select the right one depending on the target voltage, but there is no platform function for that (it returns a string).


src/target/stm32f4.c
```diff
@@ -355,3 +370,30 @@ static bool stm32f4_cmd_option(target *t, int argc, char *argv[])
 	}
 	return true;
 }
+
+static bool stm32f4_cmd_psize(target *t, int argc, char *argv[])
+{
```
gsmcmullin: Please put braces on the same line as the if and else, and add braces to the ifs below (strcmp's)

src/target/stm32f4.c
```diff
@@ -355,3 +370,30 @@ static bool stm32f4_cmd_option(target *t, int argc, char *argv[])
...
+			psize = 8;
+		else if (!strcmp(argv[1], "x32"))
+			psize = 32;
+		else
+		{
+			tc_printf(t, "usage: monitor psize (x8|x32)\n");
+			return false;
+		}
+		for (struct target_flash *f = t->flash; f; f = f->next) {
+			((struct stm32f4_flash *)f)->psize = psize;
```
gsmcmullin: This works for now, but if at some stage we support external flashes this could come back and bite us.

schodet: Any idea on where I could store this information? A global?

gsmcmullin: I'm happy to leave it here for now.  This is better than a global or static.

If we support external flashes in future, I think the solution would be to have some signature in the `target_flash` structure so that flashes belonging to this driver can be identified.

schodet:
I added a commit for style fixes.

The second commit is a proposition for a kind of "signature".

I can apply fixups if you agree or you can do it before merging.

src/target/stm32f4.c
schodet: Done.

gsmcmullin:
This looks great.  Please apply the fixups and I'll merge it.

schodet:
Done.

PR #257, opened 2017-07-12
PSoC4: Initial additions/adjustments

mryndzionek:
This is the beginning of series of additions/adjustments aiming add adding basic Cypress PSoC4 chips support.

### DIFF
--- a/src/target/stm32l4.c
+++ b/src/target/stm32l4.c
@@ -131,7 +131,8 @@ enum ID_STM32L4 {
 	ID_STM32L47  = 0x415u, /* RM0351, Rev.5 */
 	ID_STM32L49  = 0x461u, /* RM0351, Rev.5 */
 	ID_STM32L4R  = 0x470u, /* RM0432, Rev.5 */
-	ID_STM32G07  = 0x460u, /* RM0444/454, Rev.1 */
+	ID_STM32G03  = 0x468u, /* RM0444/454, Rev.2 */
+	ID_STM32G07  = 0x460u, /* RM0444/454, Rev.2 */
 };
 
 enum FAM_STM32L4 {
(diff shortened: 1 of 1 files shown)


mryndzionek:
Has this pull request a chance to be accepted? The first commit just adds a new command that could be useful elsewhere. The second shouldn't break anything.

esden:
Thank you for your submission.

I am not comfortable with adding a function that retries 10 times to connect. This should never be necessary if the target is configured and set up correctly. Is there a specific reason why that would be necessary in any other context than problematic targets?

Also this should be possible to be done by running `mon swdp` and then pressing enter in GDB 10 times. Is Psoc that problematic to connect to?

mryndzionek:
Please look at the PDF I linked in the commits:
http://www.cypress.com/documentation/programming-specifications/cybl10x6x-cy8c4127bl-cy8c4247b-programming-specifications
'4.3 Step 1 – Acquire Chip'

> The times of internal reset (<1 ms) and boot code (<100 µs)
are not specified exactly. Because they depend on the CPU
clock and the size of the code, they can vary in different revisions
of the chip.
In this case, the recommended way to enter Test mode is to
start sending an acquire sequence right after XRES is toggled
(or power is supplied in Power Cycle mode). This
sequence is sent iteratively until it succeeds; that is, all SWD
transactions are ACKed and all conditions are met. 

So yes, it's problematic. I know that there is another way. OpenOCD tries to use it, but it's also complex and less reliable. PSO4 is strange. For info you can look at:
http://dmitry.gr/index.php?r=05.Projects&proj=24.%20PSoC4%20confidential

UweBonnes:
I don't have a PSOC4. Is your problem solved? If yes, how?

mryndzionek:
Oh sorry. I forgot that this branch is a pending PR. I've forced pushed to sync and pull from master and this closed the PR. This isn't urgent right now. I have the changes on my clone. I'll re-open the PR if I figure out some other solution.

PR #258, opened 2017-07-14
Make Cortex M driver write DCCIMVAC prior to reading or writing each 32 bytes of RAM

nickd4:
Single stepping / tracing did not work properly with STM32F767, debugger couldn't see values of local variables etc. This turns out to be because the data cache was enabled so memory was out of date.

This gist demonstates the problem, I captured it before I debugged the issue and reported it on gitter:
https://gist.github.com/nickd4/ec8628c2ce90658c3efe44ead3d2116e
Looking at the register dump, it had just entered the function InitGraphics() and stored the value from r0 into pNewLCD (the name of the function parameter), r0 contained the correct value and r7+4 contained the correct address of pNewLCD and pc was the instruction just after the store... so p/x pNewLCD should have shown the same value as in r0, but it showed 0 instead, which was clearly incorrect.

The problem went away when I disabled the D-cache but this was clearly unsatisfactory so I looked up in the Cortex-M documentation how to flush the cache prior to accessing memory, here it is here:
http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0646b/BABIGDDC.html
Out of this list, the most appropriate command in this situation seemed to be writing DCCIMVAC. So I tried this in gdb and it appeared to fix the problem, here is the gdb session demonstrating this:
https://gist.github.com/nickd4/222ea242febde0c6c35b8bffbea9ca75

Therefore, I incorporated this change into the Black Magic firmware, with several optimizations, firstly it only does it for RAM regions (I figure flash will never change and memory mapped registers are not cached), secondly it does 32 byte blocks since apparently only bits 31..5 of DCCIMVAC are significant.

I then captured a final gdb session with the revised firmware showing that it definitely fixes the problem:
https://gist.github.com/nickd4/bdb13234db7d3a12b60eeb27cc7606f5

### DIFF
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -181,11 +181,43 @@ ADIv5_AP_t *cortexm_ap(target *t)
 
 static void cortexm_mem_read(target *t, void *dest, target_addr src, size_t len)
 {
+	/* flush data cache for RAM regions that intersect requested region */
+	target_addr src_end = src + len; /* following code is NOP if wraparound */
+	/* requested region is [src, src_end) */
+	for (struct target_ram *r = t->ram; r; r = r->next) {
+		target_addr ram = r->start;
+		target_addr ram_end = r->start + r->length;
+		/* RAM region is [ram, ram_end) */
+                if (src > ram)
+			ram = src;
+		if (src_end < ram_end)
+			ram_end = src_end;
+		/* intersection is [ram, ram_end) */
+		for (ram &= ~0x1f; ram < ram_end; ram += 0x20)
+			adiv5_mem_write(cortexm_ap(t), CORTEXM_DCCIMVAC, &ram, 4);
+	}
+
 	adiv5_mem_read(cortexm_ap(t), dest, src, len);
 }
 
 static void cortexm_mem_write(target *t, target_addr dest, const void *src, size_t len)
 {
+	/* flush data cache for RAM regions that intersect requested region */
+	target_addr dest_end = dest + len; /* following code is NOP if wraparound */
+	/* requested region is [dest, dest_end) */
+	for (struct target_ram *r = t->ram; r; r = r->next) {
+		target_addr ram = r->start;
+		target_addr ram_end = r->start + r->length;
+		/* RAM region is [ram, ram_end) */
+                if (dest > ram)
+			ram = dest;
+		if (dest_end < ram_end)
+			ram_end = dest_end;
+		/* intersection is [ram, ram_end) */
+		for (ram &= ~0x1f; ram < ram_end; ram += 0x20)
+			adiv5_mem_write(cortexm_ap(t), CORTEXM_DCCIMVAC, &ram, 4);
+	}
+
 	adiv5_mem_write(cortexm_ap(t), dest, src, len);
 }
 
--- a/src/target/cortexm.h
+++ b/src/target/cortexm.h
@@ -37,6 +37,9 @@
 #define CORTEXM_DCRDR		(CORTEXM_SCS_BASE + 0xDF8)
 #define CORTEXM_DEMCR		(CORTEXM_SCS_BASE + 0xDFC)
 
+/* Data cache clean and invalidate by address to the PoC=Point of Coherency */
+#define CORTEXM_DCCIMVAC	(CORTEXM_SCS_BASE + 0xF70)
+
 #define CORTEXM_FPB_BASE	(CORTEXM_PPB_BASE + 0x2000)
 
 /* ARM Literature uses FP_*, we use CORTEXM_FPB_* consistently */


src/target/cortexm.c
```diff
@@ -181,11 +181,43 @@ ADIv5_AP_t *cortexm_ap(target *t)
 
 static void cortexm_mem_read(target *t, void *dest, target_addr src, size_t len)
 {
+	/* flush data cache for RAM regions that intersect requested region */
```
gsmcmullin: I'd pull this whole blob out into a `cortexm_cache_clean` function that can be called from read and write to avoid code duplication.

src/target/cortexm.c
```diff
@@ -181,11 +181,43 @@ ADIv5_AP_t *cortexm_ap(target *t)
...
+		target_addr ram = r->start;
+		target_addr ram_end = r->start + r->length;
+		/* RAM region is [ram, ram_end) */
+                if (src > ram)
+			ram = src;
+		if (src_end < ram_end)
+			ram_end = src_end;
+		/* intersection is [ram, ram_end) */
+		for (ram &= ~0x1f; ram < ram_end; ram += 0x20)
+			adiv5_mem_write(cortexm_ap(t), CORTEXM_DCCIMVAC, &ram, 4);
```
gsmcmullin: No need to invalidate on read; this could be `DCCMVAC`.  I'd make the new `cortexm_cache_clean` function take a boolean invalidate parameter.

src/target/cortexm.c
```diff
@@ -181,11 +181,43 @@ ADIv5_AP_t *cortexm_ap(target *t)
...
+	for (struct target_ram *r = t->ram; r; r = r->next) {
+		target_addr ram = r->start;
+		target_addr ram_end = r->start + r->length;
+		/* RAM region is [ram, ram_end) */
+                if (src > ram)
+			ram = src;
+		if (src_end < ram_end)
+			ram_end = src_end;
+		/* intersection is [ram, ram_end) */
+		for (ram &= ~0x1f; ram < ram_end; ram += 0x20)
```
gsmcmullin: Cache line length should be detected, not hard-coded.

nickd4:
Yes good idea, I will do those things, I haven't had time but I will get to it.

One question: Is there an easy way to enumerate devices in a Cortex-M microcontroller? I can't think how we would determine whether the microcontroller has the D-cache. The registers like DCCIMVAC seem to be write only so, I don't think it's easy to tell if they exist. The only way I can see is to laboriously go through the datasheet for all support devices, did I miss something?

My suggestion is I make it configurable on a device level but I then set it to "true" in all devices, others can check the datasheet if they want to optimize the Black Magic Probe's performance on their device.

Also, I honestly don't think it really hurts too much to just do it for all devices, the RAM regions optimization was supposed to make this less painful. It's certainly better to be slower on some devices than to make errors on other devices.

nickd4:
Just rereading the comments I see that the suggestion is to auto detect cache presence and line length.

I think the only way we can do this is to load some code into the micro and execute it, since the debugger can't actually see the cache, it can only indirectly see the effect of the cache via getting the micro to do something. But I think this would be kind of risky because we'd have to overwrite some target RAM or otherwise interfere. Or is there some backdoor way of getting the micro to execute an arbitrary instruction?

I also don't think the line length is going to change in any Cortex M microcontroller, the Cortex M architecture manual is completely clear on which bits of DCCIMVAC and friends are significant.

gsmcmullin:
I haven't had a chance to read over this in enough detail, but the ARMv7-M Architecture Reference Manual (DDI 0403E.b), section B4.8 describes 'Cache Control Identification Registers': 

> The Cache Control Identification Registers are used to determine if a closely-coupled cache is implemented and determine the features of the cache.

If the cache line length is fixed, I haven't seen it in the docs.

nickd4:
I am looking at DDI0489D "ARM Cortex M7 Processor Revision r1p1 Technical Reference Manual", in section 3.2 it shows this:

Address Name Type Reset Description
0xE000ED78 CLIDR RO - i Cache Level ID Register on page 3-9
0xE000ED7C CTR RO 0x8303C003 Cache Type Register
0xE000ED80 CCSIDR RO - j Cache Size ID Register on page 3-10
0xE000ED84 CSSELR RW UNPREDICTABLE Cache Size Selection Register on page 3-12

So I think we should be looking at CLIDR bit 1 as follows:

Bits Name Function
[2:0] - Level 1 cache type:
0b001 Instruction cache is implemented.
0b010 Data cache is implemented.
0b011 Instruction and data cache are implemented.
0b000 No instruction or data cache are implemented.

And also CCSIDR bits [2:0] as follows:

Bits Name Function
[2:0] LineSize Indicates the number of words in each cache line.
0x1 Represents 32 bytes.

Although, since only one value is defined I guess it cannot change.

The only problem here might be if these registers are only defined for Cortex M7 and not other Cortex M.

Strangely in DDI0403D "ARM v7-M Architecture Reference Manual" there is no section B4.8, it just goes from B4.7 "Floating point feature identification registers" to B5.1 "About the ARMv7-M system instructions". So I guess the cache control identification stuff was added in a later revision of the spec (DDI0403E.b which I will try to download when I get time). But this makes me think they may not be reliably present in all Cortex-M microcontrollers. I suppose we could check for Cortex M7 somehow.

cheers, Nick

PR #264, opened 2017-07-29
Add efr32 Blue Gecko family support

dholth:
In progress, still having trouble getting it to work, but it was able to flash given a gdb-driven load & call of stub.


dholth:
By adding a linker script that puts the stub at the start of RAM, it is possible to use a single blackmagicprobe to load and debug the resulting .elf file without requiring the probe to understand how to flash the target.

dholth:
An error here is that the objcopy -O binary should happen against the .elf, otherwise _estack will be loaded as 0.

dholth:
I am now happy with this pull request. It works and it keeps support for both the EFM32 and EFR32 families.

ryankurte:
Hey, this is super cool / glad to find someone has a better grasp on this than I ^_^

Would you be open to including the EFM32(J|P)G(1|12|13)B families also? 
They're the same architecture as the FG and MG, so adding their definitions pretty much just works (in fact, it works as is just without a processor ID / custom commands).
If not, no stress, and I can open a PR when this one is closed.

Also had a [go](https://github.com/ryankurte/blackmagic/blob/feature/efm32-draco-target/src/target/efm32.c#L310) at a different way of representing the cores to try to make it easier to add new ones if that's interesting to you (though my branch is a bit out of date now).

Cheers,

Ryan

dholth:
@ryankurte very glad to find someone else interested in these chips.  I've stumbled quite a bit developing this but eventually got it to flash. It would make sense to support all similar EFM32's here, my list came from the reference manual for the BGM113 I'm playing around with. Your code looks cleaner than the original and you've done a better job than I at recognizing that more than just the MSC has changed between generations. Maybe it is foolish to try to share the flashing functions between the generations.
My code needs to be cleaned up and simplified before it can be accepted into upstream. I'd be delighted to work on this with you in my fork if you'd like to submit PR's.
Do you have a cool EFM32 application or PCB project?

dholth:
Another thing I've noticed is that many important locations in device info are actually the same across generations.

ryankurte:
Yeah, I use them quite a lot. Have some stuff for [building](
https://github.com/ryankurte/efm32-base) for them and a couple of other
public projects.

There's a bunch of commonality (most of the difference seems to be that
they zero-indexed the layout in the new manual), but also a few things like
the MSC that are not. Seems like the more that can be kept common the
better for maintenance.

Haven't had a chance to look at it again yet, but if/when I find the time I
will open a PR against yours ^_^

dholth:
Notably all the chips appear to store the flash page size, flash and ram size in the same places. Makes your table smaller. All you really have to hardcode is which gen it is.

dholth:
Unfortunately the stub has resisted several attempts at simplification. Haven't been able to write a simpler non-crashing version yet.

esden:
Hi @dholth thank you for trying to add support for this. Unfortunately we can not merge it with the size and complexity of your submission. I recommend a complete from scratch rewrite of the stub without the use of any vendor supplied code.

In the form that it is now it would violate the way we like to keep the codebase small and maintainable.

dholth:
Please be more specific. How many lines of code would be small enough? If we ignored the obviously ugly stub, what specifically about the target/efm32.c implementation is objectionable? Could you rank these attributes of the code?

* Use of pointer arithmetic
* Many cases in switch statement
* Extra reformatting
* Strings
* include statement

gsmcmullin:
Superceded by #372

PR #274, opened 2017-09-06
add new platform: Baite ST-Link V2 clone

tsaitgaist:


tsaitgaist:
TLDR: for consistency across ST-Link V2 clones you can switch SWCLK from PB13 to PA5 and SWDIO from PB12 to PB14.

PA5 is connected to PB13 on both the Aluminium ST-Link V2 clone (for which there is already a BMP firmware) and Baite ST-Link V2 clone (supported by this commit). Thus it does not matter which pin is used and you can interchange both. I think at the time I did not check what the settings for the existing ST-Link V2 clone firmware was and just picked a signal connected to the right pin on the connector.
On both adapters there is a 220R in-line resistor on the SWCLK output, but this is just a protection (current limiting) resistor so the dongle can't provide up to 40mA at 3.3V in case a connection is wrong, preventing damaging the hardware you are debugging.
Regarding PB12 and PB14, I picked PB12 because it has an in-line 100R protection resistor while PB14 doesn't. But my schematic/reverse engineering was wrong. There is also a 220R resistor. Thus choosing PB14 just adds protection, but too much protection could weaken the signal.

JojoS62:
thanks for your feedback, I got it.
I can also recommend to remove the 620 R pull up below the LED, then the VCP can be tested with a nullmodem connection between Tx and Rx pins.

JojoS62:
another problem: I've tried the dfu upgrade via the stm32_mem.py script. This fails when I use the binaries created like this:
```
make clean
make
cd src
make PROBE_HOST=baite
```

the dfu update in stlink version works, and it must be created without building the native probe first:
```
make clean
make PROBE_HOST=stlink
```

I think the prblem is in the makefile, I haven't compared everything. Do you have an idea why the build must be done in different ways?

JojoS62:
found something... I've merged your baite branch into the actual master, there the usb driver name has changed:
```
$ git diff
diff --git a/src/platforms/baite/platform.h b/src/platforms/baite/platform.h
index 0200691..e3f5847 100644
--- a/src/platforms/baite/platform.h
+++ b/src/platforms/baite/platform.h
@@ -85,7 +85,7 @@
 #define UART_PIN_SETUP() \
        gpio_set_mode(USBUSART_PORT, GPIO_MODE_OUTPUT_2_MHZ,  \
                      GPIO_CNF_OUTPUT_ALTFN_PUSHPULL, USBUSART_TX_PIN);
-#define USB_DRIVER stm32f103_usb_driver
+#define USB_DRIVER st_usbfs_v1_usb_driver
 #define USB_IRQ    NVIC_USB_LP_CAN_RX0_IRQ
 #define USB_ISR    usb_lp_can_rx0_isr
 /* Interrupt priorities.  Low numbers are high priority.
diff --git a/src/platforms/baite/usbdfu.c b/src/platforms/baite/usbdfu.c
index efd2923..f525ab4 100644
--- a/src/platforms/baite/usbdfu.c
+++ b/src/platforms/baite/usbdfu.c
@@ -68,7 +68,7 @@ int main(void)
        systick_interrupt_enable();
        systick_counter_enable();

-       dfu_init(&stm32f103_usb_driver, DFU_MODE);
+       dfu_init(&st_usbfs_v1_usb_driver, DFU_MODE);

        dfu_main();
 }
```

The dfu update works also, but I have to unplug the baite after calling the update script. After plug in again, the usb update device installs and the I can run the python script with success. There is another diff in the usb_detach() procedure, I will try the different verstion from native/stlink platform.

JojoS62:
Made more progress, the update via stm32_mem.py python scripsts works now, I had to start the app in usbdfu.c / main conditionally:
```
int main(void)
{
	if((GPIOA_CRL & 0x40) == 0x40)
		dfu_jump_app_if_valid();
	dfu_protect(DFU_MODE);
```

tsaitgaist:
Thanks for all the efforts.
- about the USB driver: I've updated the branch to the upstream master and updated the driver name as you mentioned (else compilation would fail)
- I used the opportunity to match the pins to the stlink one (see above)
- some information/explanation about DFU: I only checked if DFU works after flashing the bootloader, which it did since the main application was not flashed and the bootloader would stay in bootloader mode instead of starting the main application since this wasn't valid. The problem is when the application is valid, the DFU bootloader would always jump to it, preventing you to use the bootloader to update the application. To indicate to the bootloader it should not jump to the application, a small trick is used. In src/platforms/common/cdcacm.c dfu_detach_complete() would call platform_request_boot(). This sets GPIOA_CRL to 0x.......8. (enable pull-up on GPIOA1). Then scb_reset_core() is called. This only resets the CM3 core, but not the STM32 peripherals (i.e. leaving the register values intact). When the STM32F103 is fully reset this register defaults to 0x44444444 (see RM0008 reference manual). Thus if the value is not the default one, the bootloader will know not to start the application. This is the check you added. After that the bootloader uses scb_reset_system(), which does a full reset (including the peripherals), allowing the application to start.
- About stm32_mem.py: I wasn't aware of this DFU flashing script. I only flashed the Baite dongle using the serial bootloader since dfu-util (the usual DFU flashing application) could not be used to flash over DFU because of the low advertised flash size (too low for the blackmagic firmware). This size is ignored by stm32_mem.py, which allows to flash over DFU. I mentioned it in the README.

tsaitgaist:
Thanks for all the efforts.
- about the USB driver: I've updated the branch to the upstream master and updated the driver name as you mentioned (else compilation would fail)
- I used the opportunity to match the pins to the stlink one (see above)
- some information/explanation about DFU: I only checked if DFU works after flashing the bootloader, which it did since the main application was not flashed and the bootloader would stay in bootloader mode instead of starting the main application since this wasn't valid. The problem is when the application is valid, the DFU bootloader would always jump to it, preventing you to use the bootloader to update the application. To indicate to the bootloader it should not jump to the application, a small trick is used. In src/platforms/common/cdcacm.c dfu_detach_complete() would call platform_request_boot(). This sets GPIOA_CRL to 0x.......8. (enable pull-up on GPIOA1). Then scb_reset_core() is called. This only resets the CM3 core, but not the STM32 peripherals (i.e. leaving the register values intact). When the STM32F103 is fully reset this register defaults to 0x44444444 (see RM0008 reference manual). Thus if the value is not the default one, the bootloader will know not to start the application. This is the check you added. After that the bootloader uses scb_reset_system(), which does a full reset (including the peripherals), allowing the application to start.
- About stm32_mem.py: I wasn't aware of this DFU flashing script. I only flashed the Baite dongle using the serial bootloader since dfu-util (the usual DFU flashing application) could not be used to flash over DFU because of the low advertised flash size (too low for the blackmagic firmware). This size is ignored by stm32_mem.py, which allows to flash over DFU. I mentioned it in the README.

JojoS62:
Thanks for regarding the changes. Thanks also to @UweBonnes, I found the bootloader trick by comparing your branch to his stlink version. Both could be merged into one version, but it becomes confusing with too much different hardware in one source. I'm happy with this version, updates for new targets should be indendent of the probe hardware and are now easy to install.

UweBonnes:
The patches for a user defined dfu-util size string have not been applied to dfu-util.

JojoS62:
Ooh, got that wrong. But stm32_mem.py works fine.

DieHertz:
Works like charm for me, thank you!

tink3r:
Works perfectly here as well, many thanks for the effort. SWD+USART in a handy little package - beats STlink any day :-) My hardware version is based on the STM32F101 with 128K flash and unoficial USB. No problems so far and especially no issues with the bootloader as opposed to f103 64K versions.

hydra:
what's required to get this merged into master?

UweBonnes:
Well, maintaining a whole new platform is tedious. The name "baite" is also generic. Where is a schematics for that platform? On what port pins are the function pins?  From what I deciphered, "baite" can also be seen as a variant of the stlink platform with minimal changes: https://github.com/UweBonnes/blackmagic/tree/baite. However I don't have a baite platform to test. 
Applying minimal changes to the slink platform to support "baite" have better changces for adoption. Test reports for https://github.com/UweBonnes/blackmagic/tree/baite have been minimal.

hydra:
some related Baite info here:

https://wiki.cuvoodoo.info/doku.php?id=jtag#baite

schematics:

https://wiki.cuvoodoo.info/lib/exe/fetch.php?media=jtag:baite.pdf

UweBonnes:
What about testing and commenting my branch?

DieHertz:
Both this and your branches worked for me last time I checked.

hydra:
I'll try and find sometime to update my probes ASAP.

hydra:
I confirm @DieHertz 's findings. both branches work fine for me, tested only with STM32F405RG and STM32F722RE

I got a brand new Baite V2 board out of it's bag, built it, and flashed it using the SWD port on the Baite V2 and was able to scan for targets, mass-erase the above CPU's, upload code to them, step-debug in Eclipse and also attach to running targets.

I also tested the BMP USB-UART functionality that is very handy on the Baite V2 and that worked fine for connecting to my targets.

hydra:
I also tested the BMP USB-UART functionality that is very handy on the Baite V2 and that worked fine for connecting to my targets on the tsaitgait/baite-platform branch, but *NOT* on the usebonnes/baite branch.

hydra:
I just rebased the tsaitgaist/baite-platform against master today and reflashed to see if it works, it does.

It also detects the H74x now too:

```
GNU gdb (GNU Tools for Arm Embedded Processors 7-2018-q2-update) 8.1.0.20180315-git
Copyright (C) 2018 Free Software Foundation, Inc.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "--host=i686-w64-mingw32 --target=arm-none-eabi".
<http://www.gnu.org/software/gdb/bugs/>.
For help, type "help".
Type "apropos word" to search for commands related to "word".
Remote debugging using COM9
Target voltage: unknown
Available Targets:
No. Att Driver
 1      STM32H74x
```

hydra:
Today I had cause to compile the target with debug enabled, it didn't compile. I fixed it and tested debug mode, changes in this commit here:

https://github.com/hydra/blackmagic/commit/617ef41edb9200b210d1e0449e1c66d95a0bb377

UweBonnes:
The baite is not worth a new platform! It is only a variant of the Stlink platform, it is the same wiring as the white standalone STM Stlink https://www.st.com/content/st_com/en/products/development-tools/hardware-development-tools/hardware-development-tools-for-stm32/st-link.html. Only the version bits need to be detected. I will accept a rework of the stlink platform that recognizes the baite and does no harm to the other platform.

hydra:
@UweBonnes ok understood.  Perhaps @tsaitgaist can have a go at that?  In the mean time this is working code for those that want to use it.

hydra:
@UweBonnes Just doing a bit more testing today with latest code, it seems that I'm unable to restart my target (STM32H750) the way I was doing it before...

I use eclipse gdb integration to do the restart.  
With master of 14bedcc it worked fine - The source view would show me the source to the startup handler assembly code.
With master of dcd20ef it doesn't work - I get a view like this:

```
MY.elf	
	Thread #1 <main> (Suspended : Signal : SIGSEGV:Segmentation fault)	
		0x0	
```

Do you think there's something out of date in the baite patches or is this a different issue with the master code base?  would a debug log help?

UweBonnes:
This looks like a crash in the H7.  If you are sure BMP is the culprit, bisect!

hydra:
Yeah, that's what I thought, but it's not!

The *exact* same binaries I'm using for testing work fine when the probe is re-flashed with the old probe firmware.

Here's a debug log.


```
*** Unsupported packet: vMustReplyEmpty
*** Unsupported packet: Hg0
*** Unsupported packet: qTStatus
TARGETID 04500041
 AP   0: IDR=84770001 CFG=00000000 BASE=e00fe003 CSW=43800040
0xe000e000: Generic IP component - Cortex-M4 SCS (System Control Space)
-> cortexm_probe
0xe0001000: Generic IP component - Cortex-M3 DWT (Data Watchpoint and Trace)
0xe0002000: Generic IP component - Cortex-M7 FBP (Flash Patch and Breakpoint)
0xe0000000: Generic IP component - Cortex-M3 ITM (Instrumentation Trace Module)
0xe0041000: Debug component - Cortex-M7 ETM (Embedded Trace)
0xe0043000: Debug component - CoreSight CTI (Cross Trigger)
 AP   1: IDR=84770001 CFG=00000000 BASE=00000002 CSW=43800040
Fault reading ID registers
 AP   2: IDR=54770002 CFG=00000000 BASE=e00e0003 CSW=80000040
0xe00e3000: 0x0 <- does not match preamble (0xB105000D)
0xe00e4000: 0x0 <- does not match preamble (0xB105000D)
0xe00e5000: PrimeCell peripheral - Unknown (PIDR = 0x4001bb101)
Fault reading ID registers
*** Unsupported packet: qC
*** Unsupported packet: Hg0
*** Unsupported packet: qOffsets
*** Unsupported packet: qSymbol::
*** Unsupported packet: qfThreadInfo
*** Unsupported packet: qL1200000000000000000
m packet: addr = 165c, len = 4
m packet: addr = 165c, len = 2
*** Unsupported packet: qL1200000000000000000
Flash Erase 08000000 00020000
 started cr 00000034 sr 00000005
Flash Write 08000000 00000298
Flash Write 08000298 000003C8
Flash Write 08000660 000003B0
Flash Write 08000A10 000003C0
Flash Write 08000DD0 000003C0
Flash Write 08001190 000003D0
Flash Write 08001560 000003D0
Flash Write 08001930 000003B0
Flash Write 08001CE0 000003D0
Flash Write 080020B0 000003C0
Flash Write 08002470 000003C0
Flash Write 08002830 000003C0
Flash Write 08002BF0 000003C0
Flash Write 08002FB0 000003D0
Flash Write 08003380 000003D0
Flash Write 08003750 000003E0
Flash Write 08003B30 000003D0
Flash Write 08003F00 000003C0
Flash Write 080042C0 000003C0
Flash Write 08004680 000003C0
Flash Write 08004A40 000003D0
Flash Write 08004E10 000003E0
Flash Write 080051F0 000003D0
Flash Write 080055C0 000003D0
Flash Write 08005990 000003C0
Flash Write 08005D50 000003D0
Flash Write 08006120 000003C0
Flash Write 080064E0 000003C0
Flash Write 080068A0 000003C0
Flash Write 08006C60 000003D0
Flash Write 08007030 000003B0
Flash Write 080073E0 000003C0
Flash Write 080077A0 000003B0
Flash Write 08007B50 000003C0
Flash Write 08007F10 000003C0
Flash Write 080082D0 000003D0
Flash Write 080086A0 000003C0
Flash Write 08008A60 000003C0
Flash Write 08008E20 000003D0
Flash Write 080091F0 000003D0
Flash Write 080095C0 000003E0
Flash Write 080099A0 000003D0
Flash Write 08009D70 000003D0
Flash Write 0800A140 000003C0
Flash Write 0800A500 000003D0
Flash Write 0800A8D0 000003E0
Flash Write 0800ACB0 000003D0
Flash Write 0800B080 000003D0
Flash Write 0800B450 000003D0
Flash Write 0800B820 000003E0
Flash Write 0800BC00 000003D0
Flash Write 0800BFD0 000003D0
Flash Write 0800C3A0 000003D0
Flash Write 0800C770 000003D0
Flash Write 0800CB40 000003C0
Flash Write 0800CF00 000003C0
Flash Write 0800D2C0 000003D0
Flash Write 0800D690 000003C0
Flash Write 0800DA50 000003B0
Flash Write 0800DE00 000003D0
Flash Write 0800E1D0 000003D0
Flash Write 0800E5A0 000003C0
Flash Write 0800E960 000003C0
Flash Write 0800ED20 000003C0
Flash Write 0800F0E0 000003D0
Flash Write 0800F4B0 000003D0
Flash Write 0800F880 000003D0
Flash Write 0800FC50 000003D0
Flash Write 08010020 000003D0
Flash Write 080103F0 000003E0
Flash Write 080107D0 000003C0
Flash Write 08010B90 000003B0
Flash Write 08010F40 000003B0
Flash Write 080112F0 00000390
Flash Write 08011680 000003D0
Flash Write 08011A50 000003D0
Flash Write 08011E20 000003D0
Flash Write 080121F0 000003D0
Flash Write 080125C0 000003E0
Flash Write 080129A0 000003E0
Flash Write 08012D80 000003E0
Flash Write 08013160 000003E0
Flash Write 08013540 000003D0
Flash Write 08013910 00000368
Flash Write 08013C78 000003D8
Flash Write 08014050 00000278
Flash Write 080142C8 00000008
Flash Write 080142D0 00000004
Flash Write 080142D4 00000004
Flash Write 080142D8 00000170
Flash Write 08014448 00000240
Flash Write 08018000 000003C0
Flash Write 080183C0 000003D0
Flash Write 08018790 000003D0
Flash Write 08018B60 000003D0
Flash Write 08018F30 00000190
*** Unsupported packet: Pf=d8050008
*** Unsupported packet: qL1200000000000000000
m packet: addr = 80005d8, len = 4
m packet: addr = 80005d8, len = 2
m packet: addr = 80005da, len = 2
*** Unsupported packet: qL1200000000000000000
*** Unsupported packet: qL1200000000000000000
m packet: addr = 20000058, len = 4
m packet: addr = 200001d0, len = 4
m packet: addr = 8018500, len = 40
m packet: addr = 8018620, len = 4
m packet: addr = 8018620, len = 4
m packet: addr = 8018512, len = 2
*** Unsupported packet: vKill;1
*** Unsupported packet: qC
*** Unsupported packet: qAttached
*** Unsupported packet: qOffsets
*** Unsupported packet: Hg0
m packet: addr = 80005d8, len = 4
m packet: addr = 80005d8, len = 2
m packet: addr = 80005da, len = 2
m packet: addr = 8018500, len = 40
m packet: addr = 8018620, len = 4
m packet: addr = 8018620, len = 4
m packet: addr = 8018512, len = 2
m packet: addr = 8018512, len = 2
X packet: addr = 8018512, len = 0
X packet: addr = 8018512, len = 2
*** Unsupported packet: vCont?
*** Unsupported packet: Hc0
```

hydra:
I can bisect if you want, just a bit time consuming.  Is there anything from the debug log above that looks wrong to you?

UweBonnes:
Are you really sure  BMP/baite is the culprit? Can you perhaps test with a platform using master?

hydra:
Yes, sure.  I reflash the probe with new BMP code, unplug replug probe host and target, start same debug configuration and right click the debug session in eclipse and hit 'restart' on the context menu, the memory view in eclipse re-draws and then call-stack is as above.  If I then reflash back to old BMP code, re-plug, restart same debug config and do the same thing and the target restarts and shows me the reset handler source.

UweBonnes:
It also would make more sense if you discuss on gitter and summarize later your findings here.

tsaitgaist:
> The baite is not worth a new platform! It is only a variant of the Stlink platform, it is the same wiring as the white standalone STM Stlink https://www.st.com/content/st_com/en/products/development-tools/hardware-development-tools/hardware-development-tools-for-stm32/st-link.html. Only the version bits need to be detected. I will accept a rework of the stlink platform that recognizes the baite and does no harm to the other platform.

the LED and SWJ pins are the same between the generic stlink and baite platform, but there is an annoying difference: USART. stlink uses USART2, and only TX. baite used USART1 (remapped), with RX (and the pins are exposed to the connector).
the problem is that the USART port is defined in macros (USBUSART in platform.h).
there are two possibilities to overcome that:
- add #ifdefs. then the user would need to add a flag at compile time, which is a bit obscure (and somehow in contradiction with the used concept of folders for platforms), but could be documented in the readme
- don't use macros, but variables in the USART initialisation. then I need to also change the code in the stm32 folder (where it happens).

which do you prefer?

UweBonnes:
The problem withe the name "Baite" is that it is not the name of a unique board. What I know as baite is a board sold as STLINKV2 with a 10 pin 2 row connector. It offers the same functionality as the STM Stlinkv2 with JTAG/SWD and SWIM.  SWIM-RST is connected to PB6 and a 470 R to PB5. SWIM is connected PB6/7/8/9 and a pullup to VCC. As the swlink platform uses USART1 PB6/PB7 you have RX/TX on the 10 pin connector.

So where is the problem, or what board are you talking about?

tsaitgaist:
> The problem withe the name "Baite" is that it is not the name of a unique board. What I know as baite is a board sold as STLINKV2 with a 10 pin 2 row connector. It offers the same functionality as the STM Stlinkv2 with JTAG/SWD and SWIM.

you are right. what I refer as the baite dongle is simply a ST-LINK V2 clone from baite.
here a teardown of the dongle: https://wiki.cuvoodoo.info/doku.php?id=jtag&s[]=jtag#baite
here the schematic: https://wiki.cuvoodoo.info/lib/exe/fetch.php?media=jtag:baite.pdf

the difference (advantage) to other (simpler/cheaper) ST-LINK V2 clones is that it provides JRST, JTDO, and JTDI, thus offering JTAG support.
here a teardown of multiple clones: https://wiki.cuvoodoo.info/doku.php?id=jtag&s[]=jtag#st-link_v2_aluminium
here the schematic of one of then: https://wiki.cuvoodoo.info/lib/exe/fetch.php?media=jtag:alu.pdf

> SWIM-RST is connected to PB6 and a 470 R to PB5. SWIM is connected PB6/7/8/9 and a pullup to VCC.

SWIM is connected to PB7-11. I found this very confusing, but when you look at the STM8S-DISCOVERY schematic, it makes a bit more sense (some signals are just a IN return paths): https://www.st.com/resource/en/user_manual/cd00250600.pdf

Since BMP does not support SWIM, it would be nice to reuse PB6/PB7 as UART (USART1_REMAP), just as the swlink platform.

One issue is that the ST-LINK/V2-1 already provides UART, but on PA2/PA3 (USART2):
https://www.st.com/resource/en/user_manual/DM00105823.pdf

> As the swlink platform uses USART1 PB6/PB7 you have RX/TX on the 10 pin connector.

the "baite" alternative is probably closer to the stlink platform than swlink, but yes, my goal is to re-use USART1 PB6/PB7 to have RX/TX, as the swlink does.

> So where is the problem, or what board are you talking about?

I would like to modify the stlink code to be able to re-use the SWIM/RST pins as UART.
this is currently not possible since the USBUART is tied to UART2 at compile time:
`#define USBUSART USART2` (plus related definition).

adding UART support for stlinkv2 clones on SWIM pins could be done is two way:
- add #ifdefs. the user would need to add a "SWIM_UART" flag at compile time
- use variables in the USART initialisation, which would be set during version detection.

let me know if my goal is still not clear or problematic

hydra:
the discussion regarding re-using UART2 for USBUART is very valid, I would like to see this too.  However the SWD debugging support is initially more important.  Does it make sense to get full support for that first, then tackle the UART2 via USB issue?  Perhaps 2 PR's?  One for SWD support, and another for the UART2 support, just so that we can all move forward a little.  However if a simple solution presents itself quickly then both at the same time may be better for us.

UweBonnes:
Ah, now I see. Swim reuses the JTAG Pins , you want to use the standard GPIO pins as JTAG for a valid reason. So make a proposal for a new platform, document good so that outside people can understand, perhaps discuss with hydra an make a PR. It will take some time for me too comment on that PR than.

tsaitgaist:
I will close this pull request, but 2 new ones will appears, splitting both aspects: add support for the baite as ST-LINK/V2 clone, and reuse the SWIM pins as UART.

Electrobub:
Hello! Sorry to bring up an older ticket.

I am wondering as to what is the best way to get BMP onto a "baite" STLink version. As it seems these pinout definitions never made it into the repo, and the STLink ones do not have the correct UART ones that this PR had.

(I see this PR https://github.com/blackmagic-debug/blackmagic/pull/519 adds a detection for "baite" boards, but has USBUSART pins defined as A2/A3 which are not connected - as see in the schematic in the PR linked)

Is it just best to change the header file for this specific board before building/flashing? Or would it be better to add this support into the repo?

PR #275, opened 2017-09-08
fix *NULL when AP 0 is not present

nsd20463:
On my target adiv5_new_ap(, 0) returns NULL.

With this fix to what appears to be incomplete code I can access the CPU and memory. Without it the BMP locks up due to the referencing of a NULL priv->ahb


gsmcmullin:
Thanks @nsd20463.  Please confirm that PR #276 solves your problem.

PR #276, opened 2017-09-11
cortexa: Remove problematic code for AHB access.

gsmcmullin:
The old code for 'fast' memory accesses using the AHB directly has problems with data consitency.  Until this can be resolved, I'm removing the affected code.

This should also fix #275


nsd20463:
Yes removing the code fixes the bugs in it.

PR #279, opened 2017-09-19
cortexa: Check for fault on set/clear soft breakpoint.

gsmcmullin:


dpiegdon:
works:

(gdb) break *0xc010d87c
Breakpoint 3 at 0xc010d87c
(gdb) c
Continuing.
Warning:
Cannot insert breakpoint 3.
Cannot access memory at address 0xc010d87c

(gdb)

PR #285, opened 2017-09-24
F4 rework

UweBonnes:


gsmcmullin:
Please remove unrelated commit 759d2df.
This looks good, but the changes are significant and I'd like to test on actual hardware before merging.

UweBonnes:
I removed to unrelated change.

I tested with a 1.6 MiByte Hex file on a F429. I could load  at 0x8000000 and also at 0x801000 with some normal program loaded and running at 0x800000 and compare-section succeeded in both cases.
The changes probably still contain some errors but probably less than before. And the remaining are best found by using it.

UweBonnes:
Argh, at test on F746 fails...

UweBonnes:
The failing test was due to an image too large. For me,  gdb stalled silently in that case. I think the patch is ready for inclusion (famous last words :-)

gsmcmullin:
Please fix the printf formatting issues so the CI build passes.  Use the `PRIx32` for `uint32_t` formatting.

UweBonnes:
I removed noisy debug output kept the debug output on unexpected error paths. Formatting is fixed.

PR #286, opened 2017-09-24
stm32h7: Start of support.

UweBonnes:
Implement probe, memory map, erase, write, uid, crc.
Flashstub only written after erase.


UweBonnes:
Please generate stm32h7.stub before commiting!

UweBonnes:
Please do not yet apply. Problems with last minute changes...

src/target/stm32h7.c
```diff
@@ -0,0 +1,463 @@
...
+}
+
+static int stm32h7_flash_write(struct target_flash *f, target_addr dest,
+                               const void *src, size_t len)
+{
+	int res;
+	target *t = f->t;
+	struct stm32h7_flash *h7 = (struct stm32h7_flash *)f;
+	DEBUG("fw dest %08lx len %04x, ", dest, len);
+	if (target_mem_read32(t, SRAM_BASE) == 0) {
```
gsmcmullin: This seems absurd. Why would we not load the stub here? On first entry here, we have no knowledge of what's expected in the SRAM.

UweBonnes:
The flash stub is loaded for every call to xxx_flash_write(). The idea was to load it only when needed. The patch tried to set a marker that load is nescessary in flash_erase and only load it when needed.

esden:
I see you rebased your code, is it now fixed? Also why do we need the uid and crc monitor commands?

UweBonnes:
I can flash some sample code in the H7. Probably this does not test corner cases. As probably nobody will test this code as long as it is a pull request, commit should be considered to get broader test coverage.

On some occations when stepping in the H7 code, it seemed that breakpoints could only be set on the second attempt. This is not related to the flash code. Maybe this is related to #258.

Does the crc/uid code eat that much flash? Otherwise there is no real need for those commands, but they may come handy.

UweBonnes:
I think the patch is in a much better condition now. Please review, especially the behaviour of the large direct write to flash.
The reference manual tells that write memory access stalls when the write queue is full.  Does target_mem_write also stall in that situation? Or does it not stall, but transferring 32 bytes via SWD takes so long that the word (256 bit + 10 ECC bits) write time (typ 100 us, max 200 us) is met and so things work with an Stlink as BMP?

gsmcmullin:
`target_mem_write` will spin for up to 2 seconds until the write is complete or a fault is generated.  If the 2 second limit times out, a timeout exception is raised.  The low-level implementation is here: https://github.com/blacksphere/blackmagic/blob/master/src/target/adiv5_swdp.c#L139-L146

I don't know the bit timings but they're not important.

UweBonnes:
Okay, I didn't present the problem in an understandable way. There is no timeout. Write on the H7 could be 160 kByte/sec if we present data fast enough, at least from the datasheet. With this patch I reach ~35 kByte/sec.

At present I present the data to the flash write engine as fast as BMP can deliver. This works, with bitbanging SWD probably limiting the bandwidth. If data is delivered fast enough, the flash write engine would limit rate by stalling the AXI bus. My question was: Does a stall on the AXI bus also stall target_mem_write(). If yes, the patch would also work if bandwidth would not be limited by  the bitbanging SWD. I not, the patch should perhaps query Busy after each flash word (256 bits) written to be on the save side.

gsmcmullin:
So long as the bus is stalled for less than 2 seconds on each write `target_mem_write` will succeed, so your patch is ok as is.

EliasOenal:
I've successfully tested this with the STM32H743I-EVAL and would also suggest merging it.

korken89:
How is this merge coming a long? I have a H7 board that soon will land, and I'd like to use my BMPs with it :)

EliasOenal:
You can try my branch here: https://github.com/EliasOenal/blackmagic/tree/h7_update
I still believe this should be merged, no idea why it hasn't been done yet.

korken89:
Thanks, sadly I have put the H7 project on ice for now. Hope the support is added though.

PR #287, opened 2017-09-24
Update libopencm3

gsmcmullin:


UweBonnes:
The errors I saw not really happened consistant. However they happened before I applied the patch and happened not more after applying the patch at least on a first test. So please pull...

esden:
Did some minor testing and it all seems to be working fine. :) Good to merge I think. :+1:

PR #291, opened 2017-10-02
Single register access - Target methods and GDB server support

gsmcmullin:
- Add new target methods to access single registers
- Add GDB packets support for `p` and `P` packets
- Implement new methods in Cortex-M and Cortex-A drivers

### DIFF
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -207,6 +207,36 @@ int gdb_main_loop(struct target_controller *tc, bool in_syscall)
 			}
 			break;
 			}
+
+		/* Optional GDB packet support */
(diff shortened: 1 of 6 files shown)


UweBonnes:
Where is this access used?

gsmcmullin:
It's used when GDB sends us one of these packets.  It's not needed for the ARM targets, but is needed by upcoming RISC-V support where GDB queries high numbered registers that aren't present in the `g` and `G` packets.  I might argue that what's being done in RISC-V GDB is bad, but they're unlikely to care.

UweBonnes:
Perhaps helpful for those thinking about RiscV support

PR #299, opened 2017-10-09
stm32f1: Use buffered model to allow unaligned write.

UweBonnes:


src/target/stm32f1.c
```diff
@@ -108,7 +108,10 @@ static void stm32f1_add_flash(target *t,
 	f->length = length;
 	f->blocksize = erasesize;
 	f->erase = stm32f1_flash_erase;
-	f->write = stm32f1_flash_write;
+	f->write = target_flash_write_buffered;
```
gsmcmullin: Please add a comment explaining why this is needed with reference to the STM32F3 Reference Manual.  Also remove the `align` field, as it's not important when using the buffered model.

src/target/stm32f4.c
```diff
@@ -160,7 +160,10 @@ static void stm32f4_add_flash(target *t,
 	f->length = length;
 	f->blocksize = blocksize;
 	f->erase = stm32f4_flash_erase;
-	f->write = stm32f4_flash_write;
+	f->write = target_flash_write_buffered;
```
gsmcmullin: Reference manuals for F2, F4 and F7 all seem to explicitly indicate you can do repeated write cycles so long as you only set 1s to 0s.  Unless something here is actually broken I see no need to change it.

src/target/stm32l0.c
```diff
@@ -241,6 +241,7 @@ static void stm32l_add_flash(target *t,
 	f->done = target_flash_done_buffered;
 	f->write_buf = stm32lx_nvm_prog_write;
 	f->buf_size = erasesize/2;
+	f->align = 4;
```
gsmcmullin: `align` isn't needed with the buffered model.  It always writes a while sector.

src/target/stm32l4.c
```diff
@@ -139,6 +139,7 @@ static void stm32l4_add_flash(target *t,
 	f->done = target_flash_done_buffered;
 	f->write_buf = stm32l4_flash_write;
 	f->buf_size = 2048;
+	f->align = 8;
```
gsmcmullin: `align` isn't needed with the buffered model.  It always writes a while sector.

UweBonnes:
Why do you prefer unbuffered write? At least it is faster:
===git head===
(gdb) load
Loading section .isr_vector, size 0x190 lma 0x8000000
Loading section .text, size 0x3c540 lma 0x8000190
Loading section .ARM, size 0x8 lma 0x803c6d0
Loading section .pg_registry, size 0x460 lma 0x803c6d8
Loading section .pg_resetdata, size 0xed lma 0x803cb38
Loading section .data, size 0x1ec0 lma 0x803cc25
Start address 0x80327a0, load size 256741
Transfer rate: 17 KB/sec, 950 bytes/write.

==stm32f3==
(gdb) load
Loading section .isr_vector, size 0x190 lma 0x8000000
Loading section .text, size 0x3c540 lma 0x8000190
Loading section .ARM, size 0x8 lma 0x803c6d0
Loading section .pg_registry, size 0x460 lma 0x803c6d8
Loading section .pg_resetdata, size 0xed lma 0x803cb38
Loading section .data, size 0x1ec0 lma 0x803cc25
Start address 0x80327a0, load size 256741
Transfer rate: 19 KB/sec, 950 bytes/write.

gsmcmullin:
I don't necessarily prefer unbuffered writes, but in general I prefer less abstraction layers.  I also like to have reasons for bringing in changes.

Maybe buffered writes are better everywhere and that should all be handled in the target layer, but that is a separate discussion that has nothing to do with fixing unaligned writes on STM32F3.

gsmcmullin:
Setting `align` > 1 with the buffered model is not just unnecessary, it's incorrect.  `target_flash_write_buffered` uses `memcpy` to copy the padded buffer with the padding bytes set to their erased value, clobbering whatever was there already.

UweBonnes:
Gareth, "Successive write operations are possible without the need of an erase operation when changing bits from ‘1’ to ‘0’." is only possible for F4 family.

PM0075 Rev 2.1 for Stm32F1 page 14/31 reads:
In this mode the CPU programs the main Flash memory by performing standard half-word
write operations. The PG bit in the FLASH_CR register must be set. FPEC preliminarily
reads the value at the addressed main Flash memory location and checks that it has been
erased. If not, the program operation is skipped and a warning is issued by the PGERR bit in
FLASH_SR register (the only exception to this is when 0x0000 is programmed. In this case,
 the location is correctly programmed to 0x0000 and the PGERR bit is not set). 

So using unbuffered write will fail to write bytes at odd addresses when the even address is alread programmend.

I removed the buffered write for the stm32f4.c. The issue buffered vs unbuffered writes should be kept alive...

The last  patch in this series removes alignment for buffered writes. It is nowhere documented that the natural align value causes bad things to happen with buffered writes. It caused some head scratching when I debugged F3. I think it is cleaner to remove that bad consequence than to document the requirements to leave align unset for buffered writes.

B.t.w. the last patch was already there when you commented "Setting align > 1 with the buffered model is not just unnecessary, it's incorrect. "

PR #300, opened 2017-10-11
Cache support for Cortex-M7

gsmcmullin:
- On probe, read CTR for cache presence and minimum line length
- Make D-Cache clean a function
- Clean before memory reads
- Clean and invalidate before memory writes
- Flush all I-Cache before resume

This supersedes #258


nickd4:
This looks great Gareth. I apologize that I hadn't got around to making the requested changes in earlier pull request, but it seems like you have fixed the problem much more comprehensively than I did. cheers

PR #302, opened 2017-10-17
[RFC] Only support buffered flash writes

gsmcmullin:
It has been suggested that buffered writes are better in all cases, for performance and to solve potential alignment problems.

This follows from discussion of #299


ChuckM:
👍 looks good to me.

UweBonnes:
I think, buffered writes are best in most situation. But we should not drop infrastructure for non-buffered writes

UweBonnes:
Thinking again, I think it is okay to drop infrastructure for non-buffered writes.

Probably the f->erased values is also no more needed and should be removed from target_internal.h.

UweBonnes:
Another missthinking. Argh. f->erased is needed for setting up the buffer.

PR #311, opened 2017-12-06
stm32f1.c: Export CCM RAM of F303.

UweBonnes:
This allows download to CCM Ram.

### DIFF
--- a/src/target/stm32f1.c
+++ b/src/target/stm32f1.c
@@ -136,11 +136,12 @@ bool stm32f1_probe(target *t)
 		stm32f1_add_flash(t, 0x8000000, 0x80000, 0x800);
 		target_add_commands(t, stm32f1_cmd_list, "STM32 HD/CL");
 		return true;
-	case 0x422:  /* STM32F30x */
-	case 0x432:  /* STM32F37x */
 	case 0x438:  /* STM32F303x6/8 and STM32F328 */
-	case 0x439:  /* STM32F302C8 */
+	case 0x422:  /* STM32F30x */
 	case 0x446:  /* STM32F303xD/E and STM32F398xE */
+		target_add_ram(t, 0x10000000, 0x4000);
+	case 0x432:  /* STM32F37x */
+	case 0x439:  /* STM32F302C8 */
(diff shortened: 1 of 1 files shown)


src/target/stm32f1.c
```diff
@@ -142,8 +149,10 @@ bool stm32f1_probe(target *t)
 	case 0x439:  /* STM32F302C8 */
 	case 0x446:  /* STM32F303xD/E and STM32F398xE */
 		t->driver = "STM32F3";
+		target_add_ram(t, 0x10000000, 0x4000);
 		target_add_ram(t, 0x20000000, 0x10000);
-		stm32f1_add_flash(t, 0x8000000, 0x80000, 0x800);
+		flash_size= (target_mem_read32(t, FLASHSIZE) & 0xffff) * 0x400;
```
gsmcmullin: Why this change?
- It's not declared in the PR description or the commit message.
- It appears to be inconsistent with your newly added comment above, that also has nothing to do with CCM RAM.
- Put spaces on both sides of `=`.
- `FLASHSIZE` is 0x1FFFF7E0, but the register is at 0x1FFFF7CC, is it not?  Did you test this?

UweBonnes:
I wanted to use a flash like linker script to upload and run totally in RAM. The setup should get things right after a reset, so initialized data must be loaded to one region by the debugger and copied  by thecontroller on startup. CCM ram as a different memory region is needed for that task. It was not exported before, so download was refused by the debugger

When doing so, I revisited to probe(), but after some fiddling I saw that idcodes and devices are mixed up like herb and beet. I forgot to revoke these changes, perhaps with the hope that new new setup was "clearer".

In the spirit of " one patch for one task" the patch now only adds CCM ram.

B.t.w, do you think that "run from RAM, resistant to reset and without code change" is possible with only one RAM region?

gsmcmullin:
I don't see why you would need more than one ram region.  You will just need enough ram that you can afford to have 2 copies of your `.data` section, they can be located in the same physical region of memory.

UweBonnes:
As I understand the gnu linker, having two copies of a section is not possible.

PR #315, opened 2018-01-15
Another nRF52 device id.

markrages:
This is for the BGA-packaged device.


ntavish:
Also seeing this new device ID. CI build seems to have failed because of some issues with apt-get, so that can be ignored.

PR #324, opened 2018-03-23
Fixes fall through GCC V7 errors.

esden:
Note: This PR is specifically for the V1.6 branch of the code.

GCC V7 considers fallthrough that is not marked an error. It also
expects the markers to be formatted a specific way. This patch fixes
those issues. For more information refer to the -Wimplicit-fallthrough
redhat article:
https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/

### DIFF
--- a/src/platforms/common/cdcacm.c
+++ b/src/platforms/common/cdcacm.c
@@ -441,6 +441,7 @@ static int cdcacm_control_request(usbd_device *dev,
 		switch(req->wIndex) {
 		case 2:
 			usbuart_set_line_coding((struct usb_cdc_line_coding*)*buf);
+			/* Falls through. */
 		case 0:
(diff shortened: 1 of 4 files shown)


src/platforms/common/cdcacm.c
```diff
@@ -458,6 +459,7 @@ static int cdcacm_control_request(usbd_device *dev,
 
 			return 1;
 		}
+		/* Falls through. */
```
gsmcmullin: This fall through looks like a bug to me.  We put a `return 0` here on master.

UweBonnes:
Piotr, please have a look, if this still is needed. If needed, please correct.

UweBonnes:
Fixed by  #313

PR #325, opened 2018-03-24
Update to split 'special' register into its sane parts

korken89:
Hi, I took the changes @mubes did and expanded a bit, to make it a formal PR.
This will split the `special` register into its sane parts `BASEPRI`, `PRIMASK`, `FAULTMASK`, `CONTROL`.

BR Emil


gsmcmullin:
I've pulled this in, but renamed the registers to all lowercase for consistency.

PR #329, opened 2018-03-28
Fix flashing hanging or not working

UweBonnes:
If construction of the memory map needs to read target CPU register, the memory map was invalid on first attach after power on or always when connect_srst was enabled. This resulted in no real write to flash while the load command obvious succeeded, also at much higher speed. The load commands hanged on another board where nRST had a slow rise after SRST de-asserted and flashing started with the MCU still in reset.

### DIFF
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -401,7 +401,10 @@ void cortexm_detach(target *t)
 	/* Clear any stale watchpoints */
 	for(i = 0; i < priv->hw_watchpoint_max; i++)
 		target_mem_write32(t, CORTEXM_DWT_FUNC(i), 0);
-
+	/* Release exception catch .*/
+	target_mem_write32(t, CORTEXM_DEMCR, 0);
+	/* Continue running target where stopped.*/
+	cortexm_halt_resume(t, 0);
 	/* Disable debug */
 	target_mem_write32(t, CORTEXM_DHCSR, CORTEXM_DHCSR_DBGKEY);
 }
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -296,6 +296,16 @@ bool cortexm_probe(ADIv5_AP_t *ap)
 		target_check_error(t);
 	}
 
+	bool connect_srst = platform_srst_get_val();
+	/* Enable debug access for probing.*/
+	target_halt_request(t);
+	if (connect_srst) {
+		/* Connect_srst is active. Request halt on reset to inhibit
+		   start of the target CPU when releasing nRST.*/
+		target_mem_write32(t, CORTEXM_DEMCR, priv->demcr);
+	}
+	/* Always wait until nRST is released.*/
+	platform_srst_set_val(false);
 #define PROBE(x) \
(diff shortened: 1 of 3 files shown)


src/target/cortexm.c
```diff
@@ -296,6 +296,16 @@ bool cortexm_probe(ADIv5_AP_t *ap)
 		target_check_error(t);
 	}
 
+	bool connect_srst = platform_srst_get_val();
+	/* Enable debug access for probing.*/
+	target_halt_request(t);
```
gsmcmullin: Why do you halt a running target on probe?  This probe function should be as unintrusive as possible.
Halting debug will need to be enabled for the DEMCR to take effect, but this should be done inside the if statement below:
```
target_mem_write32(t, CORTEXM_DEMCR, CORTEXM_DHCSR_DBGKEY | CORTEXM_DHCSR_C_DEBUGEN);
```
There is no need to set `DHCSR_C_HALT`.

src/target/cortexm.c
```diff
@@ -386,7 +401,10 @@ void cortexm_detach(target *t)
 	/* Clear any stale watchpoints */
 	for(i = 0; i < priv->hw_watchpoint_max; i++)
 		target_mem_write32(t, CORTEXM_DWT_FUNC(i), 0);
-
+	/* Release exception catch .*/
+	target_mem_write32(t, CORTEXM_DEMCR, 0);
```
gsmcmullin: I don't know why you added this.  These DEMCR flags have no effect when halting debug is disabled.
Target execution would be resumed by the write to DHCSR below; there is no need to call `cortexm_halt_resume` explicitly.

UweBonnes:
For the construction of the memory map at least for some STM32 devices, STM32 internals registers need to be evaluated, e.g.for the bank split. This does no work in many cases with a device mostly in the idle thread,  waiting for interrupt. Above patch is a work-around for that situation.
A better solution should split "probe" and construction of the memory map in to separate items. Probing should be done as none-intrusive as possible and only later with attach the memory maps can be constructed with the device already halted.

However splitting probe and construction of the memory map is probably very intrusive to the code. Help and ideas needed!

gsmcmullin:
Suggestion to build memory map from attach function: https://github.com/blacksphere/blackmagic/compare/master...gsmcmullin:stm32_split_probe_attach

UweBonnes:
I will have a look at your proposal the next days...

UweBonnes:
Gareth, please disregard this pull request. Please have a look at https://github.com/UweBonnes/blackmagic/tree/stm32_split_probe_attach that mostly reuses your https://github.com/blacksphere/blackmagic/compare/master...gsmcmullin:stm32_split_probe_attach. One problem was on the stlnk platform running silently out of SRAM. I reduced the SWO trace buffers. Stm32L4 also needs to delay memory map setup to determine the split on some devices. F0 should also not read device  register during probe and so uses fixed flash sizes. Some L4 problems are also fixed in the patch set. Please review and possibly apply!

gsmcmullin:
Superseded by #334

PR #330, opened 2018-03-30
add Dockerfile and build.sh script

zoobab:
add Dockerfile and build.sh script, this builds and copies the artifacts *.bin to the root directory.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+FROM debian:jessie-slim
+MAINTAINER Benjamin Henrion <zoobab@gmail.com>
+LABEL Description="Firmware for the Black Magic Probe (BMP)"
+
+RUN DEBIAN_FRONTEND=noninteractive apt-get update -y -q && apt-get install -y -q sudo make python gcc-arm-none-eabi git-core libnewlib-arm-none-eabi
+
+ENV user blackmagic
+RUN useradd -d /home/$user -m -s /bin/bash $user
+RUN echo "$user ALL=(ALL) NOPASSWD:ALL" > /etc/sudoers.d/$user
+RUN chmod 0440 /etc/sudoers.d/$user
+
+USER $user
+WORKDIR /home/$user
+RUN mkdir -pv code
+COPY . ./code/
+RUN sudo chown $user.$user -R /home/$user/code
+WORKDIR /home/$user/code/
+RUN make all
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+set -e
+
+check_command() {
+PROGRAM=$1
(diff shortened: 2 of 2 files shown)


zoobab:
Any update?

gsmcmullin:
I don't really understand the point of this.
@UweBonnes This may be useful for you to get libftdi1 in CI by doing the build in Docker in Travis.

UweBonnes:
@zoobab : Can above dockerfile use a recent libftdi1? Can you help me to use libftdi1?

zoobab:
With some mods, it can use another libftdi1. Which one do you want?

UweBonnes:
A recent one, allowing asynchronous transfers.

UweBonnes:
The rationale of this PR is not clear. We have CI and a  place with daily builds. Closing for now.

zoobab:
"We have CI and a place with daily builds."

It does not work on my laptop.

And 3 years for a pull request?

UweBonnes:
CI runs on github. "make all_platforms" runs on the local CPU and fills ./artifacts. What are you missing?

sidprice:
What is missing is the environment that allows "make all_platforms" to actually do the build. The user must have all the tools installed and if on Windows that is not an insignificant task.

UweBonnes:
Please try with recent git head. HOSTED_BMP_ONLY=1 is now default for the HOSTED build and should succeed on windows too

sidprice:
What about building the BMP firmware and LOCM3? Do those "just work" on Windows?

A Docker solution would be a good step forward removing one more "barrier to entry" for others.

UweBonnes:
Okay, now there is a better rationale for this PR. Allow local docker build, while local system may have problems to  build. I would accept a PR with two constraints: - Should work on  recent git head - Pr creator should give some promise to keep things updated.

It would have helped, if more info and explication would have been given earlier. Do not expect the other side to understand condensed facts.

Regarding Sids windows argument. Please tell if hosted HOSTED_BMP_ONLY build on windows or  at least report error or better provide a PR if few things need change or discuss in advance if larger changes are neede. That way, with a recent downloaded firmware and hosted all target development can done. It this really a big hurdle?

sidprice:
@UweBonnes I just took a look at the docker file and I am not sure it does what I was expecting. 

My understanding of such files is that they encapsulate a build environment that is isolated from the host computer, and possibly in a cross platform manner. @zoobab does your file do this? A quick glance at it says probably not :(

UweBonnes:
Sid, can you verify my reasoning about current hosted being much more easier to build on windows now?

sidprice:
@UweBonnes Unfortunately, I just got a new laptop and I do not have GCC installed, only GCC ARM. I will install it and test later, in a day or two.

PR #332, opened 2018-04-20
Write access size

UweBonnes:
While a flash write is running, debug access is stalled for all STM32.  Direct write to F0/F1/F3 needs halfword writes. This series allows debug writes with byte/halfword/word size so all STM32 types can be written without a flashstub.


UweBonnes:
Close PR to allow to trigger a new travis build.

UweBonnes:
Reopen to trigger new travis build

gsmcmullin:
Getting rid of the flash stubs is a great. Thanks for the effort @UweBonnes.

I didn't like this because the new `mem_write_access` affects target behaviour outside of the flash routines where it these changes are relevant.  I much prefer the alternative branch https://github.com/blacksphere/blackmagic/compare/master...UweBonnes:write_access_size1
in this regard.  Referring to that change set, I don't think the new indirection of `target_mem_write_sized` is needed, I'd just have the flash driver call `adiv5_mem_write_sized` directly.  Other than that, it looks good to me.

I'm away for the next few weeks and have no hardware to test.  @esden Please test this if you have a chance.  I'm happy to merge to write_access_size1 branch with the change suggested above.

esden:
I have merged https://github.com/blacksphere/blackmagic/compare/master...UweBonnes:write_access_size1 into my working branch and used it to flash a batch of Black Magic Probes using the improved flashing routine. It definitely increases the flash speed from ~11kb/s to solid ~16kb/s which is noticable speed improvement.

I have not reviewed the code but I trust @gsmcmullin assesment regarding code quality and structure.

It is an OK to merge from me. :)

UweBonnes:
Superseded by #357

PR #336, opened 2018-04-22
stm32_mem.py: Guard dfu.dfu_device() with try/except.

gsmcmullin:
From @UweBonnes


gsmcmullin:
This sneaked in with #373

PR #337, opened 2018-04-24
Update maximum RAM sizes for F4 and F7 devices

adamgreig:
In commit c4d3712 the `target_add_ram` call for STM32F4 devices was changed from a size of 0x40000 to 0x10000, which means gdb by default will refuse to read variables that live in RAM beyond the first 64kB. This also impacts semihosting when the memory to read lives beyond the first 64kB. Setting `mem inaccessible-by-default off` in gdb resolves the issue, but is a workaround rather than a solution.

This PR updates the calls to use the maximum RAM sizes for the F4 and F7 devices.

I'm not clear if `target_add_ram` should give the _minimum_ or _maximum_ RAM sizes; specifically does this change how much is attempted to be written at once when writing flash? If so I guess these calls should somehow work out precisely how much RAM is available rather than the family min/max. This updated version doesn't seem to have problems programming flash even when using devices with less than the maximum RAM, though.

### DIFF
--- a/src/target/stm32f4.c
+++ b/src/target/stm32f4.c
@@ -253,7 +253,8 @@ bool stm32f4_probe(target *t)
 	uint32_t flashsize = target_mem_read32(t, flashsize_base) & 0xffff;
 	if (is_f7) {
 		target_add_ram(t, 0x00000000, 0x4000);  /* 16 k ITCM Ram */
-		target_add_ram(t, 0x20000000, 0x10000); /* 64 k DTCM Ram */
+		target_add_ram(t, 0x20000000, 0x20000); /* 128 k DTCM Ram */
+		target_add_ram(t, 0x20020000, 0x60000); /* 384 k Ram */
 		if (dual_bank) {
 			uint32_t optcr;
 			optcr = target_mem_read32(t, FLASH_OPTCR);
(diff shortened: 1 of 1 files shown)


UweBonnes:
What exact device do you have and what does "info mem" tell? commit c4d3712 tries to set the right ram sizes for the devices

adamgreig:
On an STM32F407ZE, with current master I get:
```
No. Att Driver
 1      STM32F40x
(gdb) att 1
Attaching to Remote target
(gdb) info mem
Using memory regions provided by the target.
Num Enb Low Addr   High Addr  Attrs 
0   y  	0x08000000 0x08010000 flash blocksize 0x4000 nocache 
1   y  	0x08010000 0x08020000 flash blocksize 0x10000 nocache 
2   y  	0x08020000 0x08080000 flash blocksize 0x20000 nocache 
3   y  	0x10000000 0x10010000 rw nocache 
4   y  	0x20000000 0x20010000 rw nocache 
```

the 0x10000 size for the RAM is incorrect: this part has 128K here (0x20000). If I use the version patched as per this PR, I instead get:
```
Num Enb Low Addr   High Addr  Attrs 
0   y  	0x08000000 0x08010000 flash blocksize 0x4000 nocache 
1   y  	0x08010000 0x08020000 flash blocksize 0x10000 nocache 
2   y  	0x08020000 0x08080000 flash blocksize 0x20000 nocache 
3   y  	0x10000000 0x10010000 rw nocache 
4   y  	0x20000000 0x20040000 rw nocache 
```

That's also wrong but follows what was set in this PR. So it looks like the memory map is coming from the `stm32f4.c` file...

UweBonnes:
You are right, the values I used in  c4d3712 are to small. It is bad to announce to few ram. In most cases it does not hurt to announce more ram than actual available. So your patch looks fine. I may however collide with the pending probe/memory map patch.

adamgreig:
As far as I can see it shouldn't conflict with #334 but I haven't tested them together. Would you like me to try it out or will you do it?

UweBonnes:
Eventually rebase when Gareth applies #334 first. But maybe Gareth applies your patch first...

gsmcmullin:
No conflicts, so I'm merging this.
The sizes in the memory map should be the max.
The value has no effect on flash programming.

PR #340, opened 2018-05-01
cortexm: On detach, clear Debug Exception and Monitor Control Register.

### LINKED ISSUE #339: Reset non-functional after detach
At least on STM32, after detaching a debug session, the device is unresponsiv to the reset buttom.  Clearing the Debug Exception and Monitor Control Register on reset made the reset buttom work.

UweBonnes:
Without clearing, reset (e.g. by the reset buttom) is ignored after detach
at least on STM32. Fixes #339.

### DIFF
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -379,6 +379,9 @@ void cortexm_detach(target *t)
 	struct cortexm_priv *priv = t->priv;
 	unsigned i;
 
+	/* Clear halt on reset */
+	target_mem_write32(t, CORTEXM_DEMCR, 0);
+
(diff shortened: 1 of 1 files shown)


gsmcmullin:
This also works fine for me.

Detach sets DHCSR.C_DEBUGEN to 0, and ccording to ARMv7-M ARM (ARM doc DDI 0403E), for all of the vector catch bits in DEMCR:
> If DHCSR.C_DEBUGEN is set to 0, the processor ignores the value of this bit.

UweBonnes:
The issue seems to be specific to local changes here.

PR #341, opened 2018-05-02
cortexm: Resume after detach.

UweBonnes:
After detach the program hangs, at least on STM32.

### DIFF
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -389,6 +389,9 @@ void cortexm_detach(target *t)
 
 	/* Disable debug */
 	target_mem_write32(t, CORTEXM_DHCSR, CORTEXM_DHCSR_DBGKEY);
+
+	/* Restart target. */
+	target_halt_resume(t, 0);
 }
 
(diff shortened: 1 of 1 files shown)


src/target/cortexm.c
```diff
@@ -389,6 +389,9 @@ void cortexm_detach(target *t)
 
 	/* Disable debug */
 	target_mem_write32(t, CORTEXM_DHCSR, CORTEXM_DHCSR_DBGKEY);
```
gsmcmullin: This clears the C_HALT bit, which resumes target execution.  It also clears C_DEBUGEN, following which writes to C_HALT (as done by `target_halt_resume`) are ignored.

From ARMv7-M ARM (ARM doc DDI 0403E):
> When DHCSR.C_DEBUGEN is 0, the processor ignores the values of DHCSR.C_HALT, DHCSR.C_STEP and
> DHCSR.C_MASKINTS, and these values are UNKNOWN on DHCSR reads.

I can't reproduce #342 on native BMP v2.1 with the current master.

UweBonnes:
The issue seems to be specific to local changes here

PR #354, opened 2018-06-03
Blue Pill as separate platform

fb39ca4:
I've modified the STLink platform make it use the alread-soldered SWD pins and the built-in LED. Also, since it is necessary to have a non-USB bootloader method of programming a new Blue Pill anyways, I had it build to be flashed directly rather than going through the DFU bootloader.


gsmcmullin:
Thank you for your contribution.  We have decided that we will not be merging new platforms as they have become a maintenance headache.

joric:
Why it's tagged "$3 board support" when it's always $1.80 with a free shipping? :) I bought it there https://www.aliexpress.com/item//32583160323.html

veerendra-nath:
i think for blue pill there is no use of detect revision block

UweBonnes:
Gareth (or Amar), please close

PR #356, opened 2018-06-13
Probe halted

UweBonnes:
This RP has now single issue commits with debug output removed. It  also allows the user to configure when to release SRST when asserted and how long to wait for the device halted is requested. This fixes issues with STML0 and F7 devices when WFI is used.

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -43,6 +43,12 @@ struct command_s {
 	const char *help;
 };
 
+enum assert_srst_t{
+	ASSERT_NEVER = 0,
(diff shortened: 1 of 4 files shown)


esden:
I have tested out this branch on multiple stm32 platforms that I had available. I did not run through all the eval boards that I have available.

I also did put some code using WFI onto one of the boards and the code seems to be able to connect to the device as advertised. I think it is OK to merge.

Definitely more broad testing on multiple platforms is needed. (I will be using this code on my probe from now on to see if I encounter some issues)

Issue #320, opened 2018-02-26
NRF5 stuck in reset until scan is started


UweBonnes:
Hello,

in https://github.com/sandeepmistry/arduino-nRF5/issues/252some user reports that the NRF5 does not start until a scan command is issued. As described there, SDWIO and Reset share the same line. Startup code of  the different platforms sets SWDIO low when BMP initializes the board, obvious causing this issue.

https://github.com/UweBonnes/blackmagic/tree/pin-defaults cares for the right  SDWIO and TDI defaults. 

Awaiting user reports...

probonopd:
Just a heads-up, I am currently away from the devices with no way to test. Will follow your suggestions and give an update, most likely next week. Thanks for your help.

probonopd:
Thanks @UweBonnes but after power cycling, SDWIO soldered to PB14 still is low.

rpavlik:
FYI, I have an adafruit nRF52 feather and will presumably be using the BMP v2.1 on it shortly.

flatsiedatsie:
This sounds like this problem:

https://forum.mysensors.org/topic/6961/nrf5-bluetooth-action/1583

> nrf52.cpu curstate
reset

UweBonnes:
Well, BMP should keep SWDIO tristated until needed. While tristated, pull-up internal to target SWDIO and maybe external will do the right thing.

UweBonnes:
Goggle slogans you don't understand first. Than ask more exact.

UweBonnes:
People with this problem please try #361. This pull request keeps TMS tristated until th scan happens. Please report!

joric:
Related problem. I use the following script to flash new firmware:

`arm-none-eabi-gdb.exe --quiet --batch -ex "target extended-remote \\.\%port%" -ex "mon swdp_scan" -ex "file %file:\=/%" -ex "att 1" %opt% -ex "load" -ex "run"`

Ctrl+C works fine (console closes, MCU keeps running), but -ex "quit" or -ex "reset" somehow hangs the MCU. How to fix? I want a script that flashes fimware and restarts MCU on exit.

UweBonnes:
Maybe "-ex run -ex detach" is executed to fast and detach happens while CPU is still in reset. Has gdb some mean of delay?

joric:
I can't use -ex run in a batch file because it's not interactive. It's just -ex load -ex kill. Don't know about delay.

PR #361, opened 2018-06-22
Speed up bitbanging SWD/JTAG

UweBonnes:
By pulling the bitbang action into the high level JTAG/SWD commands, substantial speedup and better setup/hold and high times are achieved. For native,stlink and swlink SWD turnaround  has also some speedup by using direct register access against the generalized libcm3 high level functions.


korken89:
Tested with the following results:

* With a 40k binary I get 33 kb/s with the patch applied vs 23 kb/s on master
* With a 100k binary I get 44 kb/s with the patch applied vs 24 kb/s on master

UweBonnes:
Close and reopen to trigger travis run

UweBonnes:
reopen to trigger travis run

gsmcmullin:
This works for me.  I think the small compromise in readability is worth the improved performance.

joric:
Came here from #320 - the TMS floating feature (now closed) works like a charm on my YJ-14015 (Core-nRF51822) module ([mitosis-bluetooth](https://github.com/joric/mitosis/tree/devel/mitosis-bluetooth)), so, thanks! Disconnecting wires every time was very frustrating, considering those bluetooth modules love powering off and detaching for every small cause.

PR #371, opened 2018-07-16
add push-pull on srst to miniv2

konsgn:
Self-explanitory I hope.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -11,6 +11,7 @@ endif
 OPT_FLAGS ?= -O2
 
 CFLAGS += -Wall -Wextra -Werror -Wno-char-subscripts\
+	-Wno-error=cast-function-type \
 	$(OPT_FLAGS) -std=gnu99 -g3 -MD \
 	-I. -Iinclude -Iplatforms/common -I$(PLATFORM_DIR)
 LDFLAGS += $(OPT_FLAGS)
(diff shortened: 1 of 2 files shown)


konsgn:
But we aren't tying high, we are just saying that when srst(stm32 state) is high, output 3.3v rather than float. When srst goes low, it will then output a 0 v which will assert reset on target too.


The problem is that with open-drain mode, on srst, we pull low and targets reset is asserted, but when we set it high, there is no pull up between the srst pin and the buffer, so it floats. That is not a condition that you should put to buffer.


What do you think?


PS: I also added a -Wignore to src/Makefile to get the code to compile under gcc8


Science!
________________________________

UweBonnes:
Well, the SRST line needs a pull up. And the SRST must be open collector. So activate the internal pull up on SRST when SRST is idle from the BMP line.

UweBonnes:
line -> side

konsgn:
The issue is that there is a buffer between the target and the bmp uC. I've had a situation where the float state of the stm32(bmp) was somehow seen as low input to the buffer, and the reset line of the target was held low regardless of the bmp's srst state. 

Crap, it may be that the buffer I use has "bus hold circuits" (fxla108) that make the internal pull-up of the chip not strong enough to switch directions on the buffer.

PR #373, opened 2018-07-18
Fixes for detach and stm32mem

UweBonnes:


esden:
I just tested this patch. It does indeed fixes detach. Before the patch when using the `kill` GDB command the target did not reset and run the target payload code.

It is annoying especially when you are mass programming devices and want to make sure the programming went well. :smile_cat:

I have not tested stm32_mem but the patch adding a try/except sounds very reasonable.

UweBonnes:
The patch for stm32_mem helps much when bisecting....

UweBonnes:
Can you test https://github.com/UweBonnes/blackmagic/tree/python3 and give feedback?  There are probably still a lot of systems out there that still use python2. Can users in a pyhton3 default distribution live with the explicit call to python3? Does anybody have an idea how to code the carriage returen so that python2 _and _ python3 understand it?
-               print ("Programming memory at 0x%08X\r" % addr),
+               print ("Programming memory at 0x%08X" % addr, end="\r"),

stefaandesmet2003:
quick test, no luck. get a ModuleNotFoundError, although pyusb is installed.
Must be something silly ..

$ sudo ../scripts/stm32_mem.py ./blackmagic.bin 
Traceback (most recent call last):
  File "../scripts/stm32_mem.py", line 26, in <module>
    import usb
**ModuleNotFoundError: No module named 'usb'**

$ pip list|grep usb
pyusb                              1.0.2 

Why not  have #!/usr/bin/env python2 in the original script? That worked for me..

PR #375, opened 2018-07-25
More FTDI work: SRST, Direct SWD wiring.

UweBonnes:
This set of patches touches only FTDI devices.
It allows to assert SRST and handle direct FTDI<->SWD wiring like:
https://www.allaboutcircuits.com/uploads/articles/FT2232H_Wiring.jpg
As FTDI/DO is tristated when reading, the resistor is not needed, but no error.
Asynchronous reading is also activated , when recent libftdi defines are available. Travis does not have this defines available, so travis tests only synchronous transfers.


gsmcmullin:
I tried this with the "flossjtag" cable:
SWD no longer works. "SWD not possible or missing item in cable description."
JTAG is extremely slow.  `mon jtag_scan` takes 2m29s connected to an STM32F4 target.

UweBonnes:
Even so there was few feedback, I finnaly merged. Changes are restricted to the hosted branch so chances of collateral damage a re low.

PR #383, opened 2018-08-26
Remove the sleep from the platforms and add it to the needed places

rikvdh:
Hi all,

I ran into an issue with my ST-Link where the reset (via SRST) was not working. I found out that the `platform_srst_set_val` was missing the 1000-nop-sleep.

Instead of adding it there I think it would make more sense to add it in places where the sleep is actually needed. This prevents other developers from forgetting this in other platforms.

From my opinion this is also more intuitive since adding a 1000-nop-sleep when the value is true doesn't make any sense.

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -170,8 +170,11 @@ static bool cmd_jtag_scan(target *t, int argc, char **argv)
 
 	if(assert_srst != ASSERT_NEVER)
 		platform_srst_set_val(true);
-	if(assert_srst == ASSERT_UNTIL_SCAN)
+	if(assert_srst == ASSERT_UNTIL_SCAN) {
+		for (int i = 0; i < 10000; i++)
+			asm("nop");
 		platform_srst_set_val(false);
+	}
 
 	int devs = -1;
 	volatile struct exception e;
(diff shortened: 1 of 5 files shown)


rikvdh:
I don't really agree with you. What the delay tries to achieve is to wait long enough to ensure a proper reset. When someone has enough capacitance on the reset-line (due to long wires or a capacitor on the target) the target doesn't reset. I have seen this with my ST-Link which has no delay currently, my target didn't reset because the time is too small. All chip suppliers give you reset charateristics which must be adhered to allow for a proper reset.

For example the EFM32 requires a 50 ns reset pulse before the reset takes effect. This means that if you have no delay at all the target doesn't reset or goed into an weird semi-reset state. Other chip-suppliers have other delays. When you don't wait, the STM32 on 72 MHz needs at least 4 clockcycles to achieve this (beside GPIO propagation)(not taking into account any bus capacitance).

But from my opinion this shouldn't be responsibility of the platform. When implementing a platform it is odd to delay when `assert` is true, that behavior is quite unexpected for a function which only needs to toggle the SRST. From my opinion it is the responsibility of the caller to assure the correct delay to guarantee reset on all targets.

I know about the nop-loops to be a bad way for sleeping. But I searched the code and did not find any other way in place currently. I saw this happening on some other places as well this way and believed it was the way you wanted it.

rikvdh:
Perhaps we should forget about this change. However the way I've implemented the feedback of the SRST pin in #384 should be in all platforms I think.

PR #384, opened 2018-08-26
Change the ST-Link SRST reset function

rikvdh:
The reset function works, but it is very unreadable and uses the wrong PORT when you change the definitions in the header.

I suppose using the libopencm3 headers is a bit more intuitive here.

### DIFF
--- a/src/platforms/stlink/platform.c
+++ b/src/platforms/stlink/platform.c
@@ -68,9 +68,8 @@ void platform_init(void)
 	              GPIO_CNF_OUTPUT_PUSHPULL, TCK_PIN);
 	gpio_set_mode(TDI_PORT, GPIO_MODE_OUTPUT_50_MHZ,
 	              GPIO_CNF_OUTPUT_PUSHPULL, TDI_PIN);
-	gpio_set(SRST_PORT, srst_pin);
-	gpio_set_mode(SRST_PORT, GPIO_MODE_OUTPUT_50_MHZ,
-	              GPIO_CNF_OUTPUT_OPENDRAIN, srst_pin);
+
+	platform_srst_set_val(false);
 
 	gpio_set_mode(LED_PORT, GPIO_MODE_OUTPUT_2_MHZ,
 	              GPIO_CNF_OUTPUT_PUSHPULL, led_idle_run);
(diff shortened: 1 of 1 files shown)


rikvdh:
An explanation to why I removed the polling-loop: When you assert the SRST, you set it as output and cannot rely on the gpio_get to return the actual pin-state. It can also read the written state above. On the de-asserted state, you don't need to wait. All other platforms omit any delay as well since the target is coming out of reset and they can do that themselves.

This section (like any other platform) need the delay when asserting, please keep #383 in mind. If we agree on not merging that one, this piece of code needs the delay-loop added.

UweBonnes:
I don't understand what you mean by "When you assert the SRST, you set it as output and cannot rely on the gpio_get to return the actual pin-state". Reading GPIO_IDR gets the state of the pin in all cases, as I read the documentation.

rikvdh:
Well, the chip may return the actual output state, but taking in mind lengths of the reset-lines and possible capacitances on this line returning to !asserted after asserted may be a tad too soon. When the actual reset pulse of a device is not adequate it can result in 'semi' reset devices or not reset devices (the latter was true for my case). Most devices also require minimum reset times which can't be guaranteed with this manner.

rikvdh:
Ok, perhaps my test was wrong. I'm sure my target didn't reset when I asked it to. I've added the loop, but changed it a little. Since the return-value for `gpio_get()` is not boolean and a comparison with `assert` is always false when assert is true.

This is gpio_get in libopencm3:
```
 70 uint16_t gpio_get(uint32_t gpioport, uint16_t gpios)
 71 {
 72         return gpio_port_read(gpioport) & gpios;
 73 }
```

So when you provide `srst_pin` as the argument `gpios`, the return value when the pin is set will be: `srst_pin`, which may not be equal to boolean `true`.

UweBonnes:
Okay, good analysis. I did not read the libopencm3 api carefully enough. From the ethernut API I was used to get a boolean.

I think you should either merge both commits, or split them different: -Cleanup of platform_init(), - Fix platform_srst_set_val()

rikvdh:
Thanks, I've squashed the commits to one and changed the commit message.

UweBonnes:
Gareth, please merge.

Issue #385, opened 2018-09-02
Connecting under reset not working


anyn99:
I have a custom board with an STM32L432KB.
There was some trouble when trying to attach to it, scanning worked fine most of the time.
I did some, well quite a lot of debugging, and it seems that the issue is related to my board.
It has only two layers and the SWDCLK and SDWIO traces are routed all over the place - they are simple button inputs when not debugging. The BMP code as it is, is just to fast for this! :-)

Would it be possible to add some kind of speed setting that can be set in the monitor menu?

I did something like this: 
```
void delay(unsigned int ticks)
{
volatile unsigned int i;
for (i = 0; i < ticks; i++)
  {
    asm("nop");
  }
}
```
And then at every clock out in /platforms/common/swdptap.c
```
gpio_set(SWCLK_PORT, SWCLK_PIN);
delay(DEFAULT_DELAY);
gpio_clear(SWCLK_PORT, SWCLK_PIN);
delay(DEFAULT_DELAY);
```
Actually with a DEFAULT_DELAY of 1, the code already worked flawlessly for me, so I am not completly sure if it isn't really a timing (compiler over-optimizing?!) issue

UweBonnes:
Are you using git head? Git Head has no sequence to 
bla()
gpio_set()
gpio_clear()
bla()

Either gpio_set() is already doubled or there are other commands in between. So the time the signals are high is already extended as before e54a826745ae (loop unrolling). Even before with the short high times, BMP worked for many boards.

Adding a function call in the unrolled loops with result in substantial longer programming times.

Did you look at the signals on your board with a scope? Can you compare against against e.g. a BMP modified Nucleo Board?

anyn99:
yes i use the current head, but modified it until it works for me, sorry for being unclear about that.
Here some more info. I actually don't use original BMP, but ST-LINK v2 "Baite" clone.
Sorry that i can't compare to a different BMP probe....

I started using the code as is, with gpio_set() doubled in swdptap.c, scanning worked fine, but when attaching the probe failed everytime at the first adiv5_swdp_low_access. After the register write (8 bits) the ack was always 111 (0x7) instead of 100 or 010 or at least 001. So i think it was a timing issue on the bus.
Maybe i will try to doube the gpio_clear()s as well and see if it works.

UweBonnes:
Did you try "mon assert_srst attach"

anyn99:
[...]

PR #387, opened 2018-09-03
Allow Cortex-M to connect under reset

anyn99:


anyn99:
#385

UweBonnes:
Scan should, if possible, not change the state of the scanned CPUs. In the case of assert_srst = none, your code writes CORTEXM_DEMCR even if we later do not attach to that CPU. What about writing DEMCR before scannning only if assert_srst != none and keep the late write with assert_srst = none?

Removing target_halt_resume from the probe function is also very intrusive. This will keep all CPUs in reset after scanning with assert_srst = scan. Remember, the code you touch may also be called in a JTAG chain with several CPUs.

If you can, test your code with a well behaved CPU  like STM32F3 that does not sleep with assert_srst = none. Scanning should not interrupt/reset/hold the programm running on that CPU.

anyn99:
Hm, of course your remarks are right.
I changed the code to read the value of srst line, only if srst is asserted i issue a write to DEMCR.

For the other part i haven't found a good soultion so far. cortexm_forced_halt() now gets every scanned device in a halting state, that's already intrusive. But we kind of need to do that to get the IDCODE and find out what devices are there, am i right? Now after probing, how do i determine if the target should be resumed or stay halted? If i resume a halted device which disables SWD lines in its code, i have a device in my scan list that is lost and cannot be attached. If i keep the device halted after scanning, and don't wan't to attach to it, it's just stalled and doesn't run it's program. Maybe it's time to rethink the connect under reset feature?

In OpenOCD they have the connect_assert_srst option to specify that the device is to be held under reset for connecting, and there is no distingushing between reset until scan and reset until attach. 
And there are the "reset halt", "reset init", "reset run"(default) and also "soft_reset_halt" commands that the user can issue to manually reset and define behaviour AFTER reset.
Maybe BMP should also behave that way?
That would mean that a cpu in some exotic state (e.g. swd disabled) could always be accessed with "connect under reset" and "reset halt" enabled. 

Actually i have to admit that i don't know if it really makes sense... but i will try if you think so.

UweBonnes:
Yes, halting all targets to scan them is intrusive, But at least it does not change state of the device, like a reset. I have a little hope that the need to halt targets to read the ID, as seen on stm32L0 is a pitfall of the missing two SWD cycles now added with #391. I will try to test asap.

B.t.w, I had a hell lot of problems to connect to my STM32 targets running ethernut with WFI in the idle loop with OpenOCD. Drowning in the hell lot of possible configuration switches of OpenOCD, BMP became my favorite.

Can you get OpenOCD to connect to your boards in stop mode? What arguments are needed?

anyn99:
I can't test the connection with OpenOCD atm, because i only have my BMP now, the ST-Link v2 i had is broken, but new one will arrive within a week or so.
As far as i remember it worked, but of course i had to enable debug during low power mode with HAL_DBGMCU_EnableDBG***Mode, with *** being SLEEP, STOP or STANDBY. This command is from the STM32 HAL-Library, basically its just setting a bit in DBGMCU register. Did you that?

anyn99:
I will test your probe_rework pull request and see if that works for me, so this pull request is useless now.

Issue #389, opened 2018-09-05
EFM32G222F64 detection in efm32_probe doesn't work


rikvdh:
I found out that the reading of the `efm32_v1_read_eui64` works, but after shifting and masking there is something completely different than `EFM32_V1_DI_EUI_SILABS`. The resulting value from the shift and mask (`((efm32_v1_read_eui64(t) >> 40) & 0xFFFFFF)`) of the EUI64 is in my case: `0x0024bcdc`.

Also the format of this register is not documented by Silabs, so I think it is a bad idea to rely on it.

rikvdh:
I found some knowledge base article on the unique-id registers from early 2017.
https://www.silabs.com/community/mcu/32-bit/knowledge-base.entry.html/2017/11/01/efm32_and_efr32_seri-uVaX

When looking through the public IEEE listing for OUI's (http://standards-oui.ieee.org/oui/oui.txt) I find several entries for Silabs including the current one. Also the energy micro variant is there. But my variant (see below) is not listed.

My part is the EFM32G222F64G-E-QFP48R (via digikey ordered, so I don't expect counterfeits) https://www.digikey.nl/product-detail/en/silicon-labs/EFM32G222F64G-E-QFP48R/336-4245-1-ND/7805193

```
EFM32 EUI64 0x24bcdc 055a 02d28e
EFM32 EUI48 0x574e2c 1cffff
```

UweBonnes:
Can you use the  Simplicity Commander “device info” command on your device? What does it tell?

rikvdh:
I can't use it, it needs a Segger J-Link which I don't have. I shall try to find someone in my area which can lend it to me.

UweBonnes:
The EFM demo boards have a Segger J-Link on-board. It can also be used for external devices. Do you perhaps have such a device?

rikvdh:
I don't have such board. I shall arrange something to test this with. Perhaps someone of my collegues has a J-Link.

N-Storm:
I have a custom EFM32**T**G222F32 board. And I can't even swdp_scan it:
```
Remote debugging using /dev/ttyACM0
(gdb) monitor swdp_scan
Target voltage: unknown
Ignoring packet error, continuing...
Ignoring packet error, continuing...
^C
(gdb) monitor targets
Available Targets:
No. Att Driver
No usable targets found.
```

My BMP are running on "Blue Pill" and firmware are build from recent git master branch:
```
[...]

PR #390, opened 2018-09-05
Move the LPC17xx probe function down

### LINKED ISSUE #388: Probe function for LPC17xx gets stuck
Hi,

I have a EFM32G222F64 attached to my ST-link (with BMP, latest master and some patches for my target).  When I probed my device with `mon sw` in GDB the BMP gets stuck. After some debugging, it seemed that the `lpc17xx_probe` gets stuck. For now I've commented it out because I don't need it.

I shall perform some more debugging on my BMP to exactly pin-point the culprit in `lpc17xx_probe`.

Edit:
I've already found out that the ap_idcode is the same for both LPC17xx and EFM32:

```
target/lpc17xx.c
40:#define LPC17xx_SWDP_IDCODE                  0x2BA01477

target/efm32.c
567:    if (ap_idcode == 0x2BA01477) {
```

Also the core-check passes, the detection gets stuck in the `lpc17xx_iap_call(t, &param, IAP_CMD_PARTID);` (here: https://github.com/blacksphere/blackmagic/blob/master/src/target/lpc17xx.c#L96)

In the IAP call it resets the MCU and waits for the MCU to stop by itself, but it doesn't since the IAP is not valid for my EFM32. It is now stuck in the while-loop there forever. (here: https://github.com/blacksphere/blackmagic/blob/master/src/target/lpc17xx.c#L196)

rikvdh:
since it performs an IAP call which can hang when performed on other devices than a LPC17xx. Fixes #388 

I've checked the other LPCxx probes, but they don't use IAP for probing.

### DIFF
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -335,7 +335,6 @@ bool cortexm_probe(ADIv5_AP_t *ap)
 	PROBE(stm32l4_probe);
 	PROBE(lpc11xx_probe);
 	PROBE(lpc15xx_probe);
-	PROBE(lpc17xx_probe);
 	PROBE(lpc43xx_probe);
 	PROBE(sam3x_probe);
 	PROBE(sam4l_probe);
@@ -345,6 +344,7 @@ bool cortexm_probe(ADIv5_AP_t *ap)
 	PROBE(kinetis_probe);
 	PROBE(efm32_probe);
 	PROBE(msp432_probe);
+	PROBE(lpc17xx_probe);
 #undef PROBE
 
 	return true;


UweBonnes:
Gareth, please apply!

rikvdh:
I think your device is not detected and the Blackmagic reaches the lpc17xx_probe. It may be something else, but the behavior you described was similar to mine. You can try to remove the lpc17xx_probe in order to see if the BMP does not get stuck. Can you also try to use the Simplicity Commander “device info” command?

N-Storm:
I've commented out every probes except for stm32f1, f4 and efm32 targets, i.e.:
```
#define PROBE(x) \
        do { if ((x)(t)) {target_halt_resume(t, 0); return true;} else target_check_error(t); } while (0)

        PROBE(efm32_probe);
        PROBE(stm32f1_probe);
        PROBE(stm32f4_probe);
#undef PROBE
```

It doesn't hangs now but doesn't recognizes target as EFM32:
```
(gdb) monitor swdp_scan
Target voltage: unknown
Available Targets:
No. Att Driver
 1      ARM Cortex-M
(gdb) monitor tar
Available Targets:                                                                                                                                                                                                                           
No. Att Driver                                                                                                                                                                                                                               
 1      ARM Cortex-M                                                                                                                                                                                                                         
(gdb) monitor att 1                                                                                                                                                                                                                          
Target does not support this command.  
```

rikvdh:
That at least confirms my suspicion. The LP17xx probe tries some calls which get stuck. Probably the EUI readout goes wrong like in my case. Can you read the EUI? Perhaps with OpenOCD at register 0x0fe081f4 & 0x0fe081f0 or 0x0fe081dc & 0x0fe081d8.

N-Storm:
```
(gdb) monitor mdw 0x0fe081f4 2
0x0fe081f4: 24c6da05 00040020 
(gdb) monitor mdw 0x0fe081f0 2
0x0fe081f0: 54bd830b 24c6da05 
(gdb) monitor mdw 0x0fe081dc 2
0x0fe081dc: 898a9ab6 4cff8b84 
(gdb) monitor mdw 0x0fe081d8 2
0x0fe081d8: 2c1c7569 898a9ab6 
(gdb) monitor targets
    TargetName         Type       Endian TapName            State       
--  ------------------ ---------- ------ ------------------ ------------
 0* efm32.cpu          cortex_m   little efm32.cpu          halted
```

rikvdh:
As I expected. You now have the same issue as I have. The EUI doesn't start with 0xd0cf5e nor with 0x000b57. Therefore it is not detected. What you can do (as a *temporary* workaround), at this part https://github.com/blacksphere/blackmagic/blob/master/src/target/efm32.c#L576-L588 remove this and just set `di_version = 1;` or `= 2` depending on your device.

This is the same issue as #389 and apparently it also applies to your device. I shall try to come up with a fix because perhaps other people shall face this issue shortly.

UweBonnes:
Sorry, I mixed up NRF51 and EFM32. Make it at least active as DEBUG() so that people that compiled with ENABLE_DEBUG=1 can see it. B.t.w. libftdi has always prints out DEBUG() on the contolling console.

N-Storm:
I've added this dirty hack as @rikvdh suggest to see. Here is what I get with di_version explicitly set to V1:

```
(gdb) mon swdp_scan
Target voltage: unknown
Available Targets:
No. Att Driver
 1      EFM32TG 222 F32 Tiny Gecko
(gdb) att 1
Attaching to Remote target
0x000005a8 in ?? ()
(gdb) mon tar
Available Targets:
No. Att Driver
 1   *  EFM32TG 222 F32 Tiny Gecko
(gdb) mon efm_info
DI version 1 (silabs remix?) base 0x0fe08000

EFM32TG 222 F32 = Tiny Gecko 32kiB flash, 4kiB ram
Device says flash page size is 512 bytes, we're using 512 bytes

(gdb) mon serial
Unique Number: 0x000000000000000lx
```

And with V2:
```
(gdb) mon swdp_scan
Target voltage: unknown
Available Targets:
No. Att Driver
 1      EFM32TG 222 F32 Tiny Gecko
(gdb) att 1
Attaching to Remote target
0x000005b2 in ?? ()
(gdb) mon efm_info
DI version 2 (energy micro remix?) base 0x0fe081b0

EFM32TG 222 F32 = Tiny Gecko 32kiB flash, 4kiB ram
Device says flash page size is 512 bytes, we're using 512 bytes

Package WLCSP 0 pins
Temperature grade -40 to 85degC

(gdb) mon serial
Unique Number: 0x000000000000000lx
```

rikvdh:
@UweBonnes I'm working on a fix for #389, I shall include the debug statement.

@N-Storm For your EFM32TG it should be V1. It appears to attach to your device properly now. :-)

The real fix is work-in-progress.

PR #391, opened 2018-09-06
adiv5_swdp: Add extra idle cycles with write transactions.

UweBonnes:
These extra cycles are needed by some CPU, e.g. STM32L0x1 to cross the SWCLK
/HCLK domains. Revert insufficient #373 also tackling that problem.

Thanks to Thorsten von Eicken for pointing out.

### DIFF
--- a/src/target/adiv5_swdp.c
+++ b/src/target/adiv5_swdp.c
@@ -159,6 +159,17 @@ static uint32_t adiv5_swdp_low_access(ADIv5_DP_t *dp, uint8_t RnW,
 			raise_exception(EXCEPTION_ERROR, "SWDP Parity error");
 	} else {
 		swdptap_seq_out_parity(value, 32);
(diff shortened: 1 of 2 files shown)


tve:
LGTM. This is what I did in my local fork. I only have stm32l0's to test with.

tve:
BTW, I just came across the following passage in https://static.docs.arm.com/ihi0031/c/IHI0031C_debug_interface_as.pdf page 4-89 (intro of sec 4.2), which explains why the code you originally removed had 8 cycles (notice the last bullet point):

> To ensure that the transfer can be clocked through the SW-DP, after the data transfer phase the host must do one of
the following:
• Immediately start a new SWD operation with the start bit of a new packet request.
• Continue to drive the SWD interface with idle cycles until the host starts a new SWD operation.
• If the host is driving the Serial Wire Debug clock, continue to clock the SWD interface with at least eight idle cycles. After this it can stop the clock.

PR #392, opened 2018-09-08
improve SWD timing

tve:
I found the SWD timing to be very tight, in particular, there was very little set-up time for SWDIO before the rising edge of SWCLK. I switched things around so the clock remain high when idle such that when outputting a bit SWDIO can be driven, then the clock pulsed low, then high again. The result is a ~3.5Mhz clock with quite even high and low times, plus a good set-up time. I do not know how this compares from a freq perspective with the old code, I have to admit.
I realize this PR comes a bit out of the blue, so if this doesn't look right or good, please let me know!

### DIFF
--- a/src/platforms/common/swdptap.c
+++ b/src/platforms/common/swdptap.c
@@ -23,6 +23,8 @@
 #include "general.h"
 #include "swdptap.h"
 
+//#define DEBUG_SWD_BITS
+
 enum {
 	SWDIO_STATUS_FLOAT = 0,
 	SWDIO_STATUS_DRIVE
@@ -47,9 +49,9 @@ static void swdptap_turnaround(int dir)
 
 	if(dir == SWDIO_STATUS_FLOAT)
 		SWDIO_MODE_FLOAT();
-	gpio_set(SWCLK_PORT, SWCLK_PIN);
-	gpio_set(SWCLK_PORT, SWCLK_PIN);
 	gpio_clear(SWCLK_PORT, SWCLK_PIN);
+	gpio_clear(SWCLK_PORT, SWCLK_PIN);
+	gpio_set(SWCLK_PORT, SWCLK_PIN);
 	if(dir == SWDIO_STATUS_DRIVE)
 		SWDIO_MODE_DRIVE();
 }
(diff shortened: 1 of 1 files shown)


UweBonnes:
SWD timing before loop unrolling was very tight. However I don't remember any PR stating timing problems. Loop unrolling improved timing and increased speed. 

Your changes leave the SWD clock high at the exit of the functions. I don't see when you return to the idle level. Idle level is low, as all SWCLK pins have pull-downs.

I only know the STM32H7 datasheet where SWD timing requirements are given, with some error MHz vs ns. It seems we keep the timing with much margin.

Do you really have a problem with timing? A less intrusive way would be to extend the high times in the recent cold by duplication more GPIO_SET() calls. Probably this will have few impact on programming speed, as we mostly wait for the flash programming time, but give us more margins. Programming speed test on different chips will then show the influence.

tve:
I was having timing issues due to cables. Some boards I have don't have an SWD connector for a pretty flat cable and I have to rig up probes. Even when only 20cm long the reflections are visible. I also noticed that st-link does exactly what my PR does, i.e., set SWDIO on the falling edge of SWDCLK and keep SWDCLK high when idle. I could not find any requirement to keep SWDCLK high or low when idle, in fact, if anything keeping SWDCLK idle goes against the spec...
But I understand the "if it ain't broke don't fix it" approach. I'm happy to keep my fork and work on increasing the clock rate there :-).

gsmcmullin:
This seems like a bad wiring problem rather than a firmware problem.  If there is indeed a problem with the signal timing, please reopen.

jelson:
I just wanted to give this a +1 -- I had a very frustrating few hours trying to get my BMP to connect to an STM32F031 and believe I have this tracked down to the slow rise time of SWDIO compare to the clock time. The clock pulses are apparently being driven well because I see nice, settled square waves. However SWDIO is barely getting to 2.5v (in a 3.3v circuit) before falling again; I get shark-tooth waves rather than square waves.

My STM32F0 is a TSSOP-20 package that I have soldered to a 20-pin-DIP breakout board and plugged into a solderless breadboard. 

I see SWDIO rise times I see are 50ns to get to 2.5V (in a 3.3V circuit) before starting down again. 

I'd vote strongly to slow the clock down.

UweBonnes:
Please look at my gitter post.

UweBonnes:
Your patch gives real improvement in setup time. Please fix style issue like missing spaces between item e.g. in " i<ticks". Additional margins could be gained when in-sequences would set SWDCLK high  last in the loop, if I understand things right. I experimented with adding a DELAY(x) function with user settable x,  like 
 #define DELAY() do {volatile int d = delay; while (d--);} while(0)
but that has substantial effect on F7 programming speed. Speed got down from about 60 to about 40 kBytes/s even with x = 0. Would this be worth the effort?

tve:
Thanks for taking another look! I will fix the style issues.

jelson:
Regarding DELAY, if reducing speed from 60kbps to 40kbps is not acceptable, it would be great if at least we had the option at compile-time to insert a delay. For example:

```c++
// Uncomment this if you want the highest-speed communication possible
#define DELAY() {}

// Uncomment this if you want to slow down the clock
//#define DELAY() do {volatile int d = delay; while (d--);} while(0)
```

In other words, it would be great to put DELAY() statements in the code in the right places, but at compile-time make it easy to either include them or disable them completely.

jelson:
Or, perhaps more elegantly:

```c
// Define 0 to bit-bang as quickly as possible; >0 to slow down the SWD clock.
#define NUM_DELAY_TICKS 3

#if NUM_DELAY_TICKS > 0
#define DELAY()  do {volatile int d = delay; while (d--);} while(0)
#else 
#define DELAY() do {} while(0)
```

UweBonnes:
Jeremy, You still did not answer my question about the driver chip on your BMP. Did you follow Piotrs posting to gitter? I am quite sure your have one of those screwed BMPs with the TXS driver. Perhaps a replacement applies. I am quite sure your problems are caused by the BMP hardware.

jelson:
I did see the posts on Gitter - but it seemed the problem may have been that I was using a breadboard. My hardware has silkscreen that says "BMPM v2.1c". If you tell me how to identify the driver chip, I have it here and can look at it under a magnifier.

UweBonnes:
Can you post a macro photo, best with light coming from the side. Otherwise, do some math and put the time constant of your sawtooth in relation to output resistance of your driver and the expected capacitance of the breadboard.

jelson:
Photo here: https://imgur.com/a/dRmpas1

UweBonnes:
Okay, V2.1  seems not to have TXS, also I can not decipher the chips. So carefully look at other reasons for those screwed waveforms.

jelson:
My current theory is just that it's from all the extra resistance and capacitance of the breadboard I was using as an ersatz target board. I am going to prototype on the breadboard using a slower clock (by adding DELAY() during bit-banging), and if that works, I will look at the waveforms again after I get a proper PCB made and see how it looks there. Thank you for your help.

UweBonnes:
Did you perhaps use some coax cable for probing?Coax at 50 Ohm has  100 pF/m.

EliasOenal:
I too think the BMP needs a way to slow down SWD. It would be most convenient if this was configurable at runtime. (e.g. we keep the current implementation and also offer a slower one with variable delay.) Last time I tried it couldn't reliably communicate with the official STM32H743I-EVAL evaluation board. I had to manually add delays for a stable operation.

UweBonnes:
Superseded by #783

PR #395, opened 2018-09-30
Bring swlink up-to-date and allow reuse for "bluepill" and other boards

UweBonnes:
The swlink platform missed things like reset handling, dfu-upgrade etc. This patchset brings the swlink platform up-to-date and allows to reuse it for other boards with the JTAG connections exposed, like the "bluepill" #354.


justinkb:
for blue pill, am I crazy or should SWO/RX2 say "P3/8 (13)" instead of "P3/7  (13)"

UweBonnes:
Good catch. Will you do a PR?

justinkb:
done

PR #397, opened 2018-10-02
Add target support for LPC11U3X devices.

ijager:
The DeviceID register has a different address from the LPC11XX devices: `0x400483F8` instead of `0x400483F4`. Memory and flash is compatible.

Tested on a `lpc11u34/421`


UweBonnes:
LPC11U3X and LPC8 have the Device ID at the same place. I propose you define the address only once, read it only once and use one switch case.

UweBonnes:
Looks good. Gareth, pleas merge!

PR #399, opened 2018-10-04
Probe rework

UweBonnes:
Remove the "halt always before probe" and restore the changed connect_srst syntax mess. Restore DBGMCU_CR register on detach if changed by attach. Release SRST only when target is already set up to halt on reset.


kabbi:
One of extra delays actually fixes nrf52 not starting when detached. Tested on bluepill + RedBear Nano 2 board.

korken89:
This solved my issues with an nRF52 always crashing after running `load`, thanks!

anyn99:
Hi there,

i tested with my setup (STM32L432 with binary almost always in WFI) and found that i can no longer flash my device with this probe_rework... with the master from my fork it works... after looking at the code i can see no obvious reason for this. i probe successfully and attach, everything is fine, just the "load" hangs. btw i tried with connect_srst enabled and disabled.

UweBonnes:
Cab you test what happens with the line
	if (forced && !ap->dp->targetid)
removed?

anyn99:
Then everything works as before!
I don't understand why i have to halt before probing and that solves the issue, but probing succeeds halted or not...?
My quick and dirty hack is now:
´´´
uint32_t dhcsr = target_mem_read32(t, CORTEXM_DHCSR);
	DEBUG("FORCED = %d TARGETID %08" PRIx32 " DHCSR %08" PRIx32 " \n", forced, ap->dp->targetid, dhcsr);

	if ( (forced && !ap->dp->targetid) || (dhcsr & CORTEXM_DHCSR_S_SLEEP) )
´´´

UweBonnes:
When the CPU is sleeping, access to the MCU register is not possible, so the Debug accesses fail. Your "quick and dirty" seems to me like the right solution. Can you provode a pull request.

anyn99:
I could do that, but i think a found a better solution! 
After a night of sleep i found the real problem is that we shouldn't try to access flash registers before 
attaching! I wondered why the stm32h7, nrf51 and others do not need to be halted before probing, but my stm32l4 does?
 The problem is that stm32l4_probe() reads things like flashsize from the registers, which it cannot do when in sleep and not halted. The other target implementations do that during attaching. Which is of course the right time to do this, since the device is halted anyway: cortexm_attach() calls cortexm_forced_halt() in the beginning. 
Simple solution:
probe functions must not read anything else than targetid! 
everything else is done during attach. This way the probing is least invasive. 
I will implement the simple changes for stm32l4 for sure and also have a glance at other targets if possible.

UweBonnes:
Your stm32l4 is doing long sleep. I use NutOs on my chips that do 1000 wakeups per second. As we hammer the mcu with register access requests during probe, probably I catch the CPU sometimes not sleeping and access succeeds.
Probably most people with their program doing long sleep use "connect_srst enable". They do not care that they loose the program state that way.
Only for cortex-m7 you can read the mcu ID from the access point. For other STM32 the chip id is a register access like the register access for flashsize. So check that read IDCODE read succeeds while read flashsize read fails on your target during sleep. I wonder why both accesses should work different?

For F4 , probe only reads the IDCODE. For F7/H7 no register read is needed for the IDCODE, as the DAP already provides the IDCODE. F4/7/H7 splits probe and attach. so that other register are only accessed when halted.

Probably we should remove all other register access beside IDCODE access during probe. Announcing  mode flash and ram while the chip announces will do no harm beside some users wondering. This also allows to programm the upper 64 kbytes of a STM32F103x8. We need to query the chip however if the blocksize and flash layout has some fuse. That should be done in a separate attach() like on F4/H7.

So clearly L4 should  spilt attach and probe! L4 can switch blocksize/flash layout.
 
Another problem arises when the program remaps SWD pins and "connect_srst enable" is needed. For 
some STM32L1 or maybe STM32L0 every DAP access fails during reset.

So fiddling with the probe code still has a way to go and offers the possibility for bad side effects.

anyn99:
Yes when i had my program sleep for 5s and be awake for 5s, i mostly succeeded connecting. But now i sleep 5s and am awake only 50ms.
I tryed the changes and it works flawlessly, i can access the dbgmcu register with the idcode during sleep without halting the mcu. Accessing flashsize is not possible.
The stm32l4 is a cortex-m4, but the architecture is armv7, correct me if i got this wrong, i am not sure. So maybe thats why accessing dbgmcu works?

To have it nice and tidy we could set the flashsize as a constant (maybe with possibility to override so that one can use more flash than supported officially), specific things like dualbank setting could then be corrected during attach?

anyn99:
Ah and i think i can access dbgmcu register with connect_srst enabled or disabled, at least on stm32l4.

anyn99:
I will complete the pull req soon, probably this evening

anyn99:
I will complete the pull req soon, probably this evening

anyn99:
Ok just had a read at https://de.m.wikipedia.org/wiki/ARM-Architektur#ARMv6
So cortex-m0, cortex-m0+, and cortex-m1 are armv6. So stm32l0 stm32l1 are v6 and stm32l4 is v7

UweBonnes:
"To have it nice and tidy we could set the flashsize as a constant" -> Do not set sizes during probe! Do it with attach. Look at STM32F4/H7 how to implement the spilt.

anyn99:
Ok thats what i did already, pull req coming soon.

PR #402, opened 2018-10-20
[efm32] Assume Device Identification (DI) version 1 if we don't know the OUI

richardeoin:
Silabs are using some additional OUIs we don't know about. Reported in issue #389
These should use a DI version 1 layout, so assume version 1 layout for OUIs we don't
know. However do print a notice about this on DEBUG() as suggested by @UweBonnes

The IDCODE value is sufficient to make a positive identification of an EFM32 device.
See AN0062 Section 2.2. Therefore accepting any OUI is reasonable behaviour.
Additionally the part familiy is checked, and the target rejected if not in the
`efm32_devices` table. This commit makes that rejection explicit, although it does
not change the logical behaviour here.

Note that the important registers (part number, part family, flash size) are at the
same addresses in both layouts anyhow. Currently only `efm32_cmd_serial` and
`efm32_cmd_efm_info` functions use registers that differ between DI versions.

### DIFF
--- a/src/target/efm32.c
+++ b/src/target/efm32.c
@@ -572,7 +572,8 @@ bool efm32_probe(target *t)
 		return false;
 	}
 
-	/* Check the EUI is silabs. Identify the Device Identification (DI) version */
+	/* Check the OUI in the EUI is silabs or energymicro.
+	 * Use this to identify the Device Identification (DI) version */
 	if (((efm32_v1_read_eui64(t) >> 40) & 0xFFFFFF) == EFM32_V1_DI_EUI_SILABS) {
 		/* Device Identification (DI) version 1 */
 		di_version = 1;
@@ -580,7 +581,10 @@ bool efm32_probe(target *t)
 		/* Device Identification (DI) version 2 */
 		di_version = 2;
 	} else {
-		/* unknown device */
+		/* Unknown OUI */
+		DEBUG("EFM32: Unknown EUI 0x%016llx", efm32_v1_read_eui64(t));
+		DEBUG("EFM32: Assuming V1 Device Identification (DI) page (silabs remix?)");
+		di_version = 1;
 		/* sprintf(variant_string, */
 		/* 		"EFM32 DI Version ?? 0x%016llx 0x%016llx", */
 		/* 		efm32_v1_read_eui64(t), efm32_v2_read_eui48(t)); */
@@ -590,6 +594,10 @@ bool efm32_probe(target *t)
 	/* Read the part number and family */
 	uint16_t part_number = efm32_read_part_number(t, di_version);
 	size_t device_index  = efm32_lookup_device_index(t, di_version);
+	if (device_index == 9999) {
+		/* unknown device family */
+		return false;
+	}
 	if (device_index > (127-32)) {
 		/* too big to encode in printable ascii */
 		return false;
--- a/src/target/efm32.c
+++ b/src/target/efm32.c
@@ -563,7 +563,7 @@ bool efm32_probe(target *t)
 	ADIv5_AP_t *ap = cortexm_ap(t);
 	uint32_t ap_idcode = ap->dp->idcode;
 
-	/* Check the idcode is silabs. See AN0062 Section 2.2 */
+	/* Check the idcode. See AN0062 Section 2.2 */
 	if (ap_idcode == 0x2BA01477) {
 		/* Cortex M3, Cortex M4 */
 	} else if (ap_idcode == 0x0BC11477) {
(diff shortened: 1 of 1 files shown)


UweBonnes:
The  clause with the test for 9999 seem superfluous.  Portability needs to be in the print format string, not the print argument. Rereading the device registers takes much more cycles than using CPU registers. What about following reformulated patch?

diff --git a/src/target/efm32.c b/src/target/efm32.c
index c0383e8..64e0540 100644
--- a/src/target/efm32.c
+++ b/src/target/efm32.c
@@ -572,19 +572,24 @@ bool efm32_probe(target *t)
                return false;
        }
 
-       /* Check the EUI is silabs. Identify the Device Identification (DI) version */
-       if (((efm32_v1_read_eui64(t) >> 40) & 0xFFFFFF) == EFM32_V1_DI_EUI_SILABS) {
+       /* Check if OUI in the EUI is silabs or energymicro.
+        * Use this to identify the Device Identification (DI) version */
+       uint64_t eui64 = efm32_v1_read_eui64(t);
+       uint64_t eui48 = 0;
+       if (((eui64 >> 40) & 0xFFFFFF) == EFM32_V1_DI_EUI_SILABS) {
                /* Device Identification (DI) version 1 */
                di_version = 1;
-       } else if (((efm32_v2_read_eui48(t) >> 24) & 0xFFFFFF) == EFM32_V2_DI_EUI_ENERGYMICRO) {
+       } else {
+               eui48 = efm32_v2_read_eui48(t);
+               if (((eui48 >> 24) & 0xFFFFFF) == EFM32_V2_DI_EUI_ENERGYMICRO) {
                /* Device Identification (DI) version 2 */
                di_version = 2;
-       } else {
-               /* unknown device */
-               /* sprintf(variant_string, */
-               /*              "EFM32 DI Version ?? 0x%016llx 0x%016llx", */
-               /*              efm32_v1_read_eui64(t), efm32_v2_read_eui48(t)); */
-               return false;
+               } else {
+                       /* unknown device */
+                       DEBUG("EFM32 eui64: 0x%016" PRIx64 " eui48: 0x%012"PRIx64 "\n",
+                                 eui64,  eui48);
+                       return false;
+               }
        }
 
        /* Read the part number and family */

richardeoin:
@UweBonnes Okay, a comment is fine instead of the 9999 clause

I can't read the patch - github has broken the formatting perhaps

UweBonnes:
The patch only does the register reads once and uses "PRIx64" as format specifier.

richardeoin:
@UweBonnes I've reworked to reduce the number of reads before identifying the device. I've also given up on identifying based on the OUI, since we can't track which values silabs are using.

This is what the probe code looks like now. Are you happy for this to be merged? 

```
    /* Check the idcode. See AN0062 Section 2.2 */
    if (ap_idcode == 0x2BA01477) {
        /* Cortex M3, Cortex M4 */
    } else if (ap_idcode == 0x0BC11477) {
        /* Cortex M0+ */
    } else {
        return false;
    }

    /* Check the OUI in the EUI is silabs or energymicro.
     * Use this to identify the Device Identification (DI) version */
    uint64_t oui24 = ((efm32_v1_read_eui64(t) >> 40) & 0xFFFFFF);
    if (oui24 == EFM32_V1_DI_EUI_SILABS) {
        /* Device Identification (DI) version 1 */
        di_version = 1;
    } else if (oui24 == EFM32_V2_DI_EUI_ENERGYMICRO) {
        /* Device Identification (DI) version 2 */
        di_version = 2;
    } else {
        /* Unknown OUI - assume version 1 */
        di_version = 1;
    }

    /* Read the part family, and reject if unknown */
    size_t device_index  = efm32_lookup_device_index(t, di_version);
    if (device_index > (127-32)) {
        /* unknown device family */
        return false;
    }
    efm32_device_t const* device = &efm32_devices[device_index];
    if (device == NULL) {
        return false;
    }

    uint16_t part_number = efm32_read_part_number(t, di_version);
```

The probe is relatively low down in the [priority list](https://github.com/blacksphere/blackmagic/blob/master/src/target/cortexm.c#L345), so overall I think this is a good change in order to close [#389](https://github.com/blacksphere/blackmagic/issues/389).

gsmcmullin:
This looks good to me.  @UweBonnes, if you're happy I'll merge it.

UweBonnes:
I have two remarks:
- It is always a good idea to print sensible debug messages if things go wrong. It looks like Richard did tabula rasa with the old debug messages.
- A lot of register handling here is done during probe. As the discussion in #399 shows, often register access fail when the device is not halted. Richard, did you test probing a sleeping device withoy reset?

But maybe that is stuff for the next patch.

gsmcmullin:
I'm merging this and closing #389.  If these register accesses aren't possible unless halted that can be fixed in another PR, this changeset won't make things worse than before.

richardeoin:
The MSC Flash pages can be accessed by the AHB-AP even when the core is stopped.

Here's my test method (on a `EFR32BG13P532F512GM32`)

* Load a test binary to the main flash
```
/* Chip init */
/* DC-DC converter init */
/* Clock Init */

volatile int i = 0;
while (1) {
  EMU_EnterEM3(true);
  i++;
}
```
* Power down the target
* 1 second
* Power up the target
* 1 second
* `monitor swdp_scan`
* `attach 1`
* Confirm `i = 0` (ie. the core hasn't woken up before)

The core has special EM2 and EM3 behaviour when the debugger is attached (see the various reference manuals). I guess the MSC flash memory and other blocks remain powered up.

More Issues & PRs for this family incoming!

src/target/efm32.c
```diff
@@ -572,33 +572,33 @@ bool efm32_probe(target *t)
...
 
-	/* Check the EUI is silabs. Identify the Device Identification (DI) version */
-	if (((efm32_v1_read_eui64(t) >> 40) & 0xFFFFFF) == EFM32_V1_DI_EUI_SILABS) {
+	/* Check the OUI in the EUI is silabs or energymicro.
+	 * Use this to identify the Device Identification (DI) version */
+	uint64_t oui24 = ((efm32_v1_read_eui64(t) >> 40) & 0xFFFFFF);
+	if (oui24 == EFM32_V1_DI_EUI_SILABS) {
 		/* Device Identification (DI) version 1 */
 		di_version = 1;
-	} else if (((efm32_v2_read_eui48(t) >> 24) & 0xFFFFFF) == EFM32_V2_DI_EUI_ENERGYMICRO) {
```
mmmspatz, line 579: This change made it impossible for version 2 to ever be detected.

`efm32_v1_read_eui64()` and `efm32_v2_read_eui48()` read from different addresses, so checking `efm32_v1_read_eui64() == EFM32_V2_DI_EUI_ENERGYMICRO` is nonsensical and will never be true.

Unless I'm missing something?

PR #403, opened 2018-10-23
libftdi: Allow to compile with mingw and cygwin.

UweBonnes:
Tested with x86_64-w64-mingw32-gcc-8.2.0 and cygwin gcc (GCC) 7.3.0.
Use libftdi1 unconditionally.
Setup travis to use libftdi1, as done by
  https://github.com/cyrozap/ftdi-unbricker/blob/master/.travis.yml


UweBonnes:
Slight change in commit remarks.

UweBonnes:
Travis makes me crasy! Why is the changes commit not picked up!

PR #425, opened 2019-01-05
Update nrf51.c

jtam007:
Add another device ID


UweBonnes:
Merger PR #434 should fix that issue.

PR #426, opened 2019-01-06
nrf51: Add nRF51802 QFAA A0

miek:


UweBonnes:
As my comment for #425:
Please try https://github.com/UweBonnes/blackmagic/tree/staging
This recognizes NRF chips during probe in a more general way, so there should be no need to enter individual IDs. It alos improves programming speed.

UweBonnes:
Merger PR #434 should fix that issue.

PR #428, opened 2019-01-07
Stm32g0

UweBonnes:


src/target/stm32l4.c
```diff
@@ -186,11 +206,15 @@ bool stm32l4_probe(target *t)
 		return false;
 	}
 	t->driver = designator;
-	target_add_ram(t, 0x10000000, sram2_size << 10);
-	/* All L4 beside L47 alias SRAM2 after SRAM1.*/
-	uint32_t ramsize = (idcode == ID_STM32L47)?
-		sram1_size : (sram1_size + sram2_size + sram3_size);
-	target_add_ram(t, 0x20000000, ramsize << 10);
+	if (is_stm32g0)
+		target_add_ram(t, 0x20000000, sram1_size << 10);
```
gsmcmullin: Please add braces around this.  In general, if braces are used around one part of the if/else they should be used around the other.

src/target/stm32l4.c
```diff
@@ -384,12 +413,16 @@ static bool stm32l4_cmd_option(target *t, int argc, char *argv[])
 	int len;
 	bool res = false;
 
+	const uint8_t *i2offset = l4_i2offset;
 	if (t->idcode == 0x435) /* L43x */
 		len = 5;
-	else
+	else if (t->idcode == 0x460) {/* G07x */
```
gsmcmullin: Please add braces around the preceding statement.

src/target/stm32l4.c
```diff
@@ -384,12 +413,16 @@ static bool stm32l4_cmd_option(target *t, int argc, char *argv[])
...
+		i2offset = g0_i2offset;
+		len = 7;
+	} else
```
gsmcmullin: Please add braces around the statement following the `else`.

src/target/stm32l4.c
UweBonnes: Coding style, argh. For NutOS, where I code a lot too, braces are requested even for for on-line clauses.  I remembered that BMP did not like, so now I learned about the extemption that the braces are wanted for one-liner when other clauses have it too . Using braces everywhere is clearer to me, as when the code in a one-line clause is extended, there is  no need to add the braces. I have adapted the code as requested.

PR #429, opened 2019-01-07
Nrf5

UweBonnes:
Recognize the NRF by common feature, not by individual IDs. No need to add zillions of IDs. Flash direct and remove flash stub. Show some device info thanks to Boris Sorochkin <boris@nestlogic.com>.

At least for the NRF52840, flash protection by ACLs is not yet understood. Erase/flash write probably does not work when ACLs are active. Connect under reset does not work, at least for the ACL protected NRF dongle software.


gsmcmullin:
Looks great.  If this doesn't break any existing functionality I'm happy to merge.

PR #430, opened 2019-01-09
fix: compile error debug print STM32F1

vidavidorra:
The wrong type was used to print a variable, which caused compile error in GCC compiler.

### DIFF
--- a/src/target/stm32f1.c
+++ b/src/target/stm32f1.c
@@ -214,7 +214,7 @@ static int stm32f1_flash_erase(struct target_flash *f,
 	/* Check for error */
 	sr = target_mem_read32(t, FLASH_SR);
 	if ((sr & SR_ERROR_MASK) || !(sr & SR_EOP)) {
-		DEBUG("stm32f1 flash erase error 0x%" PRIx32 "\n", sr);
+		DEBUG("stm32f1 flash erase error 0x%" PRIx16 "\n", sr);
 		return -1;
 	}
 


UweBonnes:
We read a 32-bit variable and we print a 32-bit variable. So the original code seems right too me. What compiler showed the error?

vidavidorra:
I haven't looked much at the code what was going on with the variable, but I saw that the declaration of the variable that is being printed is `uint16_t` (see [here](https://github.com/blacksphere/blackmagic/blob/ab91ab16fa07fd897d471f24d707c4e3c38a5168/src/target/stm32f1.c#L191)).

See below the relevant section of the compile command (`make ENABLE_DEBUG=1`) before this commit:
```
  CC      target/stm32f1.c
target/stm32f1.c: In function 'stm32f1_flash_erase':
target/stm32f1.c:217:9: error: format '%lx' expects argument of type 'long unsigned int', but argument 2 has type 'int' [-Werror=format=]
   DEBUG("stm32f1 flash erase error 0x%" PRIx32 "\n", sr);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

And compiler/version:
```shell
arm-none-eabi-gcc --version
arm-none-eabi-gcc (GNU Tools for Arm Embedded Processors 7-2018-q2-update) 7.3.1 20180622 (release) [ARM/embedded-7-branch revision 261907]
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

To me it looks like either one of the below, but I don't know what is being read at `FLASH_SR` and whether that should actually be 32 or 16 bits.
- the declaration should be `uint32_t` and the print staying as it is (`PRIx32`)
- the read should be using `target_mem_read16()`

vidavidorra:
Yeah, is okay for me I think. So this PR can be closed than right, since you've 'merged' it with your STM32G0 PR?

vidavidorra:
@UweBonnes Did you happen to include this in the STM32G0 patch you mentioned in your comment? If so, this can be closed.

hydra:
I think I might have just created a duplicate of this in #448

vidavidorra:
I've just check the master branch (dcd20ef08c93276b6c28ca4c152e36a658b563af) and this fix still needs to be applied there. Based on @UweBonnes prvious comment, eithe rthat PR is not merged or this fix was not added.
Also I'm not sure if the change I made is desired or the declaration should be 32-bit. I think the latter as this is potentially unsafe in reading the memory (passing 16-bit into a function that reads 32-bit).
If that is the case, I'd be happy to change it.

UweBonnes:
Try "make all_platforms ENABLE_DEBUG=1" and more error pop up. A PR fixing all these problems woyld be find.

UweBonnes:
Fixed by #449

PR #431, opened 2019-01-10
Implement watchpoints on Cortex-A

gsmcmullin:
The implementation is incomplete, as `cortexa_halt_poll` will never return `TARGET_HALT_WATCHPOINT` and the watchpoint address.  It is unclear how, if possible, to get the address from the hardware.


dragonmux (approved):
Given the current state of our Cortex-A support and that this fills a fairly big hole in that support (even if not actually enabled), we're going to merge this so work on implementing more/complete Cortex-A support can continue

PR #432, opened 2019-01-16
stm32l4 improvements to allow better probing and flashing

anyn99:
+separate attach function for this target
+flash accesses moved to attach function. now flashing works again.
+some defines/enum changes.
+save dbgmcu control register on attach, allow debug in all sleep modes,
restore dbgmcu control reg on detach.

### DIFF
--- a/src/target/stm32l4.c
+++ b/src/target/stm32l4.c
@@ -59,7 +59,7 @@ static int stm32l4_flash_write(struct target_flash *f,
 #define FPEC_BASE			0x40022000
 #define FLASH_ACR			(FPEC_BASE+0x00)
 #define FLASH_KEYR			(FPEC_BASE+0x08)
-#define FLASH_OPTKEYR		(FPEC_BASE+0x0c)
+#define FLASH_OPTKEYR			(FPEC_BASE+0x0c)
 #define FLASH_SR			(FPEC_BASE+0x10)
 #define FLASH_CR			(FPEC_BASE+0x14)
 #define FLASH_OPTR			(FPEC_BASE+0x20)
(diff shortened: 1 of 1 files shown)


gsmcmullin:
Please use the Linux kernel coding style: https://www.kernel.org/doc/html/latest/process/coding-style.html

UweBonnes:
I see  also code shuffled around (enum ID_STM32L4) and trailing white space. Otherwise I compiled for libftdi and attach to a Nucleo_L4R without problem.

anyn99:
You are perfectly right about coding style and trailing white space! 
I am not used to adhere to community coding styles, will try to improve! Will fix this pull req!

The DBGMCU register should be taken from libopencm3 anyway i think, the ifndef makes no sense, didn't think about it enough. I was confused by the stm32G0 register there, maybe this target should be moved to stm32L0 target, it seems to be the same architecture?

UweBonnes:
No, the flash on G0 resemble much die Flash on L4. Flash write happens in 64-bit words. Do not rely on the libopencm3 definition. When compiling for libftdi, libopencm3 is not used. Define in the source with a name that does not collide with the names in libopencm3.

anyn99:
Ok

UweBonnes:
This is my opinion. Don't know what Gareth thinks:
Introducing some new feature in a series of patches, where there is some fiddling around with the feature obscures the real changes. Merge your patches into one.

anyn99:
think i got it right now.
Uff git has such a steeep learning curve ;-)

UweBonnes:
II can't find the patch anymore. Is the problem still visible?

anyn99:
Hmm i had some git related difficulties.. then squashed some commits together, maybe that why you dont see the patch? 
Its 7d5a87821749614a574e95cacad07eeda6fe218b

newbrain:
I'm working on adding STM32WB55 support.
ATM, I'm testing option writing and extra commands, as probing and flashing seem OK.
WB55 is very similar to L4, and #432 make interesting changes in the probing/attaching structure.

If this PR is merged, I think the WB55 specific parts can be restricted to a simplified probing (with a couple extra target commands) and minor to medium changes can be made to the attach function proposed here (WB55 has a single bank, 4KiB page flash).

This would also call for the attach function to be be made external rather than static, so it can be referenced by my code (and maybe a more generic name than stm32l4_attach, though I would not know what to suggest).

If the orientation is to go for a merge of #432, I can start re-working the code in this light.

UweBonnes:
Yes, I intend to apply #432

newbrain:
Thanks for the info!
I also agree it's an improvement, so I'll start working on that base straight away.

UweBonnes:
anyn99 : I cleaned up code shuffling and style issues in https://github.com/UweBonnes/blackmagic/tree/#432. If you are okay with the changes, either replace your patch with my changes in your branch with this open pull request or give me a go to push my branch.

newbrain, when working on STM32WB start with https://github.com/UweBonnes/blackmagic/tree/#432

anyn99:
Sure have a go with your branch, im fine with it!

UweBonnes:
Replaced by #445

PR #434, opened 2019-01-18
Nrf5

UweBonnes:


src/target/nrf51.c
```diff
@@ -103,87 +105,40 @@ static void nrf51_add_flash(target *t,
...
-	case 0x0085: /* nRF51422 (rev 3) QFAC A0 */
-	case 0x0086: /* nRF51422 (rev 3) QFAC A1 */
-	case 0x0087: /* nRF51822 (rev 3) CFAC A0 */
-	case 0x0088: /* nRF51422 (rev 3) CFAC A0 */
+	} else {
 		t->driver = "Nordic nRF51";
+		/* Use the biggest RAM size seen in NRF51 fammily.
+		 * IDCODE is kept as '0', as deciphering is hard and
+		 * there is later no usage.*/
 		target_add_ram(t, 0x20000000, 0x8000);
```
miek: This should use `NUMRAMBLOCK` and `SIZERAMBLOCKS` from the FICR on nRF51

src/target/nrf51.c
```diff
@@ -103,87 +105,40 @@ static void nrf51_add_flash(target *t,
 
 bool nrf51_probe(target *t)
```
miek: I think it should be mapping the FICR/UICR (as it did before) - it's really useful to be able to use `print` in gdb to examine these registers.

UweBonnes: FICR is some non-changeable memory area. No need to map it. You can always use examine  "x ..." to print. I re-added UICR, as it might get flashed.

src/target/nrf51.c
gsmcmullin: I am happy for it to use the largest possible ram size, rather than probing for the exact size.

jonas-schievink:
This PR doesn't build for me with clang:

```
command.c:67:14: error: cast between incompatible function types from '_Bool (*)(void)' to '_Bool (*)(target *, int,  const char **)' {aka '_Bool (*)(struct target_s *, int,  const char **)'} [-Werror=cast-function-type]
  {"version", (cmd_handler)cmd_version, "Display firmware version info"},
              ^
command.c:68:11: error: cast between incompatible function types from '_Bool (*)(target *)' {aka '_Bool (*)(struct target_s *)'} to '_Bool (*)(target *, int,  const char **)' {aka '_Bool (*)(struct target_s *, int,  const char **)'} [-Werror=cast-function-type]
  {"help", (cmd_handler)cmd_help, "Display help for monitor commands"},
           ^
command.c:70:16: error: cast between incompatible function types from '_Bool (*)(void)' to '_Bool (*)(target *, int,  const char **)' {aka '_Bool (*)(struct target_s *, int,  const char **)'} [-Werror=cast-function-type]
  {"swdp_scan", (cmd_handler)cmd_swdp_scan, "Scan SW-DP for devices" },
                ^
command.c:71:14: error: cast between incompatible function types from '_Bool (*)(void)' to '_Bool (*)(target *, int,  const char **)' {aka '_Bool (*)(struct target_s *, int,  const char **)'} [-Werror=cast-function-type]
  {"targets", (cmd_handler)cmd_targets, "Display list of available targets" },
              ^
command.c:72:12: error: cast between incompatible function types from '_Bool (*)(void)' to '_Bool (*)(target *, int,  const char **)' {aka '_Bool (*)(struct target_s *, int,  const char **)'} [-Werror=cast-function-type]
  {"morse", (cmd_handler)cmd_morse, "Display morse error message" },
            ^
command.c:75:16: error: cast between incompatible function types from '_Bool (*)(void)' to '_Bool (*)(target *, int,  const char **)' {aka '_Bool (*)(struct target_s *, int,  const char **)'} [-Werror=cast-function-type]
  {"hard_srst", (cmd_handler)cmd_hard_srst, "Force a pulse on the hard SRST line - disconnects target" },
                ^
```

This is with clang version 5.0.2.

Rebasing on master fixes it.

UweBonnes:
So after rebase, did it work as expected?

jonas-schievink:
Unfortunately I could neither connect to an nRF52810, nor to an nRF51822 with this branch (both fail with "SW-DP scan failed!"). I might be doing something wrong though.

chocol4te:
I can scan, attach and load on my nRF52810, but I am getting segmentation faults when I try to run my program, even though I know it works fine using OpenOCD+STLink

EeuweQ:
With these changes the NRF52840 works correctly!

UweBonnes:
Negative remarks on this branch seen to be other errors.

Issue #436, opened 2019-01-18
EFM32 driver doesn't support recovering locked parts 


richardeoin:
Parts are locked by clearing the Debug Lock Word (DLW)

perigoso:
should be closed as per
5943552a6b1f54ab9add80a87f64945db00b7769
and
#567

perigoso:
thats because I deleted it, this one is actually not implemetned (doesnt look like it to me at least), sorry for the confusion

dragonmux:
Oh, huh - indeed, the linked PR doesn't look to fix this. Re-opening till we can otherwise prove it's implemented

perigoso:
according to Silabs, the correct method to clear the debug lock word is to issue a mass erase through the AAP, which is in fact implemented by `efm32_aap_cmd_device_erase`, so this too should be closed.

dragonmux:
Thank you for checking this. Closing.

Issue #435, opened 2019-01-18
EFM32 driver doesn't support recent parts PG12B, JG12B, GG11B, TG11B, GG12B


richardeoin:

perigoso:
This should be closed as per
e85df763b78c179543672902949e871e42e705be
and
#567

dragonmux:
Thanks for spotting that Perigoso!

PR #438, opened 2019-01-18
Additional EFM32 support

richardeoin:
Add recent parts PG12B, JG12B, GG11B, TG11B, GG12B (Issue [#435](https://github.com/blacksphere/blackmagic/issues/435))

Support recovering locked parts (Issue [#436](https://github.com/blacksphere/blackmagic/issues/436))


UweBonnes:
Adding all variants may be a tedious job. Is there perhaps a more general way to recognize ESP32 chips?

richardeoin:
This pull adds some new *families*, not variants. Tracking all variants across this product line would indeed be a sisyphean task - there are 430 distinct parts in stock at DigiKey right now, and many hundreds more available from factory stock.

Unfortunately we can't check the TARGETID register. The DP is a DPv1, so there is no TARGETID register.

Silabs change the MSC address, MSC layout, user data region size and bootloader region size between families. We need to track these in order to write to flash correctly.

UweBonnes:
Did anybody work with this patch?

UweBonnes:
Rebased and applied as #567

PR #442, opened 2019-02-14
Added ability to lock and unlock boot rom for samd controllers

r0ck3tAKATrashPanda:
Just added commands to lock and unlock the boot rom for SAMD microcontrollers. This is useful for working with the Feather M0 since the bootloader section in flash is locked.


UweBonnes:
The changes are well confined and I see no obvious style problems.

PR #450, opened 2019-02-25
Improve parsing of commands that require enable or disable arguments

jelson:


UweBonnes:
Did you scan the code for similar places?

jelson:
Yes, although now that I'm thinking about it, I'd like to modify the PR before you merge it so that it explicitly scans for either "enable" or "disable" and returns an error if neither is recognized.

The first couple weeks I used BMP I got burned over and over by typing "mon conn ena" to enable connect-under-reset - the BMP would silently accept this command and then not actually enable the feature.

jelson:
The PR has been updated and improved to ensure 
* All prefixes of the words 'enable' and 'disable' are accepted
* Any parse failures are reported, rather than silently ignored
* Status is printed after any status change
* All commands will either confirm the flag change or report a parse error

UweBonnes:
To answer my question "Did you scan the code for similar places?" myself:

 > grep -r '"en'
...
 t: arget/nxpke04.c                        unsafe_enabled ? "enabled" : "disabled");
 target/kinetis.c:                         unsafe_enabled ? "enabled" : "disabled");
 
I think you should export parse_enable_or_disable. Remove also duplicated empty lines.  I do not have a decive answer if you should also use that function in the other places or if we should hope that the original authors will pick them up.

jelson:
I found yet another one in cortexm.c searching only for 'e' instead of 'en'. I will upload a new version of the PR.

jelson:
OK, the currently-pushed commit should now use the new parser function everywhere in the codebase that 'enable' or 'disable' is parsed.

PR #451, opened 2019-02-25
Fix missing makefile dependency

jelson:
Add blackmagic.bin as dependency of bindata.o so that running make
in upgrade will correctly rebuild the upgrade tool if the main binary
has changed.

### DIFF
--- a/upgrade/Makefile
+++ b/upgrade/Makefile
@@ -12,6 +12,8 @@ OBJ =	bindata.o	\
 
 all: $(OUTFILE)
 
+bindata.o: ../src/blackmagic.bin
+
 $(OUTFILE) $(OUTFILE).exe: $(OBJ)
 	$(CC) $(CFLAGS) -o $@ $^ $(LDFLAGS)
 


jelson:
Just wanted to ping this, the test failure seems to be a framework failure and not a problem with the code

UweBonnes:
Shouldn't we deprecate "upgrade" versus stm32mem.py?

jelson:
I am not familiar with that script, I was just following the [Hacking page update instructions](https://github.com/blacksphere/blackmagic/wiki/Hacking#updating-firmware) which say to type "cd upgrade; make". 

If those instructions are still current, this simple PR that fixes the dependencies should be merged. If that is no longer the suggested upgrade procedure, the docs should be updated and the 'upgrade' directory deleted.

UweBonnes:
Updated the update instructions. stm32mem.py has seen much more care and should probably be easier. For me make in upgrade shows> env LANG=C make
gcc -Wall -Wextra -std=gnu99 -O0 -g -MD -mno-ms-bitfields   -c -o dfu.o dfu.c
dfu.c:24:13: fatal error: usb.h: No such file or directory
Should we still apply #451

jelson:
Thanks for updating the docs. I think it makes sense either to apply 451 or to delete the upgrade directory. It seems dangerous to have code around that is broken.

UweBonnes:
Did upgrade reallu work for you? For me on a F3 Disco board, it overwrote the bootloader. https://github.com/UweBonnes/blackmagic/tree/451 has your fix and made blackmagic_upgrade work fot me. Can you please test and report?

UweBonnes:
Fixed by #495

PR #452, opened 2019-02-28
Add compile-time-configurable optional delay to slow down the SWD clock.

jelson:
In some situations we want to slow down the SWD clock, e.g. long wires connecting the probe to the target. This PR allows the SWD clock to be slowed down. It uses compile-time configuration so there is no performance penalty to people who want the maximum-speed clock; the DELAY() function is eliminated at run-time unless NUM_DELAY_TICKS is greater than zero.

### DIFF
--- a/src/platforms/common/swdptap.c
+++ b/src/platforms/common/swdptap.c
@@ -23,6 +23,15 @@
 #include "general.h"
 #include "swdptap.h"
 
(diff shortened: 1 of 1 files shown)


UweBonnes:
Please also start and leave with CLK high as discussed earlier as discussed in #392

UweBonnes:
Superseded By #783

Issue #455, opened 2019-03-02
PR #445 introduces attach-detach fault


newbrain:
Pull request #445 introduces a serious problem when repeatedly attaching and detaching from a target.
As the memory information is stored again in the attach function, gdb will see multiple overlapping areas and ignore them.

Software used:
- latest master (da62cba), stlink platform
- Windows 10 1809
- arm-none-eabi-gdb version 8.01

Hardware:
- STM32l476 Nucleo as target
- STM32WB55 dongle (on my dev branch)
- ST-Link V2.1 taken from another Nucleo.

Log:
```
PS C:\Users\newbrain> C:\SysGCC\arm-eabi\bin\arm-eabi-gdb.exe
GNU gdb (GDB) 8.0.1
Copyright (C) 2017 Free Software Foundation, Inc.
[----------8<--snip some gdb messages--8<---------]
This GDB was configured as "--host=i686-w64-mingw32 --target=arm-eabi".
[----------8<--snip more gdb messages--8<---------]
(gdb) target extended-remote COM4
Remote debugging using COM4
(gdb) monitor swdp_scan
Target voltage: unknown
Available Targets:
No. Att Driver
 1      STM32L47x
(gdb) attach 1
Attaching to Remote target
warning: No executable has been specified and target does not support
determining executable automatically.  Try using the "file" command.
0xfffffffe in ?? ()
(gdb) info mem
Using memory regions provided by the target.
Num Enb Low Addr   High Addr  Attrs
0   y   0x08000000 0x08080000 flash blocksize 0x800 nocache
1   y   0x08080000 0x08100000 flash blocksize 0x800 nocache
2   y   0x10000000 0x10008000 rw nocache
3   y   0x20000000 0x20018000 rw nocache
(gdb) detach
Detaching from program: , Remote target
(gdb) attach 1
Attaching to Remote target
warning: No executable has been specified and target does not support
determining executable automatically.  Try using the "file" command.
warning: Overlapping regions in memory map: ignoring          <=== Here is the rub
0xfffffffe in ?? ()
(gdb) info mem
Using memory regions provided by the target.
There are no memory regions defined.
(gdb)
```

The problem is not limited to an explicit attach detach sequence, also interrupting a running program with Ctrl-C has the same outcome:
```
(gdb) file c:/Users/newbrain/Git/l476BMtest/VisualGDB/Debug/l476BMtest
A program is being debugged already.
Are you sure you want to change the file? (y or n) y
[...]

PR #456, opened 2019-03-02
Correction of #445 attach-detach problem

newbrain:
This commit should solve issue #455, introduced by PR #445.

Main changes, all in stm32l4.c:
- Both RAM and Flash are added only in the attach function, not in the probe one.
- The memory map is freed with `target_mem_map_free()` before adding the new map.
- To avoid proliferation of `switch{}` statements and slightly reduce `if` complexity, a new info data structure has been introduced. This is also propaedeutic to adding new similar targets without large changes in the code.
- Minor editing

Tested with STM32L476 Nucleo, Windows 10, gdb 8.0.1, stlink platform.

Especially the third point, will allow me to finally add the WB series without the need of a separate probe function (and additional source files, let's rework it once again).
BTW, the RM has now been published: [STM32WB55xx RM](https://www.st.com/content/ccc/resource/technical/document/reference_manual/group0/83/cf/94/7a/35/a9/43/58/DM00318631/files/DM00318631.pdf/jcr:content/translations/en.DM00318631.pdf) and also the Cube tools are available.

### DIFF
--- a/src/target/stm32l4.c
+++ b/src/target/stm32l4.c
@@ -124,6 +124,108 @@ struct stm32l4_flash {
 	uint32_t bank1_start;
 };
 
+enum ID_STM32L4 {
+	ID_STM32L41  = 0x464u, /* RM0394, Rev.4 */
+	ID_STM32L43  = 0x435u, /* RM0394, Rev.4 */
+	ID_STM32L45  = 0x462u, /* RM0394, Rev.4 */
+	ID_STM32L47  = 0x415u, /* RM0351, Rev.5 */
+	ID_STM32L49  = 0x461u, /* RM0351, Rev.5 */
+	ID_STM32L4R  = 0x470u, /* RM0432, Rev.5 */
+	ID_STM32G07  = 0x460u, /* RM0444/454, Rev.1 */
+};
+
+enum FAM_STM32L4 {
+	FAM_STM32L4xx = 1,
+	FAM_STM32L4Rx = 2,
+	FAM_STM32G0x = 3,
+	FAM_STM32WBxx = 4,
+};
+
+#define DUAL_BANK	0x80u
+#define RAM_COUNT_MSK	0x07u
+
+struct stm32l4_info {
+	char designator[10];
+	uint16_t sram1;
+	uint16_t sram2;
+	uint16_t sram3;
+	enum ID_STM32L4 idcode;
+	enum FAM_STM32L4 family;
+	uint8_t flags;
+};
+
+struct stm32l4_info const L4info[] = {
+	{
+		.idcode = ID_STM32L41,
+		.family = FAM_STM32L4xx,
+		.designator = "STM32L41x",
+		.sram1 = 32,
+		.sram2 = 8,
+		.flags = 2,
+	},
+	{
+		.idcode = ID_STM32L43,
+		.family = FAM_STM32L4xx,
+		.designator = "STM32L43x",
+		.sram1 = 48,
+		.sram2 = 16,
+		.flags = 2,
+	},
+	{
+		.idcode = ID_STM32L45,
+		.family = FAM_STM32L4xx,
+		.designator = "STM32L45x",
+		.sram1 = 128,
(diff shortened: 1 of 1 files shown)


UweBonnes:
Can you please merge into one patch and diminish unrelated whitespace changes?

src/target/stm32l4.c
```diff
@@ -141,61 +243,62 @@ static void stm32l4_add_flash(target *t,
...
-		break;
-	case ID_STM32G07:
-		idcodereg = STM32G0_DBGMCU_IDCODE_PHYS;
-		break;
-	}
+	/* Retrive chip information, no need to check return */
+	struct stm32l4_info const *chip = stm32l4_get_chip_info(t->idcode);
+
+
+	uint32_t idcodereg = (chip->family == FAM_STM32G0x)
```
UweBonnes: This looks strange! You check for G0 and select something else,

src/target/stm32l4.c
```diff
@@ -141,61 +243,62 @@ static void stm32l4_add_flash(target *t,
...
+	} else {
+		target_add_ram(t, 0x10000000, chip->sram2 << 10);
+		/* All L4 beside L47 alias SRAM2 after SRAM1.*/
+		uint32_t ramsize = (t->idcode == ID_STM32L47)?
+			chip->sram1 : (chip->sram1 + chip->sram2 + chip->sram3);
+		target_add_ram(t, 0x20000000, ramsize << 10);
+	}
+
+	/* Add the flash to memory map. */
+	/* TODO: This should really be more data driven... */
```
UweBonnes: What do you mean bt more data driven?

src/target/stm32l4.c
```diff
@@ -124,6 +124,108 @@ struct stm32l4_flash {
...
+	FAM_STM32L4Rx = 2,
+	FAM_STM32G0x = 3,
+	FAM_STM32WBxx = 4,
+};
+
+#define DUAL_BANK	0x80u
+#define RAM_COUNT_MSK	0x07u
+
+struct stm32l4_info {
+	char *designator;
```
UweBonnes: The designator has constant length. In my understanding, char designator[10] will save us a pointer for every entry. Will fill up flash fast...

UweBonnes:
Please comment on my remarks.

src/target/stm32l4.c
newbrain: Expletives deleted.
Definitely my fault. The true and false branch should be reversed.

src/target/stm32l4.c
newbrain: I really don't like the jumble of nested ifs, I would prefer something that takes the right action with fewer conditions (at the expense of more data, yes).
Maybe my working with the maintenance of SW with a Mc Cabe complexity of 183 has tainted me...
I've seen other todos here and there, but if you think this is not worth the reminder, I'll remove it.

src/target/stm32l4.c
newbrain: Yes, very good observation.
I don't think there'll be a designator longer than 9 char, so I agree.
More or less 3 bytes per entry saved, no change to the code.

newbrain:
Shall I provide a new commit (squashed with the previous one, so to only show one)?

src/target/stm32l4.c
newbrain: Note that my solution was already less expensive than the original (+data -code)!
```
Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----       2019-03-03     14:42          72284 blackmagic.bin <- master branch
-a----       2019-03-03     14:40          72192 blackmagic.nb  <- newbrain, char*
-a----       2019-03-03     14:38          72228 blackmagic.old <- newbrain char[10]
```

newbrain:
Well, I went ahead and did it.
Eager to hear more comments, if any!

UweBonnes:
Perhaps explain after  the commit message what was wrong, something like: With separate attach function, construction of memory map must happen during attach.

PR #460, opened 2019-03-18
Add new platform: Black Pill

skotopes:


UweBonnes:
Did you try to integrate the blackpill into one of the existing platforms?

skotopes:
> Did you try to integrate the blackpill into one of the existing platforms?

I've been checking possibility of using stlink(v1) code, but haven't found simple way to distinguish this board from stlink v1.

UweBonnes:
What about requiring to put some string into flash?

skotopes:
> What about requiring to put some string into flash?

String? Didn't saw anything in documentation. Do I miss something?

UweBonnes:
Yes, some creative ideas.

skotopes:
> Yes, some creative ideas.

Why make things complex if they can be simple? Shared code can be moved outside platform implementation, like it's already made with stm32 folder. Less code, less rom/ram usage, easier to document and support.

UweBonnes:
This is a bluepill board with LED on PB12 instead of PC13. Bluepill is handled with swlink platform. Add a compile time option e.g. ENABLE_BLACKPILL=1 and use PB12 instead of PB13 for the LED and you are done.

skotopes:
> This is a bluepill board with LED on PB12 instead of PC13. Bluepill is handled with swlink platform. Add a compile time option e.g. ENABLE_BLACKPILL=1 and use PB12 instead of PB13 for the LED and you are done.

My bad, didn't saw blue pill mentions in context of swlink.

Couple questions:

- How about ADC that's not usable on Pill boards, should it be left as it is for STM8S UM0817?
- Also UART2 pins are not 5v tolerant, should it be used instead of UART3?
- Also what to do with pinout for each board: should it be convenient or unified?

PS: Looks like many online manual using stlink platform as a base for blue pill, I guess current set of documentation and platforms implementations is not obvious. So maybe separating platforms from each other is a not bad idea? Or I miss something? Also it will save some memory.

UweBonnes:
Regarding ADC: There is a target voltage detection implemented in swlink.
Regarding uart2/3: bmp/swlink/bluepill was trying to be compatible with some exixting bluepill bmp forks
Perhaps add another compile time option to use usart3 vs usart2.
Separated platform tend to bitrot even more than "ifdef"ed code.

UweBonnes:
Please review https://github.com/UweBonnes/blackmagic/tree/blackpill

UweBonnes:
Feedback needed!

PR #462, opened 2019-03-24
Removed redundant "-f" options from main Makefile and native/Makefile…

sidprice:
GNU make defines the "RM" variable as "rm -f". In the two edited files there is an extra "-f" for each usage of "RM".

The extra "-f" does not bother Linux builds, however, while working towards a Windows build the "RM" variable needs change and the added "-f" causes "clean" to fail.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -86,8 +86,8 @@ blackmagic: include/version.h $(OBJ)
 
 clean:	host_clean
 	$(Q)echo "  CLEAN"
-	-$(Q)$(RM) -f *.o *.d *~ blackmagic $(HOSTFILES)
-	-$(Q)$(RM) -f platforms/*/*.o platforms/*/*.d mapfile include/version.h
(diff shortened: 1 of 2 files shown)


UweBonnes:
Sid, add your backtick substitution as another commit. I will try tomorrow. Our other way of code exchange might have introduced errors.

sidprice:
Uwe, I will try to do that. I am still trying to get the “clean” to work. PowerShell is a “bear”, very challenging. 

Do you want the backtick change as a pull request onto your blackstlink branch?

sidprice:
Uwe, I will try to do that. I am still trying to get the “clean” to work. PowerShell is a “bear”, very challenging.

Do you want the backtick change as a pull request onto your blackstlink branch?

UweBonnes:
Sending a patch in private mail would be best.

sidprice:
I have changed these two makefiles because the rm command in PowerShell just cannot be made to work in them. So, I am going to close this PR and submit a new one with the approach that I have working now.

PR #463, opened 2019-03-28
Native Windows builds

sidprice:
THese changes allow for the building of the Black Magic Probe for all the current platforms, except the launchpad-icdi, on a Windows development machine. This is done without the need for Cygwin, using only PowerShell.

Note that the  "all_platforms" target is not tested and will not work. If the launchpad-icdi host is fixed then "all_platforms" may be added.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -86,8 +86,12 @@ blackmagic: include/version.h $(OBJ)
 
 clean:	host_clean
 	$(Q)echo "  CLEAN"
-	-$(Q)$(RM) -f *.o *.d *~ blackmagic $(HOSTFILES)
-	-$(Q)$(RM) -f platforms/*/*.o platforms/*/*.d mapfile include/version.h
+ifeq ($(OS), Windows_NT)
+	-$(Q)$(shell del *.o, *.d, blackmagic*, $(HOSTFILES))
+else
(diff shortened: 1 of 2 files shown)


sidprice:
Any chance of getting this pull request merged, I would like to continue along the path I am on and not having these changes make that difficult.

UweBonnes:
Please correct the commit message. Otherwise I still wonder why redefining RM for the Windoes_NT case does not work and you need to change all invokations. Did you investigate?

sidprice:
Yes I did try the suggestion you made about redefining RM, it does not work, with only a "file not found" error being emitted. I now remember that was the solution I tried originally, but I had tried so many things I did not remember why it was rejected.

While my solution is a little more intrusive, it works well, so, having it merged would be excellent.

UweBonnes:
Commit message still " bjuilds"

sidprice:
Should be fixed now

UweBonnes:
Applied. I still had hoped for an explanation why the less intrusive approach did not work ...

sidprice:
Uwe, I too wish I knew why it fails too, however, sometimes we just have to accept what we have and move forward. Not my favorite way of working, but in this case I think it is the right way to go.

PR #464, opened 2019-03-30
Make the detection of protected samd working again

zardam:
Probing of protected samd21 was failing, due to entering the STM32L0 cause, and stopping probing there.

### DIFF
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -318,16 +318,24 @@ bool cortexm_probe(ADIv5_AP_t *ap, bool forced)
 		target_check_error(t);
 	}
 
+#define PROBE(x) \
+	do { if ((x)(t)) {target_halt_resume(t, 0); return true;} else target_check_error(t); } while (0)
+
 	/* Only force halt if read ROM Table failed and there is no DPv2
 	 * targetid!
 	 * So long, only STM32L0 is expected to enter this cause.
 	 */
 	if (forced && !ap->dp->targetid)
-		if (!cortexm_forced_halt(t))
+		if (!cortexm_forced_halt(t)) {
+			/* Protected Atmel SAM D also enters this clause, so probing for it is done here.
+			 * target_check_error is needed to clear pending error, then samd_probe checks
+			 * for the right manufacturer/part number in CoreSight ROM first and exits if
+			 * the correct part is not detected.
+			 */
+			target_check_error(t);
+			PROBE(samd_probe);
 			return false;
-
-#define PROBE(x) \
-	do { if ((x)(t)) {target_halt_resume(t, 0); return true;} else target_check_error(t); } while (0)
+		}
 
 	PROBE(stm32f1_probe);
 	PROBE(stm32f4_probe);


UweBonnes:
The patch has unrelated whitespace changes (#define probe), obstruvting the real change. Please rework and commit --amend and add reference to PR so I can merge in a single commit.

zardam:
I have done the modifications. Do not hesitate if a problem remains.

UweBonnes:
You still move the definition of probe around! I do not see any change that this move affects. So keep the definition in place, only add relevant changes!

zardam:
PROBE definition is raised, because it is called instead of the "return false" (just before its original definition).

UweBonnes:
Please experiment and document the behavior of the SAMD with different setups: Protected and not protected, erased and not erased, with normal firmware, firmware that sleeps deep prolonged and firmware that remaps the SWD pins. This results in a lot of possible combinations, but perhaps after some test you can give already some resumee.

Your patch only cures one problem, but not the cause. Can you try #454. This seems to be a better approach, also at the moment it is too intrusive, leaving the CPU in reset/halt when only probing.

zardam:
From what I understand from the [datasheet](https://cdn.sparkfun.com/datasheets/Dev/Arduino/Boards/Atmel-42181-SAM-D21_Datasheet.pdf) :
- when protected, hot plugging of a debugger is forbidden (p55)
- the CPU is held in reset state after a reset and when SWCLK is low (p52). This is the only way to connect a debugger when the device is protected (and you cannot resume the CPU execution when connected).
- probing under reset is not supported (p52)

So, from my point of view, #454 is not related to this problem, and I'm not sure that testing all this cases is useful, as the only way to connect the debugger is to do it right after a reset (but not under reset).

This case is specifically handled here : https://github.com/blacksphere/blackmagic/blob/21434d6fbb6ded21b1e7870c52c24488069ab705/src/target/samd.c#L268-L277

And I think the test I modified is failing, because the CPU is held reset by the "Device Service Unit" :
https://github.com/blacksphere/blackmagic/blob/21434d6fbb6ded21b1e7870c52c24488069ab705/src/target/cortexm.c#L330-L336

Maybe there is an other way to get the samd specific probe called when the device is protected ? I tried to check if there is some specific information available during this test to detect a protected samd, but I found nothing useful when the test is done (apart from forcing a samd probe).

UweBonnes:
Can you have a look if the device identification (12.10, p 55) is visible with the protected device? This is either done by recompiling BMP with ENABLE_DEBUG=1, enabling debug output "mon debug_bmp enable" and looking at the output of the serial console or by using the libftdi platform or the stlinkv2 platform (https://github.com/UweBonnes/blackmagic/tree/stlinkv2) with output on the console where the BMP PC program is started. That way, samd probe can perhaps be started more specific.

zardam:
I have the following debug output when probing a protected samd21: 
```
 AP   0: IDR=04770031 CFG=00000000 BASE=41003003 CSW=03000040

ROM: Table BASE=0x41003000 SYSMEM=0x0
0 Entry 0x9f0fc002 -> Not present
1 Entry 0x3002 -> Not present
ROM: Table END

-> cortexm_probe forced
```
From reading the probe code of samd target, the identification must be present, as it is checked before deciding if the device is protected:
https://github.com/blacksphere/blackmagic/blob/21434d6fbb6ded21b1e7870c52c24488069ab705/src/target/samd.c#L368-L389

So, event if the device is protected, samd probe will only success if a samd is present.

zardam:
I have updated the pull request. The "!cortexm_forced_halt(t)" test was never returning false, and in fact only the second call to samd_probe was returning true. The first call failed because errors where not cleared (I added a call to target_check_error just before probing).

UweBonnes:
Please add some comments/explanation in the code. This will help to keep this exception when sometimes probe restructuring happens. I am available today and then next week again.

zardam:
I have updated the pull request.

UweBonnes:
Can you test with #466?

zardam:
I tried https://github.com/UweBonnes/blackmagic/tree/stlinkv2

SAMD21 is not detected, either protected or unprotected. I have this debug ouput (if it may help) :
```
AP   0: IDR=04770031 CFG=00000000 BASE=41003003 CSW=03000040
Halt failed after 2100 ms
```

UweBonnes:
Did you try with "monitor  connect_srst enable"? This will try a cold plug. Without that, hot-plug is done at that is expected to fail according to the docs.

zardam:
Sorry, but still not working. I'm only using two wires for SWD interface (no reset line), so I think connect_srst does nothing, but I still tried before a reset, during a reset and after a reset.

Before and after a reset, I have this debug output with a protected samd and "monitor connect_srst enable"
```
AP   0: IDR=04770031 CFG=00000000 BASE=41003003 CSW=03000040
DEMCR 0-> 0
Halt failed after 2100 ms
```
and with an unprotected one:
```
AP   0: IDR=04770031 CFG=00000000 BASE=41003003 CSW=03000040
DEMCR 0-> 1000401
Halt failed after 2100 ms
```

During a reset, the scan fails quickly with no debug output.

UweBonnes:
Sorry, I lost track. As the datasheet says: "when protected, hot plugging of a debugger is forbidden (p55)". So how can it work wothout the reset line?
And:
- did it work before in some version or with your patches?

zardam:
I'm manually resetting the board while the probe is connected. After a reset, the CPU is held in "extended reset" if a debugger is connected, allowing it to be probed. See datasheet section "12.6.2 CPU Reset Extension".

With a stock blackmagic firmware, probing of a protected device reports only a Cortex M, not a samd21 (and is unable to mass erase the device to unprotect it). I did not try previous versions, but I confirm that probing a protected samd21 is working with my branch.

UweBonnes:
should be fixed with #511

UweBonnes:
@Zardam: Does it still work on git head? Thanks!

zardam:
I tested on the head of master branch (f9414d58265093bf8a4ae59d45c3d1ac441d6ed2). Probing a protected samd fails.

I have this debug output from the probe:

```DPIDR 0x0bc11477 (v1 MINDP rev0)
RESET_SEQ failed
AP   0: IDR=04770031 CFG=00000000 BASE=41003003 CSW=03000040 (AHB-AP var3 rev0
Halt via DHCSR: Failure DHCSR 00000000 after  2000ms
Try again, evt. with longer timeout or connect under reset
```

UweBonnes:
Please state (perhaps again) what SAMD device you use and the steps needed to protect the software in the way you see the problem, best done with BMP monitor commands. What BMP variant do you use?

zardam:
The SAMD board is an arduino zero clone (Wemos D1 or something like that) with a samd21. I'm locking the device with "monitor set_security_bit", then resetting it.

My probe is a bluepill (swlink), using swd interface.

With my old branch, unlocking the device requires connecting the probe, then manually resetting the board. Unlock is done with "monitor erase_mass"

PR #472, opened 2019-05-17
This order change is needed for STM32F103C8 to be treated as F103Cb

mryndzionek:
This order change is needed for STM32F103C8 to be treated as F103Cb in DFU, because dfu-util doesn't take into account the 'force' option for writes


UweBonnes:
This fakes an identification string. I think this is something fundamental  different than to try to ignore some identification string and test if the device can act different as scripts/stm32mem.py does. Closing. Explain and reopen if you have strong reasons for your fake.

mryndzionek:
I thought that it's not a big deal if we already have `/* Handle F103x8 as F103xC! */`. This is the only way I can get original dfu-util to write an image bigger than 64K to STM32F103C8-based boards. I'll look at scripts/stm32mem.py to see if I can handle this there. Thanks!

PR #474, opened 2019-05-21
[adiv5] Improvements in ADIv5

richardeoin:
* Reference latest version of the ARM specification
* ROM tables - more debug information, including printing SYSMEM bit
* MEM-AP - reject when Debug Base Address entry is not
  present/invalid. These would only have errored in
  adiv5_component_probe.
* Resolve note in STM32H7 driver with explaination

### DIFF
--- a/src/target/adiv5.c
+++ b/src/target/adiv5.c
@@ -18,8 +18,10 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-/* This file implements the transport generic functions of the
- * ARM Debug Interface v5 Architecure Specification, ARM doc IHI0031A.
+/* This file implements the transport generic functions.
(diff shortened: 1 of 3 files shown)


UweBonnes:
Please have a look at the conflicts and ann PR number.

UweBonnes:
Can you explain, why now  adiv5_component_probe probes from 0 to 960 and only 256 before?

richardeoin:
Please see
ARM Debug Interface v5 Architecture Specification **ARM IHI 0031E**
**Table D3-1** ROM Table register summary

Not sure why it was 256 before. Perhaps a (small) oversight.

UweBonnes:
What about adding a reference in the patch?

richardeoin:
That's a good idea. I didn't see it when writing the original commit

PR #475, opened 2019-05-21
Generate DEBUG warnings and return if `malloc`/`calloc` fail.

richardeoin:
This is will make debugging earier if this does happen, rather than
dereferencing the null pointer (or passing it to memcpy, or worse).

### DIFF
--- a/src/target/adiv5.c
+++ b/src/target/adiv5.c
@@ -372,6 +372,11 @@ ADIv5_AP_t *adiv5_new_ap(ADIv5_DP_t *dp, uint8_t apsel)
 
 	/* It's valid to so create a heap copy */
 	ap = malloc(sizeof(*ap));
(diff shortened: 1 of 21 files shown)


UweBonnes:
It seem you still work on the patch. I will have a look the next days.

richardeoin:
The CI caught that `__func__` can't be stringified by mainline gcc (although arm-none-eabi-gcc can?!). Fixed now and ready for review.

UweBonnes:
Please amend the PR number so I can rebase and merge to have only one commit.

UweBonnes:
git log --check  shows whitspace errors.

UweBonnes:
There are still whitespace errors. Run "git rebase --whitespace=fix HEAD~1", amend and force push.

richardeoin:
Ok, cleaned up the extra spaces. Thanks for the neat command

PR #476, opened 2019-05-21
[stm32h7] add revision command which reads the `DBGMCU_IDC` register

richardeoin:


UweBonnes:
Could you please amend  and change the commit message to contain the PR number and force push. That way with "rebase and merge" we only get one commit. This applies for your other PRs too, if the only contain one commit.

richardeoin:
I've added "blackmagic PR #476" to the commit message

PR #477, opened 2019-05-22
gdb_main: Always release CPU when detaching.

### LINKED ISSUE #413: hang after swd upload on swlink
I'm running into an issue with the swlink build (latest). Context:

* h/w is blue pill
* 6 wires to target: gnd, 5v, swdio, swclk, rx/tx (b6/b7)
* target is either another blue pill, or an stm32f407 board
* using macos and platformio (both latest)

Upload (via platformio, set to blackmagic) works.
Serial console debug als works fine.

But after each upload, I have to _unpower_ the bmp + target to get the uploaded code to start. A plain reset of either bmp or target is not enough - it seems there is some sort of electrical latchup?

Any tips in how to chase this down further would be helpful.

UweBonnes:
Fixes #413, was PR #477

### DIFF
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -227,8 +227,10 @@ int gdb_main_loop(struct target_controller *tc, bool in_syscall)
 
 		case 0x04:
 		case 'D':	/* GDB 'detach' command. */
-			if(cur_target)
+			if(cur_target) {
+				SET_RUN_STATE(1);
 				target_detach(cur_target);
+			}
 			last_target = cur_target;
 			cur_target = NULL;
 			gdb_putpacketz("OK");


markrages:
I'm having trouble following this patch.

Commit 703f88a only makes the "run" LED blink after detaching, no?  The only effect of `SET_RUN_STATE()` is to update `running_status`, which value is only used once to blink the LED in `timing_stm32.c`.

Does this patch do anything else?

UweBonnes:
Yes, this patch does _not_ do do what it announces. It only changes the behaviour of the LED in some way, where I even do not know if this is intended way. Otherwise it seems to do no other collateral damage, so I think there is no need to revert. Do you have another opinion?

markrages:
OK.  I just updated firmware and I noticed that "detach" was blinking the LED after I was dione debugging, but "killl" was not.

I the behavior of the "run" LED documented anywhere?  If not, it is hard to call it a bug.

PR #484, opened 2019-06-06
Improve Flashsize_103 documentation

### LINKED ISSUE #471: Clarify Flashsize_f103
Since using bluepills seems to be a popular issue, and they will not be receiving a separate platform as per https://github.com/blacksphere/blackmagic/pull/354, I'd like to make some clarifications to:
https://github.com/blacksphere/blackmagic/blob/master/src/platforms/stlink/Flashsize_F103
Which documents how to force larger flash sizes onto the F103C8T6

The objectives and implementation on this page is not very clear. 
```
script/stm32_mem.py does not care for the announced size:
 > ../scripts/stm32_mem.py blackmagic.bin
 ...
 USB Device Firmware Upgrade - Host Utility -- version 1.2
 ...
 Programming memory at 0x08010800
 All operations complete!

Get length of binary
  > ls -l blackmagic.bin
 -rwxr-xr-x 1 bon users 59712 21. Sep 22:47 blackmagic.bin
Actual file size may differ!

Upload binary from flash with the exact size
  > dfu-util -s 0x08002000:leave:force:59712 -U blackmagic.bin.1

Compare
  > diff blackmagic.bin*
No differences should get reported!
```

What is the distinction between the python script and the later use of dfu-util? What is the blackmagic.bin.1 file and how is it generated? I think this page would greatly benefit from a short paragraph describing purpose of the steps and why they occur in this order. 

Lastly, I notice this file is in the st-link, as opposed to swlink which was the target of https://github.com/blacksphere/blackmagic/pull/395. Why is this? 

I'm writing an article on using Black Magic on the bluepill/other targets, so I'm hoping to clear this up. There are a LOT of old article/tutorials out there that simply don't work because they assume outdated smaller flash sizes and I'd like to clean up that misinformation.

hierophect:
Clarifies the steps required for upload to the upper flash of the F103C8, which is a common concern for Bluepill users. 

Resolves #471

### DIFF
--- a/src/platforms/stlink/Flashsize_F103
+++ b/src/platforms/stlink/Flashsize_F103
@@ -2,52 +2,63 @@ Announced versus available Flash size on F103
 ============================================
 Up to Stlink V2, the CPU soldered on the board was a F103C8 with 64 kiByte
 flash. Up to about version 280 of BMP, this limit was not hit when linked
-against nanolib.
-
-StlinkV2-1 has a STM32F103CB, like a genuine BMP.
+against nanolib. 
 
 However with more and more devices supported, BMP at about version 282 hit
-this limit. There are two ways to work around:
-- Branch STlink V2-1 as separate platform and and care for the original STlink
-  platform by restricting/ommitting features.
+the 64kb flash limit. It will still fit on a StlinkV2-1, which uses a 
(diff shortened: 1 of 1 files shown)


UweBonnes:
Should be fixed with #511

UweBonnes:
PR #507 should fix it. Please test.

PR #487, opened 2019-06-24
fix overwriting of still-needed value

dpiegdon:
this fixes a bug that was introduced in blackmagic PR #475 which lead to firmware crashes when connecting to a BMP more than once without a power-cycle.

### DIFF
--- a/src/target/target.c
+++ b/src/target/target.c
@@ -110,8 +110,9 @@ void target_add_commands(target *t, const struct command_s *cmds, const char *na
 	}
 
 	if (t->commands) {
-		for (tc = t->commands; tc->next; tc = tc->next);
-		tc->next = tc;
+		struct target_command_s *tail;
+		for (tail = t->commands; tail->next; tail = tail->next);
+		tail->next = tc;
 	} else {
 		t->commands = tc;
 	}


UweBonnes:
What about going back to the old structure before #475 and checking both mallocs? Please also insert #487 in your commit message so that I can merge as a single commit.

dpiegdon:
the old structure makes error checking and exit overly compilcated (i.e. duplicate). allocing and checking in the beginning is a good idea. but i suppose the code could be made much less confusing.
how about this one?
also in the long term it might be useful to pick an allocator (or wrap it) that asserts the allocation. unless there are situations where malloc may explicitly fail, of course.

UweBonnes:
"How about this one?" makes me ask: Did you test the changed patch and did it fix the error?  It looks good, as it changed less. Sorry for this repeated question.

dpiegdon:
Yes of course, the changes have been tested both times, work properly and fix the error.

Also, you wrote in gitter: 
> I am running blackmagic_stlinkv2 most of the time. I can reconnect "tar ext :2000" and scan "mon s" multiple time with no problem. Why does us the #475/487 problem not visible?

Unsure. To reproduce you can use a nRF52840 and the master from https://github.com/dpiegdon/nordic-template .
In the makefile is a target flash-bmp that does not work repeatedly with the bug, but works with the fixed or before introduction.
Maybe the nRF target does something differently here that triggers the bug?

dpiegdon:
What happens if you also attach to the target and then quit?

richardeoin:
Nice find and thank you for the fix. I didn't appreciate the multiple uses of `tc` in the original function.

As has been said, the bug is only triggered after `target_add_commands` is called multiple times without a call to `target_list_free`. Therefore it isn't triggered for some use cases.

PR #466, opened 2019-04-13
Stlinkv2

UweBonnes:
Add a new platform, running on the USB host and talking to an unmodified StlinkV2/3

PR #488, opened 2019-07-05
Add support for target STM32G4[78]x

sjroe:
Tested with STM32G474RE via NUCLEO-G474RE development board.

### DIFF
--- a/src/target/stm32l4.c
+++ b/src/target/stm32l4.c
@@ -28,6 +28,7 @@
  * RM0394 STM32L43xxx STM32L44xxx STM32L45xxx STM32L46xxxx advanced
  *  ARM®-based 32-bit MCUs Rev.3
  * RM0432 STM32L4Rxxx and STM32L4Sxxx advanced Arm®-based 32-bit MCU. Rev 1
+ * RM0440 STM32G4 Series advanced Arm®-based 32-bit MCU. Rev 1
  *
  *
  */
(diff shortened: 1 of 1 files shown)


sjroe:
This doesn't include the STM32G431 & STM32G441 which are covered by a separate device ID, 0x468.  I could include the work for them too but I do not have any hardware to test against.

Device ID 0x468 is already listed in stm32l4.c for a different range of devices (STM32G03).  This was unexpected so I've cross referenced the reference manuals (RM0444 Rev 2 and RM0454 Rev 2) and this should in fact be 0x466.  Do you want me to correct this within this PR or set up another?

For the STM32G47, the flash block size doubles from 0x800 to 0x1000 when working in single bank mode compared to dual.  This is not currently the case in the code for the equivalent STM32L47 which was unexpected as I believe the STM32G47 to essentially be the same as the STM32L47 but with different peripherals.  I've checked the reference manual of the L4[79] (RM0351) and it does not appear to explicitly detail the page size for single bank operation.  Unfortunately I don't have any L47 hardware to verify the block size.

UweBonnes:
#466 additional to the stlinkv2 work also contains the fix for the wrong Idcode for  the G0 and adds th G47. Can you have a try?

sjroe:
Okay, I've just quickly reviewed stm32l4.c from #466 and with regards to the G47 it's missing:
+ Block size (page size) is incorrect when not in dual bank mode (should be 0x1000)
+ Two extra option uint32_t values for the in stm32l4_cmd_option

The above are covered as part of this PR.  If you want to roll those changes into #466 then I will happily test.

Note as mentioned above, I think it is highly likely that the L4[79] will have a page size of 0x1000 in single bank mode but I do not have HW to verify this.

UweBonnes:
Please use what is fixed in #466 in your PR and rebase with a hint on this PR in the commit message. Have a look with git log --check for whitespace problems.

UweBonnes:
Closed with =#491

PR #493, opened 2019-07-18
Weak

UweBonnes:
Remove weak attribute and backticks to help windows compile.

PR #494, opened 2019-07-18
Fix build errors for libftdi under Windows

sidprice:
Made the "weak" attributes in fdiv5.c conditional so they are not used when building libftdi host.


sidprice:
Addressed with #493 - Weak

PR #497, opened 2019-08-03
add support for baite stlink-v2 clone

cuvoodoo:
betemcu/baite offers a ST-LINK/V2 clone with SWD, JTAG, and SWIM
support.
this dongle was detected as ST-LINK/V1 because PC13 was not pulled
low.
this change verifies if the LED is on PA9 to correct this wrong
identification.

### DIFF
--- a/src/platforms/stlink/stlink_common.c
+++ b/src/platforms/stlink/stlink_common.c
@@ -46,9 +46,23 @@ uint32_t detect_rev(void)
 	gpio_set(GPIOC, GPIO14 | GPIO13);
 	for (int i = 0; i < 100; i ++)
 		res = gpio_get(GPIOC, GPIO13);
-	if (res)
-		rev = 0;
-	else {
+	if (res) {
+		/* Some clones (e.g. betemcu/baite) forgot to pull PC13 down,
+		 * but still use V2 connections.
+		 * Check using PA9, unconnected on V1, connected to LED cathode on V2
+		 */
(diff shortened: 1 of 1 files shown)


UweBonnes:
I think, checking PB11 is pulled high is a better test, as discussed in #504. The pullup is 620 Ohm, so it will win in all cases against the built-in pulldown when activated. As the LED has a quite high flow voltage, the voltage on th baite PB9 is probably in the dis-allowed range.

UweBonnes:
Can you please check https://github.com/UweBonnes/blackmagic/tree/baite?

cuvoodoo:
> I think, checking PB11 is pulled high is a better test, as discussed in #504. The pullup is 620 Ohm, so it will win in all cases against the built-in pulldown when activated. As the LED has a quite high flow voltage, the voltage on th baite PB9 is probably in the dis-allowed range.

Agreed. Checking PB11 would identify ST/LINK V2 boards with SWIM support.
This includes the baite, and would work for any mini stlink v2 (the small clone dongles with aluminium case) which would not already be detected as v2.
This is also the check I do in PR #498.

cuvoodoo:
> Can you please check https://github.com/UweBonnes/blackmagic/tree/baite?

Looks good.
I also does not mis-identify the ST-Link V1
I will close this PR once you opened one with your changes.

cuvoodoo:
the alternative change has been committed in #519

PR #498, opened 2019-08-03
use SWIM pins as UART

cuvoodoo:
some STM ST-LINK/V2 also provide SWIM functionality to flash STM8.
the SWIM pins are connected to UART (USART1 REMAP):
- SWIM_RST is connected to PB6/USART1_TX
- SWIM is connected to PB7/USART1_RX (pulled up by 680R)

cheap ST-LINK/V2 clones in form of USB dongle with aluminium case
provide SWD and SWIM signals.
since blackmagic does not support SWIM functionality, the pins can
be re-used as UART port for the USB to UART bridge.

STM ST-LINK/V2 can also offer UART, using USART2, but then the
SWIM signals are not present.
this changes verifies if the SWIM signals are present and remaps
them to USART1 if this is the case.

### DIFF
--- a/src/platforms/stlink/platform.c
+++ b/src/platforms/stlink/platform.c
@@ -40,6 +40,12 @@ uint16_t led_idle_run;
 uint16_t srst_pin;
 static uint32_t rev;
 
+uint32_t USBUSART;
+uint32_t USBUSART_IRQ;
+uint32_t USBUSART_CLK;
+uint32_t USBUSART_PORT;
+uint32_t USBUSART_TX_PIN;
+
 int platform_hwversion(void)
 {
 	return rev;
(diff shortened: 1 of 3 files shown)


UweBonnes:
Did you check that your patches do not collide with the SWO functionality?

cuvoodoo:
I did not test it.

For the stlink platform SWO is defined on UART1_RX/PA10
On the baite and mini/aluminium ST-Link/V2 clones with SWIM, PA10 is not connected.
The baite has JTDI, but it cannot be reused as SWO since it's only on the GPIO PA4.
Thus the clones do not support SWO.

The ST-Link/V2-1 supports SWO on UART1_RX/PA10, but since it is not detected as clone with SWIM support, the UART1 is not re-used on the SWIM pin.
Thus this functionality should not be affected.

Ideally for the clones with SWIM, PLATFORM_HAS_TRACESWO should be set to 0.
But since we are trying get use the same platform this should not be redefined with less support.
The clone user should just not use the SWO functionality.

PR #499, opened 2019-08-05
Minor enhancements

smunaut:
Various minor enhancements, mostly to prepare a new host that's not ARM or PC based.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,15 +4,18 @@ Q := @
 endif
 
 PC_HOSTED =
+NO_LIBOPENCM3 =
 ifeq ($(PROBE_HOST), libftdi)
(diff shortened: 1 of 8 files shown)


UweBonnes:
How do PC_HOSTED and NO_LIBOPENCM3 differ?

smunaut:
Well, at the moment, they are enabled at the same time.

But as mentionned in the PR comment, this is to prepare for an architecture that's neither ARM based (and so doesn't use libopencm3) and is not PC based (and so shouldn't enable the "non-embedded" stuff that PC_HOSTED enables like main() args/argv or the PC specific platform_buffer_flush call).

PR #501, opened 2019-08-18
Windows Build Fixes

sidprice:
After update of GCC "make" the tests for Windows OS are no longer needed.

Tested "clean" and also built all platforms except "hyrdabus"

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -99,12 +99,8 @@ $(TARGET): include/version.h $(OBJ)
 
 clean:	host_clean
 	$(Q)echo "  CLEAN"
-ifeq ($(OS), Windows_NT)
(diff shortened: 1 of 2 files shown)


UweBonnes:
Do the additional test hurt? Otherwise let us keep them for some time so that windows users with some older  GNU make do not stumble upon that.

sidprice:
Unfortunately the "clean" fails with the latest BMP master and the old Make, FYI the version I was using was released in 2010 and the one I now use was released in 2014 I think. I hope nobody else has such a VERY old version of make.

UweBonnes:
Please note the version and perhaps the date of make you use in your PR and also the number of this PR so I can merge with a single commit.

sidprice:
I will make a new PR, having trouble with my repo this morning :(

PR #502, opened 2019-08-19
Windows Build Fixes GNU make 4.2.1 (#502)

sidprice:
After update of GNU make the tests for Windows OS are no longer needed. Note that GNU make 4.2.1 is used to build BMP under Windows OS.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -99,12 +99,8 @@ $(TARGET): include/version.h $(OBJ)
 
 clean:	host_clean
 	$(Q)echo "  CLEAN"
-ifeq ($(OS), Windows_NT)
-	-$(Q)$(shell del *.o, *.d, blackmagic*, $(HOSTFILES))
-else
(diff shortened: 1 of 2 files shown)


UweBonnes:
Please change your commit message to include #502 so that I can commit in a single commit.

UweBonnes:
Have a look at 9e898cc4b8f3e1a5844a3693412a9977a65a0a63

UweBonnes:
Sid, you changed the title of this PR but not the commit messages. There are several ways to commit at PR: Create a merge commit (resulting in two commits for this mini commit), Squash and merge and rebase and commit. The later does not keep the reference to this PR, so the commit message of the (single) PR commit should carry the #502 as reference. But to keep a long discussion short, I will try the second way. Maybe it keeps the PR  and the commit messages,

UweBonnes:
Look good too. Will use more often.

PR #509, opened 2019-08-29
Allow STM32 flash script to be used with python3

KillerLink:
I modified the flash script and its dependency to have braces for print statements, allowing execution with python3.

Because every set of braces only includes 1 value (by keeping the string concatenation with % in place), even if python2 interpretes it as a tuple, it still prints it as previously. Therefore, this should be compatible with python2 and python3 at the same time.

Tested by flashing various stlink / stm32 boards with blackmagic probe firmware using python3.7.

### DIFF
--- a/scripts/dfu.py
+++ b/scripts/dfu.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 #
 # dfu.py: Access USB DFU class devices
-# Copyright (C) 2009  Black Sphere Technologies 
+# Copyright (C) 2009  Black Sphere Technologies
 # Written by Gareth McMullin <gareth@blacksphere.co.nz>
 #
 # This program is free software: you can redistribute it and/or modify
@@ -17,12 +17,12 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
- 
+
 import usb
 
 DFU_DETACH_TIMEOUT = 1000
 
-# DFU Requests 
+# DFU Requests
 DFU_DETACH = 0x00
 DFU_DNLOAD = 0x01
 DFU_UPLOAD = 0x02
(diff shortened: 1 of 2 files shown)


UweBonnes:
git log --check finds trailing whitespace and the first changed line is unrelated whitespace change. Please review.

Otherwise the script worked on my OpenSuse Python2 system.

KillerLink:
Ah you are right, i forgot to mention the reason for these changes. Both the whitespace changes as well as the seemingly unrelated change in the first line (also whitespace) are necessary, due do python3 checking for uniform identation (mixing whitespace and tabs is not allowed).

UweBonnes:
Okay, so it was functional whitespace change.
Add a short note in the commit message starting line 3.

But there is still the trailing whitespace:
scripts/dfu.py:188: trailing whitespace.
+               print("Device %s: ID %04x:%04x %s - %s" % (dfu[0].filename,

UweBonnes:
Please rebase to we have only one commit message

KillerLink:
Updates made to pull request:
-) Update to commit message, add clarification and explain leading whitespace changes.
-) Since both files appear to contain more trailing whitespaces to begin with, including the one in scripts/dfu.py:188 (which is the only one that became visible to you with  because it is in a line i actually changed) i removed all trailing whitespaces.

Since i'm still learning advanced git usage, i hope i'll be doing this right , otherwise sorry for inconvenience. If you want me to do something to be done in a specific way, please feel free to tell me, im happy to learn. Would it (have been)/be better to close this pull request and open a new one with everything done right from the start or should we continue with this?

UweBonnes:
No. Just get your local branch right.  If your local branch has this empty second commit, remove with "git reset --hard HEAD~1". Then change the commit messages as you like with "commit --amend". Than force push to github "git push -f"

KillerLink:
Did not know force push was acceptable, everything i read (including a github help page) said to avoid force push and that it could break pull requests.
Anyways, I hope its good now.

PR #510, opened 2019-09-01
stlink-ud for unidirectional isolators or level translators.

sinihappo:
I implemented an stlink variant named stlink-ud, which can be used with unidirectional digital isolators or level translators.  This is useful to flash or debug eg. 5 volt targets or targets in a different supply voltage domain.  Just compile with make PROBE_HOST=stlink-ud .  This pull request only add a directory  src/platforms/stlink-ud and doesn't touch anything else.

I will provide a suitable schematics later.

### DIFF
--- a/src/platforms/stlink-ud/Bootloader_Upgrade
+++ b/src/platforms/stlink-ud/Bootloader_Upgrade
@@ -1,0 +1,22 @@
+Bootloader Upgrade on Stlink
+============================
+
+Beside accessing the SWD pins direct like explained on
+https://embdev.net/articles/STM_Discovery_as_Black_Magic_Probe
+an updated bootloader can also be programmed via DFU. This requires three
+steps:
+1. Prepare bootloader update
+Normal BMP has to be replaced by the upgrade programm:
+ dfu-util -s 0x08002000:leave -D dfu_upgrade.bin
+Wait until Dual color led flashes red, indicating DFU is active for the
+bootloader.
+
+2. Flash new bootloader
+ dfu-util -s 0x08000000 -D blackmagic_dfu.bin
+Wait until Dual color led flashes green, indicating bootloader is active.
+
+If not, unplug USB, keep black reset button pressed, replug USB.
+Wait until Dual color led flashes green.
+
+3. Flash BMP
+ dfu-util -s 0x08002000:leave:force -D blackmagic.bin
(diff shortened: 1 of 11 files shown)


sinihappo:
Relevant schematics are [here](https://github.com/sinihappo/blackmagic/tree/stlink-ud-sq-schematics/Schematics)

UweBonnes:
Adding a new platform needs a good reason. Good reasons are: it is the native platform, there are many boards out there (Stlink, Swlink, PC-Stlink), it shows the handling on another CPU than STM32F103. What applies for stlink-ud?
- are there many boards out there
- doen't native BMP do level translation too>

The schametics also does not show where the signals come from the F103. I guess X1/SV1 are the  debuggee connections and SV3 goes to the F103. But where do the SV3 signals come from.

It  seems also a variant of the Stlink. Probably you only need SWDIR as additional signal. So why a whole new platform?

In short, a lot is needed to justify inclusion.

Why not keep it as branch on your github?

sinihappo:
> Adding a new platform needs a good reason. Good reasons are: it is the native platform, there are many boards out there (Stlink, Swlink, PC-Stlink), it shows the handling on another CPU than STM32F103. What applies for stlink-ud?

>     * are there many boards out there

No.

>     * doen't native BMP do level translation too>

Latest version (V2.1) seems to have level translation.  But it doesn't provide galvanic isolation and that is the thing which is really useful.
 
> The schametics also does not show where the signals come from the F103. I guess X1/SV1 are the debuggee connections and SV3 goes to the F103. But where do the SV3 signals come from.

That is defined in stlink-ud/platform.h and can be easily changed.

> It seems also a variant of the Stlink. Probably you only need SWDIR as additional signal. So why a whole new platform?

> In short, a lot is needed to justify inclusion.

I understand.
 
> Why not keep it as branch on your github?

That is also OK.

We are going to make a batch of the PCB and add some documentation, too.  I think you are right and I will check if the native BMP v2.1 platforms is adaptable for our hardware.  When I started this, V2.1 wasn't available at all.

UweBonnes:
Regarding the Silab isolation parts. Minimum supply voltage is 2.5 V. If you mass erased a NRF52xxx, IO voltage will be 1.8 V. So you setup will not be able to access the NFR in that situation. I have aFTDI adapter mit the Silab parts and hit that problem. A native BMP helped me in that situation.

sinihappo:
> Regarding the Silab isolation parts. Minimum supply voltage is 2.5 V. If you mass erased a NRF52xxx, IO voltage will be 1.8 V. So you setup will not be able to access the NFR in that situation. I have aFTDI adapter mit the Silab parts and hit that problem. A native BMP helped me in that situation.

You are correct. Thank you for your insight and patience!

UweBonnes:
The Analog devices digital isolators can work don to 1.62 Volt.

PR #507, opened 2019-08-28
Cortexm romtable

UweBonnes:
Rework probing on Cortex M.

Attention to users! Some devices where Romtable PIDR has unexpected values will not get recognized until the right values are put in the code. Please help to fix! Otherwise, only devices with their Romtable PIDR known can be probed selectiv. Please report these values.

- With connect under reset, activate hardware reset
- In DHCSR, enable DEBUG and try hard to halt device, if we are not in hardware reset
- Set DEMCR early to see all ROMTABLE entries and request halt on reset
- With connect under reset, release reset now
- Eventually apply device specific work-arounds, eg. for STM32F7 already set DEBUGMCU_CR
- Scan Romtable
- Many devices are now already known by their Romtable PIDR.
  Probing can now be much more selective with less wild memory reads
- In the device probe, set auxilary bits to eases later attach,
  e.g. the DBGMCU_CR sleep bit for STM32
- After successful probe, get the device going again by cleaning DHCSR
- On detach, clean auxillary bits, DEMCR and DHCSR

PR #511, opened 2019-09-01
Pc stlinkv2

UweBonnes:
Mostly pc-stlinkv2 changes:
- detect stlink attach/detach
- fix crash with some dongle connected but dongle with other serial requested
- Abort early when setting up AP0 fails
- Abort with DP errors
- Fix read/write CPU register

### DIFF
--- a/src/morse.c
+++ b/src/morse.c
@@ -59,8 +59,14 @@ static char morse_repeat;
 
 void morse(const char *msg, char repeat)
 {
-	morse_msg = morse_ptr = msg;
+#if defined(PC_HOSTED)
+	if (msg)
+		DEBUG("%s\n", msg);
+	(void) repeat;
+#else
+morse_msg = morse_ptr = msg;
 	morse_repeat = repeat;
+#endif
 }
 
 bool morse_update(void)
--- a/src/platforms/pc-stlinkv2/platform.h
+++ b/src/platforms/pc-stlinkv2/platform.h
@@ -37,7 +37,8 @@
 
 #define PLATFORM_IDENT "StlinkV2/3"
 #define SET_RUN_STATE(state)
-#define SET_IDLE_STATE(state)
+void stlink_check_detach(int state);
+#define SET_IDLE_STATE(state) stlink_check_detach(state)
 //#define SET_ERROR_STATE(state)
 
 void platform_buffer_flush(void);
--- a/src/platforms/pc-stlinkv2/stlinkv2.c
+++ b/src/platforms/pc-stlinkv2/stlinkv2.c
@@ -440,6 +440,7 @@ static int stlink_usb_error_check(uint8_t *data, bool verbose)
 		case STLINK_SWD_DP_ERROR:
 			if (verbose)
 				DEBUG("STLINK_SWD_DP_ERROR\n");
+			raise_exception(EXCEPTION_ERROR, "STLINK_SWD_DP_ERROR");
 			return STLINK_ERROR_FAIL;
 		case STLINK_SWD_DP_PARITY_ERROR:
 			if (verbose)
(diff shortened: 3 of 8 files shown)


zardam:
I'm sorry, but I still cannot connect to a protected SAMD21.

When protected : 

```
(gdb) mo sw
Target voltage: ABSENT!
Available Targets:
No. Att Driver
 1      ARM Cortex-M M0+
(gdb) at 1
Attaching to Remote target
Attaching to Remote target failed
```

Debug output :

```
AP   0: IDR=04770031 CFG=00000000 BASE=41003003 CSW=03000040
ROM: Table BASE=0x41003000 SYSMEM=0x0
�?0 Entry 0x9f0fc002 -> Not present
�?1 Entry 0x3002 -> Not present
�?ROM: Table END
-> cortexm_probe forced
```

UweBonnes:
As far as I understand tje protected SAMD, you need to connect under reset. I do not see that in the trace above. Nor what Platform you use.

Please try with #507, as that reworks connect under reset and other things.

zardam:
With #507, I cannot even list the board:

```
(gdb) mo sw
Target voltage: ABSENT!
Ignoring packet error, continuing...
Ignoring packet error, continuing...
Ignoring packet error, continuing...
Ignoring packet error, continuing...
```

```
AP   0: IDR=04770031 CFG=00000000 BASE=41003003 CSW=03000040
cortexm_prepare
```

I'm using PROBE_HOST=swlink on a "blue pill". The board under debug is an Arduino zero clone (https://startingelectronics.org/articles/arduino/wemos-arduino-m0/).

Connect under reset is not needed. When the SAMD is protected and a debugger is connected, the CPU is held in "extended reset" at startup, and you can talk with the "Device Service Unit".

The only thing I have to do with my working branch is to manually reset the board after connecting the debugger, just before trying to list it.

UweBonnes:
Zardam,

can you comment out the "cortexm_prepare()" call?

Then look if something like
`ROM: Table BASE=0xe00ff000 SYSMEM=0x1
0 0xe000e000: Generic IP component - Cortex-M3 SCS (System Control Space) (PIDR = 0x4001bb000)-> cortexm_probe`

appears. If we can read  the PIDR at this stage and decode right, we will know it is a Atmel chip and do the right thing.

zardam:
I changed cortexm_prepare() to always return true. With the protected SAMD:

```
(gdb) mo sw
Target voltage: ABSENT!
SW-DP scan failed!
```
```
AP   0: IDR=04770031 CFG=00000000 BASE=41003003 CSW=03000040
AP0 PIDR 0x0009fcd0, designer 0x1f
ROM: Table BASE=0x41003000 SYSMEM=0x0
�?0 Entry 0x9f0fc002 -> Not present
�?1 Entry 0x3002 -> Not present
�?ROM: Table END
cortexm_release, DEMCR 0, apsel 0
```

UweBonnes:
This looks good!
We can read 12.13.13 Peripheral Identification 0 and  1 as we expect for a SAMD (0x1f and 0xcd0). As the device is  protected, we will not see the normal cortexm parts, adiv5_component_probe will fail nd sampdprobe not called in the normal way. But in the failing  case, we can check for the PIDR of a SAMD and  call samd_probe() explicit.

UweBonnes:
Please update #507. It should (famous last words) now finally find the protected SAMD. If it still does not, have a look at the code. I have no SAMD her to test myself, so small error may still be there,

zardam:
Sorry for the bad news, I just tried, and it does not work. I also tried to probe a non protected SAMD without success. I will try to check the code and get it working, maybe this week end.

zardam:
Ignore my previous statement about the failure with the unprotected SAMD, it is working. I'm still working on a fix for the protected case.

Issue #307, opened 2017-11-23
STM32L0 model not detected when connecting under reset


schodet:
I have a STM32L072 doing WFI. I need to connect under reset. When I do it the microcontroller model is not detected (only ARM Cortex-M). I need to disable reset and connect again:

```
(gdb) mon connect_srst enable
(gdb) mon swdp_scan
Target voltage: 3.3V
Available Targets:
No. Att Driver
 1      ARM Cortex-M
(gdb) attach 1
Attaching to program: test.elf, Remote target
warning: while parsing target memory map (at line 1): Required element <memory> is missing
0x0801ddcc in ?? ()
(gdb) mon connect_srst disable
(gdb) mon swdp_scan
Target voltage: 3.3V
Available Targets:
No. Att Driver
 1      STM32L0x
(gdb) attach 1
Attaching to program: test.elf, Remote target
0x0801ddcc in ?? ()
(gdb)
```

UweBonnes:
This should be fixed with the recent commit of #334. Please retest and report.

gsmcmullin:
@schodet Please reopen the issue if you still have the problem.

gsmcmullin:
Ok, this is an STM32L0 for which we didn't split the probe/attach, so no surprises it didn't get fixed.

The fact that we can't even read `DBGMCU_IDCODE` when in reset is a concern. That suggests that there is no way to identify this device while it's in reset.

UweBonnes:
Trying to reproduce on the L053-DISCO with master. It is even worse!
The running  program hangs mostly in WFI but awakes for a 1 kHz timer tick.
There is no way to connect.!
Neither with "mon connect_srst enable" nor by pressing the reset buttom. I must plug a jumper between boot0 and VCC  and reset to enter the bootloader.  Now I can connect.

This is different from Nicolas experience, as for him at least the Cortex-M core appears.

I can reproduce Nicolas behaviour with eth STM32L073RZ.

schodet:
The chip is using stop mode and it wakes up every second. I do not have a reset button on this board.

UweBonnes:
In your situation , where scan detects the Cortex-M, we probably should add an option inside the Cortex-M to reprobe. So you would scan first,  attach to the cortex and do some kind of rescan. While halted and attached to the Cortex-M core, I think the STM32L07 will be responsive. 

How to attach to the L05 I don't have any idea yet.

UweBonnes:
[...]

PR #531, opened 2019-10-11
Add support for LPC1114/333 (LPC1100XL series)

TomCrypto:
... and add support for LPC1114/333 (LPC1100XL series), similar to how it was done for the LPC17xx. This series requires making an IAP call to determine the part ID, it seems that there is no other way.

I've tested this PR on an LPC1114/333 and it detects it, ~~and the PR doesn't alter any existing code paths so there should be no regressions~~; I did a test with an LPC1114FN28 and it seems all good. We can now easily add other LPC1100XL devices in the new switch statement. But we need to check that the IAP call can be made safely or it might break other devices (see comments below).

One other question I have with this PR is that I had to save the target registers prior to the IAP call and then restore them afterwards before resuming otherwise the IAP call corrupts the program state. But I don't see the registers being saved and restored in the LPC17xx code. Can anyone shed some light on this? Does the target actually resume properly with an LPC17xx during detection, is there some difference? Thanks.

### DIFF
--- a/src/target/lpc11xx.c
+++ b/src/target/lpc11xx.c
@@ -147,6 +147,7 @@ lpc11xx_probe(target *t)
 		target_add_ram(t, 0x10000000, 0x2000);
 		lpc11xx_add_flash(t, 0x00000000, 0x20000, 0x1000);
 		return true;
+	case 0x00040070:	/* LPC1114/333 */
 	case 0x00050080:	/* lpc1115XL */
 		t->driver = "LPC1100XL";
 		target_add_ram(t, 0x10000000, 0x2000);


TomCrypto:
Sorry this is a little over my head, I'm not familiar at all with the details of the JTAG/SWD protocol and googling is not helping. What should I do to get the PIDR of the device? I do not have another type of probe available.

UweBonnes:
If you have latest, compile latest firmware "make". Flash it. Now "make PROBE_HOST=pc-hosted". If you are on linux, that should work out of the Box. Windows with mingw and the appropriate devel packages should also work.Other hosts may differ. If compiling succeeds, run the resulting executable. Connect gdb to :2000 and with "mon s" the PIDR is shown on the console starting blackmagic_hosted. Or compile for debug " make DEBUG_ENABLE=1", flash the firmware. Open a terminal and connect to the second ACM modem. Now start gdb and with "mon s" the same information should be visible on the serial console.

TomCrypto:
Okay I got this out:

```
Unref AP
AP   0: IDR=04770021 CFG=00000000 BASE=e00ff003 CSW=03000040
ROM: Table BASE=0xe00ff000 SYSMEM=0x1
0 0xe000e000: Generic IP component - Cortex-M0 SCS (System Control Space) (PIDR = 0x4000bb008)-> cortexm_probe
LPC11/84: IDCODE 0x00040070
1 0xe0001000: Generic IP component - Cortex-M0 DWT (Data Watchpoint and Trace) (PIDR = 0x4000bb00a)
2 0xe0002000: Generic IP component - Cortex-M0 BPU (Breakpoint Unit) (PIDR = 0x4000bb00b)
ROM: Table END
```

The IDCODE 0x00040070 is the same as the part ID returned by the IAP call it seems; does that mean we can use it or is it just a coincidence?

UweBonnes:
Use the second solution. And make it a fall-through of the lpc1115xl case, to save program space. I do not see any profit in using the complicated way presented above. Are the register  accessed documented for the lpc1115xl ? If not, we already rely on undocumented behaviour.

TomCrypto:
Does it matter that the flash capacity is different? It seems like it's not that important as there's a lot of fallthrough? I'll amend the PR in a bit

TomCrypto:
I might've spoken too soon, it seems the IDCODE reverts to 0x00000000 sporadically. I'm trying to track the problem down now.

TomCrypto:
Ok, I think I understand the problem. I think it's likely similar to #307, the IDCODE read succeeds only if the target is halted. So on a cold boot of the BMP, the target is running, doing ` mon s` only shows a generic `ARM Cortex-M M0` because the IDCODE read returned 0x00000000. If I then attach to the target, halting it, the read will succeed and a subsequent `mon s` will correctly identify the target (but this will resume execution, so that a further `mon s` will fail again by reading an all zero IDCODE, until I attach again and so on).

The program I am working with is interrupt-driven and spends most of its time in WFI, I'm not sure if that is the culprit; I will test with a blinky later to see if I get the same outcome.

UweBonnes:
Proper device detection is always a problem. It is a chicken egg problem. You can not read the device ID as long as the device is sleeping /not halted. To do the right action to halt the device you must know the device.
#507 works on that problem. Perhaps have a look and comment. Maybe some lpc_prepare function like the stm3_prepare can be invoked at the right time.
Announcing to much flash is no problem i.m.o. We doe it for a lot of devices. So change this PR to a one line with some comment and I will apply.

B.t.w. I do not see any comment on the memset.

TomCrypto:
@UweBonnes Ok I've updated the PR, now it's just a single line.

Regarding the memset, I had it initializing some fields before but CI was erroring out because not every field of the struct was explicitly initialized, and there were a lot of fields so a memset seemed better (especially considering the `lpc11xx_add_flash` function eventually does a `calloc` to allocate its own `lpc_flash` structure).

I will take a look at #507 . Thanks a lot for your help on this!

PR #533, opened 2019-10-12
Fix buffer overflow in adiv5_component_probe()

kh90909:
This shows up when printing the ROM table to the debug console as garbage characters before indented table entries. The buffer holding the spaces used for the indent is one byte too short.


UweBonnes:
Thanks for catching! And argh, that I did not get it right the first time.

PR #534, opened 2019-10-13
Microchip SAM D51 / E5x support

kh90909:
Adds a target driver for Microchip SAM D51 / E5x family.

Tested on SAMD51G19A and SAMD51J19A. According to the datasheet, the D51 / E5x family share the same core feature set, differing only in the addition of CAN (E51) or ethernet controllers (E53/54). All members of the family should be equivalent from a debug and programming perspective.

In testing, I've observed two things that I don't believe are related to the code in this commit:
1. Unsupported packet errors showing up on the debug console that don't appear to affect any of the functionality: 
   ```*** Unsupported packet: qC
   *** Unsupported packet: Hg0
   *** Unsupported packet: qfThreadInfo
   *** Unsupported packet: qL1200000000000000000
   
   *** Unsupported packet: qSymbol::
   
   *** Unsupported packet: vKill;1
   *** Unsupported packet: qC
   *** Unsupported packet: qAttached
   *** Unsupported packet: qOffsets
   *** Unsupported packet: Hg0
   *** Unsupported packet: vCont?
   *** Unsupported packet: Hc0
   *** Unsupported packet: qL1200000000000000000
   ```
2. When running the `erase_mass` command, the "Erase successful!" message is printed twice. Haven't been able to figure this out. If this `tc_printf` call is modified, the message is still repeated, and if additional `tc_printf` calls are added, only the first one is repeated.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -48,6 +48,7 @@ SRC =			\
 	sam3x.c		\
 	sam4l.c		\
 	samd.c		\
+	samx5x.c	\
 	stm32f1.c	\
 	stm32f4.c	\
 	stm32h7.c	\
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -370,6 +370,7 @@ bool cortexm_probe(ADIv5_AP_t *ap, bool forced)
 	PROBE(sam4l_probe);
 	PROBE(nrf51_probe);
 	PROBE(samd_probe);
+	PROBE(samx5x_probe);
 	PROBE(lmi_probe);
 	PROBE(kinetis_probe);
 	PROBE(efm32_probe);
--- a/src/target/samd.c
+++ b/src/target/samd.c
@@ -41,7 +41,7 @@ static int samd_flash_erase(struct target_flash *t, target_addr addr, size_t len
 static int samd_flash_write(struct target_flash *f,
                             target_addr dest, const void *src, size_t len);
 
-static bool samd_cmd_erase_all(target *t, int argc, const char **argv);
+bool samd_cmd_erase_all(target *t, int argc, const char **argv);
 static bool samd_cmd_lock_flash(target *t, int argc, const char **argv);
 static bool samd_cmd_unlock_flash(target *t, int argc, const char **argv);
 static bool samd_cmd_unlock_bootprot(target *t, int argc, const char **argv);
@@ -178,8 +178,7 @@ uint32_t samd_read_cid(target *t)
  * Overloads the default cortexm reset function with a version that
  * removes the target from extended reset where required.
  */
-static void
-samd_reset(target *t)
+void samd_reset(target *t)
 {
 	/**
 	 * SRST is not asserted here as it appears to reset the adiv5
@@ -273,7 +272,7 @@ samd20_revB_halt_resume(target *t, bool step)
  * function allows users to attach on a temporary basis so they can
  * rescue the device.
  */
-static bool samd_protected_attach(target *t)
+bool samd_protected_attach(target *t)
 {
 	/**
 	 * TODO: Notify the user that we're not really attached and
@@ -524,7 +523,7 @@ static int samd_flash_write(struct target_flash *f,
 /**
  * Uses the Device Service Unit to erase the entire flash
  */
-static bool samd_cmd_erase_all(target *t, int argc, const char **argv)
+bool samd_cmd_erase_all(target *t, int argc, const char **argv)
 {
 	(void)argc;
 	(void)argv;
--- a/src/target/samx5x.c
+++ b/src/target/samx5x.c
@@ -1,0 +1,1198 @@
+/*
+ * This file is part of the Black Magic Debug project.
+ *
+ * Copyright (C) 2019  Ken Healy
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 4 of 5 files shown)


UweBonnes:
Is the samx5x.c really that different from other Sam that a whole new file is needed? Your code adds about 5 k to the binary.  I looked at sam.._cmd_erase_all() and the code seems to be the same. Please either have good arguments for a new file or fold it into the existing files. We should be carefull with our flash space.

kh90909:
> Is the `samx5x.c` really that different from other Sam that a whole new file is needed?

In short, yes. There are some minor savings that can be made (see below), but the NVM controller operates differently, many of the registers are in different locations, and many of the register bits are different.

The NVM functions make up the majority of both target drivers. It would take major refactoring of the existing `samd.c` functions to make them work for both families. Even then, the savings might not be very much. The code would get more complex to handle the differences in operation, and to choose between the two sets of registers. We'd also need to store these register addresses.

In any case, I don't have any SAMD1x/2x device, and I don't think it would be feasible to to make such major changes without having a device to test with. Unfortunately, even if I did, I don't have the time.

However, we can save a little bit of space easily:

* The DSU operation is the same for both families, so there are five functions that are identical for both targets. In fact, I'm surprised the linker didn't automatically remove the duplicates. To eliminate them, I used `extern` declarations to import the `samd.c` versions into `samx5x.c`.

* Some of the user-defined commands are not essential and could be omitted to save space. There is a memory built-in self test function that is also available for the SAMD 1x/2x, which I think few BMP users would need. In addition, I added functions for 8/16/32 bit writes anywhere in the address space, as the the gdb `set` command cannot write to many locations. While I think these commands are useful, they could be made optional via a preprocessor symbol, or included only for a debug build.

The space savings for the two cases above are as follows:

&nbsp;| ENABLE_DEBUG=0 | ENABLE_DEBUG=1 
-|----------------------|---------------------
Flash bytes for the original version of the samx5x target (commit 3d533ef1c304ce736df89a66fe1d138fce3d0b9b)|4924|5832
Use common functions from `samd.c`|-344|-344
Remove `mbist` and `write8/16/32` commands|-884|-1172
**Final size**|**3696**|**4316**

I haven't had a chance to test the modified code. I hope to do so tomorrow and will add it to the pull request so you can review and decide the tradeoff between functionality and size.

UweBonnes:
Where is e.g. monitor command "write8 <addr> <value>" different from gdb command "p *(unsigned char*) addr = value"?

I will have a look at your reworked PR.

kh90909:
The changes are pushed to the PR now. It required some small edits to `samd.c` to remove `static` from the declarations of the common functions, so they can be called from `samx5x.c`.


> Where is e.g. monitor command "write8 " different from gdb command "p _(unsigned char_) addr = value"?

The difference is that the gdb `print`/`set` commands cannot write to addresses in the flash memory space:
```
(gdb) p {unsigned char}0x00000000 = 0xab
Writing to flash memory forbidden in this context
```
When you write to these addresses, you are actually writing to a page buffer. This buffer is then committed to the flash if you initiate a page write.

Thinking about it further, I guess few people would need this capability. I used it for debugging the  flash write functions.

UweBonnes:
Writing at arbitrary flash cell is nothing SAMx5x specific. To my knowledge, no other device implemented it. So implement it in a global way, if if is of general interest. But having a define for it is probably the best short time solution to keep the code visible.

Otherwise I still see a lot of duplication, making long time maintainance of the code harder . I strongly suggest to merge samx5x into  samd. In probe, set a variable what device you have and handle the devices different when needed. The introduction of many  SAMX5X_ defines that only duplicate  SAMD_  defines obscurse the code and a possible merge.

kh90909:
My opinion is the opposite. For a new developer to the project, as I was when I decided to write this SAM D51/E5x target driver, having a single source file supporting more than one target family makes it more complicated to understand. You will need to understand both families and refer to both datasheets to make sense of the conditionals in the code, especially if troubleshooting.

However, this is your call as a project maintainer, not mine.

In any case, I've done all I can. I don't have the hardware or the time to undertake merging samd with samx5x, but I hope my code will serve as a good starting point for someone to do this in future.

UweBonnes:
There is only one conditional to test where the other family are involded: During probe. Then a variable is set and clean different code pathes can be taken after testing that variable.

Fabien-Chouteau:
If this can help, I successfully programmed and debugged the Adafruit Metro M5 Express (`ATSAMD51J19A`) using this patch on a BMP v2.1.

At first loading didn't work:
```
(gdb) load
```

I then used the commands `(gdb) monitor unlock_flash` and `(gdb) monitor unlock_bootprot`, reset the board, and I could load my program. I don't know if both unlocks are strictly necessary.

Thanks for the patch :+1:

UweBonnes:
Fabien: Can you please tell if now the first load succeeds? Make sure the program you load does not lock flash. If that works, I will apply, even with that lot of duplication. What also worries me is that constanst with the same value in the other SAMD driver have different names...

Fabien-Chouteau:
> Can you please tell if now the first load succeeds?
It does work all the times now.

kh90909:
Hi Fabien,

Glad it was useful.

The vFlashErase error is expected when programming Adafruit boards the
first time, as they (correctly) use the bootprot bits to protect their
bootloader during a chip erase.

The unlock_bootprot command is all you need to clear this protection.

Unfortunately, the error message is quite cryptic. I looked into providing
a clearer message in my patch, but it's buried deep in the core code. I
didn't find an easy way to display a message specific to the SAM D51/E5x.

Best regards,
Ken

PR #536, opened 2019-10-13
Set the accepted to socket to be blocking as that is what the code expects

DanielO:
This is necessary on OSX (and probably BSDs) where they default to non blocking which causes recv() to return -1 with errno set to EAGAIN.


UweBonnes:
Wow, that was quick. I tested with pc-stlinkv2 and it still worked.

PR #539, opened 2019-10-18
BUGFIX: tokenization of gdb monitor commands

ruabmbua:
I had issues with the `mon jtag_scan <ir_list>` feature, and discovered, that the GDB monitor command parsing is buggy.

* The *estimated* argv count was wrong
* The tokenizing loop probably triggered some sort of UB, I do not really understand it. My version does not cause issues with libc stack protectors and -fsanitize=address

**How to reproduce:**

Compile PROBE_HOST=pc-hosted with a recent version of GCC (I used 9.2), and
add `-fsanitize=address` to *C_FLAGS* and *LD_FLAGS*.

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -94,8 +94,9 @@ long cortexm_wait_timeout = 2000; /* Timeout to wait for Cortex to react on halt
 int command_process(target *t, char *cmd)
 {
 	const struct command_s *c;
-	int argc = 0;
+	int argc = 1;
 	const char **argv;
+	const char *part;
 
 	/* Initial estimate for argc */
 	for(char *s = cmd; *s; s++)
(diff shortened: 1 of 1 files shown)


UweBonnes:
What is "UB"?

I have compiled pc-hosted on your branch  with fsanitize=address and -lasan as first library to link. Running pc-hosted  and doing "mon jtag_scan 1" still give an report:

Change state to Shift-IR
=================================================================
==25924==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffcb51ecbc0 at pc 0x00000042ad6b bp 0x7ffcb51ec9b0 sp 0x7ffcb51ec9a8
WRITE of size 8 at 0x7ffcb51ecbc0 thread T0
    #0 0x42ad6a in jtagtap_tdi_tdo_seq platforms/pc-hosted/jtagtap.c:116
    #1 0x429348 in jtag_scan target/jtag_scan.c:83
    #2 0x40617f in cmd_jtag_scan /home/bon/devel/blackmagic/src/command.c:183
    #3 0x4065d8 in command_process /home/bon/devel/blackmagic/src/command.c:118
 
Removing your patch while still compiling with asan and running as above still gives the same report.

UweBonnes:
Looking at the code before, there were always two items more written than allocated. So yes, the patch makes sense.

ruabmbua:
Sorry, I forgot to mention, that after this fix you will trigger another bug.

PR #540, opened 2019-10-18
BUGFIX: invalid write to stack

ruabmbua:
jtag_scan defines uint32_t irout, and calls jtagtap_tdi_tdo_seq().
pc-hosted variant casts the pointer argument irout to uint64_t and writes to it.

Fixed by using uint32_t in pc-hosted jtagtap.c

**How to reproduce:**

Compile PROBE_HOST=pc-hosted with a recent version of GCC (I used 9.2), and
add `-fsanitize=address` to *C_FLAGS* and *LD_FLAGS*.

### DIFF
--- a/src/platforms/pc-hosted/jtagtap.c
+++ b/src/platforms/pc-hosted/jtagtap.c
@@ -93,14 +93,14 @@ void jtagtap_tdi_tdo_seq(uint8_t *DO, const uint8_t final_tms, const uint8_t *DI
   uint8_t construct[PLATFORM_MAX_MSG_SIZE];
   int s;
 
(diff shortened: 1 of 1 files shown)


UweBonnes:
Did you test? For me, while with a given setup "mon jtag_scan" shows the device with master, it does not work with #540

UweBonnes:
Can you also tell what stack checking setup you use?

ruabmbua:
Strange, for me it works with this PR. Maybe something relies on the 64bit integer to be available?

ruabmbua:
Closing, because there are cases, where more than 32 bits are required. I will open a ticket instead.

https://github.com/blacksphere/blackmagic/issues/542

PR #552, opened 2019-11-10
pc-hosted/jtagtap.c: Handle DO as bytes (PR #552)

UweBonnes:
Fixes #540/#542


src/platforms/pc-hosted/jtagtap.c
```diff
@@ -112,8 +111,14 @@ void jtagtap_tdi_tdo_seq(uint8_t *DO, const uint8_t final_tms, const uint8_t *DI
       exit(-1);
     }
 
-  if (DO)
-    *DOl=remotehston(-1,(char *)&construct[1]);
+  if (DO) {
+	  int i = 1;
+	  while (ticks) {
```
ruabmbua: I think this is broken. The loop will never exit. Segfaults on my machine

UweBonnes:
Can you try
https://github.com/blacksphere/blackmagic/pull/559?

PR #560, opened 2019-11-25
Fix infinite loop in jtagtap_tdi_tdo_seq

ruabmbua:


UweBonnes:
The function should take buffers of arbitary length  and cope with DI == NULL. What about:
void jtagtap_tdi_tdo_seq(uint8_t *DO, const uint8_t final_tms, const uint8_t *DI, int ticks)
{
  uint8_t construct[PLATFORM_MAX_MSG_SIZE];
  int s;

  if(!ticks) return;
  if (!DI && !DO) return;
  while (ticks > 0) {
	  uint64_t DIl = (DI) ? *(uint64_t *)DI : (uint64_t)-1;
	  /* Reduce the length of DI according to the bits we're transmitting */
	  if (ticks < 32)
		  DIl &= (1L << (ticks+1)) - 1;

	  s=snprintf((char *)construct, PLATFORM_MAX_MSG_SIZE, REMOTE_JTAG_TDIDO_STR, final_tms ? REMOTE_TDITDO_TMS : REMOTE_TDITDO_NOTMS, ticks, DIl);
	  platform_buffer_write(construct, s);

	  s = platform_buffer_read(construct, PLATFORM_MAX_MSG_SIZE);
	  if ((!s) || (construct[0]==REMOTE_RESP_ERR))
		  {
			  fprintf(stderr,"jtagtap_tms_seq failed, error %s\n", s ? (char *)&(construct[1]) : "unknown");
			  exit(-1);
		  }
	  if (DO) {
		  uint64_t Dol;
		  Dol = remotehston(8 , (char *)&construct[1]);
		  memcpy (DO, &Dol, (ticks < 32) ? ((ticks + 7) / 8) : 4);
	  }
	  ticks -= 32;
  }
}

UweBonnes:
You missed the case when DI == NULL. It is a valid request to read a big chunk of data, without providing a dummy write buffer of same big size.

UweBonnes:
The header file tells:
* - DO may be point to the same address as DI.
So this can be used to read big junks of data and my objection was wrong.

PR #562, opened 2019-11-30
Fix SAMx5x build failure - MinGW - Windows

sidprice:
The format specifier "z" causes a build failure for the SAMx5x target MCU. This happens when pc-stlinkv2 is built.

Please test and comment.

### DIFF
--- a/src/target/samx5x.c
+++ b/src/target/samx5x.c
@@ -31,6 +31,8 @@
  * particularly Sections 12. DSU and 25. NVMCTRL
  */
 
+#define __USE_MINGW_ANSI_STDIO 1
+
 #include <ctype.h>
 
 #include "general.h"


UweBonnes:
The portability of the format specifiers is a pain. Will look later.

UweBonnes:
Did you try
include/general.h:#  define PRI_SIZET "zu"
or perhaps add a definition for something similar for decimal output?

sidprice:
No I didn't try that.

What is wrong with the proposed solution, does it not work?

UweBonnes:
The definition should probably go into general.h, before stdio.h, so that other files can profit. This can probably also make the  PRI_SIZET definition go away and %z can be used. Probably also all PRIx can be go away too...

sidprice:
I will check that out and update the PR if it appears to work.

UweBonnes:
Adopted in a generalized form with #561

PR #572, opened 2019-12-20
stlink: update docs on how ti reflash BMP back to STLink original firmware

hiviah:
Instruction on how to reflash STLink from BMP with ST bootloader back to STLink firmware.


UweBonnes:
Please specify stlink version. V2 != V21.

UweBonnes:
Please squash in a single commit with a single, sensible message.

PR #574, opened 2019-12-26
Wait after reset (fixes #573)

### LINKED ISSUE #573: Firmware upload on Arduino Nano 33 BLE fails every other time
I'm using the BMP with PlatformIO and an Arduino Nano 33 BLE (nRF52840). Debugging works but firmware upload reliably fails every other time. In particular, it fails after resetting the Arduino (by power cycling or pressing the button on the board).

The executed command is:
```
arm-none-eabi-gdb -nx --batch -ex "target extended-remote /dev/cu.usbmodemC1CA9EF71" -ex "monitor swdp_scan" -ex "attach 1" -ex load -ex compare-sections -ex kill firmware.elf
```

And the output is:
```
Target voltage: 3.3V
Available Targets:
No. Att Driver
 1      Nordic nRF52 M3/M4
hal_sleep () at ./mbed-os/targets/TARGET_NORDIC/TARGET_NRF5x/TARGET_NRF52/sleep.c:87
87    ./mbed-os/targets/TARGET_NORDIC/TARGET_NRF5x/TARGET_NRF52/sleep.c: No such file or directory.
```

The error message on the last line is from the `load` command.

With a debug version of BMP I have narrowed down the problem to this line:

https://github.com/blacksphere/blackmagic/blob/master/src/target/nrf51.c#L165

Further investigations showed that the error will always occur on the first `target_mem_...` call. In fact, it looks like it's already pending before the function is called. It goes away with a call to `target_check_error()` before line 160.

I'll will further investigate it and come up with a PR. So stay tuned.

manuelbl:
Fix for #573 

On the Arduino Nano 33 BLE, the reset signal takes a long time to return to the high state. During that time the MCU is stopped and all memory operations in cortexm_reset() fail. So wait for 1ms after toggling SRST.

### DIFF
--- a/src/platforms/native/platform.c
+++ b/src/platforms/native/platform.c
@@ -213,6 +213,7 @@ void wait_for_srst(void)
 	while (platform_srst_get_val() && !platform_timeout_is_expired(&timeout));
 
 	if (platform_timeout_is_expired(&timeout)) {
+		DEBUG("Timeout waiting for SRST to go high\n");
 	} else {
 		/* Wait for the same duration again as the high/low
 		   thresholds of the BMP and the target might not match.
--- a/src/platforms/native/platform.c
+++ b/src/platforms/native/platform.c
@@ -38,6 +38,7 @@
 
 static void adc_init(void);
 static void setup_vbus_irq(void);
+static void wait_for_srst(void);
 
 /* Pins PB[7:5] are used to detect hardware revision.
  * 000 - Original production build.
@@ -190,7 +191,33 @@ void platform_srst_set_val(bool assert)
 		gpio_set_val(SRST_PORT, SRST_PIN, !assert);
 	}
 	if (assert) {
+		/* Hold SRST low for about 1ms */
 		for(int i = 0; i < 10000; i++) asm("nop");
+	} else {
+		/* Wait for SRST to go high */
+		wait_for_srst();
+	}
+}
+
+void wait_for_srst(void)
+{
+	uint32_t start = platform_time_ms();
+	/* Prerequisite: timer must have been initialized */
+	if (start == 0)
+		return;
+
+	/* Wait for SRST to go high
+	   but no longer than 400ms */
+	platform_timeout timeout;
+	platform_timeout_set(&timeout, 400);
+	while (platform_srst_get_val() && !platform_timeout_is_expired(&timeout));
+
+	if (platform_timeout_is_expired(&timeout)) {
+	} else {
+		/* Wait for the same duration again as the high/low
+		   thresholds of the BMP and the target might not match.
+		   Increase by 1ms to make up for the timer granularity.  */
+		platform_delay(platform_time_ms() - start + 1);
 	}
 }
 
--- a/src/platforms/stm32/timing_stm32.c
+++ b/src/platforms/stm32/timing_stm32.c
@@ -23,13 +23,14 @@
 #include <libopencm3/cm3/scb.h>
 
 uint8_t running_status;
+uint8_t led_update_counter;
 static volatile uint32_t time_ms;
 
 void platform_timing_init(void)
 {
 	/* Setup heartbeat timer */
 	systick_set_clocksource(STK_CSR_CLKSOURCE_AHB_DIV8);
-	systick_set_reload(900000);	/* Interrupt us at 10 Hz */
+	systick_set_reload(9000);	/* Interrupt us at 1000 Hz */
 	SCB_SHPR(11) &= ~((15 << 4) & 0xff);
 	SCB_SHPR(11) |= ((14 << 4) & 0xff);
 	systick_interrupt_enable();
@@ -45,12 +46,16 @@ void platform_delay(uint32_t ms)
 
 void sys_tick_handler(void)
 {
-	if(running_status)
-		gpio_toggle(LED_PORT, LED_IDLE_RUN);
+	time_ms++;
+	led_update_counter++;
+	if (led_update_counter == 100) {
+		led_update_counter = 0;
 
-	time_ms += 100;
+		if(running_status)
+			gpio_toggle(LED_PORT, LED_IDLE_RUN);
 
-	SET_ERROR_STATE(morse_update());
+		SET_ERROR_STATE(morse_update());
+	}
 }
 
 uint32_t platform_time_ms(void)
--- a/src/platforms/swlink/platform.c
+++ b/src/platforms/swlink/platform.c
@@ -40,6 +40,8 @@ uint16_t led_error_pin;
 static uint8_t rev;
 
 static void adc_init(void);
+static void wait_for_jrst(void);
+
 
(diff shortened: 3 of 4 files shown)


sidprice:
Would it be possible to monitor the "iRst_SENSE" input to wait until the reset input is high, with a timeout in case the reset line is held by the target? Delays make me nervous that a different target would need a longer delay.

manuelbl:
I guess that's possible.

Is it ok to put it into `platform_srst_set_val`? It already contains a delay for holding down the reset line for about 1ms. So it's more than just a *set GPIO function*.

That way we can take care of all platform and won't run into troubles when using `platform_srst_get_val` (the *libftdi*, *pc-hosted* and *hydrabus*, *f4discovery* platform implementations always return false).

manuelbl:
After a few more test, it seems that monitoring SRST_SENSE doesn't easily work. The BMP does not wait long enough, probably due to different thresholds for distinguishing between high and low. So the only feasible approach would be to wait until SRST_SENSE is high and then wait about the same time again to be on the safe side.

This leads to the next issue I've discovered. The native code for measuring time seems to have a granularity of 100ms. That's incredibly coarse. So while I thought that `platform_delay(1)` would wait for 1ms, it in fact waits for a random time between 0 and 100ms. And whoever put the same statement into `cortexm_reset` probably didn't get what he/she intended either.

What's the reason for updating the time only 10 times a second? Why not 1000 times like most platform do (incl. the code generated by STM32CubeMX)? Or use a timer whose value is only read on demand. That way, µs resolution can easily be achieved.

I see two options for going forward:
- Improve the timing granularity to 1ms and implement the solution outlined in the first paragraph
- Resort to a simple solution with `platform_delay(200)`

What would you prefer?

sidprice:
I am not too familiar with that part of the BMP code, perhaps @UweBonnes can comment?

manuelbl:
The latest commits include:

- Improved timer resolution (1ms instead of 100ms)
- Sensing until SRST goes high and then wait the same duration again (in `platform_srst_set_val()` for the native platform)
- Bring swlink platform in line with native platform (hold JRST low for 1ms, JRST sensing and waiting for low to high transition)

I've conducted some tests with different targets and both the native platfrom and swlink platform.

The stlink platform now deviates from stwlink and native. Should I make it the same even though I might not be able to test it?

UweBonnes:
I think each platform should provide pure get/set SRST and a common file provides a decorated get/set SRST that waits until the requested status is read back and also provides a timeout when the SRST is stuck.

The rest of the code should use this  decorated set srst function.

manuelbl:
Just to make sure we're on the same page:

- Many implementations of `platform_srst_set_val` already contain code to hold the reset for 1ms and/or read back the state of the reset signal. This would be removed to make the functions *pure*.
- All wait and read back code goes into *decorated* functions that are neither platform nor target specific.
- The *connect under SRST* and *hard SRST* commands/options can no longer have target or platform specific behavior (waiting or read back) as there is no platform specific code to do so.
- The timing accuracy will need to be improved to about 1ms in order to implement it reasonably.
- For the PC hosted and FTDI platforms, the wait is executed on the host and is likely useless as the communication already adds considerable delays.

If this is how it's supposed to be I would say it's simple and straight-forward. I don't see any immediate problem but it comes with some restrictions that we should be aware of.

UweBonnes:
Point 3: Is there any platform or target specific code for connect under SRST and hard SRST? I do not think so. What specific code do you see?

Point 4: For uc and linux pc-hosted platforms, time accuracy is 1 ms to my knowledge. Where did you see the 100 ms accuracy?

Point 5: USB Full speed ping-pong time frame is 125 us. I do not consider that as considerable with regard to reset timing.

manuelbl:
**Point 3**

Connect under SRST and hard SRST are implemented with `platform_srst_set_val`. The new implementation would also need `platform_srst_get_val`. So they inherit all platform specific code in these functions. And currently that's:

- *native*: SRST is asserted for 1ms. Sensing is completely separate from output.
- *swlink*, *stlink*: It waits until the read back value is the same as the desired value (*1, *2); `platform_srst_get_val` likely works in released state only.
- *f4discovery*, *hydrabus*, *libftdi*, *pc-stlinkv2*: functions are no-ops.
- *launchpad-icdi*: SRST is asserted for 1ms; `platform_srst_get_val` just reads GPIO output register.

I suppose the new implementation will use a decorated function that holds the reset for 1ms and reads back when releasing it (plus some additional wait time). If so, nothing is lost but the behavior might slightly change on some platform. I see it more as a gain than as a problem.

*Notes*:
- *1: I've experienced that reading back the value is not sufficient. And additional wait period is needed as the target might have different thresholds to distinguish between high and low.
- *2: Reading back the value when the GPIO pin is in output mode is likely nonsense as the STM32 just reads the output register, not the real value.

**Point 4**

It's here: https://github.com/blacksphere/blackmagic/blob/master/src/platforms/stm32/timing_stm32.c#L32

The comment (... 10 Hz ...) is correct. The clock source is the main clock (72MHz) divided by 8 and then again divided by 900000 resulting in 10Hz or 100ms accuracy.

**Point 5**

That's better than expected and therefore no problem.

UweBonnes:
Point3 *1: Yes, different thresholds may happen. How to cope with that? Perhaps measure the time from release to valid readback and wait that time again?
Point3 *2: SRST is suppose to be either open drain or a seperate pin. Using the IDR value in both case will read back the actual line value?
Point 4: Even a granularity of 100 ms is no big problem. The 100 ms are nneded for the morse code.

B.t.w. I do not see any user of platform get srst

manuelbl:
Point 3 *1: Yes, exactly. That's what I have already implemented in this pull request.

Point 3 *2: It doesn't seem to be configured as open-drain for `launchpad-icdi`. And *swlink* and *stlink* switch between open-drain and input. Are you sure that the IDR register measures the external value if the GPIO is configured as output? If so, why do *swlink* and *stlink* switch to input mode? I will test it.

Point 4: At a granularity of 100ms seconds, all timeouts will need to use at least 200ms, resulting in an effective wait time somewhere between 100ms and 200ms. (A timeout of 100ms can result in an effective timeout of close to 0ms, resulting in a failing command. I've experienced it.) I don't like a minimum timeout of 200ms. And it's easy to change it to 1ms resolution and still retain the morse code (see this pull request).

`platform_srst_get_val` is used in:
- https://github.com/blacksphere/blackmagic/blob/master/src/target/cortexa.c#L395
- https://github.com/blacksphere/blackmagic/blob/master/src/target/samd.c#L544
- https://github.com/blacksphere/blackmagic/blob/master/src/target/samx5x.c#L447
- https://github.com/blacksphere/blackmagic/blob/master/src/remote.c#L224

It will mainly be needed for the read-back and it must be separate from GPIO output registers.

I don't see any blockers. So I'll go ahead and implement it, most likely in a new pull request.

manuelbl:
Closing this pull request in favor of #576

PR #575, opened 2019-12-28
Adding the ctxLink probe host

sidprice:
This PR has the code to support the ctxLink probe. As part of these additions files are included that support compilation of any of the BMP platforms on Windows using VSCode.


UweBonnes:
As explained, please merge in a single commit. Consider to keep a private branch keeping the commits.

sidprice:
I thought you said that you would do that when you merged it, that is why I left it the way it is.

UweBonnes:
Please rebase to master. This will be tedious, but is needed. Your commits are not only on top of master, but mixed with other. That way, squashing is even more tedious .

Please also compile with a recent GCC. "make all_platforms " chokes like
command.c:93:14: error: cast between incompatible function types from '_Bool (*)(void)' to '_Bool (*)(target *, int,  const char **)' {aka '_Bool (*)(struct target_s *, int,  const char **)'} [-Werror=cast-function-type]
   93 | { "battery", (cmd_handler)cmd_battery, "Read the battery voltage" },

For a cure, see 
  1cf0b8ac1336d4bf46c475e64a0110447731e245

sidprice:
@UweBonnes What version of GCC are you using please?

UweBonnes:
gcc version 7.4.1 20190905

sidprice:
The latest  7.x.x available on the ARM download site is the one I have "7-2018-q2-update"

Do you have a URL for the one you have?

UweBonnes:
The gcc problem is minor, but as I noticed it, I listed it. The gcc is that, that  comes with my OpenSuse Leap .  The major thing is to squash!

sidprice:
Tried to create a better PR, see PR #579

PR #576, opened 2019-12-30
Handle slow reset

manuelbl:
New pull request replacing #574. Fix for #573. 

On the Arduino Nano 33 BLE, the reset signal takes a long time to return to the high state. During that time the MCU is stopped and all memory operations in `cortexm_reset()` fail.

This pull request is the result of discussions with @UweBonnes:

- New functions (platform agnostic):
    - `platform_srst_assert`: assert SRST for 1ms
    - `platform_srst_release`: release SRST and wait until the target is likely responsive again
    - `platform_srst_reset`: perform reset (assert & release)
- Replace the low-level functions with the new ones where appropriate
- Improve timing functions (timeouts, time measurement, delay) to a minimum precision of 1ms (relevant for STM32-based hosts and Launchpad ICDI)
- Removed timing and read-back code from `platform_srst_get_val()` as it has been moved to the new platform agnostic functions 

Aside from the cleanup, the main addition is `platform_srst_reset`. After releasing SRST, it waits until reading back the GPIO value shows the desired effect and measures the duration. It then waits the same duration again as the target might have different thresholds to distinguish low/high and as the target might need a short time to become responsive again.

This should solve the problem observed on the Arduino Nano BLE 33, which is not immediately responsive after the reset signal is released since it takes about 1ms to return to the high level.

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -179,7 +179,7 @@ static bool cmd_jtag_scan(target *t, int argc, char **argv)
 	}
 
 	if(connect_assert_srst)
-		platform_srst_set_val(true); /* will be deasserted after attach */
+		platform_srst_assert(); /* will be deasserted after attach */
 
 	int devs = -1;
 	volatile struct exception e;
@@ -196,7 +196,7 @@ static bool cmd_jtag_scan(target *t, int argc, char **argv)
 	}
 
 	if(devs <= 0) {
-		platform_srst_set_val(false);
+		platform_srst_release();
 		gdb_out("JTAG device scan failed!\n");
 		return false;
 	}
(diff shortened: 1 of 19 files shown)


UweBonnes:
Sorry, I have to request more patience from you, much other distraction

UweBonnes:
Thinking again, I do not like the behavior
"platform_srst_release: release SRST and wait until the target is likely responsive again"
platform_srst_release() should only release SRST. On most parts we can already do a lot while the target still under reset. 
For CortexM, we know we are out of Reset when two consequitive reads of DHCSR read S_RESET_ST as 0.
Only if we ask explicit platform_srst_get_val() we should be more carefull.

Only

UweBonnes:
B.t.w,
https://arm-software.github.io/CMSIS_5/Pack/html/debug_description.html

has a nice description of Debug concepts.

manuelbl:
@UweBonnes I see your point and will take some time to look into it in more detail. It looks like a major change, and my knowledge of the BMP code base is rather limited...

manuelbl:
I won't find the time to implement the proposed approach.

PR #579, opened 2020-01-06
Added ctxLink platform

sidprice:
Squashed ctxLink branch into this new branch to hopefully make the merge into master easier


sidprice:
@UweBonnes I hope this works better for you. I still need to make the change for command.c, just want to know that my merge with squash did what you expected.

UweBonnes:
The patch is very intrusive. I would expect one new platform directory and some read explaining the needed PROBE_HOST. Why are there ctxLink and winc1500?
- Consider to move the cases designe file directory to an external repo as BMP Hardware does. 
- Try to add as few as possible files/directories to ./ and ./src.
- Check for trailing whitespace. Check with git log --check and fix with git rebase --whitespace=fix HEAD~1
- Make the usbauart changes a separate patch, as it may cause colateral damage. Check for unindented whitespace changes. Look e.g. at platforms/stm32/usbuart.c with git log p and git log -p 
Changes in src/gdb_main.c are also unwanted. Perhaps invert the logic of the network client check to return immediate so you do not need to indent the rest of the code.
- Make the battery command a separate patch.
- Look for unwanted added file, like 
 src/blackmagic.dfu                                           |   Bin 0 -> 125873 bytes
- Why do you use #ifdef ctxLink and #ifdef PLATFORM_HAS_BATTERY on command.c
- Get rid of the "Merge remote-tracking branch 'upstream/master'" patch probably by "git rebase master"
- At lease for files in the common directories, use the BMP coding style:
 if (a)
     bla;
if (b) {
    bla;
} else {
bla1;
bla2;
}
- use tabs as in the rest of the code.

You make also consider to make a PR to include needed infrastructure (battery, network client check) in main BMP and use it as a submodule in your repo.

sidprice:
Many  thanks for your time spent on this and your comprehensive comments. I will close this PR while I think about moving forward with this merge into the BMP repo or continuing with maintaining my separate fork.

UweBonnes:
Please at least use BMP as a submodule. The needed infrastructure in common code is small.

sidprice:
Thanks for the suggestion, I don't know how I would do that. My Git skills are growing but that is not something I have done or know how to do.

sidprice:
BTW: There should be no invasive changes to gdb_main.c, did you look at the code of this pull ;request? I see only what look like whitespace issues that probably arose after I removed the previous invasive changes.

sidprice:
I am trying to study and consider you comments above and I think I will take a look at implementing them. My plan is to work my way through your list and squash and push each one to my branch so that I can get your feedback as I go. Is this OK with you? I would not want to expend time if that approach is not acceptable.

PR #582, opened 2020-01-18
fixup F103x8 flash size before build dfu iface string

zvova7890:
This fixes incorrect flash size in dfu iface string


UweBonnes:
Argh, this code and this error is from me. Looking at the code now, I think this  fake deserves a compile time option with default: Compile the string with the flash size the chip announces and only if the user activates an additional option, change the the string for a 64 k device. Until now I have not heared of problems with the upper 64 k flash on a 64 k device, but there is no guarantee.

Any opinion?

zvova7890:
Not sure. Because blackmagic is not feet in 64k, so user understand this in any way. After that he flash via SWDIO bootloader+firmware and then just want update firmware through DFU. I think if this can be an option, by default must allow "oversize"

UweBonnes:
Please also fix the comment as proposed by Jiri.

UweBonnes:
Please rebase into one commit.

PR #585, opened 2020-01-23
Fix for mistaking EFR32 for nRF51

kanflo:
I got surprised when when my EFR32FG1V following a reset was detected as an Nordic nRF51 when running ```mon swdp_scan```. Turns out this has been along since ```a336ac20847b6f2059a8f77f774362699cfbd32b - NRF5: New detection scheme.```. The nRF probe function reads registers such as ```NRF51_FICR_CODEPAGESIZE``` mapped to ```0x10000010``` which happens to be the address where the EFR32 has its SRAM.

I am not familiar with the nRF51 but it does not seem to have a part ID register in the FICR area (as the nRF52 has) which would explain why the probe code is written the way it is. There used to be [relevant information](https://devzone.nordicsemi.com/f/nordic-q-a/5923/sdk8-0-0-what-happened-to-nrf_ficr_type-info/20639#20639) (```NRF51_FICR_DEVICE_INFO_PART```) but Nordic Semi axed this in newer silicons as "It was decided that it would take too much effort to populate that region at production" :D

This PR is for discussing how to fix this, it is not for merging as I have not tested it on an nRF51 (I have none ;)

A more simpler fix would of course be to move the nRF probing down the list.

### DIFF
--- a/src/target/nrf51.c
+++ b/src/target/nrf51.c
@@ -86,6 +86,13 @@ const struct command_s nrf51_read_cmd_list[] = {
 #define NRF51_FICR_DEVICE_INFO_RAM		(NRF51_FICR_DEVICE_INFO_BASE + 12)
 #define NRF51_FICR_DEVICE_INFO_FLASH	(NRF51_FICR_DEVICE_INFO_BASE + 16)
 
+/* nRF51 temperature sensor */
+#define NRF51_TEMP					0x4000c000
+#define NRF51_TEMP_START			(NRF51_TEMP + 0x000)
+#define NRF51_TEMP_STOP				(NRF51_TEMP + 0x004)
+#define NRF51_TEMP_DATARDY			(NRF51_TEMP + 0x100)
+#define NRF51_TEMP_TIMEOUT			10000
+
 #define NRF51_FIELD_UNSPECIFIED				(0xFFFFFFFF)
 
 /* User Information Configuration Registers (UICR) */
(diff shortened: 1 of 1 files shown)


UweBonnes:
Argh, probing.! In an ideal world, AP V2 devices tell vendor and Device ID already in DP_TARGETID. STM32F7 and H7 do fine, STM32L5 read NULL there! Otherwise for Cortex-M, devices, SYSROM PIDR should tell designer and device id. However reading rom table may fail, as the device is in sleep. And waking up needs device specific things. Argh. Or the device missed PIDR or has wrong values.

Moving devices up and down the probe chain while probe targets blindly poke into the memory will give more collateral damage. Probing needs to be put on more solid feet with hints given by sysrom pidr..

https://github.com/UweBonnes/blackmagic/tree/pidr prints the PIDR ,if ROMTABLE read succeeds. Can you try and report back. Either compile the firmware with the debug option set and look an the second UART for output or compile any of the PC hosted platforms, e.g. pc-hosted for native and likes.

Probably I also will merge the PIDR printout branch.

kanflo:
My EFR reports PIDR ```0x300f38c1```. It would be nice to see what the nRF51 has to say about itself. I will take a dumpster dive in the lab and see if I cannot find one. Googling hints at ```0x4000bb9a3```.

I agree that bad things could happen if one pokes around too much. This is especially true if a supported micro requires a certain level of silliness to reveal itself because the designer left out vital information about the silicon. Perhaps this could be leveraged with a 'hint' when scanning. Eg. ```mon swdp_scan efr32``` could ask the BMP so scan for EFx32s and fail if none found. This would of course litter the code base with tons of strcmps...

kanflo:
A somewhat related thought is how to wake certain MCUs in deep sleep. For the EFM32 a [special sequence can be sent on RESET, SWDIO and SWCLK (page 14)](https://www.silabs.com/documents/public/application-notes/an0062.pdf) to increase the time window the debugger has to halt the CPU before it enters deep sleep by the flashed application. For this to work reliable and without side effects I would think a command like ```mon swdp_scan efr32```would be needed. This is based on me noticing that my BMP could not find my deep sleeping EFM32.

UweBonnes:
Cortex-M has a generic way to halt the CPU after reset. Did Energie Micro find a way to screw up that generic way?

kanflo:
According to Silabs the "MCU is completely shut down and the only way to wake up is with a reset" in EM4. As the chip lacks something akin to ST's BOOT0, Silabs introduced the "AAP Expansion Sequence" to hold off booting an "EM4 eager" application.

kanflo:
I will close this and hopefully one day get around to raise a PR with something like ```mon swdp_scan efr32```.

UweBonnes:
Name a cheap EFR32 board I can source in Germany expedited from the EU. Provide am elf binary I can test on that booard. Maybe I find time to test...

PR #586, opened 2020-01-31
Symbolic definition and remapping of USB endpoints

sidprice:
All USB endpoints for BMP are symbolically defined.

The addresses for the endpoints are remapped to place the GDB and TraceSWO endpoints in the first four endpoints of the platform MCU.

The USB UART endpoints are mapped after the above endpoints.

This arrangement allows platforms based upon the STM32F4xx to support GDB and SWO ports and allows the native BMP platform to additionally offer the USB UART port.

The native platform was tested on Windows and Linux (Ubuntu), however an F4Discovery platform was not available for test.

Please verify F4Discovery if one is available.


UweBonnes:
Sid, as Dave pointed out, your change will be a pita for existing applications using SWO on the BMP. So your change may only thouch your platform and the f4discovery.

sidprice:
Understood. 

However, what I propose is that the symbolic definitions of the endpoints are retained and that they be defined in platform.h. Would that be acceptable?

UweBonnes:
Has this patch still use for you? If yes, rebase to master so that the merge commit messages disappear. otherwise  the tmc4 platform compiled for me with make all_platforms. What is needed to get that error?

sidprice:
@UweBonnes This so long ago, I need to revisit the code base. I will let you know if this is still a valid PR.

UweBonnes:
Please close if no longer needed. Reopen if sometimes needed in the future.

sidprice:
OK, will do.

UweBonnes:
This PE is still open...

sidprice:
Yes, I have not had time to double check. Please give me some time.

sidprice:
I am going to close this and if I really need something like it I will create a new PR.

PR #587, opened 2020-02-01
STM32F1 High density devices DFU support

EXTREME-KEFIR:
Current version of BMP DFU is not working on STM32F103RC chip. The reason is that DFU interface is defined as 8x001Ka + 248x001Kg
In this interface page size is 1K which is correct for medium and low density 103xC devices but high density 103xC devices, such as 103RC have 2K page size. Because of that, dfu-util treats each 2K page as two 1K pages resulting in alternating blank/correct pages

I've replaced interface string with 4x002Ka + 124x002Kg and it works now. Here is small example of possible solution attached - additional define of STM32F1HD is needed in makefile in order to tweak the  interface string

### DIFF
--- a/src/platforms/stm32/dfucore.c
+++ b/src/platforms/stm32/dfucore.c
@@ -20,9 +20,14 @@
 #include "general.h"
 
 #include <string.h>
-#if defined(STM32F1)
+#if defined(STM32F1HD)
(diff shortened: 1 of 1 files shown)


UweBonnes:
Can you correct the subject and add i commit message?

UweBonnes:
The commit message still says: STM32F1 High denisty devices DFU support

PR #589, opened 2020-02-06
fix -Werror,-Wmissing-field-initializers

EliasOenal:
Building pc-stlinkv2 on MacOS Catalina with clang I got:
```
  CC      platforms/pc/cl_utils.c
platforms/pc/cl_utils.c:244:37: error: missing field 'printf' initializer [-Werror,-Wmissing-field-initializers]
        struct target_controller tc = {NULL};
                                           ^
1 error generated.
```
This pull request explicitly initialises all the fields of the target_controller struct and thus fixes the error.

### DIFF
--- a/src/platforms/pc/cl_utils.c
+++ b/src/platforms/pc/cl_utils.c
@@ -241,7 +241,7 @@ int cl_execute(BMP_CL_OPTIONS_t *opt)
 		DEBUG("Given target nummer %d not available\n", opt->opt_target_dev);
 		return res;
 	}
-	struct target_controller tc = {NULL};
+	struct target_controller tc = {NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, 0, NULL};
 	target *t = target_attach_n(opt->opt_target_dev, &tc);
 	if (!t) {
 		DEBUG("Can not attach to target %d\n", opt->opt_target_dev);


UweBonnes:
I think
+++ b/src/platforms/pc-stlinkv2/stlinkv2.c
@@ -691,7 +691,7 @@ static void stlink_resetsys(void)
 
 void stlink_init(int argc, char **argv)
 {
-       BMP_CL_OPTIONS_t cl_opts = {0};
+       BMP_CL_OPTIONS_t cl_opts = {};
--- a/src/platforms/pc/cl_utils.c
+++ b/src/platforms/pc/cl_utils.c
@@ -241,14 +241,14 @@ int cl_execute(BMP_CL_OPTIONS_t *opt)
                DEBUG("Given target nummer %d not available\n", opt->opt_target_dev);
                return res;
        }
-       struct target_controller tc = {NULL};
+       struct target_controller tc = {};
        target *t = target_attach_n(opt->opt_target_dev, &tc);
        if (!t) {
                DEBUG("Can not attach to target %d\n", opt->opt_target_dev);
                goto target_detach;
        }
        int read_file = -1;
-       struct mmap_data map = {0};
+       struct mmap_data map = {};

is a shorter solution. Can you try and update the PR?

EliasOenal:
Your suggestion works on my system with the current Makefile, yet it is a GNU extension and doesn't comply with the C standard. I added the -pedantic flag got:
```
platforms/pc/cl_utils.c:244:32: error: use of GNU empty initializer extension [-Werror,-Wgnu-empty-initializer]
        struct target_controller tc = {};
```
I personally would prefer to stay standards compliant, but I'll make the change if that's your preference.

UweBonnes:
If you really want to be that pedantic, do an explicit memset to 0 instead of the initialization. I do not like the long list of arguments...

UweBonnes:
Bt,w, can you explain why 
> env CC=clang make all_platformsBuilding for hardware platform: libftdi
  CLEAN
 GIT include/version.h
  CC      target/adiv5.c
In file included from include/general.h:35,
                 from target/adiv5.c:26:
platforms/libftdi/platform.h:24:10: fatal error: libftdi1/ftdi.h: No such file or directory
   24 | #include <libftdi1/ftdi.h>
      |          ^~~~~~~~~~~~~~~~~
fails, while
> env CC=clang make PROBE_HOST=libftdi
succeeds?

UweBonnes:
Should be fixed by #594. Reopen if still pending.

PR #599, opened 2020-03-06
Tidying of common swdp access functions

mubes:
Some tidying of the common swdp access functions. This will change the timing slightly so does need to be tested against relevant targets, but is fine on BMP.

### DIFF
--- a/src/platforms/common/swdptap.c
+++ b/src/platforms/common/swdptap.c
@@ -24,18 +24,40 @@
 #include "swdptap.h"
 
 enum {
-	SWDIO_STATUS_FLOAT = 0,
-	SWDIO_STATUS_DRIVE
+       SWDIO_STATUS_FLOAT = 0,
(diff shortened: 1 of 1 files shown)


UweBonnes:
Unrelated whitespace changes and changes in High and setup times to smaller margins not justified.

mubes:
Timings extended to align with previous version ( although they were already in line with the ARM timing specified in http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0517b/CIHBIEHF.html ). Whitespace should be addressed through a tool. I will submit a patch for that.

UweBonnes:
Also timing is adequate and even reliable when going off-board with jumper wire, people often misstake problems by flacky connections and/or  flaky hardware as timing problems, e.g. see #452 and #392. The slightly increased margins only slow down a little but for sure increases margins.

UweBonnes:
Superceeded by #721

PR #602, opened 2020-03-06
Prettifier support

mubes:
Support for automatic code prettification. It'll never be perfect but it can always be standard.  Please feel free to suggest or make changes to the scripts/astyle.conf file...I don't care what we use, as long as we use something. The 'opening' script is from Orbuculum, which has served me well enough.

Once it's agreed we can run a onetime through the repository, and then the code will always be the same.


UweBonnes:
The changes by the given astyle,conf are enourmous. Applying is (make pretty") will defeat applying and outstanding patch and require heavy rework. Please provide a much less intrusive configuration.

mubes:
Once the one time hit has been taken then there are no further changes that
will be made to the files.  You've basically got to look at the output
files from the prettifying process and decide if they are acceptable or
not. The one time massive change really doesn't matter - only if the
resulting output is acceptable to your eye.

It is not possible for any prettifier to reflect current formatting
unambiguously, especially when that formatting hasn't been explicitly
documented in the first place. I'm sure it could be made closer to what's
in the majority of the files already - that's just a case of editing the
config file.

PR #606, opened 2020-03-09
Erase range

UweBonnes:
This is adopted from #458 .Syntax has been changed to "mon erase_range start (len)" and will erase from start to start + len - 1 plus anything that is on the same page before start and anything that is on the same page behind start + len - 1.

Please give feedback if these additional 0x312 bytes in flash are usefull.

### DIFF
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -38,6 +38,7 @@
 
 #include <unistd.h>
 
+#include <errno.h>
 #if PC_HOSTED == 1
 
 /*
@@ -48,7 +49,6 @@
  */
 
 #define TARGET_NULL ((target_addr)0)
-#include <errno.h>
 #include <time.h>
 #include <sys/time.h>
 #include <sys/stat.h>
@@ -58,8 +58,10 @@
 static const char cortexm_driver_str[] = "ARM Cortex-M";
 
 static bool cortexm_vector_catch(target *t, int argc, char *argv[]);
+static bool cortexm_cmd_erase_range(target *t, int argc, char *argv[]);
 
 const struct command_s cortexm_cmd_list[] = {
+	{"erase_range", (cmd_handler)cortexm_cmd_erase_range, "Erase pages including start and end"},
 	{"vector_catch", (cmd_handler)cortexm_vector_catch, "Catch exception vectors"},
 	{NULL, NULL, NULL}
 };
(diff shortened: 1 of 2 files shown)


src/target/cortexm.c
```diff
@@ -1122,3 +1124,61 @@ static int cortexm_hostio_request(target *t)
...
+{
+	if ((argc != 2) && (argc != 3))
+		{
+			tc_printf(t, "usage: monitor erase_range <start> [<end>]\n");
+			return false;
+		}
+
+	char *eos;
+	uint32_t addr = strtoul(argv[1], &eos, 0);
+	if (*eos)
```
TDHolmes, line 1145: Looking at the docs for [`strtoul`](https://www.tutorialspoint.com/c_standard_library/c_function_strtoul.htm), looks like checking `addr != 0` should be the error check. Dereferencing `eos` just makes me paranoid.

src/include/platform_support.h
```diff
@@ -35,6 +35,7 @@ void platform_timeout_set(platform_timeout *t, uint32_t ms);
 bool platform_timeout_is_expired(platform_timeout *t);
 void platform_delay(uint32_t ms);
 
+extern bool connect_assert_srst;
```
TDHolmes, line 38: this doesn't seem to be used for anything

src/target/cortexm.c
UweBonnes: Do you think 
	errno = 0;
	uint32_t addr = strtoul(argv[1], NULL, 0);
	if ((addr == 0) && (errno))
is the right thing?

src/include/platform_support.h
UweBonnes: It is needed for hosted and command.c. Have to think aboyur that

UweBonnes:
Most important comments needed: Does this make sense?

src/target/cortexm.c
TDHolmes: Personally I think
```c
if ((addr == 0) || (errno))
```
makes more sense and is a bit more safe since either or is an error condition, but I think what you've done should work too

TDHolmes:
I don't have a ton of context on the command and why it's useful (not sure what is meant by "Please give feedback if these additional 0x312 bytes in flash are useful.") but it seems reasonable enough to me...

UweBonnes:
Flash on the BMP/firmware is not endless. New part need more algorithms, riscv would be nice and so some thoughts should be made before adding new things. 0x312 =  786 Bytes and that is about 2.5 % of the space left. sorry if I was to tense.

TDHolmes:
ohhh got it, the feature adds that many bytes of flash usage. Well if no
one is asking for it then no need 🤷🏼‍♂️ could be left on a branch if
needed later. Agreed that adding targets is more useful than adding
unnecessary features.

UweBonnes:
Blackmagic hosted already provides this feature with the -E -a<start address> -S <lenght> arguments. Closing for now until somebody reopens when needed.

canatella:
I was looking at the linked PR and issue here. I'm working daily with nrf52 and that is the one thing I still need to unplug the bmp from the board to do. It is useful when you want to erase some user data that is stored in flash. I also had to do that because I'd have written to the UICR, a configuration register stored in the flash and I needed to reset it. It might make sense to make this a compilation option if this is using too much flash ?

UweBonnes:
Cannetella, why the using hosted not an option for you?

canatella:
I'm not sure what you mean by hosted ?

UweBonnes:
https://github.com/blacksphere/blackmagic/tree/master/src/platforms/hosted

canatella:
Ah right, but to me that defeats the whole point of buying the bmp hardware which was to remove the gdb server / jlink stuff out of the way. I'd like to be able to do everything from my gdb prompt. It also simplifies the CI system: I don't have other process to manage, gdb just works.

UweBonnes:
Okay. I rebased the branch to master. Can you test if this does what you need?

canatella:
```
(gdb) x/8xub 0x76000
0x76000:	0	8	114	26	112	18	36	10
(gdb) mon erase_range 0x76000 8
(gdb) x/8xub 0x76000
0x76000:	255	255	255	255	255	255	255	255
```
Looks like it does!

UweBonnes:
Huch, arguments are describe as "start end" but actual are "start length". Correcting...

PR #613, opened 2020-03-19
dfucore: Reject erase outside the allowed range

UweBonnes:
An illegal erase request may get emitted as the DFU program may erase
first. Checking range before flashing is too late.

### DIFF
--- a/src/platforms/stm32/dfucore.c
+++ b/src/platforms/stm32/dfucore.c
@@ -176,9 +176,14 @@ usbdfu_getstatus_complete(usbd_device *dev, struct usb_setup_data *req)
 
 		flash_unlock();
 		if(prog.blocknum == 0) {
-			int32_t addr = get_le32(prog.buf + 1);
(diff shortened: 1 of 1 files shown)


src/platforms/stm32/dfucore.c
```diff
@@ -176,9 +176,13 @@ usbdfu_getstatus_complete(usbd_device *dev, struct usb_setup_data *req)
 
 		flash_unlock();
 		if(prog.blocknum == 0) {
-			int32_t addr = get_le32(prog.buf + 1);
+			uint32_t addr = get_le32(prog.buf + 1);
 			switch(prog.buf[0]) {
 			case CMD_ERASE:
+				if ((addr <  app_address) || (addr >= max_address)) {
+					usbdfu_state = 	STATE_DFU_ERROR;
+					return;
```
mnmoron: I noticed that on line 177 there is a flash_unlock().
Do you need to do a flash lock restore before returning at this point?

PR #615, opened 2020-03-24
BMP/PC: Allow to compile with mingw64.

UweBonnes:
__USE_MINGW_ANSI_STDIO 1 must be set befor any windows specific included.

### DIFF
--- a/src/include/general.h
+++ b/src/include/general.h
@@ -22,7 +22,9 @@
 #define __GENERAL_H
 
 #define _GNU_SOURCE
-#define __USE_MINGW_ANSI_STDIO 1
+#if !defined(__USE_MINGW_ANSI_STDIO)
+# define __USE_MINGW_ANSI_STDIO 1
+#endif
 #include <stdint.h>
 #include <stdbool.h>
 #include <stdlib.h>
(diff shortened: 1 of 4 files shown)


sidprice:
What does this change address? Is it a problem building on Linux?

UweBonnes:
This cure compilation problems seen when I use
> mingw64-make PROBE_HOST=libftdi
on linux. 
-- 
In file included from platforms/pc/gdb_if.c:44:
include/general.h:25: error: "__USE_MINGW_ANSI_STDIO" redefined [-Werror]
   25 | #define __USE_MINGW_ANSI_STDIO 1
      | 
In file included from /usr/x86_64-w64-mingw32/sys-root/mingw/include/windows.h:9,
 
and later the use of %z to print size_t entities.

Do you have  mingw64-libftdi1-devel installed? That is where usr/x86_64-w64-mingw32/sys-root/mingw/lib/pkgconfig/libftdi1.pc comes from.

sidprice:
OK, thanks for the explanation.

I do not have mingw64-libftdi1-devel installed, just the header files and libraries. Will check that out thanks.

PR #622, opened 2020-03-27
stm32/timing_stm32.c: Remove fixed constants and direct SCB access

UweBonnes:


sidprice:
@UweBonnes This looks like a good change, removing "magic numbers" is a good thing.

PR #641, opened 2020-04-19
Semihosting

koendv:
Implement additional semihosting calls.

### DIFF
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -25,6 +25,7 @@
  */
 
 #include "general.h"
+#include "ctype.h"
 #include "hex_utils.h"
 #include "gdb_if.h"
 #include "gdb_packet.h"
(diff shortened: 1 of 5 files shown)


UweBonnes:
The commit history is very uninstructive. If you revert something introduced outside master, fix when introducing the error and rewrite the history. Make also sure every single commit succeeds in "make all_platforms"

UweBonnes:
Multiple one word commit messages are not instructive.

UweBonnes:
Squashing all commits into a single commit makes thinsg much clearer, beside the points to review.

koendv:
On Tue, 21 Apr 2020 10:32:24 +0000 (UTC)

> @UweBonnes requested changes on this pull request.
> 
> Why not __builtin_bswap?
> 
OK. Using __builtin_bswap32, and rebased. Looks good.


koen

UweBonnes:
Argh, pc-hosted platforms though:
 CC      platforms/pc/gdb_if.c
target/cortexm.c: In function ‘probe_mem_read’:
target/cortexm.c:1043:17: error: cast to pointer from integer of different size [-Werror=int-to-pointer-cast]
  uint8_t *src = (uint8_t *)target_src;
                 ^
target/cortexm.c: In function ‘probe_mem_write’:
target/cortexm.c:1052:17: error: cast to pointer from integer of different size [-Werror=int-to-pointer-cast]
  uint8_t *dst = (uint8_t *)target_dest;
                 ^
target/cortexm.c: In function ‘cortexm_hostio_request’:
target/cortexm.c:1160:39: error: cast from pointer to integer of different size [-Werror=pointer-to-int-cast]
   int rc = tc_fstat(t, params[0] - 1, (target_addr)fio_stat); /* write fstat() result in fio_stat[] */
                                       ^
target/cortexm.c:1181:31: error: cast from pointer to integer of different size [-Werror=pointer-to-int-cast]
   int rc = tc_gettimeofday(t, (target_addr) fio_timeval, (target_addr) NULL); /* write gettimeofday() result in fio_timeval[] */
                               ^
target/cortexm.c:1181:58: error: cast from pointer to integer of different size [-Werror=pointer-to-int-cast]
   int rc = tc_gettimeofday(t, (target_addr) fio_timeval, (target_addr) NULL); /* write gettimeofday() result in fio_timeval[] */
                                                          ^
target/cortexm.c:1199:38: error: cast from pointer to integer of different size [-Werror=pointer-to-int-cast]
   int rc = tc_read(t, params[0] - 1, (target_addr) &ch, 1); /* read a character in ch */
                                      ^

UweBonnes:
Have a loo if my push works for you, then squash again.

koendv:
On Tue, 21 Apr 2020 07:59:06 -0700

> Reply to this email directly or view it on GitHub:
> https://github.com/blacksphere/blackmagic/pull/641#issuecomment-617234041

OK, done. Obviously I did not catch this. I'm using a 32-bit raspberry
pi 4 as desktop, so pointer sizes of target, probe and host match here.

koen

UweBonnes:
After some mangling, i see that your last change did more than sqush my proposal into the PR. At least it removed some magic constants and probably does more error checking, as far as I compared target/cortexm.c. So if we have a common base and you change something, do that  temporary as a new commit to I can easily review the changes and only squash later or  I can squash when I merge.

As the changes seems this is still WIP, I will not merge today so you can still work with and on  your code base.

koendv:
On Wed, 22 Apr 2020 03:17:48 -0700
> ... So if we have a common base and you change
> something, do that  temporary as a new commit to I can easily review
> the changes and only squash later or  I can squash when I merge.
> 
OK. WRITE0 now uses one gdb fileio call for the whole string instead of
one fileio call per character. 

> As the changes seems this is still WIP, I will not merge today so you
> can still work with and on  your code base.

Thank you. My wife is becoming jealous that I am spending more time
with you than with her.

koen

PR #642, opened 2020-04-19
Add 'launchpad-target' platform

WallaceIT:
Add 'launchpad-target' platform, which uses the target part of the [TI Tiva Launchpad](http://www.ti.com/tool/EK-TM4C123GXL) as probe host.

Pins have been chosen in order to use only the outer part of the header (i.e. to use two 1x10 strips instead of two 2x10):

- TMS: GPIOA3
- TCK: GPIOA2
- TDI: GPIOA5
- TDO: GPIOA4
- SWO: GPIOE0
- SWDIO: GPIOA3
- SWCLK: GPIOA2
- SRST: GPIOA6

Moreover, UART5 can be used as debug port and on-board RGB LED as ERROR, IDLE and RUN LEDs.

Compiles with:

```
make PROBE_HOST=launchpad-target
```

The BMP firmware can be loaded to the Tiva Launchpad using the [lm4flash](https://github.com/utzig/lm4tools) utility, without any additional hardware needed.


src/platforms/launchpad-target/platform.c
```diff
@@ -0,0 +1,156 @@
+/*
+ * This file is part of the Black Magic Debug project.
+ *
```
UweBonnes, line 3: Please add your name and contact email  in the copyright.

That means you do at least for now some long time commitment to care for this platform

src/platforms/launchpad-target/platform.h
```diff
@@ -0,0 +1,155 @@
...
+
+inline static void gpio_set_val(uint32_t port, uint8_t pin, uint8_t val) {
+	gpio_write(port, pin, val == 0 ? 0 : 0xff);
+}
+
+inline static uint8_t gpio_get(uint32_t port, uint8_t pin) {
+	return !(gpio_read(port, pin) == 0);
+}
+
+#define disconnect_usb() do { usbd_disconnect(usbdev,1); nvic_disable_irq(USB_IRQ);} while(0)
```
UweBonnes, line 148: Consider writing this as a function. Functions are clearer to read and do more checking.

WallaceIT: Function removed, as it was not used (copy-paste leftover)

src/platforms/launchpad-target/platform.c
WallaceIT: Done. Hope I can find enough time in the future, though.

UweBonnes:
Can I propose following:
- I merge all patches beside the last one that creates the new platform.
To avoid having to care for to many platforms, I propose
- Yyou have a look if you can create the platform as a compiled time option for the existing lm4c platform
- Or your patch gets some dumbs up in the still pending PR  so I see that there is more general need for another platform
If you agree, make another PR without the creation of the new platform, I will merge and than you keep the existing PR open and updated.

WallaceIT:
Got the point about yet another platform, and actually agreeing. What if I merge the existing _launchpad-icdi_ and the new _launchpad-target_ platforms in a single _launchpad_ platform (renaming is required, IMHO) and make the selection between ICDI and target a compile-time option?

UweBonnes:
Can eventually some auto detection been made?

WallaceIT:
> Can eventually some auto detection been made?

I could not find any mean of doing an auto-detection.

BTW, I updated the pull request with compile-time customization of the _launchpad_ platform. Sub-platform selection (so-called 'variant') can be made through the PROBE_HOST_VARIANT make parameter:

- PROBE_HOST=launchpad PROBE_HOST_VARIANT=icdi builds for launchpad-icdi (existent platform)
- PROBE_HOST=launchpad PROBE_HOST_VARIANT=target builds for launchpad-target (new platform)


Let me know what do you think of this solution.

Thank you

UweBonnes:
Please add a Readme.md in the launchpad directory descring things.

UweBonnes:
I do not see explanatorial file in neiter tmc4 nor launchpad. Please add something that others can get an idea of what is implemented there.

WallaceIT:
> I do not see explanatorial file in neiter tmc4 nor launchpad. Please add something that others can get an idea of what is implemented there.

I'm working on it, will push it ASAP. Latest forced push was only for some code cleanup.

UweBonnes:
No need to hurry.

UweBonnes:
Should I merge now or do you still want to work with this profuct for some time and you will trigger me when to merge?

WallaceIT:
Since this is not (by far) a blocking feature, let's wait some more time to double-check the platform merging is all correct. Thank you!

WallaceIT:
Closing, as I did not have time to rebase on current master.

PR #643, opened 2020-04-19
Fixed overrun, when dumping data in swolisten.

wingunder:
The swolisten program failed to print the cbw buffer correctly while
in dump mode. As printf() is used to print the dump, it is expected
that the cbw buffer is zero-terminated, which would only be the case,
if the cbw buffer is initialized with zeros, before filling it with
new data. One could set the entire cbw buffer to zero, but it will be
more efficient to only set the size-th byte to zero.

This patch fixes the above problem, by using the size variable to set
the sixe-th byte of the cbw buffer to zero, before passing it to
printf().

### DIFF
--- a/scripts/swolisten.c
+++ b/scripts/swolisten.c
@@ -442,7 +442,10 @@ int usbFeeder(void)
 	{
 	  unsigned char *c=cbw;
           if (options.dump)
-              printf(cbw);
+          {
+              cbw[size] = 0;
+              printf("%s", (char*)cbw);
+          }
           else
               while (size--)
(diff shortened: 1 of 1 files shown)


UweBonnes:
Can you please sqush into s single commit with propper message? Thanks

wingunder:
Hi @UweBonnes,
> Can you please sqush into s single commit with propper message? Thanks

No problem. It's squashed and the messages are merged into one.
Hope it's ok now.

Regards

wingunder:
Thanks for merging.

PR #644, opened 2020-04-20
Special case value-line F750 flash mapping

rschlaikjer:
The value-line F750 part has the same idcode as a F745, but only has 64KiB of
flash. If we use the same flash mapping calculation for both, the
`remains` value will **underflow**, resulting in a flash size of
`0xfffd'0000` for the region starting at `0x8040'0000`. This causes GDB
to warn

    Overlapping regions in memory map: ignoring

and causes programming of the F750 part to fail.

This patch adds a check on the flash size of the device, and in the case
of the F750 adds the two smaller sectors.
Memory difference information can be seen on page 77 of ST manual RM0385
(Rev 8).

This patch has been verified to work on an F750 part (STM32F750V8T6), and a true F745 part (STM32F745VET6).

### DIFF
--- a/src/target/stm32f4.c
+++ b/src/target/stm32f4.c
@@ -331,29 +331,14 @@ static bool stm32f4_attach(target *t)
 	else
 		banksize = flashsize << 10;
 	if (large_sectors) {
-		// The F750 has the same id code as a F74x, but has a significantly
-		// different memory capacity. If we treat it the same, we will underflow the
-		// size of the final memory region, and generate a close to 4GiB mapping.
-		// To avoid this, match on it's flash size (64KiB)
-		const bool is_stm32f750 = flashsize == 0x40;
-		if (is_stm32f750) {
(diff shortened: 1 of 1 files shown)


UweBonnes:
Can you please try to ignore to read out what the device says and test to use the full flash size of the F745? I have come to the opinion that the user via his linker script should care for the flash size, not the programmer. Probably the whole flash is there on a F750, but untested.

rschlaikjer:
Looks like it does program and compare-sections happily past the supposed flash size on the F750.
However, I do still need the underflow check to prevent gdb getting upset at the overlapping memory regions and failing to program at all.

UweBonnes:
Can you please test #645

rschlaikjer:
#645 seems to work fine for me as well, feel free to close this and merge that.

UweBonnes:
Superseeded by #645

Issue #635, opened 2020-04-06
It is nescessary to fine decode FLASH size?


UweBonnes:
Hello,

today, two PRs popped up that tackle with flash size for different chip variant in some chip family. We go into great deapth to decode these size. However most vendors produce only one chip and on the variants sold with less flash, a smaller flash size is announced and the flash is probably untested but mostly functional. 
By setting up the memory map according to the sizes announced by vendor, we inhibit eventual use of that flash, if the user decides to use this flash at his own risk.

I propose that we announce the highest available size for a variant and drop the fine decoding for the memory sizes.

Any opinions?

sidprice:
I think allowing access to the full flash is a good idea. Ideal would be to check the vendor size and if flashing fails let the user know this could be the reason. However, I don't think there is really a way to do this, other than through debug messages.

UweBonnes:
Done with #645. StTM32L4 still does flashsize decoding.

PR #633, opened 2020-04-06
Allow flashing all of the H750's first flash bank.

hydra:
See also #632

PR #645, opened 2020-04-21
stm32f4/7: Always use largest flashsize for device family (#633, #635…

UweBonnes:
…, #644)

Do not care for the FLASHSIZE register. Leave it up to the user to abuse
flash area the ST did not announce.


hydra:
tested on STM32H750VBT6 OK.

PR #651, opened 2020-04-30
semihosting

koendv:
Remaining semihosting calls: SYS_CLOCK, SYS_EXIT_EXTENDED, SYS_HEAPINFO, SYS_ISERROR, SYS_TMPNAM and a few bugfixes.

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -59,6 +59,7 @@ static bool cmd_target_power(target *t, int argc, const char **argv);
 #ifdef PLATFORM_HAS_TRACESWO
 static bool cmd_traceswo(target *t, int argc, const char **argv);
 #endif
+static bool cmd_heapinfo(target *t, int argc, const char **argv);
 #if defined(PLATFORM_HAS_DEBUG) && !defined(PC_HOSTED)
 static bool cmd_debug_bmp(target *t, int argc, const char **argv);
 #endif
@@ -83,6 +84,7 @@ const struct command_s cmd_list[] = {
 	{"traceswo", (cmd_handler)cmd_traceswo, "Start trace capture, Manchester mode" },
 #endif
 #endif
+	{"heapinfo", (cmd_handler)cmd_heapinfo, "Set semihosting heapinfo" },
 #if defined(PLATFORM_HAS_DEBUG) && !defined(PC_HOSTED)
 	{"debug_bmp", (cmd_handler)cmd_debug_bmp, "Output BMP \"debug\" strings to the second vcom: (enable|disable)"},
 #endif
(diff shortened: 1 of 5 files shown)


koendv:
Note: if the target wants to read the special filename ":semihosting-features" to know what semihosting features are supported, it's easiest to create that file on the host in the directory where gdb runs:
```
$ echo -e 'SHFB\x03' > ":semihosting-features"
$ chmod 0444 ":semihosting-features"
```

UweBonnes:
I lost track of your patches. Is this the only PR I should apply or are there more? Furthermore the commit message should tell about the errors corrected introduced with older patch I aoplied.

koendv:
On Fri, 01 May 2020 02:11:37 -0700

> I lost track of your patches. Is this the only PR I should apply or
> are there more? Furthermore the commit message should tell about the
> errors corrected introduced with older patch I aoplied.
> 

This is the only PR you should apply.

The errors corrected are:
SYS_READ: return value incorrect if at eof.
SYS_WRITE: return value incorrect if nothing was written.
SYS_SEEK: return value now 0 if success, -1 if error.
SYS_RENAME, SYS_REMOVE, SYS_SYSTEM: arguments are strings, not file handles.
SYS_READC: always read from STDIN, not from file handle. (bug of mine)

The following additions:
SYS_CLOCK: return time in 1/100 of seconds. code for SYS_CLOCK uses same code as SYS_TIME (time in seconds); more compact.
SYS_EXIT_EXTENDED: _exit() with 64-bit value.
SYS_ISERROR: true if argument is error code. error numbers in target.h synced with gdb. 
SYS_TMPNAM: return temporary file name.
SYS_HEAPINFO: add "heapinfo" command to set heap and stack values in newlib/libc/sys/arm/crt0.S 

This ends adding functionality to semihosting. I hope the focus is now on using instead of patching.

Thank you.

koen

UweBonnes:
Does SYS_HEAPINFO also cope with nanolib?

koendv:
On Fri, 01 May 2020 04:50:17 -0700

> Does SYS_HEAPINFO also cope with nanolib?
> 

If you look at crt0.S, does it have code like this?

#ifdef ARM_RDI_MONITOR
	/*  Issue Angel SWI to read stack info */
	movs	r0, #AngelSWI_Reason_HeapInfo
	adr	r1, .LC0	/*  point at ptr to 4 words to receive data */
#ifdef THUMB_V7M_V6M
	bkpt	AngelSWI
#elif defined(__thumb2__)
	/*  We are in thumb mode for startup on armv7 architectures. */
	AngelSWIAsm	AngelSWI
#else
	/*  We are always in ARM mode for startup on pre armv7 archs. */
	AngelSWIAsm	AngelSWI_ARM
#endif

That is the code to call SYS_HEAPINFO, for the different arm architectures. 
Needs a define to get compiled.
 

koen

koendv:
Answer: yes, it'll work with newlib-nano. To use heapinfo you need to link with  rdimon-crt0.o instead of crt0.o.

koendv:
Thanks. Hope the users find it useful.

PR #652, opened 2020-04-30
cdcacm: Fix UART interface OUT endpoint MPS in descriptor.

### LINKED ISSUE #538: BMP serial interface ignores write (PC->device) bigger 16 byte.
When writing blocks of > 16 bytes without delay between character, BMP ignores the whole block.

zyp:
3e3e450 reduced the buffer size for the UART interface OUT endpoint to
32B, but left wMaxPacketSize at 64B. This effectively configures the
hardware to receive packets up to 32B in size while telling the host it
can send larger packets. On reception of a larger packet, the usb core
treats it as an error, drops the packet and replies with a STALL
handshake.

This commit fixes the issue by updating the descriptor to match the
buffer configuration.

Fixes #538

### DIFF
--- a/src/platforms/common/cdcacm.c
+++ b/src/platforms/common/cdcacm.c
@@ -188,7 +188,7 @@ static const struct usb_endpoint_descriptor uart_data_endp[] = {{
 	.bDescriptorType = USB_DT_ENDPOINT,
 	.bEndpointAddress = 0x03,
 	.bmAttributes = USB_ENDPOINT_ATTR_BULK,
-	.wMaxPacketSize = CDCACM_PACKET_SIZE,
+	.wMaxPacketSize = CDCACM_PACKET_SIZE / 2,
 	.bInterval = 1,
 }, {
 	.bLength = USB_DT_ENDPOINT_SIZE,


UweBonnes:
The old commit had a remark about problems with SWO with the normal CDC ACM buffer size. Did you have a lot at that ?

zyp:
No, I haven't looked at the actual buffer allocations, I just assumed that the old commit fixed the problem it claimed to.

Since this commit only changes the endpoint descriptor to match the new buffer size, it's just telling the host not to send packets larger than what can fit. I'm not changing any actual buffer sizes and therefore there's no risk of regressions elsewhere.

zyp:
Looked over the buffer sizes now:

The buffer memory is 512B in total and the first 64B is consumed by the buffer descriptor table. In the original setup 2x64B is consumed by the control endpoint, each CDC ACM interface uses 2x64B+16B and the trace interface uses 64B, bringing the total to 544B.

After 3e3e450 reduced one of the bulk endpoints to 32B everything exactly fits and all that is required now is this fix to tell the host not to send packets larger than 32B on that endpoint.

UweBonnes:
Please ammend this explanation to the commit message.

zyp:
I feel that the commit message already contains a thorough explanation of the current problem and the provided solution. As the current buffer sizes are not a problem and this commit doesn't change them, the commit message is not an appropriate place to elaborate on them.

UweBonnes:
Sorry, did nit read carefully enough.

PR #659, opened 2020-05-09
Rename variant_string in efm32 samd samx5x

alexxy:
Files efm32 samd samx5x uses same function name that collides during
linking (checked with gcc10)

Signed-off-by: Alexey 'Alexxy' Shvetsov <alexxyum@gmail.com>


UweBonnes:
Shouldn't those stringa be also static?

alexxy:
> Shouldn't those stringa be also static?

Yep it should. And now they are =D

PR #660, opened 2020-05-11
Port gdb.py and hexprog.py to Python 3.

markrages:
Unfortunately, the serial data are not 7-bit clean (see write_mem()).
So getpacket() and putpacket() use bytes objects rather than strings.

### DIFF
--- a/scripts/gdb.py
+++ b/scripts/gdb.py
@@ -24,313 +24,313 @@ import struct
 import time
 
 def hexify(s):
-    """Convert a bytes object into hex bytes representation"""
-    return s.hex().encode()
+	"""Convert a bytes object into hex bytes representation"""
+	return s.hex().encode()
 
 def unhexify(s):
-    """Convert a hex-encoded bytes into bytes object"""
-    return bytes.fromhex(s.decode())
+	"""Convert a hex-encoded bytes into bytes object"""
+	return bytes.fromhex(s.decode())
 
 class FakeSocket:
-    """Emulate socket functions send and recv on a file object"""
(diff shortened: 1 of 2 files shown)


UweBonnes:
Why all these double braces with the print arguments? I do not like if wide spread whitespace fixes are mixed with code changes. Anyway, I will checkout the patch and test an if successfull I tebd to apply as is.

markrages:
The hexprog.py is a quick and dirty port with `2to3`.  Some of the extra parens are added by `2to3`.

This port doesn't attempt to address any of its shortcomings of `hexprog.py`.  It has a three-second delay hardcoded into its startup. It uses a hand-rolled hex parser, rather than something like https://pypi.org/project/IntelHex/.  The feeble hex parser cannot read all the hex record types emitted by modern binutils.

Python 3 code is also just uglier than Python 2 code, and it really shows in lowlevel programs like these.

The indentation changes are courtesy of my editor.  Python 3 has disallowed smart tabs, so any indentation is a compromise.  I will update the PR to use tabs, which appears to be the local style for blackmagic Python scripts.

UweBonnes:
I would prefer intelligent porting, not porting by some dump translator. dfu-convert.py should be the first target, as make all-platforms invoked it before 4ede113ca76d8c84b3e73a5b1b2f7b21a76fb719

esden:
I have merged this PR with some minor changes with f43101fd9fc4a9aa67e24a9b54f1780b39c7724b
Thank you very much for your contribution @markrages :D

PR #662, opened 2020-05-13
set semihosting sys_clock time origin

koendv:
The semihosting sys_clock is a 32-bit timer that counts in 1/100 second. This patch sets the time origin of sys_clock to the second of first use. This way the counter wraps around after 16 months of uptime; sufficient.

### DIFF
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -71,6 +71,10 @@ static target_addr cortexm_check_watch(target *t);
 
 static int cortexm_hostio_request(target *t);
 
+#if !defined(PC_HOSTED)
+static uint32_t time0_sec = UINT32_MAX; /* sys_clock time origin */
+#endif
+
 struct cortexm_priv {
 	ADIv5_AP_t *ap;
 	bool stepping;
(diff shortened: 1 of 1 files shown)


koendv:
On Wed, 13 May 2020 06:04:10 -0700

> @UweBonnes commented on this pull request.
> 
> Please use 
> #if PC_HOSTED == 0
> or 
> #if PC_HOSTED == 1
> to avoid clash with pending patches, here
> https://github.com/UweBonnes/blackmagic/tree/multi
> 

OK, done.

UweBonnes:
Sorry, I gave the wrong advice. PC_HOSTED  is not yet defined with a value. I have to do that change after merging your original patch when I rebase my patch. Please drop the second patch.

Otherwise, if a branch adds new features and steps forward and perhaps back around that feature, please squash into one commit.

koendv:
OK, done. 
BTW: What do you suggest if a patch looks promising, but needs to be tested whether the patch works on platforms you don't have?

UweBonnes:
One test it to run "make all_platforms" This ensures that things compile at least and bisect can run clean. Otherwise, ask for review and test on discord or in the issues/PRs. If I notice, that the patch touches fundamental things, I hopefully will notice and have a look too. If a patch only touches a peripheral part of BMP, requirements are relaxed.

koendv:
OK, thanks. Opened issue; https://github.com/blacksphere/blackmagic/issues/663#issue-617593459

PR #664, opened 2020-05-14
traceswo decoding

koendv:
This patch decodes traceswo in the probe itself. The output goes to the usb serial port.
Adds "decode" option to "traceswo" command.
Increases stlink firmware size by 360 bytes, or by 464 bytes if ENABLE_DEBUG=1.
Note: only tested on stm32f103/stlink.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,9 @@ JTAG or Serial Wire Debugging (SWD) port and on-chip debug logic provided
 by the microprocessor. The probe connects to a host computer using a
 standard USB interface. The user is able to control exactly what happens
 using the GNU source level debugging software, GDB.
+Serial Wire Output (SWO) allows the target to write tracing and logging to the host
+without using usb or serial port. Decoding SWO in the probe itself
+makes [SWO viewing as simple as connecting to a serial port](https://github.com/blacksphere/blackmagic/wiki/Serial-Wire-Output).
 
 See online documentation at https://github.com/blacksphere/blackmagic/wiki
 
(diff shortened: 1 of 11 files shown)


UweBonnes:
Probably a naive question: Why does the decoded output go to the serial port and not the trace port?

koendv:
No device driver or special program is needed to read text from a serial port.
You could send the decoded swo stream to the trace port, but then you would still need a program to read the text.

Some answers to questions:
Decoding in the probe itself also lowers usb traffic. If the target sends characters one at a time (common), usb traffic drops 50%. If the target sends 4 characters at a time, usb traffic still drops 20%.

Throughput. A simple loop printing the numbers from 1000000 to 2000000 at 2250000 baud shows no errors. Throughput is 180 kbyte/s. 

Receiving 360 kbyte/s of text continuously at 4500000 baud, distributed over 8 stimulus channels, is too much for a blue pill. (traceswo packets of 1 byte header + 4 bytes payload, back to back)  
Results depend upon how much you want to decode. Measurement result:
#channels decoded |%correct
---|---
1|99.97
2|98.87
3|97.96
4|98.25
5|97.38
6|87.44
7|87.88
8|95.65

UweBonnes:
Please take above patch as aa example for a  start and add more where usefull. Squash when ready and request merge and get me to notice.

koendv:
I've added a page to the wiki (without anything linking to it yet):
[Serial Wire Output](https://github.com/blacksphere/blackmagic/wiki/Serial-Wire-Output)

PR #668, opened 2020-05-27
traceswo flush

koendv:
This patch implements the following: when decoding traceswo, a write to stimulus channel 30 flushes usb output buffers.
- Text written to channel 30 itself is printed as soon as it is received. This makes channel 30 "unbuffered", suitable for e.g. error messages. 
- This allows implementing flush() by writing zeroes to channel 30. 

Because the usb packets sent are smaller than maximum size, low latency comes at the cost of lower throughput.


UweBonnes:
Can that eventually be handled by a timeout?

koendv:
With different interrupt levels for timer and swo serial port dma? Yes, that would be an interesting solution. This patch does the job in 88 bytes.

UweBonnes:
I have the remarks from Mube in mind.

PR #672, opened 2020-06-03
Add support for swlink for f4 boards

alexxy:
Currently black pill aka WeAct f411 tested

Signed-off-by: Alexey 'Alexxy' Shvetsov <alexxyum@gmail.com>


UweBonnes:
Does this platform introduce any new concepts? If not, I tend to not apply to master to avoid having too may similar platforms. If you notice their is substantial interest in your platform, please  let me know.

alexxy:
Actualy there are two concepts:
1. allow to use [cheap f4 board](https://github.com/WeActTC/MiniF4-STM32F4x1) (~2-3$) for debugger. Its became as popular as bluepill
2. I plan to use similar chip for debugger board (actulay this platform was used for testing). This board will be released as opensource (GPLv3)

In principle i can merge this platform to swlink one (with many additional ifdefs).

alexxy:
Another possibility is to merge it together with f4discovery platform into swlink-f4

UweBonnes:
The two concept you mention are not new in BMP:
1. The Bluepill is also a cheap board. What advantage has the  F4 board against a bluepill beside more flash?
2. What disadvantage is it when your special interest board is only available in your fork? If more interest in it evolves, we can than pull that fork into the main branch.
 
Too many (similar) platform make it harder for beginners to start. So I still have no good reason to merge.

alexxy:
Good teason to merge, is ability to consolidate 3 platforms under one hood. That i mean saying this: ican merge into swlink-f4 f4disco and hydrabus. Since they are similar. And add blackpill

PS can you describe what is a good reason to merge?

UweBonnes:
A good reason for a new platform is to implement something new improving BMP speed/functionality or to support something used by many other people.
Wanted new features are:
- USB HS. The STLinkV3 mini could be a good target. TinyUsb STM32 HS support is not yet ready.
- Fast JTAG/SWD with low impact timing option. Using STM32 SPI up from STM32F3 in 4 bit bi-directional mode could perhaps be used for SWD. For JTAG I have no good idea yet
- FAST SWO: You be done with fast timers/uart
- Multi Bit SWO: No idea yet beside by using a small FPGA.

Each existing platform should be treated somehow as legacy, so merging exising platforms should have good reasons. I think replacing the different PC-Hosted platforms be a single. mor capable one had good reasons.

Adding a new platform should also be a long-time commitment by the author.

B.T.W: Why do you mention blackpill here. What is wrong treating it as a bluepill with swlink?

alexxy:
Blackpill from weact is f411 based board, not f103

tzarc:
Haven't yet tried this PR on hardware, but one other aspect to this could be that the F411's have 512kB flash and onboard ROM USB DFU, negating the need to mess about with the libopencm3-based DFU bootloader.

Given that the F103 target has sometimes required slimming down the binary to get it to flash (especially on ones that actually only have 64kB flash), the 512kB should allow for the extra headroom for other chip support. In the past I've had to rip out specific targets to get the binary to fit on the only-64kB bluepill clones, and this would've been a godsend at the time.

UweBonnes:
No failure has been reported yet when flashing a STM32F103x8 as a STM32F103xB with scripts/stm32_mem.py. And binary size is still well below the flash limit having still rooms for improvement.

As this PR does not introduce new concepts, I will not add it as a separate platform to master.

Merging this branch with F4discovery would be a good option, best with auto detection of the platform, so that make all_platforms checks the code for both platforms. F4discovery was meant as a bootstrap solution to get BMP-stlink on the stlink of the Disco board and so probably is not used often. However it shows how to handle the internal bootloader and so has some reason to exist.

Autodetect could look for Pullup on PC0 (OTG_FS_PowerSwitchOn)  or I2C or I2S rwaction from the chip on the disco board.

tzarc:
That's curious -- I've previously needed to produce binaries for the ST-Link target, explicitly removing things like JTAG and Cortex-A to get the binary to actually fit. I guess I'll have to revisit this at some point instead.

UweBonnes:
Read platforms/stlink/Flashsize_F103

For StlinkV21 I propose:
- Compile  "make PROBE_HOST=stlink ST_BOOTLOADER=1"
-  Clone and compile https://github.com/jeanthom/stlink-tool.git
- Run "stlink-tools blackmagic.bin" to flash the stlink. Run "stlink-tools " to activate the ST-Link.
You can now switch between original STLink firmware and BMP-STLink.

This has however the backdraw that after replug, either "stlink-tool" has to be run again or that BMP-Hosted has to be run twice to get into bmp-stlink mode.

With the BMP bootloader this additional step is not needed. If you already have the BMP loader, probaly keep it.

With BMP bootloader:
- make PROBE_HOST=stlink 
-../scripts/stm32stm32_mem.py blackmagic.bin

UweBonnes:
Another drawback of the f4 series is that F4 USB-OTGFS it is missing one USB endpoint to be fully BMP compatible. If you want to go for more flash, consider the F723 and reflashing the STLINKV3-MODS. There is work pending from from Shopov and me.

Otherwise does the patch consolidate or add a new platform?

alexxy:
I think i merge f4disco and f4swlink. 

About f723. Yep i think about board like this or may be something from h7xx

unrelentingtech:
So 5eb43a1ddbb8e5b6aeb51bc8b82bcd4649384d98 has added an f4 blackpill to f4discovery.. except for some reason using PA13 & PA14, precisely the pins missing on the WeAct boards o_0 (What other f4 black pills are there?)

But now it is enough to just change the pins to something accessible:

```diff
--- i/src/platforms/f4discovery/platform.h
+++ w/src/platforms/f4discovery/platform.h
@@ -56,8 +56,8 @@
 #define TCK_PORT JTAG_PORT
 #define TDO_PORT GPIOB
 #define TDI_PIN GPIO1
-#define TMS_PIN GPIO13
-#define TCK_PIN GPIO14
+#define TMS_PIN GPIO5
+#define TCK_PIN GPIO6
 #define TDO_PIN GPIO3
 
 #define SWDIO_PORT JTAG_PORT
```

and build with `PROBE_HOST=f4discovery BLACKPILL=1` :)

DX-MON (changes requested):
Please do not commit your Kate temporary files - src/platforms/swlink-f4xx/.platform.c.kate-swp has no business being in this PR.

Otherwise, this looks fairly reasonable to us.

UweBonnes:
The merge between f4disco and f4swlink has already happened and this request is outdated. Any more F4 variants are done best as platform variants, like the (f4) blackpill variant, with least new code added.Going to close this PR as outdated.

DX-MON:
Ah, having re-checked the history on this, yeah - ok. Closing the PR.

PR #690, opened 2020-07-12
Add ROM table support for Cortex-M33 and Cortex-M23

themadinventor:
This patch adds support for handling the DEVTYPE and ARCHID values in the ROM tables, so that ADIv6 debug components can be correctly identified.
It also updates the core identification to handle M23 and M33, and also fixes a bug where the probe would time out while polling in cortexm_forced_halt, due to new status bits (S_SDE).

This branch is NOT verified to fully support debugging in secure and non-secure mode. Additional work may be needed.

Also, flash driver support for relevant targets (STM32L5, LPC55S6) need to be implemented in separate patches.


UweBonnes:
Argh, some existing PIDRs also have DEVTYPE and ARCHID.

themadinventor:
Ouch, sorry! I was thinking about whether that would happen, but it seems like pyOCD matches all fields exactly.
Perhaps `dev_type` and `arch_id` should not be compared if they are zero in the `pidr_pn_bits` table, so it'll behave as before in those cases.

UweBonnes:
I have one patch ommiting the check if the fileds are zero and one patch adding the missing fields. Not sure, what patch to apply...

PR #700, opened 2020-07-19
adiv5: fix debug print of dev_type

WallaceIT:
Since dev_type is an 8 bit unsigned integer, use the PRIx8 macro instead of PRIx32.


UweBonnes:
For printing a byte , we do not need the UINT8_T type, normal %x is enough, to my experience.

WallaceIT:
I agree; however, PRIx32 is not always defined as  %x, but can also be %lx. In the latter case, trying to use it to print a byte triggers a compiler error.

PR #709, opened 2020-08-03
Find by serial number for hosted windows and use hidapi static for windows

UweBonnes:


sidprice:
Seems to find the BMPs and USB-connected ctxLinks on Windows 10. :)

PR #711, opened 2020-08-14
Use correct IAP entry address for LPC84x

dlaw:
LPC84x uses a different address for IAP entry function compared to other LPC microcontrollers. Flash operations will not work unless we use the correct IAP entry address.

2020-08-14-130147_1700x917_scrot

### DIFF
--- a/src/target/lpc11xx.c
+++ b/src/target/lpc11xx.c
@@ -29,7 +29,8 @@
 #define MIN_RAM_SIZE            1024
 #define RAM_USAGE_FOR_IAP_ROUTINES	32	/* IAP routines use 32 bytes at top of ram */
 
-#define IAP_ENTRYPOINT	0x1fff1ff1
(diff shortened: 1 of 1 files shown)


UweBonnes:
Can you program multiple times? The setup "Hosted/BMP-Native -> LPC845 Breakout" only flashed successful after reset. Flashing another time without reset hangs in the flash iap call. The setup "Hosted/On-Board CMSIS DAP -> LPC845 breakout" already hangs in the erase IAP call.

dlaw:
Yes, I've programmed the same board about 20 times so far. I am not sure what you mean by "flashing without reset" as I believe that in my normal GDB workflow the reset signal is automatically asserted each time the target is attached.

UweBonnes:
I tested with BMP/hosted. cl_utils.c needs to reset also for erase and flash!

PR #714, opened 2020-08-21
Fix macos compilation for hosted platform

tech2077:
A few small fixes to get BMP hosted working on macos.

- The libftdi version on brew deprecated `ftdi_usb_purge_buffers` and specified `ftdi_tcioflush` as a drop in replacement
- Added in serial auto-connection using the macos /dev/tty.usb<serial> naming scheme
- Removed unnecessary flags and libraries from darwin in hosted makefile
- Fixed top level makefile to match new platform name


UweBonnes:
ftdi_tcioflush  is not yet widely available. 
On my Opensuse 15.2:
> grep ftdi_tcioflush /usr/include/libftdi1/ -r
finds nothing.

I.m.h.o. either we pull in libftdi static or we live with the wrong ftdi_usb_purge_buffers for some time

Please remove the libdtdi thing from this PR, perhaps making it a new issue pr PR. I think "one subject in one PR".

UweBonnes:
Did you run multiple invocations of e.g. blackmagic -t with success? #728 impies that opening ttyusbmodem hangs infinite for the second time and that  cumodem  does not show that error. Comments welcom

tech2077:
I just ran it and saw the hang. I changed it to cumodem and it no longer hangs. I'll do a PR to add this fix.

esden:
@tech2077 just FYI, we got the fix submitted by @djix123 so the problem should be solved now.

PR #721, opened 2020-09-04
SWD Timing: Enhance SWD high time with low speed loss.

UweBonnes:
On F103 no high time < 80 ns and no low time < 100 ns was seen with a scope


JojoS62:
> On F103 no high time < 80 ns and no low time < 100 ns was seen with a scope

Isn't this due to the 10 MHz setting for the SWDIO_PIN ? How about using the 50 MHz setting for this pin?

UweBonnes:
No.  The MHz setting you mention changes the slew rates. And you want slew rates as slow as possible when going offboard or with linger on-boards not impedance controlled lines to avoid reflection and ringing. So 10 MHz is a good compromise.

JojoS62:
Hmm. The init sets to 50 MHz, the FLOAT/DRIVE Macros override it with 10 MHz (100 ns).

PR #735, opened 2020-09-20
Symbolically define the USB endpoint addresses and enable them to be overridden by platform if required

sidprice:
If a platform requires the use of different USB endpoints they may be defined in platform.h for the platform.

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -415,7 +415,7 @@ static bool cmd_traceswo(target *t, int argc, const char **argv)
 	traceswo_init(swo_channelmask);
 #endif
 	serial_no_read(serial_no, sizeof(serial_no));
-	gdb_outf("%s:%02X:%02X\n", serial_no, 5, 0x85);
+	gdb_outf("%s:%02X:%02X\n", serial_no, 5, USB_TRACESWO_EPT_OUT);
 	return true;
 }
 #endif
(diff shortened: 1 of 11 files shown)


Vestrel:
I think names should also reflect endpoint type (in/out)

UweBonnes:
What about e.g. xxx_EP_OUT and xxx_EP_IN?

sidprice:
@UweBonnes Good idea, just because I ALWAYS confuse In and Out, would this be correct?

#define CDCACM_GDB_DATA_EP_IN 0x81

UweBonnes:
0x80 + x seesm to be the IN-EPs. Recheck with lsusb -v

sidprice:
@UweBonnes Names changed to reflect IN/OUT

UweBonnes:
"make all_platforms" stops in launchpad-icdi

sidprice:
Sorry, I thought I had run each platform build, I was not aware of "make all_platforms", however, on Windows that does not work. I will create a new issue for that. All platforms now build.

Note I do not have hardware to test them so we should probably request someone who has launchpad-icdi, hydrabus, stlink, and swlink to check the builds.

UweBonnes:
I also do not have all platforms and do not test all platforms. But at least all platforms should compile, so that a "git bisect" can succeed it at a later point in time when somebody with that platform finds a problem.  Your code looks like  a simple substitution without code change, so chances for colateral damage are small (famous last words ;-)

sidprice:
I agree ;-)

sidprice:
I would like to try and prepare a PR for ctxLink and having this PR merged is necessary before that because ctxLink uses the end point names.

sidprice:
@dragonmux It looks like my PR was covered by #956 and already merged. So, this PR can be closed, I think.

dragonmux:
Yeah, seems reasonable - closing

PR #747, opened 2020-10-01
stm32h7: Don't tc_printf from flash functions

gsmcmullin:
Receving an 'O' packet while flashing confuses GDB and then weird stuff happens.


UweBonnes:
I think the message should not get dropped totally but replace against DEBUG_WARN. Will you do or should I?

And nice to hear from you again!

Issue #740, opened 2020-09-30
usb icdi error windows 10


daguirrem:
![image](https://user-images.githubusercontent.com/49079545/94641245-56dbbc80-02a6-11eb-8fd1-a60a91ea2701.png)
The launchpad-icdi version appears as unknown device on device manager on windows 10 with the latest version of firmware, i have a older version (i don't know exactly which one) that works well

UweBonnes:
Is the the same PC setup as with the succeeding tests earlier? Is the blackmagic.inf file installed? Does the icdi  work on a linux box?

daguirrem:
Yes, i test the 2 firmwares at same time (and on same pc) and the older works fine, i think that can't probe on a VM becouse Windows alerts me like when we connect a damage usb, directly can't update (or do something) with the driver and the device.

### The older version that works (with the number of firmware):
![image](https://user-images.githubusercontent.com/49079545/94694221-3a696f80-02fa-11eb-919f-57ee99282697.png)

### The new version (unknow device) (with windows event report) :
![image](https://user-images.githubusercontent.com/49079545/94694818-dabf9400-02fa-11eb-8e10-2b985ec15b2f.png)
![image](https://user-images.githubusercontent.com/49079545/94695084-1ce8d580-02fb-11eb-96a2-5f1a1b81f90d.png)
![image](https://user-images.githubusercontent.com/49079545/94695109-22462000-02fb-11eb-99b5-2e78a99d5d1d.png)

UweBonnes:
As asked above, check that driver/blackmagic.inf is installed

daguirrem:
is the .inf necesary for windows 10? when i try install appears error: "Third-party INF does not contain digital signature information" and the wiki just specify the driver for windows 7 and older

UweBonnes:
Probably not, but I do not work on windows. Compiles some old firmware, load it and check if that works. If you are sure it is an error with the firmware, do a bisect.

esden:
As the icdi usb error patch is merged now I assume this issue is resolved? Closing.

Issue #693, opened 2020-07-14
Platform "stlink" for ST-Bootloader has problems enumerating with bMaxPacketSize0 = 8


UweBonnes:
When compiling "make PLATFORM=stlink ST_BOOTLOADER=1" , un/replug or "dfu-util -e" brings up the bootloader. After flashing a new firmware with "stlink-tool blackmagic.bin" or starting the old firmware witk "stlink-tool" , reenumeration took some time. Bisect showed that [34a13723d86cdf4f75182d751149445f24bfa07f] Decrease the control USB pipe size on f103 blackmagic probes to 8 bytes started the problem. ".bMaxPacketSize0 = 16" does now work, ".bMaxPacketSize0 = 32" works.

UweBonnes:
Fixed by #694

PR #748, opened 2020-10-07
Fix icdi usb error

daguirrem:
(https://github.com/blacksphere/blackmagic/issues/740)


UweBonnes:
Are you sure this is an error with BMP and not the libopencm3 implementation for USB? Did you have a look at #693 and  #697?

daguirrem:
well, i don't sure, but before commit (https://github.com/blacksphere/blackmagic/commit/34a13723d86cdf4f75182d751149445f24bfa07f) the icdi works fine, after that, the system don't recognize the icdi-bmp, i understand that the commit above mentioned is for "save" memory for usb traffic, in f103 case it have a 'only' 512 bytes, on the otrher hand, the tm4c123 have up to 4kb of dedicated memory for endpoints and maybe this decrease is not necesary for icdi

UweBonnes:
Please amend the reference to this discussion in your commit message additional to #740

PR #752, opened 2020-10-11
Added option to control halt on attach

sidprice:
This option allows BMP to attach to a running target without halting the target.

The default is for the option to be disabled, giving the current behavior without user intervention.

Syntax is mon no_halt [enable|disable]

Command "mon no_halt" reports the state of the option.


UweBonnes:
How is that used? Some examples? And, excuse my pickuness, unrelated whitespace changes in the patch. Check with "git log -p" versus "git log -p -w"

sidprice:
>>> How is that used? 

It permits BMP to att5ach to a running target, memory locations on the target may be non-invasively read. If you look at the tool STM32Monitor, it shows how ST use this feature in their STLink to monitor a running system.

>>> unrelated whitespace

Sorry about the whitespace, I need to find the option on VSCode that turns off code reformatting on save.

UweBonnes:
We need feedback from gdb gurus...

sidprice:
Sure, I will be continuing with a cross-platform UI that uses this. Having used the feature with STLink and STM32Monitor I find it of great value and one I would like to have for BMP/ctxLink.

UweBonnes:
Please describe how to use. I tried in in some ways, but can not get to the gdb prompt with the target running.

sidprice:
tar ext <port>
mon no_halt en
mon swd
...
att 1
...

Now target memory may be read with target running

Then you can read memory from the target.

UweBonnes:
Running with stlink/firmware, I can not read:
(gdb) detach
Detaching from program: /devel/en_apps/gpio/nucleo_l073rz/gpio.elf, Remote target
[Inferior 1 (Remote target) detached]
(gdb) tar ext /dev/ttyACM0
Remote debugging using /dev/ttyACM0
(gdb) mon no en
Enabling no halt
(gdb) mon s
Target voltage: 3.30V
Available Targets:
No. Att Driver
 1      STM32L0x M0+
(gdb) att 1
Attaching to program: /devel/en_apps/gpio/nucleo_l073rz/gpio.elf, Remote target

Program stopped.
0x00000000 in ?? ()
(gdb) x /4x 0x0000000
0x0:    0x00000000      0x00000000      0x00000000      0x00000000

(gdb) mon no dis
Disabling no halt
(gdb) mon s
Target voltage: 3.30V
Available Targets:
No. Att Driver
 1      STM32L0x M0+
(gdb) att 1
Attaching to program: /devel/en_apps/gpio/nucleo_l073rz/gpio.elf, Remote target

Program stopped.
NutThreadYield () at /devel/ethernut/nut/os/thread.c:247
247     {
(gdb) x /4x 0x0000000
0x0:    0x200002c0      0x08000081      0x20000433      0x20000439

What was your setup?

sidprice:
@UweBonnes I am busy on another project at this time, I will get back to this as soon as I can and let you know what I find, So far, all my testing was done with BMP native.

UweBonnes:
No hurry!

UweBonnes:
Seeing the https://embdev.net/topic/505517#new announcement, I wonder if bmp/hosted is a better  place for online variable view than your approach. 
With e.g. -O 4@0x20003854 - O 12@0x20004310 arguments when started (O for on-line view), hosted would open another socket and when idle would read these values via SWJ and exposes them on the socket. Or perhaps another socket listens to the data packets like the -O packets above. Perhaps threads would be helpfull with that task
Does that idea make sense?
However: I have few experience  with socket programming and even less with PC-thread

sidprice:
@UweBonnes This sounds like a good idea because, if I understand what you wrote, it would mean that one could have a GDB debug session open to the BMP with full debug active and a monitor app connected using the live memory features.

I need to study this a little and think about it ... thanks.

UweBonnes:
What is the update rate of those variable you like to watch?

sidprice:
I do not have specific variables I am monitoring. I am try to write an application that, using ctxLink/BMP, monitors and displays variables from the target MCU. The extensive STM32CubeMonitor is an example of such a program, it is specific to ST MCUs of course. My application will, at least initially, be much simpler and without the graphic displays.

My initial application offers a ranger of sampling speeds from 10Hz and slower. Later I may test it at higher rates. For now I have other work on it that needs to be done. It is a Python application and once I have a stable prototype I will probably open source it.

sidprice:
I am not sure that what you suggest is the way forward. I will do some more research. 

I have a prototype UI written and basically working. It needs more work before it is ready for general usage. However, such a tool is invaluable.

UweBonnes:
As I understand you want to have something like openocd has on the telnet port 4444 e.g. "mww" for memory write word and friends. After probing, BMP can already access memory  and DP and AP registers, without the need to halt the target. A set of commands like "monitor mwv <target> <address> <value>" would be needed. BMP hosted can also get extended to act as a server to provide the needed access. Finally implementing GDB MI interface would also be a clean approach, but probably implementing MI interface is a major effort.

All this sounds clearer to me than faking a state with attach that  gdb does not expect.

sidprice:
The change proposed is quite small and non-intrusive to the current operation of BMP. It is one added command that does not affect current operation or user experience. As you say, implementing GDB MI would be a major effort.

Pus, one of the attractions of BMP is that intermediate software is not needed to be run between a client app and the debugger. Yes, it could be added to "hosted" as we previously discussed, however, after more thought I find this approach contrary to a fundamental benefit of BMP - no "man in the middle" required.

My application for this change to BMP will not use GDB, the client app directly connects to BMP.

I still have more to do on the tool to use this so no hurry to reach a decision, further discussion/comment most welcome.

sidprice:
@esden Thanks, I use this kind of monitoring with JLink and the Segger debugger all the time. For me it is a "light-weight", easy to use way of watching the target without burden. 

I have an early UI, written in Python, that uses it. It uses the ELF file to allow the user to select variables/addresses to monitor. Still needs lots of work but could be a useful, complementary tool.

Like many I am looking forward to OrbTrace :)

esden:
That sounds good. Do you have a writeup on how to use your current implementation so I can try it out and get a feel for the current state of it? This way I can give you some feedback on the way forward.

sidprice:
Here is my UI repo, probably more input than you need. It has been a long time since I worked on it and it for sure needs some love and attention.
https://github.com/sidprice/ctxMonitor

sidprice:
@esden To test, I start GDB with the ELF file on the command line then just do:
tar ext //./comx
mon s
mon no_halt en
att 1 <- Doesn't halt the target
x 0x20000000 <- Reads the location while the target is running.

As I said the monitor app I have started communicates directly with the BMP.

stoyan-shopov:
I did some quick hacking today, and this is what I did:
https://github.com/stoyan-shopov/blackmagic/tree/add-helper-thread
(there are 6 commits in this branch)
Please, note that I only tested this on a Windows 10 machine

This is what I do:
- in `src/platforms/pc/gdb_if.c`, in function **gdb_if_init()** I am deploying a thread (let us call it the **sidekick** thread), listening on port 4000; the **pthread_create()** function is used to deploy the thread
- I export the **cur_target** variable out of `gdb_main.c`, so that the **sidekick** thread can also access the target
- I add routines **target_lock()** and **target_unlock()**, for providing exclusive access to the target (i.e., **cur_target**); the lock used is a **pthread_mutex_t**
- the **sidekick** thread waits for connections (on port 4000); when a connection is established - it prints the value of the **cur_target** pointer, and in case it is not 0, the value at address 0 of the target is being read, and printed over the socket; the connection is then shut down
- I add exclusive access to the target, by employing the **target_lock()**/**target_unlock()** routines in the threads racing for access the target; the threads racing for access to the target are the main thread in **gdb_main_loop()**, and the **sidekick** thread
- A very important detail is that I release the lock, and reacquire it, when polling for the target state in the **gdb_main_loop()** routine

With these changes, it is possible to read target memory, while it is running. However, I only checked this on a Windows 10 machine

What do you think about this?

To be honest, I fear that such a step (introducing multiple threads) could turn the hosted bmp into a, so to say, Chimera (citing wikipedia: a monster from Greek mythology that breathes fire and has a lion's head, a goat's body, and a snake's tail). If done responsibly, though, I think multiple threads in the bmp can be a good thing

DX-MON:
We think that rather than judging the merits of this based on a pthreads approach, we'll consider it once you've rewritten it using C11 threading as that is standardized and so cross-platform. https://github.com/DX-MON/crunch/tree/main/crunch/threading contains a usable cross-platform compat shim for platforms that haven't yet implemented C11 threading.

sidprice:
The purpose of my original PR was to build this functionality into BMP itself and to be able to attach a client application directly to the BMP to monitor variables without a "man-in-the-middle" application like hosted needing to be run. I feel the original PR is minimally invasive for BMP.

DX-MON:
Hi Sid, sorry - should have made it a little clearer, we weren't suggesting this to you, we were suggesting it to @stoyan-shopov re their contribution - we'll give yours a look over tomorrow and see where it's at.

sidprice:
No worries, it looked like the PR was being high-jacked :)

dragonmux:
Finally managed to come back round to this. Sorry it's taken so long! @sidprice, could we bother you to rebase the change set against main? We'll then give it a review with a view to trying to merge it in.

sidprice:
@dragonmux This rebase is a little complex because of the changes around the reset commands. Will take me a little time to figure it out and do the rebase.

dragonmux:
No worries, we understand these things can take time

sidprice:
@dragonmux Doing some testing and I found a problem, may be due to the changes made to the code base since my original PR. Will hopefully chase it down later and update things here :)

sidprice:
@dragonmux I have been looking closely at the issue I have with this, and it appears when GDB attaches and BMP does not halt the target, GDB gets confused because it doesn't know the target is running. I have been using BMDA and Wireshark to capture interaction and so far, I have not found what GDB expects to see. If you have any ideas of where things may be going "pear-shaped" I would love to hear them.

sidprice:
I think I see the problem, according to the RSP the response to "$vAttach" should be sent when the target halts! I need to read a little more and maybe do a test or two.

dragonmux:
Sounds good, if it means we get to fix two bugs in one go, we support that.

If you're right we should probably do a review of the issue tracker for related issues and make sure they get linked over

sidprice:
I think you misunderstood what I wrote. The issue is with this PR.

I am beginning to think that my approach to adding this feature may be flawed.

Maybe we can take the discussion to Discord rather than muddy this PR.

sidprice:
After the discussions on Discord with @dragonmux and @esden it was decided to abandon this approach to connecting to a running target without halting it.

A new approach investigating GDB non-stop mode will be done.

PR #773, opened 2020-11-14
Make possible building 'hosted' without libusb and libftdi

stoyan-shopov:
To build the hosted variant of the BMP, without the need to
link against libusb and libftdi, run make with options:
"make PROBE_HOST=hosted HOSTED_BMP_ONLY=yes"

This minimizes external dependencies, and can be useful in
scenarious where access to the libusb and libftdi libraries
is not readily available.

The patch builds successfully on msys2, and on debian, tested with both "HOSTED_BMP_ONLY=yes", and without setting "HOSTED_BMP_ONLY".

### DIFF
--- a/src/include/general.h
+++ b/src/include/general.h
@@ -74,6 +74,7 @@ static inline void DEBUG_WARN(const char *format, ...)
 	va_list ap;
     va_start(ap, format);
     vfprintf(stderr, format, ap);
+    fflush(stderr);
     va_end(ap);
 	return;
 }
@@ -85,9 +86,15 @@ static inline void DEBUG_INFO(const char *format, ...)
 	va_list ap;
     va_start(ap, format);
 	if (cl_debuglevel & BMP_DEBUG_STDOUT)
+	{
 		vfprintf(stdout, format, ap);
+		fflush(stdout);
+	}
 	else
+	{
 		vfprintf(stderr, format, ap);
+		fflush(stderr);
+	}
     va_end(ap);
 	return;
 }
@@ -99,6 +106,7 @@ static inline void DEBUG_GDB(const char *format, ...)
 	va_list ap;
     va_start(ap, format);
     vfprintf(stderr, format, ap);
+    fflush(stderr);
     va_end(ap);
 	return;
 }
@@ -111,6 +119,7 @@ static inline void DEBUG_GDB_WIRE(const char *format, ...)
 	va_list ap;
     va_start(ap, format);
     vfprintf(stderr, format, ap);
+    fflush(stderr);
     va_end(ap);
 	return;
 }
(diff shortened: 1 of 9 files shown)


UweBonnes:
Other compile time options use '=1', e.g. "ST_BOOTLOADER=1'. Perhaps that was not the best choice but for consistency consider to also use '=1'. Thanks for the effort. It compiles here on Linux with mingw64-make too.

Playing with that I see:
no device connected:

UweBonnes:
I rebased the patch and had a deeper look, It misses the auto-detect and select by position function. I am sure, bmp autodetection can be done with the registry only, w/o libusb.

stoyan-shopov:
Thanks for looking into this.

I will explain the autodetection situation as I understand it. By only consulting the windows registry, it is possible to see which BMP probes have at some time been connected to the computer, and which serial port numbers (e.g., 'COM7', 'COM3', etc.) have been assigned to the particular probe. But I did not find some kind of key in the registry which reports if the particular probe is currently plugged in or not.

As a simple workaround, I am thinking of simply trying to open the particular serial port number, and if that succeeds - then the probe is reported as connected. That should be simple to do, I will try to test later this week.

UweBonnes:
It seem that also 
Computer\HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\COM Name Arbiter\Devices 
has a list of previous and active connected devices.
It seem that 
Computer\HKEY_LOCAL_MACHINE\HARDWARE\DEVICEMAP\SERIALCOMM
has only an entry \Device\USBSERXXX with value COMY when the device is avtive connected. But probably trying to open it is also valid.
Probably defining a lot of empty libusd/libftdi functions in the BMP_ONLY case  could keep the code changes much smaller. beside the changed detection. Should I try to do so or will you?

stoyan-shopov:
What is your idea, maybe to define some dummy headers for libusb/libftdi for the BMP_ONLY case, and some dummy functions, which would only return errors, to make the linker happy, so that the necessary changes to the rest of the code are minimized? That makes sense, having code with `#ifdefs` is hard to maintain. I can try this approach when I have the time, will try to do it this week.

I just checked, you are correct that `Computer\HKEY_LOCAL_MACHINE\HARDWARE\DEVICEMAP\SERIALCOMM` only lists the active COM ports. I also found some thread that may be useful in listing serial devices here https://stackoverflow.com/questions/304986/how-do-i-get-the-friendly-name-of-a-com-port-in-windows . But maybe the simplest approach is to somehow match the list of known BMP probes with the currently active COM ports, and make a report based upon this matching. I am inclined to use this approach. The automatic probe detection is very convenient and should be available in the BMP_ONLY case

UweBonnes:
Superceded by #805

PR #788, opened 2020-12-03
gd32f1/f3 detection and ram/flash autoset

mean00:
Hopefully this one is correct
it contains a small cosmetic fix : reorder the designer id so they are in increasing order (2 lines swapped)


UweBonnes:
Argh, travis is slow today...

PR #799, opened 2020-12-10
Adjust USB strings for BMP native.

stoyan-shopov:
This change adjusts some of the usb strings for BMP native
to include the firmware version. Also, some unused usb strings
are removed.


stoyan-shopov:
Never mind, I was not at the latest master.

PR #800, opened 2020-12-10
Add hard reset option to PC hosted platform

noahp:
Add an option `-x` to issue a reset pulse on RST before running other
commands. This can help recover a chip that's stuck in Lockup.

### DIFF
--- a/src/platforms/hosted/platform.c
+++ b/src/platforms/hosted/platform.c
@@ -333,6 +333,13 @@ void platform_init(int argc, char **argv)
 	default:
 		exit(-1);
 	}
+
+	if (cl_opts.opt_hard_srst) {
+		DEBUG_INFO("hard reset\n");
+		platform_srst_set_val(true);
+		platform_srst_set_val(false);
+	}
+
 	int ret = -1;
 	if (cl_opts.opt_mode != BMP_MODE_DEBUG) {
 		ret = cl_execute(&cl_opts);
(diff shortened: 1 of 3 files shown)


UweBonnes:
Doesn't the -R option (connect under Reset) serve this purpose even better?

noahp:
> Doesn't the -R option (connect under Reset) serve this purpose even better?

This patch is somewhat half-baked which I regret :grimacing: + apologize for. The reason I added it was using connect-under-reset from gdb extended remote fails on the LPC54628:

```
>>> mon connect_srst enable
Assert SRST during connect: enabled
>>> mon hard_srst
>>> mon swdp_scan
Target voltage: 3.3V
SW-DP scan failed!
>>> mon connect_srst disable
Assert SRST during connect: disabled
>>> mon hard_srst
>>> mon swdp_scan
Target voltage: 3.3V
Available Targets:
No. Att Driver
 1      LPC54628J512 M4
```

I believe due to the part using the ~RST line to select JTAG/SWD:
>49.6.3 Boundary scan
The RESET pin selects between the test TAP controller for JTAG boundary scan
(RESET = LOW) and the ARM SWD debug port TAP controller (RESET = HIGH). The
ARM SWD debug port is disabled while the part is in reset. A LOW on the TRST pin resets
the test TAP controller.

What's surprising is supplying the `-C` option to the PC hosted version the blackmagic seems to work OK... not sure how it's different from what's happening in gdb above :frowning_face: .

UweBonnes:
please check in the datasheet/reference manual if LPC has extended reset like LPC17.

noahp:
Will do, thanks!

PR #801, opened 2020-12-11
target/lpc546xx: fix lpc546xx flash support (#801)

### LINKED ISSUE #786: Probing LPC546xx fails
Due to attempted read of LPC43xx failing and causing SWD fault.
See #785 for a workaround.

The issue is fundamentally due to the same JEDEC ID being used on both part families by NXP. It's likely this problem could be dealt with by somehow recovering from the failed read during the LPC43xx probe attempt.

noahp:
**Summary**
Modifications to fix flash support on the lpc546xx:

- fix IAP entrypoint to be `0x03000204`, not the value at that address
- add a reset and attach sequence before erasing flash sectors. there's
little documentation around this, but experimentally, erasing sector 0
fails when the ROM bootloader is mapped to it (on reset). stepping the
chip once and attaching is enough to snap the chip out of it, permitting
flash erase on sector 0.
- add a few test commands to the lpc546xx table (read_uid, erase/write
sector, etc).
- write the magic CRC value when writing to sector 0
(`lpc_flash_write_magic_vect`).
- move the lpc546xx probe to before the lpc43xx probe, to prevent
getting the lpc546xx into Lockup when reading an illegal memory location
during lpc43xx probing

Fixes #786.

I don't 100% understand the reset/load sequence of this part, but these
changes are sufficient to program and debug the part now.

I didn't do a detailed analysis of what pyocd (via st-link hardware
adapter) and segger jlink do to handle the same, but both of those
worked without modification, so there's some difference in the
sequence they're using.

**Testing**
Verified I can now successfully erase and write an executable in sector
0 (and other sectors).
Verified the new commands work correctly.

### DIFF
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -445,8 +445,16 @@ bool cortexm_probe(ADIv5_AP_t *ap)
 			PROBE(lpc11xx_probe); /* LPC24C11 */
 			PROBE(lpc43xx_probe);
 		} else if (ap->ap_partno == 0x4c4) { /* Cortex-M4 ROM */
-			PROBE(lpc43xx_probe);
+			/* The LPC546xx and LPC43xx parts present with the same AP ROM Part
+			Number, so we need to probe both. Unfortunately, when probing for
+			the LPC43xx when the target is actually an LPC546xx, the memory
+			location checked is illegal for the LPC546xx and puts the chip into
+			Lockup, requiring a RST pulse to recover. Instead, make sure to
+			probe for the LPC546xx first, which experimentally doesn't harm
+			LPC43xx detection. */
 			PROBE(lpc546xx_probe);
+
+			PROBE(lpc43xx_probe);
 			PROBE(kinetis_probe); /* Older K-series */
 		}
 		/* Info on PIDR of these parts wanted! */
--- a/src/target/lpc546xx.c
+++ b/src/target/lpc546xx.c
@@ -27,36 +27,53 @@
 
 #define LPC546XX_CHIPID 0x40000FF8
 
-#define IAP_ENTRYPOINT_LOCATION	0x03000204
+#define IAP_ENTRYPOINT_LOCATION 0x03000204
 
 #define LPC546XX_ETBAHB_SRAM_BASE 0x20000000
-#define LPC546XX_ETBAHB_SRAM_SIZE (160*1024)
+
+/* only SRAM0 bank is enabled after reset */
+#define LPC546XX_ETBAHB_SRAM_SIZE (64 * 1024)
 
 #define LPC546XX_WDT_MODE 0x4000C000
-#define LPC546XX_WDT_CNT  0x4000C004
+#define LPC546XX_WDT_CNT 0x4000C004
 #define LPC546XX_WDT_FEED 0x4000C008
 #define LPC546XX_WDT_PERIOD_MAX 0xFFFFFF
 #define LPC546XX_WDT_PROTECT (1 << 4)
 
-#define IAP_RAM_SIZE	LPC546XX_ETBAHB_SRAM_SIZE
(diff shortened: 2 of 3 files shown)


UweBonnes:
Please provide the output of blackmagic -t even if I have asked for and received the output  before. LPC43xx part ID register is at 0x4004 3200 where LPC54 has Asynchronous system configuration (base address 0x4004 0000) but no specified register at 0x3200.

Shuffling this damned LPC probing up and down is no real solution.  Wid memory probing will always step on some parts toe. We need to find out earlier if and what LPCXX we have. If nothing else helps, user input is needed.

noahp:
No problem, here it is (with the shuffle, probing lpc546xx first):

```
❯ ./src/blackmagic -d /dev/serial/by-id/usb-Black_Sphere_Technologies_Black_Magic_Probe_*if00 -t -R
BMP hosted v1.7.1-101-g3cec441-dirty
Target voltage: 3.3V Volt
DPIDR 0x2ba01477 (v1 rev2)
RESET_SEQ failed
AP   0: IDR=24770011 CFG=00000000 BASE=e00ff003 CSW=23000040
AP#0 IDR = 0x24770011 (AHB-AP var1 rev2)
ROM: Table BASE=0xe00ff000 SYSMEM=0x00000001, designer 43b Partno 4c4
0 0xe000e000: Generic IP component - Cortex-M4 SCS (System Control Space) (PIDR = 0x04000bb00c  DEVTYPE = 0x00 ARCHID = 0x0000)-> cortexm_probe
CPUID 0x410fc241 (M4 var 0 rev 1)
Halt via DHCSR: success 00030003 after 1ms
1 0xe0001000: Generic IP component - Cortex-M3 DWT (Data Watchpoint and Trace) (PIDR = 0x04003bb002  DEVTYPE = 0x00 ARCHID = 0x0000)
2 0xe0002000: Generic IP component - Cortex-M3 FBP (Flash Patch and Breakpoint) (PIDR = 0x04002bb003  DEVTYPE = 0x00 ARCHID = 0x0000)
3 0xe0000000: Generic IP component - Cortex-M3 ITM (Instrumentation Trace Module) (PIDR = 0x04003bb001  DEVTYPE = 0x00 ARCHID = 0x0000)
4 0xe0040000: Debug component - Cortex-M4 TPIU (Trace Port Interface Unit) (PIDR = 0x04000bb9a1  DEVTYPE = 0x11 ARCHID = 0x0000)
5 0xe0041000: Debug component - Cortex-M4 ETM (Embedded Trace) (PIDR = 0x04000bb925  DEVTYPE = 0x13 ARCHID = 0x0000)
ROM: Table END
AP   2: IDR=002a0000 CFG=00000000 BASE=00000000 CSW=00000000
AP#0 IDR = 0x002a0000 (AHB-AP var0 rev0)
***  1      LPC54628J512 M4
```

I'll take another look through the LPC43xx and LPC546xx datasheets for some common memory address that we might be able to use to distinguish them. I'm not super familiar with AP ROM table contents so there may be something I'm missing there that would help tell the chips apart... unfortunately I only have an LPC54628 dev board so I can't test the LPC43xx :(.

UweBonnes:
Thanks. I stored for later reference. But the LPC43 detection is identical, beside the AP2 on the LPC54. 

I tested your #801 (lpc546xx-flash) and put the lpc43 probe after the lpc54 probe, and the lpc43 still gets detected. Do so too in your PR, perhaps describe the problem with wild probing, and put the reference to #801 in the commit and I will apply.

UweBonnes:
At some point AP2 perhaps should get exposed like the MDM APs of  kinetis, nrf5 and efm32.

noahp:
> At some point AP2 perhaps should get exposed like the MDM APs of kinetis, nrf5 and efm32.

Ah, that makes sense, thanks for the reference!

noahp:
Ok, updated, should be ready for re-review.

PR #802, opened 2020-12-11
Add .clang-format

noahp:
Add a .clang-format specification file. I tried to match as closely as
possible to the existing style, but there are a few exceptions that the
tool doesn't support:

- macro invocation puts brace on the following line (eg `TRY_CATCH`
blocks)
- inline comment wrapping in control blocks without braces:
   ```c
   	// clang-format outputs this:
   	if (connect_assert_srst)
		platform_srst_set_val(
			true); /* will be deasserted after attach */
	// instead of:
	if (connect_assert_srst)
	    platform_srst_set_val(true); /* will be deasserted after attach */
   ```
   this is due to clang-format attempting to enforce 80-character line
   limit

Otherwise the format seems to match pretty closely with the existing
source.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,10 @@
+﻿---
+Language:        Cpp
+
+BreakBeforeBraces: Linux
+ConstructorInitializerIndentWidth: '8'
+ContinuationIndentWidth: '8'
+IndentWidth: '8'
+TabWidth: '8'
+UseTab: Always
+...


UweBonnes:
I checked out your branch, renamed .dir_local and opened src/command.c in emacs. Emacs tell in the status line clang-format somewhere.

Now I go to line 41 and type "TAB" . This results in an ident change. If I keep .dir_local , no ident change happend.
Is .clang-format supposed to work together with .dir-local?

noahp:
I'm not very experienced in emacs :grimacing: but for example vim has plugins like https://github.com/rhysd/vim-clang-format that will detect the presence of the `.clang-format` file in a parent directory to the current buffer and provide editor commands for formatting ranges or entire files. Similar for VS Code. Seems like this might be the way to enable it in emacs? https://clang.llvm.org/docs/ClangFormat.html#emacs-integration

The idea behind this patch is providing a format specification local to this repo (similar to how the linux tree is set up).

UweBonnes:
But I need to be sure that your proposed file matches the existing style. Most important is I think what .dir-local tells:
((c-mode . ((c-file-style . "bmp")
            (c-basic-offset . 4)
            (tab-width . 4)
            (indent-tabs-mode t))))
Brace after condition e.g. if should be on same line
 if ( bla) {

How can I check?

And the patch contains a lot of cruft possible under ForEachMacros

noahp:
Ahh gotcha, yeah I assumed based on the mixed tabs-and-spaces that this was linux style, let me condense the .clang-format to better fit the existing style, sorry about that!

UweBonnes:
A good test would be some setup that would apply the choosen style to all files. I could then diff and see if the changes are acceptable, e.g. Stoyan opening the brace after if() on the next line, and Mubes changes having his own mind.

noahp:
It's possible to apply the format to all files with eg `clang-format -style=file -i $(find . -name '*.[ch]')` (or pass a single file to reformat it). The updated config I pushed looks a lot better, but still a few inconsistencies with the existing style; it's as close as I could get it with clang-format, which I think is the best supported c styler right now.

noahp:
Going to close this for now, I don't think it's super useful in its current state.

PR #804, opened 2020-12-11
Fix building with newer GNU make

noahp:
More recent versions of GNU make remove the `-` prefix from
`$(MAKEFLAGS)`, which breaks the build:

```bash
 # output truncated to interesting parts
> make -C src --trace all_platforms
set -e ;\
mkdir -p artifacts/v1.7.1-91-g98b4ec5 ;\
echo "<html><body><ul>" > artifacts/index.html ;\
for i in platforms/*/Makefile.inc ; do \
        ...\
        make w --trace --no-print-directory clean ;\
        make w --trace --no-print-directory;\
        ...\
done ;\
echo "</ul></body></html>" >> artifacts/index.html ;\
cp artifacts/*.bin artifacts/v1.7.1-91-g98b4ec5
Building for hardware platform: f4discovery
make[1]: *** No rule to make target 'w'.  Stop.
```

Per the
[documentation](https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html),
there's no need to manually apply insert `$(MAKEFLAGS)`, it should be
automatically included when using `$(MAKE)` when recursing.

I tested on my machine (make 4.2.1) and a docker image based on
ubuntu:trusty (make 3.81), both behaved the same (directories were not
printed when recursing by default);

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -126,8 +126,8 @@ all_platforms:
 		export PROBE_HOST=`basename $$DIRNAME` ;\
 		export CFLAGS=-Werror ;\
 		echo "Building for hardware platform: $$PROBE_HOST" ;\
-		$(MAKE) $(MAKEFLAGS) clean ;\
-		$(MAKE) $(MAKEFLAGS);\
+		$(MAKE) clean ;\
+		$(MAKE);\
 		if [ -f blackmagic.bin ]; then \
 			mv blackmagic.bin artifacts/blackmagic-$$PROBE_HOST.bin ;\
 			echo "<li><a href='blackmagic-$$PROBE_HOST.bin'>$$PROBE_HOST</a></li>"\


UweBonnes:
What is a make version that stumbles with the present makefile? 4.2.1 does work at least for me with the present makefile.

noahp:
The version of make installed from Ubuntu 20.04 ppa's is 4.2.1 and exhibits the problem for me. This reduced makefile repros the problem:
```makefile
.PHONY: yolo

MAKEFLAGS += --no-print-dir

yolo:
	$(MAKE) $(MAKEFLAGS)
```
If you call it with `make -C .` (assuming the Makefile above is in the cwd).

Here's a Docker file you can use to reproduce the problem on blackmagic:
https://github.com/noahp/blackmagic/blob/da15211215f25beb4ceee6c8d0affdd9876e776c/Dockerfile#L42

Which you can run like
```bash
# build the image
docker build -t blackmagic --build-arg "UID=$(id -u)" .
# really clean the repo
git clean -dxff && git submodule foreach --recursive "git clean -dxff"
# run make on the cwd inside the docker container
docker run -v "$(pwd):/mnt/workspace" -i -t blackmagic bash -c "make -C libopencm3 -j6 lib && make -C src -j6 all_platforms"
```

I'll check if this problem occurs on GNU make upstream source (it may be a weirdness in the Ubuntu ppa version of make).

UweBonnes:
Both 
 devel/blackmagic> make -C src
and
 devel/blackmagic/src> make -C .
succeed here. 
devel/blackmagic/src> make --version
GNU Make 4.2.1
Gebaut für x86_64-suse-linux-gnu

noahp:
I built GNU make from upstream (git://git.savannah.gnu.org/make.git @ tag 4.2.1) and it has the same problem as GNU make installed from the Ubuntu PPA. Let me take a look at the SUSE version.

noahp:
Same problem on a clean opensuse:tumbleweed Docker image (`zypper install -y make` only), which is make 4.3.
Same problem on same for opensuse:leap:

```bash
2e575f96e494:~ # make -C .
make: Entering directory '/root'
make w --no-print-directory
make[1]: *** No rule to make target 'w'.  Stop.
make: *** [Makefile:6: yolo] Error 2
make: Leaving directory '/root'
2e575f96e494:~ # make --version
GNU Make 4.2.1
Built for x86_64-suse-linux-gnu
```

noahp:
Ah, on blackmagic you have to specify `all_platforms` target to run the rule with the problem:

```bash
make -C all_platforms
```

noahp:
I see that `MFLAGS` is being used in the top-level makefile, which isn't needed but GNU make supports using it for historical compatibility (see the link https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html). I'd say we drop both in favor of just appending to MAKEFLAGS and relying on `$(MAKE)` itself.

stoyan-shopov:
Just some notes. I really don't know `make` well.
I also did some testing. I have never used the `all_platforms` target, that is good to know.

I am using make 4.3 on msys2. Running, from within `blackmagic/src/`:
```
make -C . all_platforms
```
I also see these errors:
```
make: Entering directory '/c/tmp/bmp/merge/blackmagic/src'
Building for hardware platform: f4discovery
make[1]: *** No rule to make target 'w'.  Stop.
make: *** [Makefile:121: all_platforms] Error 2
make: Leaving directory '/c/tmp/bmp/merge/blackmagic/src'
```
Just running:
```
make all_platforms
```
works properly.

With the patch, proposed - building works in both cases.

Uwe, does `make all_platforms` work for you without error? That would be strange...

UweBonnes:
make -C . all_platforms did not work. With #804 apllied, it does work. I added the reference tp #804 to the patch and will apply.

PR #805, opened 2020-12-12
Provide a compile time option (HOSTED_BMP_ONLY=1) for easier compilation of hosted

UweBonnes:
Provide a compile time option (HOSTED_BMP_ONLY=1) to remove the need for libusb, libftdi and hidapi. The resulting executable will only support BMP probes. On windows, the need to announce the probe via zadig to find it without giving the device on the command line is removed. All this should lower the entry barrier for hosted considerable,

Hopefully issues and problems reported will now always include relevant console output of blackmagic hosted!

Many thanks to  Stoyan Shopov <stoyan.shopov@gmail.com> for ideas, contributions and testing!


stoyan-shopov:
Please, note - `zadig` is still needed on windows in order to use the DFU functionality, if I am not mistaken. `zadig` is also needed in order to access the trace capture interface on windows. For windows versions 7 and above, the serial interfaces are supported natively (no driver installation needed, no `zadig` needed).

UweBonnes:
Thanks for the clarification. Is there a need to point that out more prominent, e.g. in the wiki?

stoyan-shopov:
I am not sure. To be honest, I have never read the wiki. For me, the BMP has always just 'worked out of the box', which is one of the greatest things about this project, so I have had no much need to read the documentation. Nothing like the `openocd` nightmare...

I will read the wiki, I will let you know if I feel something needs adjusting

PR #808, opened 2020-12-12
add src/artifacts to .gitignore

noahp:
This folder is generated when running `make -C src all_platforms`. Add
it to ignored files.


UweBonnes:
I additional added .gdb_history

noahp:
Nice, good call.

PR #811, opened 2020-12-14
hosted: remove ftdi when built with BMP_ONLY

xobs:
Hosted mode has the option to build with `HOSTED_BMP_ONLY=1`. This
removes dependencies on libusb, as well as a dependency on libftdi.

This patch removes `#include <ftdi.h>` from the `platform.c` file under
Hosted mode, and removes all references to it when built with
`-DHOSTED_BMP_ONLY=1`.

### DIFF
--- a/src/platforms/hosted/ftdi_bmp.h
+++ b/src/platforms/hosted/ftdi_bmp.h
@@ -27,7 +27,6 @@
 #include "jtagtap.h"
 
 #include "bmp_hosted.h"
-#include <libftdi1/ftdi.h>
 
 typedef struct data_desc_s {
 	int16_t data_low;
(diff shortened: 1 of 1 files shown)


UweBonnes:
Thanks for pointing out. But instead of sprinkling the code with #idefs, I prefer moving
+#include <libftdi1/ftdi.h>
+extern cable_desc_t cable_desc[];
+extern cable_desc_t *active_cable;
+extern struct ftdi_context *ftdic;
+extern data_desc_t active_state;

to the #else ( HOSTED_BMP_ONLY == 1) branch.Will you do or should I edit your PR?

xobs:
I'm not sure where you're saying that could go. Would you mind editing the PR?

xobs:
That looks good to me! Is there anything else you would like me to add to this patch?

xobs:
I marked the patch as "Draft" because I realized it wasn't allowing me to use my device. However, after your change to the patch, I can verify that it does now work.

PR #813, opened 2020-12-16
gitattributes: add attributes file

xobs:
On different platforms, line endings can differ. Git tries to
intelligently convert source code to match native line endings when
doing a checkout, unless a particular configuration flag is set.

An alternate approach is to create a `.gitattributes` file and manually
specify various types of files. This ensures files are properly handled.
It also prevents git from creating patches that consist of nothing more
than replacing line endings.


UweBonnes:
Let's hope nothing bad happens...

PR #818, opened 2020-12-21
Initial support for the stlinkv3 target

stoyan-shopov:
This commit adds support for the stlinkv3 target as a host for
running the blackmagic firmware.

This commit still lacks proper serial number handling for the
STM32F723 device.

In short, here is what has been tested:

- DFU firmware upgrade
- SWD bus driving
- JTAG bus driving
- BMP data usb cdcacm interface
- trace capture, asynchronous mode
- target voltage measurement

Not tested:

- srst driving is available, and should work, but has not been tested
- LEDs are not driven properly

### DIFF
--- a/scripts/stm32_mem.py
+++ b/scripts/stm32_mem.py
@@ -207,8 +207,10 @@ if __name__ == "__main__":
 	if args.address :
 		start = int(args.address, 0)
 	else :
-		if "F4" in product or "STLINK-V3" in product:
+		if "F4" in product:
 			start = 0x8004000
+		elif "STLINK-V3" in product:
+			start = 0x8020000
 		else:
 			start = 0x8002000
 	addr = start
--- a/src/crc32.c
+++ b/src/crc32.c
@@ -98,6 +98,8 @@ static uint32_t crc32_calc(uint32_t crc, uint8_t data)
 	return (crc << 8) ^ crc32_table[((crc >> 24) ^ data) & 255];
 }
 
+void crc32_init(void) {}
+
 int generic_crc32(target *t, uint32_t *crc_res, uint32_t base, size_t len)
 {
 	uint32_t crc = -1;
@@ -135,7 +137,13 @@ int generic_crc32(target *t, uint32_t *crc_res, uint32_t base, size_t len)
 	return 0;
 }
 #else
+#include <libopencm3/stm32/rcc.h>
 #include <libopencm3/stm32/crc.h>
+void crc32_init(void)
+{
+	rcc_periph_clock_enable(RCC_CRC);
+}
+
 int generic_crc32(target *t, uint32_t *crc_res, uint32_t base, size_t len)
 {
 	uint8_t bytes[128];
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -560,5 +560,6 @@ handle_z_packet(char *packet, int plen)
 
 void gdb_main(void)
 {
+	crc32_init();
 	gdb_main_loop(&gdb_controller, false);
 }
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -101,7 +101,10 @@ int gdb_main_loop(struct target_controller *tc, bool in_syscall)
 	while(1) {
 		SET_IDLE_STATE(1);
 		size = gdb_getpacket(pbuf, BUF_SIZE);
-		SET_IDLE_STATE(0);
+		// If port closed and target detached, stay idle
+		if (!(pbuf[0] == 0x04) || cur_target) {
+			SET_IDLE_STATE(0);
+		}
 		switch(pbuf[0]) {
 		/* Implementation of these is mandatory! */
 		case 'g': { /* 'g': Read general registers */
--- a/src/include/crc32.h
+++ b/src/include/crc32.h
@@ -22,5 +22,5 @@
 #define __CRC32_H
 
 int generic_crc32(target *t, uint32_t *crc, uint32_t base, int len);
-
+void crc32_init(void);
 #endif
--- a/src/platforms/common/cdcacm.c
+++ b/src/platforms/common/cdcacm.c
@@ -220,7 +220,11 @@ static const struct usb_endpoint_descriptor uart_data_endp[] = {{
 	.bDescriptorType = USB_DT_ENDPOINT,
 	.bEndpointAddress = CDCACM_UART_ENDPOINT,
 	.bmAttributes = USB_ENDPOINT_ATTR_BULK,
+#if defined(USB_HS)
+	.wMaxPacketSize = CDCACM_PACKET_SIZE,
+#else
 	.wMaxPacketSize = CDCACM_PACKET_SIZE / 2,
+#endif
 	.bInterval = 1,
 }, {
 	.bLength = USB_DT_ENDPOINT_SIZE,
--- a/src/platforms/common/cdcacm.c
+++ b/src/platforms/common/cdcacm.c
@@ -526,7 +526,7 @@ static const struct usb_config_descriptor config = {
 	.bConfigurationValue = 1,
 	.iConfiguration = 0,
 	.bmAttributes = 0x80,
-	.bMaxPower = 0x32,
+	.bMaxPower = 250,
 
 	.interface = ifaces,
 };
--- a/src/platforms/common/cdcacm.c
+++ b/src/platforms/common/cdcacm.c
@@ -651,25 +651,9 @@ static void cdcacm_set_modem_state(usbd_device *dev, int iface, bool dsr, bool d
 }
 
 #if defined(PLATFORM_HAS_SLCAN)
-static volatile uint32_t count_new;
-static uint8_t double_buffer_out[CDCACM_PACKET_SIZE];
-static void slcan_usb_out_cb(usbd_device *dev, uint8_t ep)
-{
-	(void)ep;
-	usbd_ep_nak_set(dev, CDCACM_SLCAN_ENDPOINT, 1);
-	count_new = usbd_ep_read_packet(dev, CDCACM_SLCAN_ENDPOINT,
-									double_buffer_out, CDCACM_PACKET_SIZE);
-	usbd_ep_nak_set(dev, CDCACM_SLCAN_ENDPOINT, 0);
-	usbd_ep_write_packet(dev, CDCACM_SLCAN_ENDPOINT,
-						 double_buffer_out, count_new);
-
-}
-static void slcan_usb_in_cb(usbd_device *dev, uint8_t ep)
-{
-	(void) ep;
-	(void) dev;
-}
+extern void slcan_usb_out_cb(usbd_device *dev, uint8_t ep);
 #endif
+
 static void cdcacm_set_config(usbd_device *dev, uint16_t wValue)
 {
 	configured = wValue;
@@ -706,7 +690,7 @@ static void cdcacm_set_config(usbd_device *dev, uint16_t wValue)
 	usbd_ep_setup(dev, CDCACM_SLCAN_ENDPOINT, USB_ENDPOINT_ATTR_BULK,
 				  CDCACM_PACKET_SIZE, slcan_usb_out_cb);
 	usbd_ep_setup(dev, CDCACM_SLCAN_ENDPOINT | USB_REQ_TYPE_IN, USB_ENDPOINT_ATTR_BULK,
-				  CDCACM_PACKET_SIZE, slcan_usb_in_cb);
+				  CDCACM_PACKET_SIZE, NULL);
 	usbd_ep_setup(dev, (CDCACM_SLCAN_ENDPOINT + 1) | USB_REQ_TYPE_IN, USB_ENDPOINT_ATTR_INTERRUPT, 16, NULL);
 #endif
 
--- a/src/platforms/common/cdcacm.c
+++ b/src/platforms/common/cdcacm.c
@@ -24,6 +24,18 @@
  * field firmware upgrade.
  *
  * The device's unique id is used as the USB serial number string.
+ *
+ * Endpoint Usage
+ *
+ *     0 Control Endpoint
+ * IN  1 GDB CDC DATA
+ * OUT 1 GDB CDC DATA
+ * IN  2 GDB CDC CTR
+ * IN  3 UART CDC DATA
+ * OUT 3 UART CDC DATA
+ * OUT 4 UART CDC CTRL
+ * In  5 Trace Capture
+ *
  */
 
 #include "general.h"
@@ -42,7 +54,15 @@
 #include <libopencm3/cm3/scb.h>
 #include <libopencm3/usb/dfu.h>
 
-#define DFU_IF_NO 4
+#define GDB_IF_NO   0
+#define UART_IF_NO  2
+#define DFU_IF_NO   4
+#if defined(PLATFORM_HAS_TRACESWO)
+# define TRACE_IF_NO 5
+# define TOTAL_INTERFACES  6
+#else
+# define TOTAL_INTERFACES  5
+#endif
 
 usbd_device * usbdev;
 
@@ -78,7 +98,7 @@ static const struct usb_device_descriptor dev_desc = {
 static const struct usb_endpoint_descriptor gdb_comm_endp[] = {{
 	.bLength = USB_DT_ENDPOINT_SIZE,
 	.bDescriptorType = USB_DT_ENDPOINT,
-	.bEndpointAddress = 0x82,
+	.bEndpointAddress =  (CDCACM_GDB_ENDPOINT + 1) | USB_REQ_TYPE_IN,
 	.bmAttributes = USB_ENDPOINT_ATTR_INTERRUPT,
 	.wMaxPacketSize = 16,
 	.bInterval = 255,
@@ -87,14 +107,14 @@ static const struct usb_endpoint_descriptor gdb_comm_endp[] = {{
 static const struct usb_endpoint_descriptor gdb_data_endp[] = {{
 	.bLength = USB_DT_ENDPOINT_SIZE,
 	.bDescriptorType = USB_DT_ENDPOINT,
-	.bEndpointAddress = 0x01,
+	.bEndpointAddress = CDCACM_GDB_ENDPOINT,
 	.bmAttributes = USB_ENDPOINT_ATTR_BULK,
 	.wMaxPacketSize = CDCACM_PACKET_SIZE,
 	.bInterval = 1,
 }, {
 	.bLength = USB_DT_ENDPOINT_SIZE,
 	.bDescriptorType = USB_DT_ENDPOINT,
-	.bEndpointAddress = 0x81,
+	.bEndpointAddress = CDCACM_GDB_ENDPOINT | USB_REQ_TYPE_IN,
 	.bmAttributes = USB_ENDPOINT_ATTR_BULK,
 	.wMaxPacketSize = CDCACM_PACKET_SIZE,
 	.bInterval = 1,
@@ -118,7 +138,7 @@ static const struct {
 		.bDescriptorType = CS_INTERFACE,
 		.bDescriptorSubtype = USB_CDC_TYPE_CALL_MANAGEMENT,
 		.bmCapabilities = 0,
-		.bDataInterface = 1,
+		.bDataInterface = GDB_IF_NO + 1,
 	},
 	.acm = {
 		.bFunctionLength = sizeof(struct usb_cdc_acm_descriptor),
@@ -130,8 +150,8 @@ static const struct {
 		.bFunctionLength = sizeof(struct usb_cdc_union_descriptor),
 		.bDescriptorType = CS_INTERFACE,
 		.bDescriptorSubtype = USB_CDC_TYPE_UNION,
-		.bControlInterface = 0,
-		.bSubordinateInterface0 = 1,
+		.bControlInterface = GDB_IF_NO,
+		.bSubordinateInterface0 = GDB_IF_NO + 1,
 	 }
 };
 
@@ -155,7 +175,7 @@ static const struct usb_interface_descriptor gdb_comm_iface[] = {{
 static const struct usb_interface_descriptor gdb_data_iface[] = {{
 	.bLength = USB_DT_INTERFACE_SIZE,
 	.bDescriptorType = USB_DT_INTERFACE,
-	.bInterfaceNumber = 1,
+	.bInterfaceNumber = GDB_IF_NO + 1,
 	.bAlternateSetting = 0,
 	.bNumEndpoints = 2,
 	.bInterfaceClass = USB_CLASS_DATA,
@@ -169,7 +189,7 @@ static const struct usb_interface_descriptor gdb_data_iface[] = {{
 static const struct usb_iface_assoc_descriptor gdb_assoc = {
 	.bLength = USB_DT_INTERFACE_ASSOCIATION_SIZE,
 	.bDescriptorType = USB_DT_INTERFACE_ASSOCIATION,
-	.bFirstInterface = 0,
+	.bFirstInterface = GDB_IF_NO,
 	.bInterfaceCount = 2,
 	.bFunctionClass = USB_CLASS_CDC,
 	.bFunctionSubClass = USB_CDC_SUBCLASS_ACM,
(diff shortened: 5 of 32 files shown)


stoyan-shopov:
NOTE: I just found that handling of the `qCRC` remote serial protocol packet request from gdb does not work properly for this firmware.
The `qCRC` packet is sent to the BMP firmware as part of the `compare-sections` command, executed from gdb.
This is probably an issue with improper use of the CRC engine of the stm32f723.
I will debug this

UweBonnes:
Please move rcc_periph_clock_enable(RCC_CRC) from the platforms/*platform.c files to to the STM32 codepath of src/crc32.c as a seperate PR. This should fix it. Look in git log, I have done some harm with CRC32 too long time ago too.

stoyan-shopov:
Thanks!

Yes, enabling the CRC clock fixed this issue, I checked twice

stoyan-shopov:
Hm, I was too fast to respond, just saw your comment about `rcc_periph_clock_enable(RCC_CRC)`...
I will check the CRC clock enable in the other platform files

The issue definitely was that the CRC clock was not enabled

stoyan-shopov:
I see what you mean, I looked at `src/crc32.c`. Yes, that needs to be a separate pull request

If you don't mind, I would prefer, for the time being, to leave the `rcc_periph_clock_enable(RCC_CRC)` line in the `stlinkv3/platform.c` file. The last commit that I pushed, does just that, and corrects the CRC issue. Thank you for your insight!

I can prepare a request, for consolidating the `rcc_periph_clock_enable(RCC_CRC)` line, from the `platforms/xxx/platform.c` files to `src/crc32.c` at a later time

I just prefer doing one thing at a time

UweBonnes:
Yes, that is better for the moment. Putting the clock enable in crc32.c needs more changes. And probably we will not add too many stm32 platforms that could fall into the same pitfall ;-)

UweBonnes:
I force pushed. Two patches outside of stlinkv3 before, than your patches and mine on top. In the end I proposed to merge all but the first two into one, but now they are kept seperate for review

stoyan-shopov:
Thanks a lot for your work!

I am almost not getting at my machine these days, it's vacation time here. I will have a deeper look at your changes when I can spend more time on the computer.

Do you think it would be preferable to merge your changes, maybe do some testing, and create a new pull request, containing all changes as one commit?

Thanks, and happy holidays!

stoyan-shopov:
That's a good catch with relocating the vector table, it's better to do it explicitly in the firmware.
Right now, it seems to work without it, because the bootloader code in `dfu_jump_app_if_valid()` redirects the vector table, but it's better to have it explicitly in the firmware.

I need to read and learn more about caches... It looks quite suspicious why `register uint32_t sets;` would hang the code without the `register` keyword, does this hang every time? I think nowadays compilers ignore the `register` storage class specifier, so this is quite strange. You got the cache enabling code from some IAR example, probably? If so, I would expect it to work as is, but who knows...

Thanks for editing the readme. I will also test with the `stm32mem.py` script, I have only used `dfu-util` until now. I will also test flashing with the `STLinkUpgrade.jar` modification method, it is great that this actually works! I also need to edit the readme a bit, some of the notes and `TODO` items there are already done, and no longer relevant.

UweBonnes:
Yes, merging all F723 patches is intended. Only things that change something outside F723 or are more "fundamental" finding shoul be kept as seperate patch i.m.h.o

stoyan-shopov:
I updated the branch and force-pushed the changes.
Here is what I did. I merged your changes related to the stlinkv3 hardware into the main commit for the stlinkv3. I slightly edited the Readme file.

Most importantly, I discarded your change for workind around the issue with enabling the caches. This looks very suspicious, Actually, there was a slight error in handling the `NO_BOOTLOADER` define in the makefile, which makes the firmware always link at address 0x08020000 - and this will not work without a bootloader. I fixed this, I am not experiencing any issues with enabling the caches, so I discarded your change, it works on my side without it. Please, if you still experience such an issue, let me know which toolchain you are using so that I can try to reproduce this problem.

Thanks for taking care for the LEDs. Actually, LED indication has been pretty useless to me, because I could not find where the LED states are documented, so I don't know what the different patterns mean. If I am not missing it somewhere in the documents, I believe this is something worth adding to the documentation.

I have not done any other tests, I just made sure that the latest version builds, flashes and enumerates correctly, and detects an STM32F407 target device.

UweBonnes:
Did you run speed tests without caches? I guess, speed will be much slower as the loops now need to wait the FLASH wait states to rerun the loop. Frequency calculation needs to account for the missing caches. The code I entered is a (hopefully) 1:1 translation of the cmsis cache functions from core_cm7.h. So maybe with this explanations it is less "dubious".

stoyan-shopov:
Ah, caches are enabled, of course. You just had one small commit, which added the register keyword, and you noted that this is a workaround.
I am speaking of this commit - I just reverted adding the register storage class specifier, caches are of course enabled, your previous code is there, I have not touched it

stoyan-shopov:
I adjusted the copyright notes. I am not sure if that is what you meant.
Let me know if I have misunderstood something

UweBonnes:
On linux, dmesg reports:
[181204.039569] usb 2-1.1.3: new high-speed USB device number 20 using ehci-pci
[181204.148669] usb 2-1.1.3: config 1 interface 0 altsetting 0 endpoint 0x82 has an invalid bInterval 255, changing to 11
[181204.148672] usb 2-1.1.3: config 1 interface 1 altsetting 0 bulk endpoint 0x1 has invalid maxpacket 64
[181204.148673] usb 2-1.1.3: config 1 interface 1 altsetting 0 bulk endpoint 0x81 has invalid maxpacket 64
[181204.148675] usb 2-1.1.3: config 1 interface 2 altsetting 0 endpoint 0x84 has an invalid bInterval 255, changing to 11
[181204.148677] usb 2-1.1.3: config 1 interface 3 altsetting 0 bulk endpoint 0x3 has invalid maxpacket 32
[181204.148678] usb 2-1.1.3: config 1 interface 3 altsetting 0 bulk endpoint 0x83 has invalid maxpacket 64

Do you see a chance to fix?

stoyan-shopov:
Thanks for the report. Actually, the usbview utility from the windows ddk, also reports some problems with the usb descriptors on windows. It actually reports some problems with the master BMP firmware, if I remember correctly, but since it works, I have not payed much attention to this report. I will check again

This definitely needs to be fixed, I have a real linux machine here, not just a virtual machine, I will try to check the usb descriptors in the next few days

stoyan-shopov:
I am also seeing these errors on a real linux machine. The errors are not seen on a Virtualbox machine, there I see another error:
```
usb 1-2: config 1 interface 5 altsetting 0 endpoint 0x85 has invalid maxpacket 512, setting to 64
```
It looks like the stlinkv3 enumerates as a high-speed device on real hardware, and as a full-speed device on the virtual machine. 512 byte endpoints are not allowed for full-speed devices. Looks like I need to detect the enumeration speed and perform some parameterization of the USB descriptors.

It should be uncommon, and doesn't make much sense that the stlinkv3 be used in a full-speed configuration. Maybe it makes sense to just fix the descriptors for the high-speed case.

UweBonnes:
I think a USB HS device must also enumerate as FS device, e.g. when connected to a FS-only host.

stoyan-shopov:
Yes, it does. Enumeration is not the issue.
The issue is that some values in the usb descriptors are only valid for high-speed modes of operation.
For example, the USB2 specification says that the control endpoint size must be 64 bytes (smaller sizes are not compliant with the spec).
Also, 512 byte-size endpoints are not allowed for full-speed devices. This is currently the size of the trace IN endpoint 0x85, so that is why I am seeing such warning on a virtual machine, and you are not seeing it on real harware, working in high-speed mode.

It looks like there are also limitations on the polling intervals, from the USB specification:
```
A full-speed endpoint can specify
a desired period from 1 ms to 255 ms. Low-speed endpoints are limited to specifying only 10 ms to 255 ms.
High-speed endpoints can specify a desired period (2 ^ (bInterval-1) )x125 µ s, where bInterval is in the range 1 to
(including) 16
```
This is probably why the kernel gives those errors. I am not yet sure about why the other errors, about the endpoint sizes are issued, but I will investigate. A polling interval of 255 is currently used for the dummy interrupt IN endpoints used for monitoring the virtual com port status. They are not used, but must be present, if the device should be compliant with the CDCACM specification.

What do you think is a more proper solution - we can have two sets of descriptors, for full and high speed modes, but this will make the code more complicated. Or it would be safe to assume that people will not use the stlinkv3 as a full-speed device, and only have one set of descriptors, for the high-speed case?

This could also be mentioned in the documentation.

UweBonnes:
I think, two sets of descriptor are needed. E.g. when using the device in Virtualbox without the payed USB2 extension.

stoyan-shopov:
It looks like bulk endpoints should be 512 bytes in size for high-speed endpoints:
https://superuser.com/questions/80763/is-there-any-usb2-0-data-transfer-chunk-size-limit
```
At the USB bus level packet sizes are restricted by the USB Specifications — restrictions for USB 2.0 are:

    for bulk transfers — 512 bytes for high-speed endpoints, 8, 16, 32, or 64 bytes for full-speed endpoints (and low-speed bulk endpoints are not allowed at all);
```

I saw this in the USB spec: `The USB defines the allowable maximum bulk data payload sizes to be only 8, 16,
32, or 64 bytes for full-speed endpoints and 512 bytes for high-speed endpoints.` but thought that smaller sizes are also allowed. Seems like this isn't the case.

No worry, the STM32f723 has 4kBytes in total for usb memory, but allocating larger buffers will definitely limit the maximum achieved speed on the TRACE IN endpoint, which is expected to be the most heavily loaded one. I have achieved speeds of almost 35 MBytes/second with a test firmware, this speed will drop if more memory is allocated to other endpoints.

I agree that the cleanest solution is having two sets of descriptors. In order to minimize `#ifdef` handling, I will try to put the descriptors for the STM32F723 in a separate file.

UweBonnes:
It would be bad if every EP on the HS devices would need 512 bytes. Trace profites from a deep buffer.

stoyan-shopov:
I agree, but so says the specification. If we want to be compliant, we have to adjust the endpoint sizes.

The device works properly on windows, and maybe it will work properly on linux as well, despite of the warnings from the kernel log. Also, I *beleive* it would be possible to advertise the endpoints as 512 byte ones, but only allocate and send 64 byte IN packets, but this is not very good practice, I am not sure if the usb core will like this.

Actually, situation does not look that bad. There are currently 3 bulk IN endpoints in the firmware:
- the gdb data endpoint
- the UART endpoint
- the trace endpoint

Note that what really matters is the IN endpoints, because there is dedicated USB for each IN endpoint. For OUT endpoints, the USB packet memory is unified, and shared for all OUT endpoints.

Here are some calculations (I hope I am getting them right):
- we need 64 bytes of dedicated memory for the control IN endpoint
- interrupt dummy IN endpoints can be small, 8 bytes, as is the case now, so they don't matter. They are needed so that the firmware is compliant with the CDCACM specification, but are otherwise not significant or used
- if we want to be USB2 compliant, we will need 3 buffers, 512 bytes each, of dedicated memory for the three bulk IN endpoint
- As the largest possible OUT packet is 512 bytes, there should be at least 512 bytes, + a few more bytes for control information - this memory is shared between all OUT endpoints. Having more memory dedicated for OUT endpoints will make OUT transfers faster, but that is really not needed, I think communication will be fast enough without extra memory

So, we have: 64 + 2 * 8 + 3 * 512 + 512 = 2128 bytes necessary, so that the firmware is compliant. There are 1968 more bytes available, and this means they can fit 3 more 512 byte packets for the TRACE IN endpoint, so there will be 4 IN packet buffers - these may be enough for achieving good TRACE throughput, but tests are needed to confirm this. The remaining unused memory can be allocated for OUT packets.

Regarding enumeration speed, I found this in the USB spec:
```
Device_Qualifier
The device_qualifier descriptor describes information about a high-speed capable device that would
change if the device were operating at the other speed. For example, if the device is currently operating
at full-speed, the device_qualifier returns information about how it would operate at high-speed and
vice-versa.
```
It looks like a device qualifier descriptor needs to be provided. I have seen examples in the CubeMX sources, I will consult them.

UweBonnes:
Regarding the max packet size I had some discussion with zyp on discord #usb-hacking

stoyan-shopov:
Thanks for discussing that issue there! I just read the discussion.
Unfortunately, I am getting quite pressed on my day job, and have very little time to spare :-(

If I understand correctly from the discussion, my understanding looks correct - for bulk endpoints, their size must be reported as 512 bytes.
Also, this does not really matter much for OUT (host to device) endpoints, as the memory is shared, so if we configure the OUT FIFO to be 512 bytes + a few more bytes (I think, 8 bytes more were needed by the USB core), we should be fine.

Also, as zyp says, probably we may cheat - we could say endpoints are 512 bytes in size, but only send e.g. 64 bytes, on IN (device to host) endpoints. This way, we can configure IN endpoints to use only 64 bytes of memory. Actually, this is not cheating, we just declare that we *may* send 512 bytes, but only send 64 bytes maximum on IN endpoints. That will save precious USB FIFO memory. Only the TRACE IN endpoints will be large.

I investigated, and it really seems, as zyp suggests, that the device should check at what speed it enumerates, and provide two different sets of descriptors - for full and high speed. The enumeration speed can be checked in the `OTG_DSTS` register, field `ENUMSPD`, for the stm32f723. I think the documentation is wrong - it only shows a value for full speed, but I guess a value of 3 should mean high-speed, I haven't checked the register's value for this field. We can assume that all BMP probes until now are running at full speed, and only check and provide an alternate set of descriptors in the stlinkv3 case.

I was hoping I could do this the next couple of weeks, but it looks I will not have any time to spare working on this now, sorry...

UweBonnes:
I was able to load a program to a can connected device with a bootloader and socketcan, so the slcan implementation works somehow...

koendv:
usb problems.

$ uname -a
Linux raspberrypi 5.10.52-v8+ #1441 SMP PREEMPT Tue Aug 3 18:14:03 BST 2021 aarch64 GNU/Linux

[ 2506.239086] usb 1-1.3: new high-speed USB device number 8 using xhci_hcd
[ 2506.339419] usb 1-1.3: config 1 interface 1 altsetting 0 bulk endpoint 0x1 has invalid maxpacket 64
[ 2506.339440] usb 1-1.3: config 1 interface 1 altsetting 0 bulk endpoint 0x81 has invalid maxpacket 64
[ 2506.339457] usb 1-1.3: config 1 interface 3 altsetting 0 bulk endpoint 0x3 has invalid maxpacket 32
[ 2506.339472] usb 1-1.3: config 1 interface 3 altsetting 0 bulk endpoint 0x83 has invalid maxpacket 64
[ 2506.339487] usb 1-1.3: config 1 interface 5 altsetting 0 bulk endpoint 0x85 has invalid maxpacket 64
[ 2506.339502] usb 1-1.3: config 1 interface 7 altsetting 0 bulk endpoint 0x6 has invalid maxpacket 64
[ 2506.339516] usb 1-1.3: config 1 interface 7 altsetting 0 bulk endpoint 0x86 has invalid maxpacket 64
[ 2506.339923] usb 1-1.3: New USB device found, idVendor=1d50, idProduct=6018, bcdDevice= 1.00
[ 2506.339938] usb 1-1.3: New USB device strings: Mfr=1, Product=2, SerialNumber=3
[ 2506.339951] usb 1-1.3: Product: Black Magic Probe STLINK-V3 v1.6-rc0-1178-ge7730018-dirty
[ 2506.339963] usb 1-1.3: Manufacturer: Black Sphere Technologies
[ 2506.339976] usb 1-1.3: SerialNumber: 0050002F3137511233333639
[ 2506.347261] cdc_acm 1-1.3:1.0: ttyACM0: USB ACM device
[ 2506.348651] cdc_acm 1-1.3:1.2: ttyACM1: USB ACM device
[ 2506.350649] cdc_acm 1-1.3:1.6: ttyACM2: USB ACM device

Three serial ports are created, but they do not accept data. 
FYI: https://github.com/koendv/bmp/releases

koendv:
Better, but not yet:

[10487.421523] usb 1-1.3: new high-speed USB device number 26 using xhci_hcd
[10487.522308] usb 1-1.3: New USB device found, idVendor=1d50, idProduct=6018, bcdDevice= 1.00
[10487.522331] usb 1-1.3: New USB device strings: Mfr=1, Product=2, SerialNumber=3
[10487.522345] usb 1-1.3: Product: Black Magic Probe STLINK-V3 v1.6-rc0-1223-ge9a2cdee
[10487.522358] usb 1-1.3: Manufacturer: Black Sphere Technologies
[10487.522369] usb 1-1.3: SerialNumber: 0050002F3137511233333639
[10487.533044] cdc_acm 1-1.3:1.0: ttyACM0: USB ACM device
[10487.537574] cdc_acm 1-1.3:1.2: ttyACM1: USB ACM device
[10487.542482] cdc_acm 1-1.3:1.6: ttyACM2: USB ACM device

(gdb) tar ext /dev/ttyBmpGdb
Remote debugging using /dev/ttyBmpGdb
Ignoring packet error, continuing...
warning: unrecognized item "timeout" in "qSupported" response
Ignoring packet error, continuing...
Remote replied unexpectedly to 'vMustReplyEmpty': timeout
(gdb) q

I assume this works for you , so maybe it's my desktop?
koen@raspberrypi:~ $ uname -a
Linux raspberrypi 5.10.52-v8+ #1441 SMP PREEMPT Tue Aug 3 18:14:03 BST 2021 aarch64 GNU/Linux

UweBonnes:
Just for reference: A "normal" BMP works on the Rpi?

koendv:
Yes. bmp on blue and black pill are fine.

koen

UweBonnes:
Took me some time to install RPI:
```(gdb) mon 
Black Magic ProbeSTLINK-V3 v1.7.1-349-ge9a2cdee, Hardware Version 0
Copyright (C) 2015  Black Sphere Technologies Ltd.

(gdb) mon s
Target voltage: 3.3V
Please report Designer  15 and Partno   0 and the probed device
Available Targets:
No. Att Driver
*** 1 Unknown ARM Cortex-M Designer  15 Partno   0 M33
(gdb) quit
pi@raspberrypi:~/devel $ uname -a
Linux raspberrypi 5.10.63-v7+ #1459 SMP Wed Oct 6 16:41:10 BST 2021 armv7l GNU/Linux
```

koendv:
OK. This is what I do:
$ uname -a 
Linux raspberrypi 5.10.52-v8+ #1441 SMP PREEMPT Tue Aug 3 18:14:03 BST 2021 aarch64 GNU/Linux
$ arm-none-eabi-gcc --version
arm-none-eabi-gcc (GNU Arm Embedded Toolchain 9-2020-q2-update) 9.3.1 20200408 (release)

git clone -b stlinkv3 https://github.com/stoyan-shopov/blackmagic/
cd blackmagic/
make PROBE_HOST=stlinkv3
cd src
koen@raspberrypi:~/src/stlinkv3/blackmagic/src $ ls -l blackmagic*
-rwxr-xr-x 1 koen koen  112020 Dec  3 17:47 blackmagic.bin
-rwxr-xr-x 1 koen koen    8776 Dec  3 17:47 blackmagic_dfu.bin
-rwxr-xr-x 1 koen koen  434200 Dec  3 17:47 blackmagic_dfu.elf
-rw-r--r-- 1 koen koen   24740 Dec  3 17:47 blackmagic_dfu.hex
-rwxr-xr-x 1 koen koen 1543684 Dec  3 17:47 blackmagic.elf

I flash blackmagic_dfu.elf and blackmagic.elf and reset

(gdb) tar ext /dev/ttyBmpGdb
Remote debugging using /dev/ttyBmpGdb
Remote debugging using /dev/ttyBmpGdb
Ignoring packet error, continuing...
warning: unrecognized item "timeout" in "qSupported" response

Could you attach compiled firmware to your post so we can rule out a compiler problem?

UweBonnes:
I suspect /dev/ttyBmpGdb points to a bad device. Please verify the device.

koendv:
Tried /dev/ttyACM0, /dev/ttyACM1 and /dev/ttyACM2. All give timeouts.

UweBonnes:
Dies the StlinkV3 work on x86(_64) PC? What does hosted "blackmagic -t" show?

UweBonnes:
Git head shows for me"```commit e9a2cdee7bdba5d85ad2f7f637a72e4b9664bf15 (HEAD -> stlinkv3, shopov/stlinkv3)
Author: Uwe Bonnes <bon@elektron.ikp.physik.tu-darmstadt.de>
Date:   Tue Nov 30 20:36:45 2021 +0100

    stlinkv3: Added some delay needed by a Stlinkv3-Set.
``` Btw: What about contining on discord to not clutter to much?

**Comment by koendv** [comment]
On Fri, 03 Dec 2021 09:26:18 -0800
UweBonnes ***@***.***> wrote:

> Dies the StlinkV3 work on x86(_64) PC? What does hosted "blackmagic
> -t" show?
> 

***@***.***:~/src/blackmagic $ ./src/blackmagic -t
INFO: Open USB 04e8:61f5 class  0 failed
BMP in bootloader mode found. Restart or reflash!
No debugger found. Please check access rights to USB devices!

The wire-wrap wire I use to connect to swd just broke off; so I guess that's it for today.

koen

**Comment by stoyan-shopov** [comment]
Hi,

First, a big **THANK YOU!**, Uwe, for keeping this branch up to date with the BMP master firmware! You are doing an amazing work!

Personally, I had much troubles this year, and virtually no time to follow the development of the BMP firmware. I haven't visited the discord channels for several months as well. Hopefully, things will get better on my side next year...

Anyway, I tried the latest **stlinkv3** branch a few days ago, and I also experience big timeouts when connecting to the **stlinkv3** probe, with the **stlinkv3** branch firmware flashed on it. I am working on a Windows 10 machine.

The timeouts are experienced when establishing a connection to the probe from gdb.
After a brief inspection of the usb traffic with **wireshark**, my feeling is this has something to do with invoking **USB_CDC_REQ_SET_LINE_CODING**, as there is where I observe unusually large timeouts. This is just a guess, though.

Once the connection from gdb to the **stlinkv3** is established, ***NO*** timeouts are observed, and communication with the probe is smooth.

When a `disconnect` command is issued from **gdb**, the timeouts manifest them once again (**USB_CDC_REQ_SET_LINE_CODING** is being triggered at that time as well).

My very wild guess is that this anomaly may have something to do with the **SLCAN** usb interface support.
I think so, because disabling the **SLCAN** interface resolves the timeout issues for me:
```
diff --git a/src/platforms/stlinkv3/platform.h b/src/platforms/stlinkv3/platform.h
index 2ef1031b..9a4fa702 100644
--- a/src/platforms/stlinkv3/platform.h
+++ b/src/platforms/stlinkv3/platform.h
@@ -73,7 +73,6 @@ int usbuart_debug_write(const char *buf, size_t len);
 #define NUM_TRACE_PACKETS              (16)
 #define TRACESWO_PROTOCOL              2                       /* 1 = Manchester, 2 = NRZ / async */

-#define PLATFORM_HAS_SLCAN             1
 #define CAN_APB_FREQUENCY  (54*1000*1000L)
 #define CAN_RX0_IRQ   NVIC_CAN1_RX0_IRQ
 #define CAN_RX0_ISR   can1_rx0_isr
```

I thought I should let you know of these observations.

Thanks,
Stoyan

koendv:
Thank you Stoyan. 

If I set
#define PLATFORM_HAS_SLCAN             0
then stlinkv3 works fine for me.

rtt also works; printing speed from an Arduino sketch in a blue pill is 49795 characters/s.

best,
koen

UweBonnes:
Stoyan Shopov writes:
> Hi,
> 
> First, a big THANK YOU!, Uwe, for keeping this branch up to date with the BMP
> master firmware! You are doing an amazing work!

It is egoistic. I have  a test adapter with SWJ/Serial and CAN  with 3
plugs that I want to substitute with a single plug. So my need for CAN.

> 
> Personally, I had much troubles this year, and virtually no time to
> follow the 
> development of the BMP firmware. I haven't visited the discord channels for
> several months as well. Hopefully, things will get better on my side next
> year...
My best wishes that things get better for you !
>
> ...
> I think so, because disabling the SLCAN interface resolves the timeout issues
> for me:
>

I will push that change, ask for tests, try to test myself and find
the root cause...

Cheers
-- 
Uwe Bonnes                ***@***.***

Institut fuer Kernphysik  Schlossgartenstrasse 9  64289 Darmstadt
--------- Tel. 06151 1623569 ------- Fax. 06151 1623305 ---------
(Phone also available during "mobile work")

koendv:
Hi stoyan. Thanks for porting bmp to stlink v3.
I add a patch so the led blinks when bmp is active.
[stlinkv3_led.txt](https://github.com/blacksphere/blackmagic/files/7709762/stlinkv3_led.txt)

dragonmux:
Closing this in favour of #1354 - thank you for the work and contribution stoyan-shopov :) The only reason for not keeping this open is the amount of work needed to update it for the current code-base, which James Turton has been kind enough to expend.

PR #828, opened 2021-01-29
The USB CDC is no modem and does not use the AT protocol

Ho-Ro:
This prevents also the intervention of "ModemManager" that can happen on some linux systems.

Signed-off-by: Martin <Ho-Ro@users.noreply.github.com>


UweBonnes:
Good catch! Thanks.

PR #836, opened 2021-02-22
target/stm32f1: halt the target when idcode is 0

mmoskal:
This happens when the target is stuck in WFI.
target_halt_poll() waits up to 2s for the WFI to end.

I have a program on STM32F030 that goes into deep sleep (STOP mode) for up to 10ms, and then uses RTC alarm IRQ to wake up. Around half of the time, when attaching the chip is detected as generic Cortex M0. With this patch it is detected as F030 all the time.

BTW, I found the BMP to be the only debugger that connects to that target at all when sleeping. Openocd and pyocd both require a reset.

### DIFF
--- a/src/target/stm32f1.c
+++ b/src/target/stm32f1.c
@@ -149,6 +149,15 @@ bool gd32f1_probe(target *t)
 	target_add_commands(t, stm32f1_cmd_list, t->driver);
 	return true;
 }
+
+static void stm32f1_read_idcode(target *t)
+{
+	if ((t->cpuid & CPUID_PARTNO_MASK) == CORTEX_M0)
+		t->idcode = target_mem_read32(t, DBGMCU_IDCODE_F0) & 0xfff;
+	else
+		t->idcode = target_mem_read32(t, DBGMCU_IDCODE) & 0xfff;
+}
+
 /**
     \brief identify the stm32f1 chip
 */
@@ -156,10 +165,17 @@ bool gd32f1_probe(target *t)
 bool stm32f1_probe(target *t)
 {
 	uint16_t stored_idcode = t->idcode;
-	if ((t->cpuid & CPUID_PARTNO_MASK) == CORTEX_M0)
-		t->idcode = target_mem_read32(t, DBGMCU_IDCODE_F0) & 0xfff;
-	else
-		t->idcode = target_mem_read32(t, DBGMCU_IDCODE) & 0xfff;
+
+	stm32f1_read_idcode(t);
+
+	/* This likely means the MCU was in deep sleep. */
+	if (t->idcode == 0) {
+		/* Try to halt it outside of WFI. */
+		target_halt_poll(t, NULL);
+		/* Re-read the idcode. */
+		stm32f1_read_idcode(t);
+	}
+
 	size_t flash_size;
 	size_t block_size = 0x400;
 	switch(t->idcode) {
--- a/src/target/stm32f1.c
+++ b/src/target/stm32f1.c
@@ -166,12 +166,12 @@ bool gd32f1_probe(target *t)
 bool stm32f1_probe(target *t)
 {
 	uint16_t stored_idcode = t->idcode;
-	uint32_t dbgmcu = ((t->cpuid & CPUID_PARTNO_MASK)) ?
+	uint32_t dbgmcu = ((t->cpuid & CPUID_PARTNO_MASK) == CORTEX_M0) ?
 		DBGMCU_IDCODE_F0 : DBGMCU_IDCODE;
-	t->idcode = target_mem_read32(t, dbgmcu) & 0xfff;
 	ADIv5_AP_t *ap = cortexm_ap(t);
 	ap->ap_storage = target_mem_read32(t, dbgmcu + 4);
-	target_mem_write32(t, dbgmcu + 4, ap->ap_storage & 7);
+	target_mem_write32(t, dbgmcu + 4, ap->ap_storage | 0x7); /* enable debug in all sleep modes */
+	t->idcode = target_mem_read32(t, dbgmcu) & 0xfff;
 	t->detach = stm32f1_detach;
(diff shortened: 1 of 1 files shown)


UweBonnes:
Can you run hosted and show the printout of "blackmagic -t " when the device is in that sleep period.. Perhaps this at least documents the behavior.  Otherwise, what about setting the debugmcu sleep bit at the beginning of the probe function and reset with detach? I already do that for F7.

mmoskal:
I'm unable to get hosted to work with a BMP as a probe (I'm on mac). Will try harder.

With STLINK/V2 clone I randomly get a correct connection or not (regardless of my patch). When it fails, it seems the failure comes from the probe itself (No device connected).

With DAPLink, it just always hangs after `RESET_SEQ failed` when the target is in periodic sleep mode. When target is not sleeping, it scans like the STLINKV2.

Success on STLINK:
```
BMP hosted v1.7.1-128-gd70fa8c-dirty
Using 0483:3748 5?m4R775	W STMicroelectronics
 STM32 STLink
STLink firmware version: V2J37S7
Leaving DEBUG Mode
Running in Test Mode
Target voltage: 3.24V Volt
Divisor for 4.0000 MHz is 0
Speed set to  3.6000 MHz for SWD
Leaving MASS Mode
Read Core ID: 0x0bb11477
DPIDR 0x0bb11477 (v1 MINDP rev0)
RESET_SEQ failed
AP   0: IDR=04770021 CFG=00000000 BASE=e00ff003 CSW=03000040
AP#0 IDR = 0x04770021 (AHB-AP var2 rev0)
ROM: Table BASE=0xe00ff000 SYSMEM=0x00000001, designer  20 Partno 440
0 0xe000e000: Generic IP component - Cortex-M0 SCS (System Control Space) (PIDR = 0x04000bb008  DEVTYPE = 0x00 ARCHID = 0x0000)-> cortexm_probe
CPUID 0x410cc200 (M0 var 0 rev 0)
1 0xe0001000: Generic IP component - Cortex-M0 DWT (Data Watchpoint and Trace) (PIDR = 0x04000bb00a  DEVTYPE = 0x00 ARCHID = 0x0000)
2 0xe0002000: Generic IP component - Cortex-M0 BPU (Breakpoint Unit) (PIDR = 0x04000bb00b  DEVTYPE = 0x00 ARCHID = 0x0000)
ROM: Table END
***  1      STM32F03 M0
Ram   Start: 0x20000000, length    0x5000
Flash Start: 0x08000000, length    0x8000, blocksize    0x400
```

Fail on STLINK:
```
BMP hosted v1.7.1-128-gd70fa8c-dirty
Using 0483:3748 5?m4R775	W STMicroelectronics
 STM32 STLink
STLink firmware version: V2J37S7
Leaving DEBUG Mode
Running in Test Mode
Target voltage: 3.23V Volt
Divisor for 4.0000 MHz is 0
Speed set to  3.6000 MHz for SWD
Leaving MASS Mode
Read Core ID: 0x0bb11477
DPIDR 0x0bb11477 (v1 MINDP rev0)
RESET_SEQ failed
write_retry failed. No device connected
No device connected
No target found
```

UweBonnes:
Thanks for the logs. The second shows clearly that there was no contact to the STN device. The only chance to get to the device  is either connect under reset  and loose the state or wait much longer to catch the device awake with a longer "halt_timeout".  If we catch the device, we need to set DBGMCU sleep bits. I will try to come up with a solution, but if you try yourself, that would help too. Have a look how the F7 is handled. However the F7 is a ADIv5 V2 device, we knoe much earlier that it is a F7

UweBonnes:
Could you test the changes I pushed to your branch. I only compiled, did not test...

mmoskal:
Sorry for long delay. I modified your patch and now it works.

If your check-in https://github.com/UweBonnes/blackmagic/commit/91eb1206d1edda962139a8c00468cb232ff668f7 going to be merged anytime soon?

UweBonnes:
Resolved by f89b07d892c6

mmoskal:
@UweBonnes the STM32F030F4 is now detected as "STM32F05/F030x8" whereas it used to be detected as "STM32F03" (this is regardless if it sleeps or not during detection). The difference is that the first one is specified to have 64kB of flash and the second 32kB. However the actual part has 16kB, so I don't think I care much. Thanks!

UweBonnes:
There are a zillion of STM32 chip variants.  They are produced all the same, but tested different and perhaps with some flash disable or not . If we read out flash size and only announce that size in the memory map, we disable users to try to use flash above the the ST announced size. Use the right linker script and there is no chance to try to use the upper flash. I do not see a danger here. Are the argument  understandable?

mmoskal:
Makes perfect sense, thanks!

My understanding is that the partid in PIDR registers is always 0x440 for the F0xx parts (though I couldn't find anything in ST docs to back this), and then the idcode (0x40015800) differentiates further (documented by ST). With the idcode reading removed, all parts with partid==0x440 are detected as F05 which is fine by me, however I wonder if F07x and F09x are now detected correctly - they have even more flash and different block size. Unfortunately I don't have these chips to test. Also not sure about the F1 clones.

I guess we could just wait until someone complains ;)

mmoskal:
Or maybe to put it differently, the cases for 0x444 and 0x445 (I only had F030 and F042 to test) in stm32f1_probe() are now dead code. If there are others that are also dead code that might be a problem, which I guess we'll figure out in due time.

UweBonnes:
Ah: PIDR != MCUID. That needs to be considered. So create the memory map in attach when the device is halted, not in probe. Can you come up with a solution?

mmoskal:
The patch from this PR actually worked - the one you proposed with my fixes (no halt, just setting that debug-in-sleep bit). I can resurrect that.

UweBonnes:
Can you please test if #846 fixes your problem?

mmoskal:
works like a charm, thank you!

PR #840, opened 2021-03-06
Cortexm prepare

UweBonnes:
- Always halt CPU before probing. Is an option needed not to halt?
- For STM32F4/7 always set debug sleep bits with probe and attach
- Restore debug sleep bit on detach to value stored on probe.
- Provide monitor command to inhibit restore to ease next  probe/halt.


UweBonnes:
Superceded by other changes

Issue #117, opened 2016-02-21
Watchdog not paused when debugging


avrhack:
As far as I can see both from testing and from looking at the source, the watchdog is not paused when at a breakpoint, so any applications that use watchdogs have to have the source code edited before being debugged.

It's trivial for the bmp to set the two flags for the two watchdogs to pause their counters on break. 

Something that's been considered and rejected in the past or something that could be done?

gsmcmullin:
To which target are you referring?  I'd be happy to accept a patch.  We've
been using the STM32F4 with the watchdog enabled, and it is paused while
the target is halted.

On Sun, Feb 21, 2016 at 9:21 PM, avrhack <notifications@github.com> wrote:

> As far as I can see both from testing and from looking at the source, the
> watchdog is not paused when at a breakpoint, so any applications that use
> watchdogs have to have the source code edited before being debugged.
>
> It's trivial for the bmp to set the two flags for the two watchdogs to
> pause their counters on break.
>
> Something that's been considered and rejected in the past or something
> that could be done?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/blacksphere/blackmagic/issues/117>.
>



-- 
*Black Sphere Technologies Ltd.*

Web: www.blacksphere.co.nz
Mobile: +64 27 777 2182
Tel: +64 9 478 8885
Skype: gareth.mcmullin
LinkedIn: http://nz.linkedin.com/in/gsmcmullin

avrhack:
It was an STM32F303 so surprised it is fine on your STM32F4 as I can't see any code that would halt the WD. You positive the WD is working normally? Also, which WD are you using as one is synchronous and I'm using the other one :)

UweBonnes:
I also see no code to disable the watchdogs. But disabling Timers and watchdogs during debug is best done in the debugger startup file.

esden:
Do we have a conclusion on what should be done here? Do we want/need to patch this before the V1.6 release? I think stopping the asynchronous watchdog when halted is probably a good idea. I am just not sure if that can be done in a generic manner or needs to be done per chip family.

UweBonnes:
>>>>> "Piotr" == Piotr Esden-Tempski <notifications@github.com> writes:

    Piotr> Do we have a conclusion on what should be done here? Do we
    Piotr> want/need to patch this before the V1.6 release? I think stopping
    Piotr> the asynchronous watchdog when halted is probably a good idea. I
[...]

PR #841, opened 2021-03-10
Comprehensive STM32G0 driver

### LINKED ISSUE #819: STM32G070 option bytes support
Hi,
I am working on STM32G070 in my daily job and I notice the option bytes are not implemented in BMP yet.
STM32G0 support is implemented in stm32l4.c.
I actually bricked a Nucleo-G070RB by applying by mistake (before analysing BMP firmware code) STM32L4 option bytes to STM32G070.
I did a "monitor option write help" (yes it has no sense but did not throw any error) then "monitor option erase".

I have done a significant work on option bytes analysis (including PCROP_RDP option bit designed to clear the flash R/W protected pages, not much documented by the manufacturer, even in the Reference Manual Rev 5).
I would like to contribute to the project in that perspective. So my questions are:

- Is there a special meaning for STM32Gx support residing in stm32l4.c? Memory space?
- Is there a special objection if I suggest a contribution with a separate stm32g0.c file? Given that there is already some if() existing in stm32l4.c to differentiate chips and STM32G0x support could grow as:
    - STM32G0[B,C]x are arriving with more than 128 kBytes of Flash;
    - STM32G05x is missing (more RAM than G03x);
    - Other contributions possible...

Is my help welcome?
What do you suggest to me? In stm32l4.c or a separate file?

fabalthazar:
Fixes #819.

This PR provides a comprehensive STM32G0 target driver in a separate file. G0 has some divergences from L4/G4.
And may leave cleaner room in stm32l4.c for #707 and #458. Maybe...

Refer to this message for content details:
https://github.com/blacksphere/blackmagic/issues/819#issuecomment-780142079

Waiting for Dual-bank G0B/C release (in the ST's pipe) might be reasonable to fully test it.

### DIFF
--- a/src/target/stm32g0.c
+++ b/src/target/stm32g0.c
@@ -1,0 +1,710 @@
+/*
+ * This file is part of the Black Magic Debug project.
+ *
+ * MIT License
+ *
+ * Copyright (c) 2021 Fabrice Prost-Boucle <fabalthazar@falbalab.fr>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+/*
+ * This file implements STM32G0 target specific functions for detecting
+ * the device, providing the XML memory map and Flash memory programming.
+ *
+ * References:
+ * RM0454 - Rev 5 (Value line)
+ *   Reference manual - STM32G0x0 advanced ARM(R)-based 32-bit MCUs
+ *                      (STM32G030/STM32G050/STM32G070/STM32G0B0)
+ * RM0444 - Rev 5 (Access line)
+ *   Reference manual - STM32G0x1 advanced ARM(R)-based 32-bit MCUs
+ *                      (STM32G031/STM32G041/STM32G051/STM32G061/
+ *                       STM32G071/STM32G081/STM32G0B1/STM32G0C1)
+ * PM0223 - Rev 5
+ *   Programming manual - Cortex(R)-M0+ programming manual for STM32L0, STM32G0,
+ *                        STM32WL and STM32WB Series
+ */
+
+#include "general.h"
+#include "target.h"
+#include "target_internal.h"
+#include "cortexm.h"
+#include "command.h"
+
+/* FLASH */
+#define FLASH_START                     0x08000000
+#define FLASH_MEMORY_SIZE               0x1FFF75E0
+#define FLASH_PAGE_SIZE                 0x800
+#define FLASH_BANK2_START_PAGE_NB       256U
+#define FLASH_SIZE_MAX_G03_4            (64U * 1024U)    // 64 kiB
+#define FLASH_SIZE_MAX_G05_6            (64U * 1024U)    // 64 kiB
+#define FLASH_SIZE_MAX_G07_8            (128U * 1024U)   // 128 kiB
+#define FLASH_SIZE_MAX_G0B_C            (512U * 1024U)   // 512 kiB
+
+#define G0_FLASH_BASE                   0x40022000
+#define FLASH_ACR                       (G0_FLASH_BASE + 0x000)
+#define FLASH_ACR_EMPTY                 (1U << 16U)
+
+#define FLASH_KEYR                      (G0_FLASH_BASE + 0x008)
+#define FLASH_KEYR_KEY1                 0x45670123
+#define FLASH_KEYR_KEY2                 0xCDEF89AB
+#define FLASH_CR                        (G0_FLASH_BASE + 0x014)
+#define FLASH_CR_LOCK                   (1U << 31U)
+#define FLASH_CR_OBL_LAUNCH             (1U << 27U)
+#define FLASH_CR_OPTSTRT                (1U << 17U)
+#define FLASH_CR_STRT                   (1U << 16U)
+#define FLASH_CR_MER2                   (1U << 15U)
+#define FLASH_CR_MER1                   (1U << 2U)
+#define FLASH_CR_BKER                   (1U << 13U)
+#define FLASH_CR_PNB_SHIFT              3U
+#define FLASH_CR_PER                    (1U << 1U)
+#define FLASH_CR_PG                     (1U << 0U)
+
+#define FLASH_SR                        (G0_FLASH_BASE + 0x010)
+#define FLASH_SR_BSY2                   (1U << 17U)
+#define FLASH_SR_BSY1                   (1U << 16U)
+#define FLASH_SR_OPTVERR                (1U << 15U)
+#define FLASH_SR_RDERR                  (1U << 14U)
+#define FLASH_SR_FASTERR                (1U << 9U)
+#define FLASH_SR_MISSERR                (1U << 8U)
+#define FLASH_SR_PGSERR                 (1U << 7U)
+#define FLASH_SR_SIZERR                 (1U << 6U)
+#define FLASH_SR_PGAERR                 (1U << 5U)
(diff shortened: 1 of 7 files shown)


UweBonnes:
Several requests.issues:
- I am somehow torn apart by the amount of additional flash (2.6 k) needed, but I am willing to accept.
- As this is new code, documenting steps forward and backward in the git history does not help other users to understand the code. So for the PR, squash into one file or discuss reasons not to do so
- Add real usage examples of the new features you introduced, either in the code, in the wiki or  where you think it fits.

UweBonnes:
Also if you do not mind: Tag new files that you created with a more liberal license than the GNU license, e.g. BSD.

fabalthazar:
Ouch 2.6k is enormous. I suspected but not that much.
I analysed the .text section usage at the function level (after making functions non static so the global footprint is even higher but it is relevant for proportions). Here is a summary:

- Fundamental functions related to probe, attach/detach, flash erase/write => 972 bytes
And for custom commands:
- erase mass/bank => 122 bytes
- option bytes => 938 bytes => way too much!

In the option bytes part, a single function _stm32g0_is_PCROP_active()_ occupies no less than 136 bytes by itself just for combinatorial computations to know whether the Proprietary Code Read Out Protection is active or not.

So I have two thoughts:

1. The primary idea to have a separate file (as discussed in linked #819) was ambitious. From this memory analysis, one can estimate that an additional STM32 driver will occupy at least ~1k of code for fundamental features. Here, this part brings STM32G0B/C dual-bank specificity so it can be managed in existing stm32l4.c at the price of additional if() or switch/case complexity;
2. For option bytes, I tried to implement it in a generic and flexible manner. So it could make sense if it is shared with other STM32 drivers. For example in a stm32_common.c file. If this stays isolated in G0 file, this surely is overkill. So apart from PCROP stuff, default bytes can be implemented in stm32l4.c too.

Generally speaking, have you thought about the concept of STM32 API? STM32 drivers often have the same patterns.
This may not be easy for flash management due to many chips specificities. Or at the cost of versioned functions like stm32_flash_write_v1, _v2 (that must be well documented!)... with many arguments to deal with registers.
Or maybe this concept can apply to option bytes, the way they are given on the command line, parsed, validated or not, written, erased, etc. Current STM32 drivers seam to have their own way of dealing with these bytes. And this could lead to even lower flash usage.

So so so... What I propose to transform this PR into. Only implement these little features/fix (from complete list in https://github.com/blacksphere/blackmagic/issues/819#issuecomment-780142079) in stm32l4.c:

- Support for STM32G05x/G06x;
- Really enable debug in low power modes (DBG peripheral clock enabled). Current L4 implementation may not work (it does not on G0 chips);
- Watchdogs (IWDG and WWDG) are stopped when the core is halted. I found it to be the only way to program a binary file with hosted blackmagic where the debug registers cannot be written. Already appeared in #117 ;
- Lock the Flash back after a flash operation to leave the device in a state as close as possible to its previous state;
- Properly clear EOP flag after a flash erasure/write as recommended;
- EMPTY bit cleared after a successful programming to allow rebooting in Main Flash memory without a necessary power off. Particularly useful in case of virgin programming in production where subsequent actions are done automatically, for instance;
    Option bytes:
- G0-specific default values. I did not reproduce an "option erase" command on G0 with current STM32L4 values so I cannot confirm whether it is the reason for my bricked Nucleo board;
- Option bytes support for G0 dual-bank devices.

The idea behind the _irreversible (enable|disable)_ was to eventually enable OTP area programming (as CMSIS-DAP). Perhaps makes sense in a shared common file too.

This proposal may assuredly limit a lot the memory blow up.

Your advise?

fabalthazar:
Let me know if you're interested in the concept of shared functions between STM32 (globally or limited to option bytes). In case I find time, I could prototype something.
Limited to option bytes may at least unify the option bytes programming CLI in the STM32 range. At the price of breaking compatibility with current users private scripts (needs doc in Wiki and release note).

src/target/stm32g0.c
```diff
@@ -0,0 +1,758 @@
...
+	STM32G07_8 = 0x460,
+	STM32G0B_C = 0x467
+};
+struct saved_regs_s {
+	uint32_t rcc_apbenr1;
+	uint32_t dbg_cr;
+	uint32_t dbg_apb_fz1;
+};
+
+static struct saved_regs_s saved_regs = {0};
```
UweBonnes, line 148: But that as piv in the target structure

src/target/stm32g0.c
```diff
@@ -0,0 +1,758 @@
...
+static bool irreversible_enabled = false;
```
UweBonnes, line 149: Why not as additional entity in saved_regs

src/target/stm32g0.c
```diff
@@ -0,0 +1,758 @@
...
+	saved_regs.dbg_cr = target_mem_read32(t, DBG_CR);
+	target_mem_write32(t, DBG_CR, saved_regs.dbg_cr |
+			   (DBG_CR_DBG_STANDBY | DBG_CR_DBG_STOP));
+	saved_regs.dbg_apb_fz1 = target_mem_read32(t, DBG_APB_FZ1);
+	target_mem_write32(t, DBG_APB_FZ1, saved_regs.dbg_apb_fz1 |
+	                   (DBG_APB_FZ1_DBG_IWDG_STOP | DBG_APB_FZ1_DBG_WWDG_STOP));
+
+	target_mem_map_free(t);
+
+	size_t flash_size = (size_t)target_mem_read16(t, FLASH_MEMORY_SIZE);
```
UweBonnes, line 262: I think it is best to announce the biggest flash size for that chip. Smaller flash sizes only means that that flash up to the real size is not tested or perhaps disabled. By using the largest size, we give the user control what to do with flash above the announced size. Do you have G0 devices with criplled flash? Can you test?

UweBonnes:
> Ouch 2.6k is enormous. I suspected but not that much.
That is in line with the other file:
   2772       0       0    2772     ad4 stm32f1.o
   4373       0       0    4373    1115 stm32f4.o
   2856       0      16    2872     b38 stm32g0.o
   2683       0       0    2683     a7b stm32h7.o
   4007       0       0    4007     fa7 stm32l0.o
   2428       0       0    2428     97c stm32l4.o
   2560       0       0    2560     a00 stm32l4.o (in master)

There are two aspects:
- Keeps things simple and straightforward. In that way, I like to apply your PR, perhaps squashed and with some usage example. Look at the comments too. I also would not mind a more liberal license like BSD, but that is your choice.
- When we hit the 120 k (ST-Bootloader firmware) limits, it will get complicated to decide what to include in F103 firmware build and what not

Reuse is fine, when good tested so few chances for colateral damage

My advise:
- Lets get this PR applied with few changes
- I appreciate if you work out your other proposal sometimes later

src/target/stm32g0.c
fabalthazar: Are you referring to "target_storage" field in the "target_s" structure?
I attempted to store a pointer to an allocated structure in target_storage thanks to the uintptr_t standard type but after tries and tries, it fails on 64 bits hosts as uintptr_t uses 8 bytes and target_storage is 4 bytes (segfault).
Other idea than static variables?

src/target/stm32g0.c
fabalthazar: Yes. I am waiting for knowing how to properly store "saved_regs".

src/target/stm32g0.c
fabalthazar: I got an STM32G031 (32 kiB) and I can confirm that I succeeded in writing above 0x8000!
I implemented the full 64 kiB announcement for single bank devices.

fabalthazar:
Work still in progress.

I just put an hand on a fresh STM32G0B1 (512 kiB) dual-bank device! They are out.
I fixed a little thing in page erasure.
I confirm pages are organized a little bit differently than L4.

I removed the PCROP-related code that did not make sense until BMP does not support automatic re-scanning and attaching. I kept it simple(r).

I squashed the mass erase functions into a unique "erase_mass [1|2]" one to deal with the whole flash memory or specified bank number.

You might notice a little flash usage reduction.

I tested all this quite heavily.

Still to do:
- target_storage following your guidelines and help;
- Squash commits;
- Licence (I don't mind a more liberal one, I will have a look at this point);
- I noted your remarks concerning documentation too. I intend to populated the Wiki if its OK. I find it a bit skinny concerning target-specific usage. Perhaps I will clone the Wiki to work on it with a clear head.

fabalthazar:
Almost finished.

May you review `static struct stm32g0_priv_s stm32g0_private_storage` I am using to store target specific data?
As I am storing several data (uint32s + bool), I have not found a way to use the "target_storage" field in "target_s" struct since it is uint32_t.

Any idea other than current "static"?

UweBonnes:
What about extending "target_storage"? Instead of "uint32_t target_storage" "uint32_t target_storage[4]"

fabalthazar:
OK. Since it is common to all drivers, uint32_t *target_storage sounds more flexible. But needs memory allocation... source of leakage.
Or statically reserved table as you suggest but the table size is only allowed to increase in the future.
Feasible with few alignments is stm32h7 and l4.
And the drivers are responsible of having the knowledge of indices in the table.

UweBonnes:
"source of leakage" It  is allocated with the target structure and freed with the target structure. That happens already now when *target_storage" is static allocated. "drivers are responsible of having the knowledge of indices in the table" that happens with all those "priv" structures already.

But don't bang your head to much about these problems. There is a concept of "good enough"  and I think your PR is already there (when sensible squashed). But if you like to think further, you are welcome.

UweBonnes:
But what has to be kept in mind that something like 
"static struct saved_regs_s saved_regs = {0};"
will make problems when in the SWJ chain there are several targets of the same kind.

fabalthazar:
Rebased and squashed.

Target data is now allocated in target_storage and properly freed.

Small single-bank chips are announced with their max flash capacity ("overflashing" tested on STM32G031) ;
Dual bank devices get erased and flashed smoothly, including mass erase of bank 1 or 2 (tested on STM32G0B1) ;
Option bytes and full debugging experience tested on STM32G031, G070, G0B1.

Memory footprint: 2kB.

I am feeling nice with this work. Please have a look and criticize.

UweBonnes:
Thanks!

Issue #827, opened 2021-01-26
RP2040 Support


dexterfoxy:
Any idea? Is it even possible?

UweBonnes:
Debugging should work out of the box.  Detecting needs somebody to run hosted "blackmagic -t" and post the output. Flashing needs implementation of the QSPI algo, probably doable but needs real hardware and some time.

wasv:
Came here to ask exactly this question. I have some pico's and an stlink clone with black magic probe firmware. I'll poke around at debugging and detecting later this week. @UweBonnes do you know what a good starting point in the code base would be for flashing the QSPI algo?

UweBonnes:
The Rp2040 has no internal flash. Flash for user code is connected via (Q)SPI. So if BMP should be able to program this flash this must be implemented. At the moment, no BMP device supports externl flash. So there is no good starting point yet.

wasv:
Looks like the RP2040 requires multidrop SWD support. It doesn't seem to be widely supported, the Raspberry Pi devs had to patch it into openocd, and JLinks only support it after hardware revision 10. I can't find any reference to multidrop in the blackmagic repo, so I assume this project doesn't support that either? For reference, I have found that there may be a way to debug a RP2040 using openocd and an FTDI chip (see https://github.com/raspberrypi/openocd/issues/11).

UweBonnes:
I had a short look at the multidrop doc, but to implement it, real hardware is needed. If there is already support using FTDI, porting is easier, as you can switch on the command line between the already supported program and BMP hosted.

UweBonnes:
The Pico board has no dedicated debug connector and no reset buttom. Not very friendly for bring up.

wasv:
There is a debug connector at the bottom, but I was also unable to find a reset button.
![Screenshot from 2021-01-27 12-30-06](https://user-images.githubusercontent.com/1253385/106029923-9c18cb00-609b-11eb-830d-a2a623dfc024.png)

UweBonnes:
It seems, Pin30 "Run" is also "Nrst", To quote: "RUN is the RP2040 enable pin, and has an internal (on-chip) pull-up resistor to 3.3V of about ~50K Ohms. To reset RP2040, short this pin low."

UweBonnes:
Have a look at the https://github.com/UweBonnes/blackmagic/tree/rp branch. No flashing yet, but debugging and memory read.

UweBonnes:
[...]

PR #843, opened 2021-03-14
Implement SWD Multidrop and detect RP2040. Sometimes flashing works (Argh!)

UweBonnes:
Implements some aspects of #827

Contribution of a flash loader is welcome!


UweBonnes:
At least, hosted now can write flash in many cases. However it often fails in the first call to the Pico ROM _connect_internal_flash() with a hardfault. 

Any help is appreciated.

PR #846, opened 2021-03-26
Cortexm halt

UweBonnes:
Revert last patch mangling up all STM32F0.  Fix #836 by always halting CPU before romtable scan. Release from halt only once after scan


mmoskal:
works like a charm, thank you!

PR #833, opened 2021-02-16
[WIP] RTT support

katyo:
This is work in progress PR which adds support for RTT.

__NOTE:__ This PR is opened for testing and reviewing.
The implementation and/or interface still may change later.

The Segger's Real-Time Transfer is a technology for asynchronous
communication between host debugger and target platform. Because the
I/O operates with ring buffers in device memory by polling so it does
not requires additional I/O lines nor system calls nor break/watch
points.
See also https://www.segger.com/products/debug-probes/j-link/technology/about-real-time-transfer/.

To test you should build BMP with `ENABLE_SRTT`:
```
make ENABLE_SRTT=1
```

Next you should run the firmware which is built with RTT support.
(Simple examples from https://github.com/wez/jlink_rtt works fine).

Session example:
```
(gdb) monitor help
General commands:
	version -- Display firmware version info
	help -- Display help for monitor commands
	jtag_scan -- Scan JTAG chain for devices
	swdp_scan -- Scan SW-DP for devices
	frequency -- set minimum high and low times
	srtt_scan -- Scan Segger Real-Time Transfer
	targets -- Display list of available targets
	morse -- Display morse error message
	halt_timeout -- Timeout (ms) to wait until Cortex-M is halted: (Default 2000)
	connect_srst -- Configure connect under SRST: (enable|disable)
	hard_srst -- Force a pulse on the hard SRST line - disconnects target
	tpwr -- Supplies power to the target: (enable|disable)
	traceswo -- Start trace capture, Manchester mode: (decode channel ...)
	heapinfo -- Set semihosting heapinfo
ARM Cortex-M specific commands:
	vector_catch -- Catch exception vectors
STM32 LD/MD/VL-LD/VL-MD specific commands:
	erase_mass -- Erase entire flash memory
	option -- Manipulate option bytes
RTT commands:
	srtt_buffers -- Display list of available RTT buffers
	srtt_attach -- Attach to UP buffer to receive outgoing device messages
	srtt_detach -- Detach from UP buffer
	srtt_recv -- Receive text from UP buffer
	srtt_send -- Send text to DOWN buffer
	srtt_poll -- Poll attached buffers

(gdb) monitor srtt_scan <-- seeking for RTT control block in device RAM
Found RTT control block at 0x20001b70
RTT up/down buffers: 1/1
	UP buffer #1 "Terminal" at 0x20001bb9 of 1024 bytes
	DOWN buffer #1 "Terminal" at 0x20001fb9 of 16 bytes
[...]

PR #848, opened 2021-04-02
[WIP] Applet API proposal

zwizwa:
WIP: This is not intended to be merged as is, just as a starting point for discussion.

The purpose of this patch is to add a minimal change to the Black Magic Probe firmware that makes it easier to build out-of-tree firmware extensions, e.g. to include functionality in the debug probe that is very specific to a particular target board and firmware.

As a working title, these these extensions are called 'applets'.

See documentation included in applet.h comments, and the files in example_applet/


UweBonnes:
Whats the difference to #833? How to use? How to build.

zwizwa:
> Whats the difference to #833? How to use? How to build.

Thanks, I wasn't aware of #833. I'll review that first and work on top of it if possible.
If I continue on this path here, I'll submit a test so it would be easier to evaluate.

zwizwa:
Continuing with this since #833 might not be merged as-is due to rights issues.

Reworked to present a minimal change, with an example applet to demonstrate the basic idea.  The logging applet is in active use.

The thing I am not happy with is the location of app_poll(). Any suggestions?

zwizwa:
The patch excluding the example is fairly minimal now. I hope it makes the idea clear. If not I'm happy to clarify more, or reshape it if there are concerns about how this is hooked in.

TODO:
- move example out of tree
- create minimal blue pill board firmware for testing the example logger applet

dragonmux:
Seeing as BMD's architecture has drastically changed since this was proposed, and it hooks into some of the machinery (such as halt polling) improperly (meaning the concept would need re-imagining for it to become viable to even begin to entertain), we are going to close this PR as extensibility is achieved in other ways in-tree already and we want to encourage people to contribute to the tree rather than maintaining their own out-of-tree patch sets wherever possible.

It's a neat idea, but with Flash space at a premium, the reworked packet processing machinery from when this was imagined, and those other considerations in mind, this is not an avenue we wish to explore further at this time. If that changes, however, then we can always re-open this PR or otherwise draw inspiration from it with credit.

zwizwa:
Thanks for the comments. If the weather is right I might pick this up again on top of a more recent branch.

Issue #832, opened 2021-02-16
Add support for STM32WLE5JC SOC (Cortex M4 + Lora radio)


ofauchon:
Hi all, 

I'd like to use BMP for flashing/debugging new STM32WLE5 chips.  

https://www.st.com/en/microcontrollers-microprocessors/stm32wle5jc.html

I have Seeed Studio Lora E5 module on a DIY breakboard, which exposes SWDIO/SWCLK/RST pins: 

<img src="https://user-images.githubusercontent.com/561441/108035700-89a90600-7037-11eb-9421-9b152eb49c64.png" width=50%></img>

With BMP hosted ( blackmagic -t -v31), I could get traces with Read Timeout errors:  

```
CPUID 0x410fc241 (M4 var 0 rev 1)
!HM000023000040e000ed8800000004#
Timeout on read RESP
remote_ap_mem_read error -4 around 0xe000ed88
ap_memread @ e000ed88 len 4: 05 56 00 00
ap_mem_write_sized @ e000ed88 len 4, align 4: 05 56 f0 00
!Hm00002300004002e000ed88000000040556f000#
Timeout on read RESP
remote_ap_mem_write_sized error -4 around address 0xe000ed8c
!HM000023000040e000ed8800000004#
Timeout on read RESP
remote_ap_mem_read error -4 around 0xe000ed88
ap_memread @ e000ed88 len 4: 05 56 00 00
!HM000023000040e000ed7c00000004#
Timeout on read RESP
remote_ap_mem_read error -4 around 0xe000ed7c
ap_memread @ e000ed7c len 4: 05 56 00 00
swdptap_seq_out         8 ticks: 0000008d
```

As CPACR (FPU access register) is located at address 0xE000ED88, I tried to remove the FP probe code in cortexm.c
(Search for /* Probe for FP extension */ in the source)

It seems I can go further with this hack, but I still have errors: 

```
Read  AP 0x01 CSW   : 0x0100001a
!Ha000001f4#
Timeout on read RESP
Read  AP 0x01 CFG   : 0x0100001a
AP   0: IDR=00370011 CFG=0100001a BASE=0100001a CSW=01000008
AP#0 IDR = 0x00370011 (AHB-AP var1 rev0)
!HM00000100000801000ff000000004#
Timeout on read RESP
remote_ap_mem_read error -4 around 0x01000ff0
ap_memread @ 1000ff0 len 4: 00 00 00 00
!HM00000100000801000ff400000004#
[...]

PR #856, opened 2021-04-18
Add support for STM32WLxx

fabiobaltieri:
This adds support for the STM32WL series in stm32l4.c. These parts have the same flash registers layout as the L4 series, but a different base. Since there are already two sets of registers in this target file, this adds support for register maps that can be customized for each device ID.

Tested it on a Seed LoRa-E5, that's based on the STM32WLE5, the single core version of the WL55. Unfortunately it looks like ST left the second core AP configured on the single core part, and the probe hangs when probing it. I currently hacked my way around it by making it only probe the first AP and the flash and debugging capabilities seems to work fine, I'll have to figure how work around that cleanly, but this should work fine on the WL55.

### DIFF
--- a/src/target/stm32l4.c
+++ b/src/target/stm32l4.c
@@ -56,30 +56,11 @@ static int stm32l4_flash_write(struct target_flash *f,
                                target_addr dest, const void *src, size_t len);
 
 /* Flash Program ad Erase Controller Register Map */
-#define FPEC_BASE			0x40022000
-#define L4_FLASH_ACR			(FPEC_BASE+0x00)
-#define L4_FLASH_KEYR			(FPEC_BASE+0x08)
-#define L4_FLASH_OPTKEYR		(FPEC_BASE+0x0c)
-#define L4_FLASH_SR			(FPEC_BASE+0x10)
-#define L4_FLASH_CR			(FPEC_BASE+0x14)
-#define L4_FLASH_OPTR			(FPEC_BASE+0x20)
-//#define FLASH_OPTCR		(FPEC_BASE+0x14)
-
-#define L5_FPEC_BASE		0x40022000
-#define L5_FLASH_ACR		(L5_FPEC_BASE+0x00)
-#define L5_FLASH_KEYR		(L5_FPEC_BASE+0x08)
-#define L5_FLASH_OPTKEYR	(L5_FPEC_BASE+0x10)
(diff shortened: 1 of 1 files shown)


UweBonnes:
Fabio, the problem with the second AP should be solved. We discussed before and I asked you for a test, but I got no response. Otherwise , how does your PR compare to #458? Also please consider the discord channel for discussion.

fabiobaltieri:
Hi Uwe, don't you discussed it with me, this is my first patch to the project... anyway I'm running whatever is upstream now and it still hangs:

> AP   0: IDR=24770011 CFG=00000000 BASE=e00ff003 CSW=23000040 (AHB-AP var1 rev2)
> AP   1: IDR=84770001 CFG=00000000 BASE=f0000003 CSW=43800040 (AHB-AP var0 rev8)
> Timeout during scan. Is target stuck in WFI?
> SW-DP scan failed!

What do you want me to test?

As for #458, I'd say it compares pretty badly, different chip, different offset support strategy... you tell me, I'd say this one is cleaner.

UweBonnes:
Sorry, I mixed up names. There is discussion around that also in #832

UweBonnes:
Did you check that your construct does not need static ram? Does providing  those static register structures result in less flash usage than calculating the addresses on the fly? Otherwise the additional operation  will not hamper programming speed substantial and should be taken for the sake of reduced flash usage.

fabiobaltieri:
That is a very good point, they were indeed using more SRAM, I've set them as `const` and updated the PR so they don't take RAM anymore.

The flash and ram usage for native are the following, on current master:
```
Memory region         Used Size  Region Size  %age Used
             rom:      126400 B       128 KB     96.44%
             ram:        3560 B        20 KB     17.38%
```

after this patch:
```
Memory region         Used Size  Region Size  %age Used
             rom:      126188 B       128 KB     96.27%
             ram:        3560 B        20 KB     17.38%
```

It does look like flash usage goes down, stm32l4.o data goes from 3296 to 3084 bytes on my build.

As for programming speed, I don't think it's going to be affected since these are only going to be used for setting up the flash controller if I understand correctly, but I don't quite have a convenient way to compare. This is what I'm observing right now:

```
Start address 0x08002df0, load size 84422
Transfer rate: 34 KB/sec, 827 bytes/write.
```

Does it look reasonable?

PR #864, opened 2021-04-21
modify target stm32f4.c to enable register R/W from debugger for the …

alex31:
…stm32xx families

With this patch, one can observe peripheral registers which is very useful when debugging drivers.


UweBonnes:
Do you have "set mem inaccessible-by-default off" in your .gdbinit? Than the register area get visible.  Read the Wiki on that subject.

The classification for flash and ram is for binary upload of programs.

alex31:
Sorry, 
you are right, "set mem inaccessible-by-default off" make it works.

PR #871, opened 2021-04-26
CMSIS-DAP improvements

zyp:
These patches adds support for CMSIS-DAP discovery based on interface strings in addition to the manufacturer/product strings, as well as support for the v2 transport (plain bulk instead of HID).


UweBonnes:
Initialising res let the CI run. However the CMSIS debugger on IMXRT chokes with the last commit.
https://pastebin.com/USagXmrz

UweBonnes:
BMP hosted v1.7.1-192-g494524dd
Using 0d28:0204 022700001613a76e00000000000000000000000097969905 ARM
 DAPLink CMSIS-DAP
cmd :   00.03.
OUT error
IN error
cmd 03 invalid response received 00

PR #896, opened 2021-06-07
Add nRF51 erase UICR command and fix some warnings

bmellstrom-radinn:
Hi!

I'm trying to get another open source project (https://github.com/vedderb/bldc) that integrates black magic probe to more closely follow the latest version. To minimize the changes it would be nice to get a few fixes from there included in upstream. It's mostly compiler warnings due to a different set of compiler flags being used, but there is also one new command for the nRF5x.

Let me know if I should change something!

### DIFF
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -436,27 +436,27 @@ handle_v_packet(char *packet, int plen)
 	} else if (!strncmp(packet, "vRun", 4)) {
 		/* Parse command line for get_cmdline semihosting call */
 		char cmdline[83];
-		char *pbuf = cmdline;
+		char *pcmdline = cmdline;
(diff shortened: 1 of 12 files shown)


UweBonnes:
Will you care about the other -Wshadow problems?

bmellstrom-radinn:
I can do that, but it's not as straight forward to what the solution is in those cases. It's mostly local variable names conflicting with global ones, so I think some sort of naming convention to distinguish them is needed. For example, I made an experimental commit where I added a "g_" prefix to all the problematic global variables:
https://github.com/RadinnAB/blackmagic/commit/9b3d813b1c6f61db19d7365cf7279b016815cf50
If you think this is an acceptable solution I could chop it up into a few smaller commits.

PR #905, opened 2021-07-30
stm32f4x1 miniF4 aka blackpill v2 platform

aam335:
STM32F4x1CEU6 bluepill like platfom (based on f4discovery)


UweBonnes:
I am not to happy to add new platforms that do not show new capabilities. They need to be compiled for every test and if basic changes are needed in all platforms, I do not have a reference part. Furthermore, the F4 lacks USB endpoints. A variant of the F4 disco, e.g. "make PROBE_HOST=f4discovery BLACKPILL=1" leaving the f4discovery intact would however be accaptable.

PR #907, opened 2021-08-02
stm32f4x1 blackpillv2

aam335:
@UweBonnes 
> I am not to happy to add new platforms that do not show new capabilities. They need to be compiled for every test and if basic changes are needed in all platforms, I do not have a reference part. Furthermore, the F4 lacks USB endpoints. A variant of the F4 disco, e.g. "make PROBE_HOST=f4discovery BLACKPILL=1" leaving the f4discovery intact would however be accaptable.

+

### DIFF
--- a/src/platforms/f4discovery/Makefile.inc
+++ b/src/platforms/f4discovery/Makefile.inc
@@ -8,8 +8,15 @@ CFLAGS += -Istm32/include -mcpu=cortex-m4 -mthumb \
 	-DSTM32F4 -I../libopencm3/include \
 	-Iplatforms/stm32
 
+ifeq ($(BLACKPILL), 1)
+LINKER_SCRIPT=platforms/stm32/blackpillv2.ld
+CFLAGS += -DBLACKPILL=1
+else
+LINKER_SCRIPT=platforms/stm32/f4discovery.ld
+endif
+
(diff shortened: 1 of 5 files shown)


UweBonnes:
Does 
+#ifdef BLACKPILL
+       // https://github.com/libopencm3/libopencm3/pull/1256#issuecomment-779424001
+       OTG_FS_GCCFG |= OTG_GCCFG_NOVBUSSENS | OTG_GCCFG_PWRDWN;
+       OTG_FS_GCCFG &= ~(OTG_GCCFG_VBUSBSEN | OTG_GCCFG_VBUSASEN);
+#endif
really apply only to the blackpill? Otherwise I can can squash both commit and fix the things "git log --check" shows. Unrelated whitespace like
-               SYSCFG_MEMRM |=  1;
+               SYSCFG_MEMRM |= 1;
obscure your changes.

aam335:
> really apply only to the blackpill? Otherwise I can can squash both commit and fix the things "git log --check" shows. 
I checked on F401CE and F411CE, without this, the usb definitely does not work. Perhaps it will be useful for other similar architectures (F446?).

aam335:
>I can can squash both commit and fix the things "git log --check" shows

it would be great!

PR #908, opened 2021-08-06
Reset error LED after successful attach to target

compuphase:
When the BMP looses connection to the target MCU (for example, after accidentally redefining the SWCLK or SWDIO pin to GPIO or another function), the red LED of the BMP starts blinking in a morse code pattern ("TARGET LOST").

This pull request resets the LED after successfully re-attaching to the target (or attaching to some other target).


UweBonnes:
Please, rebase on master.

compuphase:
Sorry, I messed up the rebase attempt. I will make a new pull request.

PR #916, opened 2021-08-31
Escape '*' in responses of the Remote Serial Protocol

### LINKED ISSUE #915: In response packets, BMP should escape '*' character
Nitpicking, but a trivial change.

In its discussion of the "remote serial protocol", the GDB manual describes run-length encoding of packets. See https://sourceware.org/gdb/onlinedocs/gdb/Overview.html#Overview

RLE is optional, and BMP does not support it. However, the same page also has the following snippet:
"[...] The bytes 0x23 (ASCII ‘#’), 0x24 (ASCII ‘$’), and 0x7d (ASCII ‘}’) must always be escaped. Responses sent by the stub must also escape 0x2a (ASCII ‘*’), so that it is not interpreted as the start of a run-length encoded sequence (described next)."

In gdb_packet.c, '#', '$' and '}' are indeed escaped, but '*' is not.

compuphase:
Escape '*' in responses of the Remote Serial Protocol, to avoid that it is interpreted as the start of an RLE sequence.
Fixes #915.

### DIFF
--- a/src/gdb_packet.c
+++ b/src/gdb_packet.c
@@ -158,7 +158,7 @@ void gdb_putpacket(const char *packet, int size)
 			else
 				DEBUG_GDB_WIRE("\\x%02X", c);
 #endif
-			if((c == '$') || (c == '#') || (c == '}')) {
+			if((c == '$') || (c == '#') || (c == '}') || (c == '*')) {
 				gdb_if_putchar('}', 0);
 				gdb_if_putchar(c ^ 0x20, 0);
 				csum += '}' + (c ^ 0x20);


UweBonnes:
Did you hit a problem, eventually with some code you added or is it "only" a formal fix so long?

compuphase:
I did not encounter any problem.

I was reading on the remote serial protocol, because of a utility that uses the BMP directly (no GDB). Then, I checked the source code for the BMP to see whether it uses RLE (run-length encoding). If so, I would have to implement it as well in the utility. The BMP does not use RLE, and at that point I noticed that it did not escape the '*'. The GDB manual says (on the page that I quoted in #915) that it should. And it's a trivial thing to add.

That's all there is to it. I realize that one reason that this omission was not caught earlier, may be that the BMP never transmits a '*'.

PR #930, opened 2021-09-21
Avoid power conflict, test VREF before enabling target power

compuphase:
Refuse the "monitor tpwr enable" command (with an error message) if a voltage is sensed on VREF. This decreases the chance for a power conflict (when both the BMP and the target power the VREF pin, possibly with a different voltage).

For example, if 2.5V is sensed on the VCC/VREF pins and "monitor tpwr enable" is issued, this patch keeps power disabled and replies "Power conflict, power not enabled."

Caveat #1: I put the threshold for detecting external power on 0.5V. So if VREF is powered with 0.4V, "monintor tpwr enable" will proceed. The value of 0.5V is a bit arbitrary; I want to be sure it is above the noise level.

Caveat #2: This patch does not protect against "monitor tpwr enable" been given first and external power (on VREF) being applied afterwards. This case will (hopefully) be handled in a subsequent PR.

Other note: When you use hosted "blackmagic -t -p" and external power is detected, the reply is:
```
platform_target_set_power failed, error 0
Powering up device unimplemented or failed
```
Maybe we should allocate a specific error number for this case.

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -393,8 +393,15 @@ static bool cmd_target_power(target *t, int argc, const char **argv)
 	} else if (argc == 2) {
 		bool want_enable = false;
 		if (parse_enable_or_disable(argv[1], &want_enable)) {
-			platform_target_set_power(want_enable);
-			gdb_outf("%s target power\n", want_enable ? "Enabling" : "Disabling");
+			if (want_enable
+				&& !platform_target_get_power()
+				&& platform_target_voltage_sense() > POWER_CONFLICT_THRESHOLD) {
+				/* want to enable target power, but VREF > 0.5V sensed -> cancel */
+				gdb_outf("Target already powered (%s)\n", platform_target_voltage());
+			} else {
+				platform_target_set_power(want_enable);
+				gdb_outf("%s target power\n", want_enable ? "Enabling" : "Disabling");
+			}
 		}
 	} else {
 		gdb_outf("Unrecognized command format\n");
--- a/src/include/platform_support.h
+++ b/src/include/platform_support.h
@@ -37,7 +37,9 @@ void platform_timeout_set(platform_timeout *t, uint32_t ms);
 bool platform_timeout_is_expired(platform_timeout *t);
 void platform_delay(uint32_t ms);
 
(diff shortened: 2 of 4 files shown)


UweBonnes:
Can you try the approach I pushed. I do not have a native at hands, so I can not test myself. The approach is much more localized.

compuphase:
Your localized implementation works (just tested it), but...

1) It is implemented only for the gdbserver interface. When using the remote interface via hosted, the "-p" option will still enable target power, regardless of any voltage already on the pins.

2) The function I added to return the voltage as a number rather than a string, was also in preparation for the second step: if target power is enabled, periodically verify the VREF voltage and disable target power when it deviates too much. If the target power is enabled, voltage should be 3.3V; if it is measured as 3.7, there is a power conflict; if it is measured as 2.8, there is either a power conflict or the target draws too much current from the BMP. Comparing integers is easier that string parsing.

By the way, do you want localized patches because you are close to a release? This PR is not critical, it can wait until after the release.

compuphase:
Oops, I was a little too quick. There is a tiny issue in your code. You can no longer disable target power. The cause is the line
```
bool want_enable = parse_enable_or_disable(argv[1], &want_enable);
```
When the input in `argv[1]` is "disable", the function sets the second parameter to false, but returns true (because the syntax is correct). And you assign the return code to the variable again.

I also noticed that when you say `mon tpwr enable` and it was already enabled, the old code said "Enabling target power" and the new code does not say anything at all. You may have intended that, but I thought I'd mention it anyway.

UweBonnes:
The error in my code are purely unintentional. The patch was in the state "should work" and not "works for me", partly due to missing native at hand.

The power switch  together with vref measurement is only available on native. 
I was misslead by your `changes to f4disco and remote.c that all platforms would need code change.
Your code change however in fact it is constrained to native only. That's good.

Maybe  consider Millivolts  as result of your new API

I removed my "localized" patch and replaced it with a style patch. Before allpying finally,  I will squash.
I fixed these style issues:
- trailing/leading white space by git rebase --whitespace=fix
- Opening brace after "if (< on the same line as the if. Only needed when any condition has multiple code line.

compuphase:
Thanks for the style corrections. I pushed one final tiny change. I liked your warning message ("target already powered") more informative than mine ("power conflict"), so I changed the message.

PR #934, opened 2021-10-05
hosted/dap: Copy command buffer in wait/retry loop

zyp:
The same buffer is used for both command and response, causing part of the command to be overwritten before it's retried. Solve this by saving and restoring a copy of the command before retrying.

Also included a commit marking some inline functions as static to fix compile errors due to symbol conflicts when the header is included in multiple translation units.


UweBonnes:
Don't the changes in "hosted/dap: Mark inline functions as static" also apply to the other headers in hosted?

zyp:
I didn't look for others, fixing these were enough to fix the compile errors I were having.

UweBonnes:
Probably there are some more easter eggs in the cmsis-dap code...

PR #941, opened 2021-10-26
stm32/gdb_if: Use _WFI when nothing to do.

UweBonnes:


fabalthazar:
Test done on a swlink platform (Bluepill).
I see a little downside. As soon as a first GDB command is sent, the following disconnection from the probe makes the IDLE led blink forever.

Steps to reproduce:
```shell
$ blackmagic -t
```
Or:
```shell
$ arm-none-eabi-gdb firmware.elf
$ tar ext /dev/ttyACM0
$ monitor swdp_scan
$ quit
```

Looking at the commit, I would say it comes from gdb_packet.c:48 --> if (packet[0]==0x04) return 1;
The gdb_getpacket() function returns every systick interrupt then the SET_IDLE_STATE(0) and SET_IDLE_STATE(1) forever.
If my analysis is good.

src/platforms/stm32/gdb_if.c
```diff
@@ -118,8 +123,10 @@ unsigned char gdb_if_getchar_to(int timeout)
 
 	if (!(out_ptr < count_out)) do {
 		/* Detach if port closed */
-		if (!cdcacm_get_dtr())
-			return 0x04;
+			if (!cdcacm_get_dtr()) {
+				__WFI(); /* systick will wake up too!*/
+				return 0x04;
+			}
```
fabalthazar, line 129: One indentation level too many

fabalthazar:
After another look a this, I notice that without this PR (master branch), the IDLE LED is weak, no much light, which may mean a very fast blinking (duty cycle ~ 50%).
This PR maybe lowers this frequency due do WFI added delays and makes the frequency human visible!

compuphase:
Some measurements:

Original firmware (without this _WFI patch):
- GDB attached, target stopped (Activity LED full on): 42 mA
- Target running (Activity LED blinking): 39 mA
- After GDB detach (Activity LED blinking at 9.3 kHz and a 2% duty cycle): 37 mA

Firmware with _WFI
- GDB attached, target stopped (Activity LED full on): 23 mA
- Target running (Activity LED blinking): 22 mA
- After GDB detach (Activity LED blinking at 20 Hz and a 20% duty cycle): 20 mA

In conclusion: the _WFI patch reduces power consumption of the BMP by about 20 mA when it is basically idle.
The fast blinking of the Activity LED after detach, described by fabalthazar is confirmed. I also confirmed his hypothesis that this LED has always been blinking after detach --only much faster.

src/platforms/stm32/gdb_if.c
UweBonnes: The line if () do { causes the strange ident.

UweBonnes:
Platform dependency in gdb_if is removed. Strange idle led behavior on native is now more obvious, as blink frequency now is lower. See also #947.

esden:
I don't think the IDLE_TASK define is necessary. I would rather go for the simpler solution introduced in #1027. Thus I am closing this PR in favor of the #1027

PR #947, opened 2021-11-11
Do not turn IDLE LED off when port closed and target already detached

fabalthazar:
Avoids the IDLE LED to be dimmed after detaching the target. 
Should be #941 friendly as it avoids toggling the LED (human visible).

Additionally, I see no requirement in the GDB protocol to always answer "OK" on a D (detach) command when the target is already detached. I clarified a bit. 
This avoids sending "OK" forever in case of 0x04 command (port closed).

### DIFF
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -296,10 +296,11 @@ int gdb_main_loop(struct target_controller *tc, bool in_syscall)
 			if(cur_target) {
 				SET_RUN_STATE(1);
 				target_detach(cur_target);
+				last_target = cur_target;
+				cur_target = NULL;
 			}
-			last_target = cur_target;
-			cur_target = NULL;
-			gdb_putpacketz("OK");
+			if (pbuf[0] == 'D')
+				gdb_putpacketz("OK");
 			break;
 
 		case 'k':	/* Kill the target */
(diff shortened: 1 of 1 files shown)


koendv:
Using this.

fabalthazar:
Thanks for the feedback!

Your patch makes sense. Sending 'OK' when the link is still present and the target already detached is safe and harmless.
Maybe `if (pbuf[0] == 'D')` is more intuitive (reminding an actual detach command is received and waits for a reply)?

I agree I could split the commit: one for the LED, one for the detach.

koendv:
> Maybe `if (pbuf[0] == 'D')` is more intuitive (reminding an actual detach command is received and waits for a reply)?
OK.
> I agree I could split the commit: one for the LED, one for the detach.
No problem, I can read it the way it is.

fabalthazar:
Linked with #1027

dragonmux (approved):
This needs testing with #1027 but it otherwise looks good and seems ready to be merged.

esden (approved):
I have tested it to the best of my abilities. It does seem to work right as far as I can tell, and the IDLE LED is not flickering any more after detach.

esden:
Thank you very much for this contribution @fabalthazar :D Sorry it took so long for us to merge it.

PR #951, opened 2021-11-21
start_time undeclared if ENABLE_DEBUG=1

kdv-temp:
Fix 'start_time' undeclared when compiling with `make PROBE_HOST=stlink ENABLE_DEBUG=1`

target/adiv5.c:410:29: error: 'start_time' undeclared (first use in this function)


UweBonnes:
I must admit, for debug mesages I always use hosted. So much less work...

kdv-temp:
On Sun, 21 Nov 2021 03:58:20 -0800
UweBonnes ***@***.***> wrote:

> I must admit, for debug mesages I always use hosted. So much less
> work...
> 

Tested; you're right.

koen

PR #956, opened 2021-12-12
cdcacm: Lower usage of magic numbers.

UweBonnes:


compuphase:
This also makes it easier for platforms that have different endpoint definitions, see PR #735.

UweBonnes:
Test reports are welcome! This PR can do some damage, Perhaps my cursory test did  not show these problems.

compuphase:
I tested the GDB interface, UART interface and TRACESWO, and found no problems. All tests were done on the native BMP 2.1.

esden:
This looks good to me. I have tested it and compared with the original "magic" numbers. Thanks @UweBonnes

PR #401, opened 2018-10-19
support for Synwit SWM ARM chips

smdprutser:
With these changes one can program the Synwit ARM chips (TSOP8 version: http://www.synwit.cn/products_detail/productId=27.html and SSOP16 version: http://www.synwit.cn/products_detail/productId=34.html)

More detailed (in English) information about this on my blog: https://smdprutser.nl/blog/the-arm-chip-that-wont-cost-an-arm-and-a-leg/ Link to my GitHub with sourcecode for these chips: https://github.com/smdprutser/SWM050

PR #957, opened 2021-12-19
Added support for synwit ARM chips (SWm050)

esden:
This is a cleaned up and rebased pull request #401 

After evaluation, this does not seem to be complete? Is it possible to do anything more than erase and write the flash on the part? If you would like this added to BMP codebase the implementation has to be able to also debug the firmware by stepping through the code of the target. We aspire to be more than just a flash programmer. :)

As I don't have access to any of the synwit chips I can not test this. All I could do is clean up this pull request to best of my abilities.

If someone from the original contributors like @sdmprutser or @david-sawatzke can test this pull request, or provide me with a test target I can test this and merge it into master.

I am sorry this patch was sitting for so long but it would be a bummer to not get it merged if it is fully functional and useful. Otherwise I will be forced to close this pull request. :(

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -59,6 +59,7 @@ SRC =			\
 	stm32l4.c	\
 	stm32g0.c	\
 	target.c	\
+	synwit.c	\
 
 include $(PLATFORM_DIR)/Makefile.inc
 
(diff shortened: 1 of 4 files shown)


UweBonnes:
It seems there are few if any users of this chip are around. If there are requests from active users,  merging the PR should be considered. To lets keep the PR open to help those potential users find it.

esden:
I tried to find the SWM050 chips anywhere. But it seems that they have completely disappeared. They are not even available through aliexpress or alibaba. So I will not be able to test this, and that there does not seem to be anyone that currently has the part in active use. I am closing this pull request. Feel free to open again if the part comes back from the dead. :)

PR #962, opened 2021-12-28
crc32: define `start_time` when debug is enabled

xobs:
The variable `start_time` indicates when a CRC32 operation began. This
variable is used to benchmark the speed of the CRC32 function.

Currently, this is tied to `PC_HOSTED`. However, the actual usage is
tied to `DEBUG_WARN`. This means that the variable is undefined when
`DEBUG_WARN` is defined and we're not configured for `PC_HOSTED` mode.

Add macro guards around this variable so that it is defined when
debugging is enabled, rather than only when building on `PC_HOSTED`.

### DIFF
--- a/src/crc32.c
+++ b/src/crc32.c
@@ -106,9 +106,11 @@ int generic_crc32(target *t, uint32_t *crc_res, uint32_t base, size_t len)
 	 * 22 s @ 4k and 21 s @ 64k
 	 */
 	uint8_t bytes[0x1000];
-	uint32_t start_time = platform_time_ms();
 #else
 	uint8_t bytes[128];
+#endif
+#if defined(ENABLE_DEBUG)
+	uint32_t start_time = platform_time_ms();
 #endif
 	uint32_t last_time = platform_time_ms();
 	while (len) {


esden (approved):
LGTM

esden:
Just to make sure. If I understand this correctly. This issue appears when compiling for a different platform than STM32, and PC_HOSTED? It definitely makes sense to me to move the statement. I am just trying to understand why we did not catch it yet. :)

xobs:
Yes, I'm building for ESP32. The project I'm using as a starting point builds with `-DPC_HOSTED=0 -DNO_LIBOPENCM3=1` and then sets `#define ENABLE_DEBUG 1`.

Maybe it's just not common to build with `ENABLE_DEBUG=1` on embedded targets? I'm doing it on ESP32 since it has a webserver that buffers the debug log, so it's useful during bringup. 

I notice a similar construct with a different approach at https://github.com/blacksphere/blackmagic/blob/master/src/target/adiv5.c#L401, so someone seems to have run into this issue. But I think overall the reason why it hasn't been caught is that it's normally not useful to build without `PC_HOSTED=1`.

esden:
Thanks for the details. Enabling DEBUG on embedded is totally valid and I use it very often myself on STM32 targets (aka. native). We don't have CI set up for the different platforms yet so that is why this was not caught. This should be a problem of the past fairly soon. ;)

PR #967, opened 2022-01-04
Stm32 mem fixes

schodet:
Avoid crash at end of programing.

Also add minor fixes.

```
-
USB Device Firmware Upgrade - Host Utility -- version 1.2
Copyright (C) 2011  Black Sphere Technologies
-
Device ID:       1d50:6017
Manufacturer:    Black Sphere Technologies
Product:         Black Magic Probe (Upgrade)
Serial:          7BAE5BAD
-rogramming memory at 0x0801B400
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/usb/core.py", line 1021, in ctrl_transfer
    buff = util.create_buffer(data_or_wLength)
  File "/usr/lib/python3/dist-packages/usb/util.py", line 162, in create_buffer
    return array.array('B', _dummy_s * length)
TypeError: can't multiply sequence by non-int of type 'str'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/usb/_interop.py", line 92, in as_array
    return array.array('B', data)
TypeError: cannot use a str to initialize an array with typecode 'B'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/nicolas/work/stm32/blackmagic/./scripts/stm32_mem.py", line 261, in <module>
    stm32_manifest(dfudev)
  File "/home/nicolas/work/stm32/blackmagic/./scripts/stm32_mem.py", line 75, in stm32_manifest
    dev.download(0, "")
  File "/home/nicolas/work/stm32/blackmagic/scripts/dfu.py", line 95, in download
    self.handle.controlMsg(usb.ENDPOINT_OUT | usb.TYPE_CLASS |
  File "/usr/lib/python3/dist-packages/usb/legacy.py", line 205, in controlMsg
    return self.dev.ctrl_transfer(
  File "/usr/lib/python3/dist-packages/usb/core.py", line 1023, in ctrl_transfer
    buff = _interop.as_array(data_or_wLength)
  File "/usr/lib/python3/dist-packages/usb/_interop.py", line 97, in as_array
    a.fromstring(data) # deprecated since 3.2
AttributeError: 'array.array' object has no attribute 'fromstring'
```

### DIFF
--- a/scripts/stm32_mem.py
+++ b/scripts/stm32_mem.py
@@ -18,7 +18,6 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from __future__ import print_function
 from time import sleep
 import struct
 import os
--- a/scripts/stm32_mem.py
+++ b/scripts/stm32_mem.py
@@ -72,7 +72,7 @@ def stm32_read(dev):
 	return data
 
 def stm32_manifest(dev):
-	dev.download(0, "")
+	dev.download(0, b"")
 	while True:
 		try:
 			status = dev.get_status()
--- a/scripts/stm32_mem.py
+++ b/scripts/stm32_mem.py
@@ -138,7 +138,7 @@ def stm32_scan(args, test):
 		product = dfudev.handle.getString(dfudev.dev.iProduct, 96).decode('utf8')
 		serial_no = dfudev.handle.getString(dfudev.dev.iSerialNumber, 30).decode('utf8')
 		if args.serial_target:
-			if man == "Black Sphere Technologies" and serial_no ==	args.serial_target:
+			if man == "Black Sphere Technologies" and serial_no == args.serial_target:
 				break
 		else:
 			if man == "Black Sphere Technologies":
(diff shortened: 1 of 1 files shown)


esden:
Thank you very much for the contribution. Merged. :)

schodet:
Thanks :)

How about replacing python with python3 in the `#!` line? Can I do it?

UweBonnes:
Isn't python python3 implicit in` the meantime?

esden:
@schodet what is the advantage of explicitly requiring python3? Python2 is officially EOL since a while, so no systems should by default run python2 it was around for way too long already. :D

esden:
Also, we should discuss this in a separate issue. So that we don't get a topic mess. :)

DX-MON:
For some distros that are still in support, 'python' still means python2. It is correct to use python3 here no the scripts are for Python 3 and will be for some time

schodet:
I don’t know for other system, but on Linux, usually python is not python3.

esden:
Ok, fair enough. I am not here to police python version wars. ;) ... @schodet feel free to send in a PR to change `python` to `python3` in the `#!` :)

schodet:
(I takes a little bit more time than just changing the headers, I try to also test the scripts :))

PR #972, opened 2022-01-08
hosted/firmware/jtag: Fix unhandled exception introduced with da15cc3…

UweBonnes:
…cb75cc2a9f.

Write in endianess independant way.

### DIFF
--- a/src/platforms/hosted/remote_jtagtap.c
+++ b/src/platforms/hosted/remote_jtagtap.c
@@ -126,26 +126,21 @@ static void jtagtap_tdi_tdo_seq(
 			chunk = 64;
 		}
 		ticks -= chunk;
-		uint8_t di[8];
-		memset(di, 0, 8);
+		uint64_t di = 0;
 		int bytes = (chunk + 7) >> 3;
+		int i = 0;
 		if (DI) {
-			memcpy(&di, DI, bytes);
-			int remainder = chunk & 7;
-			DI += bytes;
-			DI += bytes;
-			if (remainder) {
-				uint8_t rem = *DI;
-				rem &= (1 << remainder) - 1;
-				*di = rem;
+			for (; i < bytes; i++) {
+				di |= *DI << (i * 8);
+				DI++;
 			}
-		};
+		}
 		/* PRIx64 differs with system. Use it explicit in the format string*/
 		s = snprintf((char *)construct, REMOTE_MAX_MSG_SIZE,
 					 "!J%c%02x%" PRIx64 "%c",
 					 (!ticks && final_tms) ?
 					 REMOTE_TDITDO_TMS : REMOTE_TDITDO_NOTMS,
-					 chunk, *(uint64_t*)di, REMOTE_EOM);
+					 chunk, di, REMOTE_EOM);
 		platform_buffer_write(construct,s);
 
 		s = platform_buffer_read(construct, REMOTE_MAX_MSG_SIZE);
@@ -156,8 +151,10 @@ static void jtagtap_tdi_tdo_seq(
 		}
 		if (DO) {
 			uint64_t res = remotehston(-1, (char *)&construct[1]);
-			memcpy(DO, &res, bytes);
-			DO += bytes;
+			for (i = bytes; i > 0; i--) {
+				*DO++ = res & 0xff;
+				res >>= 8;
+			}
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


src/platforms/hosted/remote_jtagtap.c
```diff
@@ -126,26 +126,19 @@ static void jtagtap_tdi_tdo_seq(
...
-			DI += bytes;
-			DI += bytes;
-			if (remainder) {
-				uint8_t rem = *DI;
-				rem &= (1 << remainder) - 1;
-				*di = rem;
-			}
+		int i = 0;
+		for (; i < bytes; i++) {
+			di = di | (*DI << (i * 8));
```
DX-MON, line 133: `|=` is a thing - as you aren't mucking with 'di' any other way, please use that instead or clarity and safety

src/platforms/hosted/remote_jtagtap.c
```diff
@@ -126,26 +126,19 @@ static void jtagtap_tdi_tdo_seq(
 			chunk = 64;
 		}
 		ticks -= chunk;
-		uint8_t di[8];
-		memset(di, 0, 8);
+		uint64_t di = 0;
 		int bytes = (chunk + 7) >> 3;
-		if (DI) {
```
DX-MON, line 132: Why the removal of the DI check? This looks like it introduces nullptr issues with its removal

src/platforms/hosted/remote_jtagtap.c
```diff
@@ -156,7 +149,7 @@ static void jtagtap_tdi_tdo_seq(
 		}
 		if (DO) {
 			uint64_t res = remotehston(-1, (char *)&construct[1]);
-			memcpy(DO, &res, bytes);
+			memcpy(DO, &res, bytes); /* Fixme: Endian dependant!*/
```
DX-MON, line 152: the memcpy is right - you can use https://github.com/bad-alloc-heavy-industries/substrate/blob/7fa29f823130c40ea0b10f393537b680c6d4f81e/impl/socket.cxx#L133 as a basis for how to fix this (yes, this is C++, but the fixed stack array and platform-independent population of it still applies here). The compiler is very good at reading through the code and generating bswap/xchng operations as required

DX-MON (changes requested):

src/platforms/hosted/remote_jtagtap.c
UweBonnes: Please explain why |=  has more clarity or  safety.

src/platforms/hosted/remote_jtagtap.c
UweBonnes: Colateral damage.

src/platforms/hosted/remote_jtagtap.c
UweBonnes: Committed not the latest version...

src/platforms/hosted/remote_jtagtap.c
DX-MON: you are or'ing a value to `di` and making it equal to that new value, that is literally what `|=` means and has the benefit that you can't typo the left hand side of the `|` or otherwise accidentally modify the value in some other way first.

src/platforms/hosted/remote_jtagtap.c
```diff
@@ -156,8 +151,10 @@ static void jtagtap_tdi_tdo_seq(
 		}
 		if (DO) {
 			uint64_t res = remotehston(-1, (char *)&construct[1]);
-			memcpy(DO, &res, bytes);
-			DO += bytes;
+			for (i = bytes; i > 0; i--) {
```
DX-MON, line 154: This results in sub-par code gen - the approach we linked was suggested specifically because you'll get the best possible code generation that way - we've already spent hours making sure that was so.

UweBonnes: Maybe this is not the optimal code, but it is clear and readable and
 1. We wasted  before a lot of cycles for the USB transaction
 2. This call can not fill up an USB request with more than 8 bytes and is tagged with "FIXME: Provide and test faster call and keep fallback for old firmware"

src/platforms/hosted/remote_jtagtap.c
UweBonnes: Changed as requested.

src/platforms/hosted/remote_jtagtap.c
UweBonnes: DI check put back in place.

esden:
As this causes `blackmagic -t -j` to crash. I am merging this commit. 

Although I agree with @DX-MON comments here. We can polish the coding style and make it more robust to future changes in a separate step.

Thanks @UweBonnes

Issue #249, opened 2017-06-19
When target stuck in WFI, synchronisation is lost between GDB and BMP


schodet:
Configuration : BMP 2.1c connected to USB, 1bitsy 1.0d connected using the debug connector, no USB on 1bitsy.

For some reason, BMP does not manage to interrupt the target, GDB tries to communicate with BMP without success.

Program is derived from 1bitsy-locm3-template, with the following diff:

```diff
diff --git a/src/main.c b/src/main.c
index e5bc2b5..302311c 100644
--- a/src/main.c
+++ b/src/main.c
@@ -42,6 +42,7 @@ static void gpio_setup(void)
 int main(void)
 {
        int i;
+       __asm__ __volatile__ ("wfi");

        clock_setup();
        gpio_setup();
```

Here is GDB session log:

```
GNU gdb (7.7.1+dfsg-1+6) 7.7.1
Copyright (C) 2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "--host=x86_64-linux-gnu --target=arm-none-eabi".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from src/1bitsy-example.elf...done.
(gdb) target extended-remote /dev/serial/by-id/usb-Black_Sphere_Technologies_Black_Magic_Probe_7BAE5BAD-if00
Remote debugging using /dev/serial/by-id/usb-Black_Sphere_Technologies_Black_Magic_Probe_7BAE5BAD-if00
(gdb) mon connect_srst enable
(gdb) mon tpwr enable
(gdb) mon swdp_scan
Target voltage: 3.3V
Available Targets:
No. Att Driver
 1      STM32F4xx
(gdb) attach 1
Attaching to program: /home/nicolas/work/1bitsy/1bitsy-locm3-template/src/1bitsy-example.elf, Remote target
0x08000544 in reset_handler () at ../../cm3/vector.c:71
71                      *dest = *src;
(gdb) load
[...]

PR #973, opened 2022-01-08
STM32F4: Move DBGMCU_handling to target specific code. Apply for F7 too.

UweBonnes:


UweBonnes:
#249 explicit states the need for this PR for correct behaviour of a STM32F4 using WFI

dragonmux (approved):
Having carefully checked the logic of this, while it looks like it'd be more correct to write `priv_storage->dbgmcu_cr | DBGMCU_CR_DBG_SLEEP | DBGMCU_CR_DBG_STANDBY | DBGMCU_CR_DBG_STOP` to DBGMCU_CR in the F4/F7 attach routine, we're happy this does what it's supposed to without extra consequences, so merging.

PR #974, opened 2022-01-13
Cortexm: With connect under reset, keep the device halted until attach.

UweBonnes:
#249 shows that the device was running after probe with attach with connect under reset was requested.

### DIFF
--- a/src/target/adiv5.c
+++ b/src/target/adiv5.c
@@ -834,9 +834,15 @@ void adiv5_dp_init(ADIv5_DP_t *dp)
 		adiv5_ap_unref(ap);
 	}
 	/* We halted at least CortexM for Romtable scan.
-	 * Release the devices now. Attach() will halt them again.*/
-	for (target *t = target_list; t; t = t->next)
-		target_halt_resume(t, false);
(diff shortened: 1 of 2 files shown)


DX-MON:
This PR was in need of review and testing to which we hadn't yet got - we don't yet understand the impact just deleting the target_halt_resume from the Cortex-M code will have on a variety of errata-having processors. Please don't merge your own PRs - we (Esden or us) will review and merge them as soon as we can.

DX-MON:
What we're pointing out is that in the Cortex-M code you deleted the target_halt_resume() call rather than making it conditional at that point too and the Git history shows that was added to fix several MCUs that aren't fully compliant - this needs properly evaluating with hardware for as many MCUs as possible, not just slap-dashed added and hoping for the best.

UweBonnes:
For a target recognized, halt_resume at end of cortexm.c was never reached, as the probe function already does the return. For an unrecognized target, it is probably not a good idea to resume while probing is still going on and resume was not done neither in that case before the halt/probe rework.  Maybe two PR would have been clearer.

PR #976, opened 2022-01-19
sscanf cleanup in GDB Z packet handler

fabalthazar:
Little cleanup of this sscanf attempt previously marked as not working.
Tested with arm-none-eabi-gcc 10.2.1 and z1/2/3/4 breakpoint and watchpoint types.


esden:
Thank you very much for the contribution. Merged! :D

PR #977, opened 2022-01-21
Cleanup for the kinetis GCC11 fixes

DX-MON:
While the original patch (#940) made compilation work, the original patch is by no means entirely correct by what the warnings in question were complaining about, as they were an attempt by GCC to warn about the anachronism that is array parameters.
(For the reader: array parameters don't really exist, the type of the parameter is lowered to a raw pointer before the function signature becomes finalised, so they're a programmer's wish at best and outright wrong at worst)

This patch cleans things up by removing the array syntax and turning it into a pointer and a count, while also making the type of that pointer align to all its usages, removing much of the cast munging going on in the rest of the code.

### DIFF
--- a/src/target/kinetis.c
+++ b/src/target/kinetis.c
@@ -373,7 +373,7 @@ do_common_s32k14x:
 }
 
 static bool
-kl_gen_command(target *t, uint8_t cmd, uint32_t addr, const uint8_t data[8])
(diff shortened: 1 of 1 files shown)


esden (commented):
LGTM

esden:
Merged with c1a12edbe9a801051a6bfdfde7264be623bd389f thank you! :)

PR #924, opened 2021-09-15
WIP: Riscv 0.13 Working on GD103VF103, debug working on ESP32-C3

UweBonnes:
Disclaimers:
- May ruin your debugger or target
- Forced push may happen

Prerequisite:
- For hosted/firmware use, compile and flash firmware  from this branch.
- Firmware should still be compatible with older hosted software

What should work:
- Core/Hart Detection and debugging
- Memory setup on GD32VF103 and flashing

What does not work (yet?)
- ESP32-C3 memory setup, help needed
- ESP32-C3 flashing: Probably needs major effort

Hint:
- Use hosted for verbose output

Speed :
- GD32FV103/firmware ~ 40 kByte/s read ~10 kByte/s write
- GD32FV103/remote ~ 9kByte/s read ~3 kByte/s write
- GD32FV103/builtin CMSIS-DAP on GD32VF103C-START: ~ 2kByte/s.
- ESP32-C3, FT2232H adapter: ~ 27 kByte/s

PR #978, opened 2022-01-22
Reversions, cleanup, and fixes to the scanchain systems

DX-MON:
This set of commits seeks to address issues in the current JTAG scanchain detection and setup code.

The first problem this addresses is the way in which the post-scanchain-detection handler init is done as the table of ID Codes to handlers is needed for a future PR to provide support for JTAG-PDI, and passing the entire jtag_devs slot structure to the handler is horridly wasteful for information that can be gleaned a different way without this layering violation.

The second problem is in how chain scanning is being performed - the old code works a very particular way and this is necessary for correctly detecting things like ATXMega devices.

The third is the conflation of SWD and JTAG scan that is occurring, which is only partially addressed by this PR as SWD being able to defer on failure to JTAG at all is a violation of the boundy set by having specific `jtag_scan` and `swd_scan` monitor commands - a future PR will be submitted that entirely removes this deferral and introduces an `auto_scan` command that allows the BlackMagic Probe Firmware to slew through a pile of scan chain possibilities and try and auto-detect what is present.


UweBonnes:
Please break up in several PRs. b502a43, 1b677c5, d693647 are mostly syntactical fixes. Removal of the "auto-scan" is user interface change . For reverting the JTAG detection scheme, I think an issue to discuss is worth the effort. I have to access the impact on the RiscV and AVR8 branch.  If jtag_scan again needs the irlens argument to scan through unknown chains, I think this would be a regression.

src/target/adiv5.h
```diff
@@ -89,23 +89,19 @@
...
- */
+/* Known designers seen in SYSROM-PIDR. Ignore Bit 0 from
+ * the designer bits to get JEDEC Ids with bit 7 ignored.*/
 #define AP_DESIGNER_FREESCALE    0x00e
 #define AP_DESIGNER_TEXAS        0x017
 #define AP_DESIGNER_ATMEL        0x01f
 #define AP_DESIGNER_STM          0x020
 #define AP_DESIGNER_CYPRESS      0x034
 #define AP_DESIGNER_INFINEON     0x041
-#define DESIGNER_XILINX          0x049
```
esden, line 102: Why are we loosing these defines? (It is missing the `AP_` prefix but it might be useful in the future. Unless it is wrong.

src/target/adiv5.h
```diff
@@ -89,23 +89,19 @@
...
 #define AP_DESIGNER_NORDIC       0x244
 #define AP_DESIGNER_ARM          0x43b
 /*LPC845 with designer 501. Strange!? */
 #define AP_DESIGNER_SPECULAR     0x501
 #define AP_DESIGNER_CS           0x555
-#define DESIGNER_XAMBALA         0x61e
```
esden, line 108: Same here as for the removed XILINX designer.

src/target/jtag_devs.c
```diff
@@ -25,35 +26,45 @@
...
+	{.idcode = 0x00000093, .idmask = 0x00000FFF,
+		.descr = "Xilinx."},
+	{.idcode = 0x0000063D, .idmask = 0x00000FFF,
+		.descr = "Xambala: RVDBG013."},
+	{.idcode = 0x000007A3, .idmask = 0x00000FFF,
+		.descr = "Gigadevice BSD."},
 /* Just for fun, unsupported */
-	{.idcode = 0x8940303F, .idmask = 0xFFFFFFFF, .descr = "ATMEL: ATMega16."},
-	{.idcode = 0x0792603F, .idmask = 0xFFFFFFFF, .descr = "ATMEL: AT91SAM9261."},
+	{.idcode = 0x8940303F, .idmask = 0xFFFFFFFF, .descr = "Atmel: ATMega16."},
```
esden, line 66: If we are removing the manufacturer description above for the ST parts then we probably should remove Atmel here? (and on the line below)

esden (commented):
Overall it looks good to me. I am happy to merge this after the other questions are answered or amended.

src/target/adiv5.h
DX-MON: Same reasoning as with the Xilinx code. We don't have any of their processors to test with though unlike Xilinx Zynq.

src/target/jtag_devs.c
DX-MON: Sure, we can do that

UweBonnes:
I still urge to break up this PR in smaller, unrelated ones.  This mixes up changes to the scan system with other things like syntactical changes and user interface changes.
The scan system targeted to revert is used in the present way in the riscv branch #924 and <https://github.com/UweBonnes/blackmagic/tree/avr> (able to flash JTAG-AVRs and some Xilinx parts). Does the present way  really not find the XMega? The present way tries more auto-detection on unknown chains, while the ability to provide scan length number is still given.

Of course  auto/swd/jtag scan is a good idea, but still worth it's own PR.

src/target/adiv5.h
esden: As discussed, this is being added back in through the upcoming scanchain-xilinx branch.

esden:
There is a bunch of patches in the pipeline that will improve things regarding device detection and supporting new platforms. They depend on this cleanup patch. I agree that we should strive to make pull requests transparent and focused on one task. But after talking to @DX-MON I think we will have to make an exception in this particular case. So I am merging this.

Thanks for the hard work @DX-MON  :)

Issue #981, opened 2022-01-25
LPC and MCULINK broken since b343ebe06f303ebe


UweBonnes:
Following hotfixes were done to get bmp/hosted working with Dragonprobe https://git.lain.faith/sys64738/DragonProbe.git

commit 166eb3ee3eb736ece3e874738e8eb564e3f08cbb
Author: Uwe Bonnes <bon@elektron.ikp.physik.tu-darmstadt.de>
Date:   Mon Oct 11 20:57:49 2021 +0200

    cmsis_dap: Always transfer 65 bytes with hid_read|write()
    
    CMSIS-DAP Dragonprobe works on some devices. E.g. it does not work
    on NUCLEO-F103

commit b343ebe06f303ebe5d46e4a05aaceafcf0a04b3d
Author: Uwe Bonnes <bon@elektron.ikp.physik.tu-darmstadt.de>
Date:   Thu Oct 14 11:45:36 2021 +0200

    cmsis-dap/linux: Link against libhidapi-hidraw
    
    libhidapi-libusb detaches a kernel module but does not reattach again.
    This may wrack interaction with other programs using hidapi.

LPC-Link and MCU Link transfer hangs. Other CMSIS probes, e.g. GD32VF103C-Start seem not affected. Reverting both patches gets things going again for the LPC- and MCU-Link

UweBonnes:
I can not reproduce.

dragonmux:
Does that mean the associated PR can be closed too, or?

PR #982, opened 2022-01-25
LPC-Link and MCU-Link support regression fixes

UweBonnes:
Revert hotfixes  to support (experimental) Dragonprobe #981.


DX-MON:
Rather than straight reverting this - would it not be more sensible to make a platform for DragonProbe specifically and make these changes depend on which platform is being built? We support DragonProbe right now, and we know there are users for it.. if we were them, we'd be pretty angry to have that support ripped back out without trying alternatives first.

UweBonnes:
I can not reproduce.

PR #984, opened 2022-02-13
Python3 fixes and shebang line update

schodet:
Make sure all scripts are using python version 3, fix some issues on the way.


esden:
Thank you very much! Down the road I think we should rather use the `b"\xfb"` type of notation for bytes rather than the int literals. And passing int tupels to `bytes()` But that is just a nitpicks. :) 
Merged.

schodet:
I can make the change if you want :)

PR #986, opened 2022-02-17
Update samd.c

arpadbuermen:
Added support for section-wise locking of bootloader and flash in lock_bootprot and lock_flash.


arpadbuermen:
Currently bootloader locking supports only locking the maximal amount of flash. 
Similarly, flash locking can lock only the complete flash memory. 
Optional argument to lock_bootprot and lock_flash allows for more fine-grained locking.

PR #987, opened 2022-02-17
Added support for fine-grained bootloader and flash locking in samd.c

arpadbuermen:
Up to now only the whole flash can be locked in one piece. Locking by segments is not supported. 
When locking the bootloader only the largest chunk (32kB) can be locked. 
By specifying an optional numerical parameter (can be given in decimal or 0x prefied hexadecimal) to lock_bootprot (0 is for 32k, 6 is for 512 bytes) and lock_flash monitor commands the extended functionality is obtained. If this parameter is not specified the commands behave as previously.

### DIFF
--- a/src/target/samd.c
+++ b/src/target/samd.c
@@ -715,11 +715,51 @@ static bool samd_set_flashlock(target *t, uint16_t value, const char **argv)
 	return true;
 }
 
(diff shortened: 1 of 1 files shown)


esden:
Thank you very much for the contribution. Merged. :)

PR #991, opened 2022-03-14
samd: Fixes for the "hosted" build

DX-MON:
The code from #987 assumed unsigned long was 32-bit, which is not true for x86_64 Linux/Mac and other potential Black Magic App targets. This generally fixes the soundness of the type punting going on.


esden:
I saw that issue. We should ask the hosted build to the github actions so we can catch these kind of problems.

Thanks for the patch! :D

PR #993, opened 2022-03-16
Native hw6

esden:
This patch set adds detection and support of native Hardware V4, 5 and 6.

This version is different than PCB revisions. This is the hardware version that used to be encoded using GPIO strapping. Now these versions are stored in user option bytes of the STM32. This future proves us and is making more GPIO available to add new features.

For the fun of it, here is a picture showing a selection of the different native Black Magic Probes throughout history.
black-magic-hardware-versions-w-legend

Before this is merged we need to test this patch on all the past hardware to hopefully avoid regression and issues:
* [x] V0 SWD tested
* [x] V0 JTAG tested
* [ ] V1 SWD tested
* [ ] V1 JTAG tested
* [x] V1 UART tested
* [x] V2 SWD tested
* [x] V2 JTAG tested
* [x] V2 UART tested
* [x] V3 SWD tested
* [x] V3 JTAG tested
* [x] V3 UART tested
* [x] V4 SWD tested
* [x] V4 JTAG tested
* [x] V4 UART tested
* [x] V5 SWD tested
* [x] V5 JTAG tested
* [x] V5 UART tested
* [x] V6 SWD tested
* [x] V6 JTAG tested
* [x] V6 UART tested

### DIFF
--- a/src/platforms/native/platform.c
+++ b/src/platforms/native/platform.c
@@ -343,7 +343,18 @@ void platform_request_boot(void)
 
 void exti15_10_isr(void)
 {
-	if (gpio_get(USB_VBUS_PORT, USB_VBUS_PIN)) {
+	uint32_t usb_vbus_port;
+	uint16_t usb_vbus_pin;
+
+	if (platform_hwversion() < 5) {
+		usb_vbus_port = USB_VBUS_PORT;
+		usb_vbus_pin = USB_VBUS_PIN;
+	} else {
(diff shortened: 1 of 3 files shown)


esden:
I have tested all the previous native hardware as far as I could.

The tests have uncovered that Hardware V1 is not working any more with the code in the main branch. Even without this patch, so it is a regression independent from this pull request.

Thus I am merging this PR. And I will open a separate issue to investigate the V1 hardware regression.

PR #995, opened 2022-03-16
Reduce stack usage

mean00:
Hi,
This a request for comments on the following changes to see if it's worth cleaning and submitting  (they are probably not following the coding standard for the moment , i.e. evil tab vs nice space etc...). 
As a result the indentation is all over the place for now, sorry for that.

The main aim is to reduce the stack usage which  is high at the moment 
The patch contains several sub changes :

- reuse the exception when they dont stack, gain of ~ 140 bytes on the stack
- split the commands beginning by q into a command engine to have several smaller functions for simplicity/clarity sake
(the same can be done for 'v' commands)
- Add a gdb_put_packet2(a,b) function that sends a packet made of 2 sub packets
- Use the above to avoid duplicating the data on the stack for the memory map /xml commands, i.e. the above is used to add the 'm' in front of the reply without having to duplicate.

I have another one i need to change before submitting where  the memory map reply is built using auto growing heap allocatio instead of the 1024 bytes allocated on the stack in the main code.

Thank you
Ps: I had to patch locally the conflicting USBUSART_DMA_TX_ISR locally when used for swlink as there seems to be a duplicate define there

### DIFF
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -56,7 +56,6 @@ static target *last_target;
 static void handle_q_packet(char *packet, int len);
 static void handle_v_packet(char *packet, int len);
 static void handle_z_packet(char *packet, int len);
-
 static void gdb_target_destroy_callback(struct target_controller *tc, target *t)
 {
 	(void)tc;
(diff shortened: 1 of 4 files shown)


DX-MON:
"i.e. evil tab vs nice space etc" - the code base actually uses tabs, not spaces, for indentation and compressed {} syntax.

The idea is certainly intriguing! One of either Esden or us will give this code a review in the next few days.

src/gdb_main.c
```diff
@@ -56,7 +56,6 @@ static target *last_target;
 static void handle_q_packet(char *packet, int len);
 static void handle_v_packet(char *packet, int len);
 static void handle_z_packet(char *packet, int len);
-
```
DX-MON, line 59: Please don't remove whitespace like this

src/gdb_main.c
```diff
@@ -330,102 +329,155 @@ int gdb_main_loop(struct target_controller *tc, bool in_syscall)
...
+    	gdb_putpacketz("E01");
+      return;
+  }
+  if(addr== strlen (str))
+  {
+    	gdb_putpacketz("l");
+      return;
+  }
+  unsigned long int outputLen=strlen(str)-addr;
+  if(outputLen>len) outputLen=len;
```
DX-MON, line 352: The basic concept looks good, the formatting, as you noted, needs re-doing here (whitespace around operators, condensed {} style as shown by the original code around this).

src/gdb_main.c
```diff
@@ -330,102 +329,155 @@ int gdb_main_loop(struct target_controller *tc, bool in_syscall)
...
+  gdb_putpacket2("m",1,str+addr,outputLen);
```
DX-MON, line 353: We'd need to check how gdb_putpacket() works, but is there a reason for not making this two gdb_putpacket() calls (with possibly a new parameter for whether to leave the packet unsent) so this could be:
```c
gdb_putpacket("m", 1, false);
gdb_putpacket(str + addr, outputLen);
```

Also, per the style for the code here, it'd be output_len.

src/gdb_main.c
```diff
@@ -330,102 +329,155 @@ int gdb_main_loop(struct target_controller *tc, bool in_syscall)
...
 
-static void
-handle_q_packet(char *packet, int len)
+typedef struct
+{
+    const char *cmdPrefix;
+    void  (*func)(char *packet,int len);
+}cmdExecuter;
+//---
+static void execqRcmd(char *packet,int len)
```
DX-MON, line 362: Ditto above about it being exec_q_Rcmd, exec_q_supported, exec_q_memory_map, etc.. looks good otherwise

src/gdb_packet.c
```diff
@@ -138,35 +138,61 @@ int gdb_getpacket(char *packet, int size)
 	return i;
 }
 
+static void gdb_next_char(char c, unsigned char *csum)
```
DX-MON, line 141: quite like this refactoring - it's a good call that might save some space if aggressively applied to the rest of gdb_packet.c

DX-MON (changes requested):
Generally looks good

mean00:
Created a new one, closing this one

esden:
@mean00 it is better to keep the thread in one PR. Unless I am missing something fundamental it would be better if you pushed the changes to address the issues to this branch/PR. So that we can keep things in one place.

mean00:
It seems I borked the original PR branch by moving it to the 2nd Pr branch  (note for self: dont do that)
Can i get a free pass for this time ? I'll be more careful in the future
Thank you

esden:
@mean00 Sure no worries. Let's continue on the other PR then. :)

PR #997, opened 2022-03-20
Reduce stack usage, Take 2

mean00:
Hi
I tried to incorporate the comments, except one :  i let the qRcmd suffix etc... because they are  the actual name of the command i.e.  "qRcmd" => exec_qRcmd(), "qCRC"' => exec_qCRC etc.... I find that much easier to deal with, just search for the command.

Assuming i didnt break anything (had to backport manually things from my working tree), there are 2 follow ups coming : 

1- Construct the memory map using the heap (+autorealloc) instead of 1k on the heap.  I have to backport it from C++ to C.
2- For arm,  the exception allocation is the default arm one i.e. 23 reg x 4 bytes. But cortex m0(+), m3, m4 etc.. are thumb(2) only, so only 10 registers. 120 bytes => 40 bytes, if/when exceptions stack, that makes a difference.

BTW, Is there a .clang-format available to format things properly before submitting ?
Thanks

### DIFF
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -48,6 +48,14 @@ enum gdb_signal {
 #define ERROR_IF_NO_TARGET()	\
 	if(!cur_target) { gdb_putpacketz("EFF"); break; }
 
+typedef struct
+{
+	const char *cmd_prefix;
+	void  (*func)(const char *packet,int len);
+}cmd_executer;
+
+
+
 static char pbuf[BUF_SIZE+1];
 
 static target *cur_target;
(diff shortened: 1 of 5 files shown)


esden:
Hi @mean00, Thank you very much for the contribution. :)

Can you push the changes that address the questions to the original feature branch? This will update the PR in question automatically. It is better than creating a new PR as we can keep the conversation in one place and it will make further review easier.

Cheers :D

DX-MON:
This generally reads a lot better on a quick review (formatting aside).

Re clang-format, this is something that needs working on and that is on the list but which has not yet been done.

We see what you're saying re the q-commands, however it is inconsistent with the rest of the code base so for consistency's sake, the request regarding those commands stands. We'd rather not set an unwanted precedent of allowing style mixing when we already have so much cleanup work to do.

If you get that fixed, we'll test it on real hardware as a double-check and get it merged ASAP.

mean00:
Hi
This should be okay now
Thank you

Ps: I  had to patch locally  src/platforms/swlink/platform.h to fix the conflicting USBUSART_DMA_TX_ISR when swlink is targeted

DX-MON (approved):
Looks good to us. We'll build and hardware test this shortly and do a post-PR cleanup commit for the formatting, but it'll be better to get this in sooner than later.

Issue #999, opened 2022-03-22
error: redefinition of 'dma1_channel7_isr'


mikeditto:
I got this error when building with `make PROBE_HOST=stlink`:

```
  CC      platforms/stm32/usbuart.c
In file included from include/general.h:39,
                 from platforms/stm32/usbuart.c:31:
platforms/stlink/platform.h:121:32: error: redefinition of 'dma1_channel7_isr'
  121 | #define USBUSART_DMA_TX_ISR(x) dma1_channel7_isr(x)
      |                                ^~~~~~~~~~~~~~~~~
platforms/stm32/usbuart.c:525:6: note: in expansion of macro 'USBUSART_DMA_TX_ISR'
  525 | void USBUSART_DMA_TX_ISR(void)
      |      ^~~~~~~~~~~~~~~~~~~
platforms/stlink/platform.h:121:32: note: previous definition of 'dma1_channel7_isr' was here
  121 | #define USBUSART_DMA_TX_ISR(x) dma1_channel7_isr(x)
      |                                ^~~~~~~~~~~~~~~~~
platforms/stm32/usbuart.c:504:6: note: in expansion of macro 'USBUSART_DMA_TX_ISR'
  504 | void USBUSART_DMA_TX_ISR(void)
      |      ^~~~~~~~~~~~~~~~~~~
make[1]: *** [Makefile:102: usbuart.o] Error 1
make: *** [Makefile:17: all] Error 2
```

mikeditto:
Fix in https://github.com/mikeditto/bmp/tree/stlink-isr-fix

I am learning how to make a pull request.

PR #1000, opened 2022-03-22
Remove duplicate definition of USBUSART_DMA_TX_ISR

mikeditto:
Fixing #999.

It looks safe for all platforms because it's literally a duplicate copy of another definition of the same function a few lines above.


DX-MON:
Thank you for your contribution! We will review the PR shortly.

DX-MON (approved):
We'll do a build and run check for native hardware and merge this assuming it all still works - it looks very curious that this builds at all for native given.

PR #1004, opened 2022-03-29
Temporarily enable DBG clock in stm32g0_detach()

### LINKED ISSUE #1003: hard fault on deep sleep with STM32G0 with DBGEN clock enabled
After I flash my STM32G030 the first `WFI` with deep sleep enabled causes Hard Fault (I think on wake up). This is fixed by power-cycling the board (but not by pulsing the NRST line). 

This wasn't the case when the L4 driver was used and started happening with the G0 driver. I found that the culprit is [enabling `DBGMCU` clock in `APBENR1`](https://github.com/blackmagic-debug/blackmagic/blob/master/src/target/stm32g0.c#L277). The code seems to disable it afterwards, but it somehow doesn't work. The hard fault only happens after flashing (not after just connecting to debugger). Seems like a really strange behavior.

@fabalthazar do you know when enabling that clock is needed? I can't find any docs in reference manual on that.

mmoskal:
fixes #1003

### DIFF
--- a/src/target/stm32g0.c
+++ b/src/target/stm32g0.c
@@ -295,6 +295,16 @@ static void stm32g0_detach(target *t)
 {
 	struct stm32g0_priv_s *ps = (struct stm32g0_priv_s*)t->target_storage;
 
+	/*
+	 * First re-enable DBGEN clock, in case it got disabled in the meantime
+	 * (happens during flash), so that writes to DBG_* registers below succeed.
+	 */
+	target_mem_write32(t, RCC_APBENR1, ps->saved_regs.rcc_apbenr1 |
+	                   RCC_APBENR1_DBGEN);
+
+	/*
+	 * Then restore the DBG_* registers and clock settings.
+	 */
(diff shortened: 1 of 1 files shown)


dragonmux (approved):
LGTM - needs @esden to do some testing on the hardware he has for this but otherwise this seems ready for merge

mmoskal:
While this fixes the immediate problem I reported I just had the same hard fault on sleep issue in other conditions that I’m not yet able to reproduce. I think we may want to always clear the debug in sleep bits when detaching, not just restore the previous state.

esden:
I have tested that this patch does not introduce a regression. I don't have a firmware to test the WFI mode and reproduce the bug that this patch is addressing at the moment. I did not have the spare cycles to put that together. If someone wants to contribute such an example with description how tot test it please feel free to open a pull request to the [test-fw-archive repository](https://github.com/blackmagic-debug/blackmagic-test-fw-archive/tree/main/stm32/g0). Ideally a simple code that does not depend on some huge framework/RTOS/HAL... locm3 based examples are preferred.

That said I think it is ok to merge this PR.

esden (approved):

PR #1006, opened 2022-04-02
CH32F103 flash support ("working" this time)

mean00:
The attached PR does two things : 
 - Group the clone function probe in the stm32f1_clones_probe function
 - Add flash support the CH32F103 chip

Without this PR, the BMP does not work at all with CH32F103 and fails after a few bytes.

The CH32F103 is similar to the STM32F103 with the following features : 
- Really does have only 64kB of very slow flash
- Have one usb device *and* one usb host/device IP
It is found in some chinese products as it is much more available than the stm32f103 chip probably.

I've put the ch32 code in a separate file included by stm32f1.c as to avoid spreading the changes all over the place.
That way, it's easy to remove if need be.
The ch32 code still contains a bit of debug stuff, as the core problem is not fully understood yet. 

What i think so far is happening  :

There is some sort of bus stall happening when the flash is programmed from the chip itself. 
That way, using the EOP/BUSY bits synchronize the operation just fine, albeit very slowly for that chip.

That synchronization does not happen completely for some reasons when the BMP probe does the flash programming.
The end result is that the programming sequence continues while the previous step is not done, resulting in incomplete/erroneous sequence.

The PR actually implements the "Fast" CH32 method not the STM32F103 method. 
To workaround the above  issue, it does 32x flash read before starting to load the flash write buffer to let the previous operation finish.
My measures showed that about 10 flash reads were needed, increased to 32 to have a bit of margin.
Yes it is hackish and the result of trials and errors/guesswork.

It is still a bit on the slow side, but it's bearable (and at least it works). My estimate is that it is ~ 2 times slower than a  stm32f103 similar chip.

For reference, openOCD using STM flash code is ~ 10 times slower with the CH32 than with a similar STM32F1 chip
It works because they do execute code on the target itself, and so are immune to the bus stalling problem.

I tried to comply to bmp coding style from the get go this time, probably missed a couple of friendly space etc...
Sorry in advance  if some got through


src/target/stm32f1.c
```diff
@@ -97,7 +97,9 @@ static int stm32f1_flash_write(struct target_flash *f,
 
 #define FLASHSIZE     0x1FFFF7E0
 #define FLASHSIZE_F0  0x1FFFF7CC
-
+//
+#include "stm32f1_ch32.c"
```
dragonmux, line 101: Rather than doing this, we'd rather see a new target in the build system even at the expense of some duplicated code.

src/target/cortexm.c
```diff
@@ -442,7 +454,7 @@ bool cortexm_probe(ADIv5_AP_t *ap)
 				PROBE(rp_probe);
 			PROBE(lpc11xx_probe); /* LPC8 */
 		} else if (ap->ap_partno == 0x4c3)  { /* Cortex-M3 ROM */
-			PROBE(stm32f1_probe); /* Care for STM32F1 clones */
+			PROBE(stm32f1_clones_probe); /* Care for STM32F1 clones */
```
dragonmux, line 457: This is an interesting approach and we can see why you've done it this way, but I think it'd be better to stick the CH32F1 probe routine here as another normal PROBE() call and treat it as its own target rather than a weird subtype of STM32F1

src/target/stm32f1_ch32.c
```diff
@@ -0,0 +1,338 @@
...
+#define FLASH_CR_FLOCK_CH32       (1<<15) // fast unlock
+#define FLASH_CR_FTPG_CH32        (1<<16) // fast page program
+#define FLASH_CR_FTER_CH32        (1<<17) // fast page erase
+#define FLASH_CR_BUF_LOAD_CH32    (1<<18) // Buffer load
+#define FLASH_CR_BUF_RESET_CH32   (1<<19) // Buffer reset
+#define FLASH_SR_EOP              (1<<5)  // End of programming
+#define FLASH_BEGIN_ADDRESS_CH32  0x8000000
+#define FLASH_MAGIC               (FPEC_BASE+0x34)
+
+static volatile uint32_t magic,sr,ct;
```
dragonmux, line 36: Is it absolutely necessary to carry these as global state? (also, spaces go around operators such as << and +, and after commas)

src/target/stm32f1_ch32.c
```diff
@@ -0,0 +1,338 @@
...
+	return 0;
+}
+
+/**
+	\fn waitFlashReady
+	\brief poll the beginning of a block till it is fffff, meaning we can proceeed
+
+*/
+
+static bool waitFlashReady(target *t,uint32_t adr)
```
dragonmux, line 208: please call this ch32f1_wait_flash_ready() to be consistent with the project naming style.

src/target/stm32f1_ch32.c
```diff
@@ -0,0 +1,338 @@
...
+{
+  // Wait a bit
+  // the # of cycles needed to do N read access over SWD
+  // the actual number is around 10
+  // then check we do read ffff (/!\ even if we read ffff it does not mean its ok
+  // these are the data from the previous operation and they could be ffff)
+  uint32_t ff;
+  for(int i=0;i<32;i++)   ff=target_mem_read32(t,adr);
```
dragonmux, line 216: Same comment as above about the lack of spaces here. Also, the body of the for loop really should be on its own line.

src/target/stm32f1_ch32.c
```diff
@@ -0,0 +1,338 @@
...
+	  return false;
+  }
+  return true;
+}
+/**
+  \fn ch32f1_flash_write
+  \brief fast flash for ch32. Load 128 bytes chunk and then flash them
+*/
+
+static int upload(target *t, uint32_t dest, const void  *src, uint32_t offset)
```
dragonmux, line 228: ch32f1_upload() please

src/target/stm32f1_ch32.c
```diff
@@ -0,0 +1,338 @@
...
+		if ((sr & SR_ERROR_MASK) ) {
+			ERROR_CH("ch32f1 flash write error 0x%" PRIx32 "\n", sr);
+			return -1;
+		}
+
+	}
+#ifdef CH32_VERIFY
+	DEBUG_CH("Verifying\n");
+	int i=0;
+	for(i=0;i<(int)len;i+=4)
```
dragonmux, line 322: use size_t for the loop counter, don't cast a size_t variable to int. The compiler warning was telling you about a bug you've created here - this is not the fix.

dragonmux (changes requested):
It's looking like a good start, but rather than having a C impl file include another C impl file, please add a new target to the build system (it's a one line change in the Makefiles).

There are a few consistency changes needed and a pile of extra newlines that could be cleaned up, but we think it's not far from being mergable.

esden:
@mean00 thank you for the contribution on this. It is very much appreciated. You did not misunderstand Uwe, but his suggestion is wrong. @dragonmux suggests a cleaner solution, so follow their lead. Looking forward to the updated PR. :slight_smile:

mean00:
Hi
I hope/think i've taken all the remarks into account
Thank you

esden (approved):
I tested this changes on an STM32F103 and it seems to still detect and work fine. (I don't have a CH32F103 test setup yet)

Unless there are some other concerns I think we can Rebase and Merge this. :)

esden (approved):

PR #1010, opened 2022-04-15
scripts: bootprog.py: use bytes literals

schodet:
As requested in #984, replace int literals with bytes literals.

Still using `bytes(…)` when all I have is a int.


schodet:
Tested on a stm32f100, by commenting out read/write protection related calls.

esden:
Looks good to me. Thank you very much for the contribution! :)

PR #1012, opened 2022-04-20
clang-format: add initial format file

xobs:
Add a format file for `clang-format`. This is based on the `.clang-format` file from Linux, adjusted to work with this project.

Notably, it opts for 120 characters, given that some functions are rather wide.

Care was chosen to adopt settings that resulted in a minimum of disruption when run through the formatter, however since no formatter has been used before this necessarily results in code that does not conform to the formatter.

This PR is made more to spur discusson on this, depending on whether or not this is a desirable thing to have.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,45 @@
+---
+AccessModifierOffset: -4
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AllowShortFunctionsOnASingleLine: None
+AllowShortIfStatementsOnASingleLine: false
+# This is currently broken:
+# https://github.com/llvm/llvm-project/issues/53442
+#AlignArrayOfStructures: Left
+BraceWrapping:
+  AfterClass: false
+  AfterControlStatement: false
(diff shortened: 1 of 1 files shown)


esden:
That is great! Thank you for the contribution, it will save a lot of time. Can't wait to incorporate it into our PR tests. I will look through this with @dragonmux and merge/modify as needed. 

Glad to hear that you already tried to make it low impact onto the codebase. I will likely have to make a bulk coding style correction commit after we solidify the coding style.

dragonmux (approved):
This looks like a good first run at a clang-format config for the repo.
There are things we'd like to tweak either because things aren't formatted in a way that works well for consistency, or as in the case of UseTab, and TabWidth the feeling between Esden and us is that a tab width of 4 and UseTab being set for AlignWithSpaces.

Our feeling personally is that's probably best done in the feature/clang-tidy-format branch and we early-merge this so we have something.

esden (approved):
Ok, I took it a bit for a spin to see how it behaves. As @dragonmux says it is a good start so we should merge this. And I agree that we can discuss and make improvements in a separate branch.

PR #1014, opened 2022-05-17
added RTS and DTR to UART interface

cyber-murmel:
The flow control lines are commonly used to bring MCU into different boot modes (e.g. ESP32)


dragonmux (approved):
It builds on native and doesn't change how we actually handle any targets, while on enabled platforms passing the host's requests to change DTS and RTS line statuses through to the micro's pins. Seems fine, merging.

PR #1015, opened 2022-05-17
add shell.nix

cyber-murmel:
This file contains a Nix function, which spawns a [nix-shell](https://nixos.org/manual/nix/unstable/command-ref/nix-shell.html) providing the tools necessary to build and interact with blackmagic.
It's like a `requirements.txt` for people using Nix or NixOS. The development environment is entered by simply running `nix-shell` in the root of the repository.

The package repository is pinned to a commit to provide reproducibility, but can be overwritten from command line.

The file may need to be updated when moving to a newer version of the underlying libraries (e.g. libus) when they bring breaking changes. Ideally this means only updating the `ref` end `ref` in line 5 and 7

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -15,6 +15,7 @@ mkShell {
     gcc-arm-embedded
     dfu-util
 
+    blackmagic
     pkg-config
     libftdi1
     libusb-compat-0_1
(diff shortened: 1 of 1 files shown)


dragonmux:
Thank you for the contribution.

Could you please edit the initial post to include a description of what this is, why we'd want it and what kind of maintenance overhead we're looking at? Thanks!

PR #1016, opened 2022-05-17
hosted: with mubes help, fixes some issues with how we talk CMSIS-DAP

dragonmux:
It seems that the addition of the DragonProbe target exposed some per-existing weaknesses in our CMSIS-DAP implementation that results in comms not working with some adaptors - this PR fixes said protocol implementation issues.


esden (commented):
LGTM

esden:
@dragonmux & @mubes Thank you for sorting it out. :) Merged.

mubes:
There may be a couple more to come...just working through some issues at the moment that might be on ORBTrace or BMP side....more likely ORBTrace in fairness because BMP is the only probe interface that seems to use DAP_Transfer and DAP_Block_Transfer properly, so it's exercising that gateware heavily. Will report back.

esden:
@mubes keep them coming! :D

PR #1018, opened 2022-05-21
update include/version.h only when content changes

TheJJ:
this adds on-demand regeneration for `include/version.h`.

without this patch, parts of the code were always recompiled, even though nothing changed at all.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -153,7 +153,13 @@ all_platforms:
 
 command.c: include/version.h
 
+GIT_VERSION := $(shell git describe --always --dirty --tags)
+VERSION_HEADER := \#define FIRMWARE_VERSION "$(GIT_VERSION)"
+
 include/version.h: FORCE
-	$(Q)echo " GIT include/version.h"
-	$(Q)echo "#define FIRMWARE_VERSION \"$(shell git describe --always --dirty --tags)\"" > $@
+ifneq ($(file <include/version.h), $(VERSION_HEADER))
+	@echo " GEN GIT  $@"
+	$(Q)$(file >$@,$(VERSION_HEADER))
+endif
+
 -include *.d


UweBonnes:
Running on opensuse and tcsh build in root directory fails for me:
  BUILD   lib/lm4f
make[2]: Nothing to be done for 'all'.
Makefile:156: *** unterminated call to function 'shell': missing ')'.  Stop.

src/Makefile
```diff
@@ -153,7 +153,12 @@ all_platforms:
 
 command.c: include/version.h
 
+VERSION_STRING := $(shell echo "#define FIRMWARE_VERSION \"`git describe --always --dirty --tags`\"")
+
 include/version.h: FORCE
-	$(Q)echo " GIT include/version.h"
-	$(Q)echo "#define FIRMWARE_VERSION \"$(shell git describe --always --dirty --tags)\"" > $@
+ifneq ($(file <include/version.h), $(VERSION_STRING))
```
dragonmux, line 159: This introduces what looks like it might be a bash-ism, it might be worth having the rule be `$(Q)./gen-version.sh` or so, and have a script based on /bin/sh to regen the header as-needed

dragonmux:
Thank you for the contribution! Reducing the amount of churn on rebuild would definitely be interesting

src/Makefile
TheJJ: not sure a separate script pays off here, i hope i've simplified the code now so it works with other shells too :)

UweBonnes:
Thanks, that works now for me!

src/Makefile
dragonmux: The way you fixed it is also valid (hopefully won't cause issues with older GNU Make and make-alternatives). The reason we suggested a standalone script is so you wouldn't tangle with other bash-alternative shells, not because it was the only workable option :)

dragonmux (approved):
LGTM, we'll merge this when the v1.9 window opens

UweBonnes (approved):

PR #1021, opened 2022-05-24
target: add samx7x probe method

### LINKED ISSUE #963: Add support for SAMS70 (C-M7)
Just showing interest for a particular family i use, and cortex-m7 support in general (no rush)

https://www.microchip.com/en-us/product/ATSAMS70N19

Dev boards for this are few, expensive, and probably out of stock, but I have a custom board and couple spare ICs if needed

Open to help support this

perigoso:
adds support for SAME70, SAMS70, SAMV71 and SAMV70

closes #963 

### DIFF
--- a/src/target/sam3x.c
+++ b/src/target/sam3x.c
@@ -178,8 +178,18 @@ struct sam_flash {
 	uint8_t write_cmd;
 };
 
+struct samx7x_descr {
+	char product_code;
+	uint8_t product_id;
+	char pins;
+	uint32_t ram_size;
+	uint32_t flash_size;
+	uint8_t density;
+	char revision;
+};
 struct sam_priv_s {
-	char sam_variant_string[60];
+	struct samx7x_descr descr;
+	char sam_variant_string[16];
 };
 
 static void sam3_add_flash(target *t,
@@ -228,6 +238,35 @@ static void sam_add_flash(target *t,
 	target_add_flash(t, f);
 }
 
+static void samx7x_add_ram(target* t, uint32_t tcm_config, uint32_t ram_size) {
+	uint32_t itcm_size = 0, dtcm_size = 0;
+
+	switch (tcm_config)
+	{
+	case GPNVM_SAMX7X_TCM_32K:
+		itcm_size = dtcm_size = 0x8000;
+		break;
+	case GPNVM_SAMX7X_TCM_64K:
+		itcm_size = dtcm_size = 0x10000;
+		break;
+	case GPNVM_SAMX7X_TCM_128K:
+		itcm_size = dtcm_size = 0x20000;
+		break;
+	}
+
+	if (dtcm_size > 0) {
+		target_add_ram(t, 0x20000000, dtcm_size);
+	}
+	if (itcm_size > 0) {
+		target_add_ram(t, 0x00000000, itcm_size);
+	}
+
+	uint32_t sram_size = ram_size - (itcm_size + dtcm_size);
+	if (sram_size > 0) {
+		target_add_ram(t, 0x20400000, sram_size);
+	}
+}
+
 static size_t sam_flash_size(uint32_t cidr)
 {
 	switch (cidr & CHIPID_CIDR_NVPSIZ_MASK) {
@@ -264,16 +303,6 @@ static size_t sam_sram_size(uint32_t cidr)
 		return 0;
 	}
 }
-
-struct samx7x_descr {
-	char product_code;
-	uint8_t product_id;
-	char pins;
-	uint32_t ram_size;
-	uint32_t flash_size;
-	uint8_t density;
-	char revision;
-};
 
 struct samx7x_descr samx7x_parse_id(uint32_t cidr, uint32_t exid) {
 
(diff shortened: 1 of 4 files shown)


dragonmux (approved):
LGTM - we don't have any hardware to test this on so if you could post the output of a successful detection and Flash just to prove that the code does what it claims to, that would be appreciated.

We'll merge this in sequence with the opening of the v1.9 merge window.

perigoso:
second commit is a bit opinionated, let me know if it's ok

dragonmux:
Think we need to come up with some kind of "3x/x7" naming convention here, but at least given a cursory glance for now, it seems reasonable enough as it's in-line with how things are named in other parts of the code base.

perigoso:
yeah the fact that there are other sam___ target files makes it non ideal

perigoso:
```
[perigoso@perigoso-desktop blackmagic]$ ./src/blackmagic -t
INFO: Open USB 046d:c31d class  0 failed
BMP hosted v1.7.1-418-g725670a
Using 1d50:6018 0669FF564949775087212225 Black Sphere Technologies
 Black Magic Probe (STLINK/V2) v1.7.1-415-gc4869a5
Running in Test Mode
Target voltage: 3.30V Volt
Speed set to  3.2727 MHz for SWD
DPIDR 0x0bd11477 (v1 MINDP rev0)
RESET_SEQ failed
AP   0: IDR=04770041 CFG=00000000 BASE=e00fd003 CSW=03000040 (AHB-AP var4 rev0
Halt via DHCSR: success 00030003 after 1ms
ROM: Table BASE=0xe00fd000 SYSMEM=0x00000001, designer  1f Partno   0
ROM: Table BASE=0xe00fe000 SYSMEM=0x00000001, designer 43b Partno 4c8
ROM: Table BASE=0xe00ff000 SYSMEM=0x00000001, designer 43b Partno 4c7
  0 0xe000e000: Generic IP component - Cortex-M4 SCS (System Control Space) (PIDR = 0x04000bb00c  DEVTYPE = 0x00 ARCHID = 0x0000)  -> cortexm_probe
Silicon bug: Single stepping will enter pending exception handler with this M7 core revision!
CPUID 0x410fc271 (M7 var 0 rev 1)
  1 0xe0001000: Generic IP component - Cortex-M3 DWT (Data Watchpoint and Trace) (PIDR = 0x04000bb002  DEVTYPE = 0x00 ARCHID = 0x0000)
  2 0xe0002000: Generic IP component - Cortex-M7 FBP (Flash Patch and Breakpoint) (PIDR = 0x04000bb00e  DEVTYPE = 0x00 ARCHID = 0x0000)
  3 0xe0000000: Generic IP component - Cortex-M3 ITM (Instrumentation Trace Module) (PIDR = 0x04000bb001  DEVTYPE = 0x00 ARCHID = 0x0000)
  4 Entry 0xfff41002 -> Not present
  5 Entry 0xfff42002 -> Not present
  ROM: Table END
 1 0xe0041000: Debug component - Cortex-M7 ETM (Embedded Trace) (PIDR = 0x04000bb975  DEVTYPE = 0x13 ARCHID = 0x4a13)
 2 Entry 0xfff44002 -> Not present
 3 Entry 0x1ff02002 -> Not present
 ROM: Table END
1 0xe0040000: Debug component - Cortex-M7 TPIU (Trace Port Interface Unit) (PIDR = 0x04000bb9a9  DEVTYPE = 0x11 ARCHID = 0x0000)
2 Entry 0x1ff03002 -> Not present
3 Entry 0x1ff03002 -> Not present
ROM: Table END
***  1      SAMS70J19A M7
RAM   Start: 0x20000000 length = 0x40000
Flash Start: 0x00400000 length = 0x80000 blocksize 0x1000
```

```
[perigoso@perigoso-desktop blackmagic]$ ./src/blackmagic openinput-sams70-generic-r334-627a730.hex
Silicon bug: Single stepping will enter pending exception handler with this M7 core revision!
Flash Write succeeded for 45199 bytes,   46.645 kiB/s
```

src/target/sam3x.c
```diff
@@ -195,9 +228,152 @@ static size_t sam_flash_size(uint32_t cidr)
 	return 0;
 }
 
+static size_t sam_sram_size(uint32_t cidr)
+{
+	switch (cidr & CHIPID_CIDR_SRAMSIZ_MASK) {
```
Syncena, line 233: You shouldn't hard-code the SRAM size here. Some of these devices have dTCM and iTCM which steal from the SRAM array. The sizes are configured by GPNVM bits in the Flash controller. Some configurations permit all SRAM to be allocated to xTCM, so SRAM size in this situation will be zero.

See [Maven's SAMx7X](https://github.com/Syncena/Maven-firmware/blob/main/src/cortexm/soc_samx7x.c) for an example.

src/target/sam3x.c
```diff
@@ -195,9 +228,152 @@ static size_t sam_flash_size(uint32_t cidr)
...
+	case CHIPID_CIDR_ARCH_SAMS70:
+	case CHIPID_CIDR_ARCH_SAMV71:
+	case CHIPID_CIDR_ARCH_SAMV70:
+		break;
+	default:
+		return false;
+	}
+
+	struct samx7x_descr descr = samx7x_parse_id(cidr, exid);
+	target_add_ram(t, 0x20000000, descr.ram_size);
```
Syncena, line 349: SRAM is at 0x20400000. You'll find dTCM at 0x20000000 only if configured in the GPNVM bits.

src/target/sam3x.c
perigoso: I'm aware of the TCM, but I'm unsure if I should do this dynamic setting of ram, if a command to set  GPNVM bits is done I would need to recalculate and reset the ram, besides physically the ram is there, it just has a different purpose, depends on what exactly blackmagic or gdb does with this

@dragonmux is re-setting the ram possible? should the ram be dynamically set?

perigoso: for reference,

TCM is "Tightly Coupled Memory", essentially a region of ram that is directly accessible by the processor at it's running clock, used for fast code execution/memory access, it is dynamically set by 2 GPNVM (like fuse bits), and makes use of the SRAM

perigoso: Hm I'm thinking about this wrong, the ram would still be mapped at 0x20400000, just with a smaller size, so you're right, I should do dynamic setting, though for reference, from he datasheet `"Accesses made to the SRAM above the size limit will not generate aborts"`

perigoso:
TCM handling is untested as of now, no access to hardware

perigoso:
so after changing the bits in the GPNVM, the TCM configuration might change, but this only takes effect after a reset

my thinking right now is to store the current TCM configuration in `target_storage` and add an `extended_reset` method that checks if the configuration has changed and does a memory reconfiguration somehow

does this sound like a good idea?

my question is if this is allowed (reconfiguring the memories), or if there some method to trigger a "reprobe", and if this would handle resets triggered by the device itself (I assume not)

perigoso:
looking through the code, two other targets have some sort of TCM, `stm32h7` and `stm32f7`, they both seem to handle the memory mapping on attach, although they don't seem to have dynamic TCM like this

perigoso:
complete, but untested due to hardware problems on my side

dragonmux (approved):
Looks good to us!

PR #1023, opened 2022-05-30
Copyright and vendor strings update

esden:
As the GitHub organization was renamed to Black Magic Debug. The copyright and USB vendor strings are not correct any more. This PR fixes this.

We are leaving the codebase Copyright header strings in place as they are correct and the code was written and copyrighted by the Black Sphere Technologies company.

The only issue that remains that the new probes might not be detected correctly by Black Magic App (BMA aka. hosted) in "serial" mode. The code for the device detection in there is so convoluted and hard to read that I decided not to touch it until it is completely rewritten in a proper way.


dragonmux (approved):
LGTM, merging.

PR #1025, opened 2022-05-31
A couple of misc fixes to BMDA ported from Uwe's RISC-V support branch

dragonmux:
This fixes a clean exit on error bug, and an issue with how BMDA detects some probes ("cables" in the current terminology used)


esden (approved):
LGTM

PR #1026, opened 2022-05-31
BMDA - Add more adaptors - ESP-Prog, Tigard, HiFive1 and ARM-USB-TINY-H

dragonmux:
In preparation for supporting RISC-V targets, this adds support for some new probes both general (Tigard) and focused on RISC-V targets (ESP-Prog, HiFive1) to the Black Magic Debug App.


esden (approved):
LGTM

PR #1027, opened 2022-05-31
stm32/gdb_if: Use _WFI when nothing to do.

dragonmux:
Ported out from Uwe's RISC-V branch, this makes use of the WFI instruction in STM32 based probes when they're idle. To be seen if this causes problems or not - this one will need some further testing before merge.

### DIFF
--- a/src/platforms/stm32/gdb_if.c
+++ b/src/platforms/stm32/gdb_if.c
@@ -24,6 +24,7 @@
  */
 #include "general.h"
 #include "cdcacm.h"
+#include <libopencmsis/core_cm3.h>
 #include "gdb_if.h"
 
 static uint32_t count_out;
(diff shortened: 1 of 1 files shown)


esden:
This requires actual hardware testing before we can merge this patch. So I am assigning it to myself, I hope I come around to it soon.

dragonmux:
ack, just noticed this duplicates #941 (minus one extra thing that introduces which.. we are less than comfortable merging) - do we want to close this one or that one, continuing with the one we don't close?

esden (approved):
LGTM, I have tested it on BMP V2.3b, and it seems to be working ok.

I was concerned that this might introduce problems debugging the BMP firmware (inception). But I seem to be able to scan and break into the firmware without problem. If this starts causing issues down the road I take the liberty to get rid of WFI again. But for now, let's save those milliwatts :D

PR #1028, opened 2022-06-03
Complete the version string migration from 5e3cade, fixing build for HOSTED_BMP_ONLY=1

Qyriad:
Looks like 5e3cade left a few string comparisons un-migrated which caused this specific build to fail. This migrates those couple instances over.


esden:
Great! Thank you very much! You get the "first v1.8" bugfix patch badge. :D

I will take care of it in the next spare work cycle. :D

esden (approved):
LGTM

PR #1030, opened 2022-06-04
build: Added all_platforms target to toplevel Makefile

esden:
Now it is possible to run the all_platforms build target from the toplevel directory.

The src level all_platforms target now also checks for locm3 existence
as well as triggers build of the needed locm3 libraries.

This prepares for switching the build tests to using the all_platforms build target instead of just the native one.


dragonmux (approved):
LGTM, merging.

PR #1031, opened 2022-06-04
96b_carbon: Updated so that it compiles in the current codebase.

esden:
The pull request we merged to add 96b carbon target was a tiny bit stale. It needed some TLC :D

### DIFF
--- a/src/platforms/96b_carbon/platform.h
+++ b/src/platforms/96b_carbon/platform.h
@@ -32,11 +32,8 @@
 #include <setjmp.h>
 
 #define PLATFORM_HAS_TRACESWO
-#define BOARD_IDENT       "Black Magic Probe (Carbon), (Firmware " FIRMWARE_VERSION ")"
-#define BOARD_IDENT_DFU   "Black Magic (Upgrade) for Carbon, (Firmware " FIRMWARE_VERSION ")"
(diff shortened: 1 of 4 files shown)


src/platforms/96b_carbon/platform.h
```diff
@@ -100,9 +97,9 @@
  * For now USART preempts USB which may spin while buffer is drained.
  * TIM3 is used for traceswo capture and must be highest priority.
  */
-#define IRQ_PRI_USB		(2 << 4)
-#define IRQ_PRI_USBUSART	(1 << 4)
-#define IRQ_PRI_USBUSART_TIM	(3 << 4)
+#define IRQ_PRI_USB		(1 << 4)
```
dragonmux, line 100: Could you update the comment above as well so it reflects the new prioritisations?

dragonmux (changes requested):
Looks decent, we have one thing that would be good to see changed though

src/platforms/96b_carbon/platform.h
esden: Good catch. Thank you. I have updated the comment. Let me know if it is correct/sufficient. :)

dragonmux (approved):
LGTM :) ready to merge if you think it's complete

esden:
Wonderful. Thanks! :)

PR #1032, opened 2022-06-04
build: Corrected the all_platform build target.

esden:
The target was building but not including the BMDA binary, as the
build system does not expect a binary that does not end with `.bin`.
Additionally this corrects the BMDA Makefile.inc that was missing the
`all` and `host_clean` targets.


dragonmux (approved):
LGTM, merging

PR #1034, opened 2022-06-04
Feature/blackpill extraction

esden:
This patch extracts the blackpill platform from f4discovery. The way our current build system works it is not possible to build multiple variants of the same platform. Without that ability we can not automatically run build tests and prevent regressions. Thus extracting the target is a better choice even though it means a bit of code duplication.

The new build system scheduled for v2.0 release cycle will hopefully address some of the code duplication and make the support of host platforms easier. Until then each new platform needs to have it's distinct sub-directory with a `Makefile.inc`.


src/platforms/blackpill/platform.c
```diff
@@ -0,0 +1,144 @@
...
+	OTG_FS_GCCFG |= OTG_GCCFG_NOVBUSSENS | OTG_GCCFG_PWRDWN;
+	OTG_FS_GCCFG &= ~(OTG_GCCFG_VBUSBSEN | OTG_GCCFG_VBUSASEN);
+}
+
+void platform_srst_set_val(bool assert) { (void)assert; }
+bool platform_srst_get_val(void) { return false; }
+
+const char *platform_target_voltage(void)
+{
+	return NULL;
```
dragonmux, line 123: This looks like it would be better implemented as returning a string like "Unknown"

src/platforms/blackpill/platform.c
```diff
@@ -0,0 +1,144 @@
...
+bool platform_target_get_power(void)
+{
+	return !gpio_get(PWR_BR_PORT, PWR_BR_PIN);
+}
+
+void platform_target_set_power(bool power)
+{
+	gpio_set_val(PWR_BR_PORT, PWR_BR_PIN, !power);
+}
+#endif
```
dragonmux, line 144: Could the newline that GH is complaining about be added as this will break older compilers and cause noise when we reformat the code base.

dragonmux (changes requested):
Mostly looks good to us, just a couple of small things that could do with some attention and then it's ready for merge.

src/platforms/blackpill/platform.c
esden: It seems that all other targets that don't have voltage measurement return NULL here. On some level it makes it easier to check if the platform has voltage support or not when calling this function instead of having to compare strings.

src/platforms/blackpill/platform.c
esden: Ok, I added the newline as requested.

esden:
I renamed the platform to blackpillv2. There is also a board called blackpill which uses an stm32f1, whereas the blackpill V2 uses an stm32f4. The naming change should remove some confusion. :)

src/platforms/blackpill/platform.c
dragonmux: Ah, hmm.. we'd seen something returning a string when it couldn't know voltage, but we buy your reasoning here.

dragonmux (approved):

PR #1035, opened 2022-06-04
Feature: BMDA better usage output and help

dragonmux:
In this PR we have: Completely rewritten the help output to make it more useful and readable. In addition, we have generated a better usage output from the help option and fixed 'hosted' appearing in the version output.

As a side benefit, this includes complete ZSH completions under contrib/zsh.

### DIFF
--- a/src/platforms/hosted/bmp_libusb.c
+++ b/src/platforms/hosted/bmp_libusb.c
@@ -28,8 +28,8 @@
 
 void bmp_ident(bmp_info_t *info)
 {
-	PRINT_INFO("Black Magic Debug App %s\n for Black Magic Probe, ST-Link V2/3, CMSIS_DAP,"
-		" JLINK and LIBFTDI/MPSSE\n", FIRMWARE_VERSION);
+	PRINT_INFO("Black Magic Debug App %s\n for Black Magic Probe, ST-Link v2 and v3, CMSIS-DAP,"
+		" JLink and libftdi/MPSSE\n", FIRMWARE_VERSION);
 	if (info && info->vid && info->pid) {
 		PRINT_INFO("Using %04x:%04x %s %s\n %s\n", info->vid, info->pid,
 			(info->serial[0]) ? info->serial : NO_SERIAL_NUMBER,
--- a/src/platforms/hosted/bmp_libusb.c
+++ b/src/platforms/hosted/bmp_libusb.c
@@ -28,10 +28,10 @@
 
 void bmp_ident(bmp_info_t *info)
 {
-	DEBUG_INFO("BMP hosted %s\n for ST-Link V2/3, CMSIS_DAP, JLINK and "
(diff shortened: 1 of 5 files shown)


src/platforms/hosted/bmp_libusb.c
```diff
@@ -28,10 +28,10 @@
 
 void bmp_ident(bmp_info_t *info)
 {
-	DEBUG_INFO("BMP hosted %s\n for ST-Link V2/3, CMSIS_DAP, JLINK and "
+	PRINT_INFO("Black Magic Debug App %s\n for ST-Link V2/3, CMSIS_DAP, JLINK and "
```
esden, line 31: I think this list should also include Black Magic Probe in that list. Currently it makes it seem that it only supports 3rd party probes.

src/platforms/pc/cl_utils.c
```diff
@@ -130,52 +130,65 @@ static void bmp_munmap(struct mmap_data *map)
...
-	DEBUG_WARN("\t-M <string>\t: Run target specific monitor commands. Quote multi\n");
-	DEBUG_WARN("\t\t\t  word strings. Run \"-M help\" for help.\n");
-	DEBUG_WARN("Flash operation modifiers options:\n");
-	DEBUG_WARN("\tDefault action with given file is to write to flash\n");
-	DEBUG_WARN("\t-a <addr>\t: Start flash operation at flash address <addr>\n"
-		"\t\t\t  Default start is start of flash in memory map\n");
-	DEBUG_WARN("\t-S <num>\t: Read <num> bytes. Default is until read fails.\n");
-	DEBUG_WARN("\t <file>\t\t: Use (binary) file <file> for flash operation\n");
+	PRINT_INFO("\n");
+	PRINT_INFO("Usage: %s [-h | -l | [-vBITMASK] [-d PATH | -P NUMBER | -s SERIAL | -c TYPE]\n", argv[0]);
```
esden, line 134: Do we need to print every single line separately? Would it not be possible to just write:
```
PRINT_INFO(
"my first line of text\n"
"my second line of text\n"
"my third line of text\n");
```
I think this is specified in C89 so it should be fine to use and in my opinion would make this look better and make it slightly easier to edit when needed.

src/platforms/hosted/bmp_libusb.c
dragonmux: Ah, good catch!

src/platforms/pc/cl_utils.c
dragonmux: We can do it that way, we didn't initially because of the extra preprocessor logic that was in there, and the command line argument, but now that's all sorted, we can do a second pass and fix this.

esden: Ohh ok, No worries, I figured that it might be something along those lines. Let's make that additional pass and then we should be good to merge this. :)

esden (approved):
Wonderful! Thanks for the additional changes. Let's merge it. :)

PR #1037, opened 2022-06-05
platforms/f3: Introduced a new platform for the STM32F3 series

dragonmux:
As part of pulling the RISC-V PR apart to its principal components and because it features in other branches on Uwe's fork, here is a platform for the STM32F3 series parts to allow BMF to run on them


esden (approved):
LGTM

PR #1038, opened 2022-06-05
platforms/f072: Introduced a new platform for the SMT32F072

dragonmux:
As part of pulling the RISC-V PR apart to its principal components and because it features in other branches on Uwe's fork, here is a platform for the STM32F072 to allow BMF to run on it.

Some notes from the commit this was rescued from:
F0 needs separation of DMA Interrupts, show problems with 128 Byte USART/DMA
buffers, perhaps caused by the st_usbfs_v2_usb_driver and has no
scb_reset_core.


scripts/dfu.py
```diff
@@ -112,7 +112,7 @@ def clear_status(self):
 			usb.RECIP_INTERFACE, DFU_CLRSTATUS,
 			"", index=0)
 
-	def get_state(self):
+		def get_state(self):
```
esden, line 115: This seems like a separate bug fix unrelated to the new platform addition?

esden (changes requested):
Except for the dfu.py change I think it looks good. I will merge as soon as we move that dfu.py change out.

scripts/dfu.py
dragonmux: Ah, yes, hadn't even spotted that got in to this - we'll rework to remove that tomorrow.

esden: Sounds good :D

esden: As discussed on "sideband" I resolved this. :D

esden (approved):
LGTM

PR #1044, opened 2022-06-08
target: add weak aliasing to allow disabling targets

perigoso:
Cherry-picked from #1033

This PR doesn't actually allow (easily) the disabling targets, just lays some groundwork for it, and cleans a couple things up

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -61,6 +61,7 @@ SRC =			\
 	stm32l4.c	\
 	stm32g0.c	\
 	target.c	\
+	target_probe.c
 
 
 include $(PLATFORM_DIR)/Makefile.inc
--- a/src/target/adiv5.c
+++ b/src/target/adiv5.c
@@ -28,6 +28,7 @@
 #include "general.h"
 #include "target.h"
 #include "target_internal.h"
+#include "target_probe.h"
 #include "adiv5.h"
 #include "cortexm.h"
 #include "exception.h"
@@ -273,8 +274,6 @@ static const struct {
 /* Used to probe for a protected SAMX5X device */
 #define SAMX5X_DSU_CTRLSTAT 0x41002100U
 #define SAMX5X_STATUSB_PROT (1U << 16U)
-
-extern bool cortexa_probe(ADIv5_AP_t *apb, uint32_t debug_base);
 
 void adiv5_ap_ref(ADIv5_AP_t *ap)
 {
@@ -677,7 +676,7 @@ static void rp_rescue_setup(ADIv5_DP_t *dp)
 		return;
 	}
 	ap->dp = dp;
-	extern void rp_rescue_probe(ADIv5_AP_t *);
+
 	rp_rescue_probe(ap);
 	return;
 }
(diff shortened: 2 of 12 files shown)


perigoso:
I don't understand the chain of includes that made it so this fails with my changes, supposedly doesn't touch any of that

perigoso:
doesn't make much sense, it should have been failing before?

src/target/target_probes.c
```diff
@@ -0,0 +1,87 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "target_probes.h"
+
+#define CORTEXA_PROBE_DUMMY __attribute__((weak, alias("_cortexa_probe_dummy")))
+#define CORTEXM_PROBE_DUMMY __attribute__((weak, alias("_cortexm_probe_dummy")))
+#define TARGET_PROBE_DUMMY __attribute__((weak, alias("_target_probe_dummy")))
+
+static inline bool _target_probe_dummy(target *t)
```
dragonmux, line 26: Rather than calling these "dummy", and as the _ on the front makes them reserved naming violating, we'd prefer, eg, `target_probe_nop` or similar.

src/target/target_probes.c
```diff
@@ -0,0 +1,87 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "target_probes.h"
+
+#define CORTEXA_PROBE_DUMMY __attribute__((weak, alias("_cortexa_probe_dummy")))
```
dragonmux, line 22: These macros are probably better named either eg `CORTEXA_PROBE_WEAK` or `WEAK_CORTEXA_PROBE`

src/target/target_probes.c
```diff
@@ -0,0 +1,87 @@
...
+/*
+ * "dummy" alias functions to allow suport for these targets
+ * to be disabled by not compiling them in.
+ */
+
+bool cortexa_probe(ADIv5_AP_t *apb, uint32_t debug_base) CORTEXA_PROBE_DUMMY;
+
+bool cortexm_probe(ADIv5_AP_t *ap) CORTEXM_PROBE_DUMMY;
+
+bool kinetis_mdm_probe(ADIv5_AP_t *ap) CORTEXM_PROBE_DUMMY;
```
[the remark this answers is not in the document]
perigoso: the functions are not compatible because of the different argument types, unfortunately

src/target/target_probes.c
perigoso: nop does is better :+1:

perigoso:
I thought about adding a debug message to warn that a probe method was called that has been disabled at build time, but left it for a future commit if there's interest to not block this

src/target/target_probes.c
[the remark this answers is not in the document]
dragonmux: Ah, fair cop.. hadn't spotted that

dragonmux (approved):
LGTM, merging.

This'll definitely make, eg, just dropping all LPC targets out of firmware for debug builds.. considerably easier - thank you!

PR #1050, opened 2022-06-11
Fix version.h generation on BSD-like OSes

### LINKED ISSUE #1043: macOS: version.h not found
I'm turning back to blackmagic again cause I'm in need to do some programming .
Don't know if I made a mistake, but I did a 

` git clone  https://github.com/blackmagic-debug/blackmagic.git`

and then cd blackmagic followed by this and got this:
```
    $ pwd
    /Users/kuku/Downloads/blackmagic
    $ make PROBE_HOST=hosted
     GEN GIT  include/version.h
      CC      target/adiv5.c
      CC      target/adiv5_jtagdp.c
      CC      target/adiv5_swdp.c
      CC      command.c
    command.c:34:10: fatal error: 'version.h' file not found
    #include "version.h"
             ^~~~~~~~~~~
    1 error generated.
    make[1]: *** [command.o] Error 1
    make: *** [all] Error 2
```
It's under macOS. But it seems version.h is totally missing compared to the previous version I'm running and where compilation worked.

dragonmux:
This PR does two things to fix #1043 - the first part is to switch to the alternative approach we indicated in #1018 as this PR caused a regression.

The second is to apply the HIDAPI pkg-config based library finding fix also presented in #1043 so we get the MacOS flavour of BMDA building again even if an outstanding issue that is the lack of implementation of find_debuggers() for MacOS is still present.


dragonmux:
We note the failing CI - frustratingly, this all works flawlessly locally, so it's going to take a bit to diagnose this build failure.

Qyriad:
Fixed the CI issues — basically Ubuntu uses dash as its `/bin/sh`, and dash doesn't support the `&>` redirection syntax that was used in the `gen-version.sh` script.

esden (approved):
LGTM

Issue #1048, opened 2022-06-11
Platform: Add support for blackpillv1/blackpillv3


Garag:
The old version of the Blackpill with a STM32F401CCU controller have only 256K flash and 64K RAM. The current linker script is for the STM32F401CEU with 512K Flash and 96K RAM . As the Stackpointer is after reset above 64K it will hard fault.
If i change the linker script to the lower values BMP works. Maybe it is possible to change the values in the linker script or use conditionally a different linker script for the old model.

During analysis of the problem i found that the pin PA1 is used twice. PA1 is used for JTAG TDI and for LED UART. I think it is better to use different pins.

Link to the old version
[https://stm32-base.org/boards/STM32F401CCU6-WeAct-Black-Pill-V1.2.html](https://stm32-base.org/boards/STM32F401CEU6-WeAct-Black-Pill-V3.0)

Link to the new version
[https://stm32-base.org/boards/STM32F401CEU6-WeAct-Black-Pill-V3.0](https://stm32-base.org/boards/STM32F401CEU6-WeAct-Black-Pill-V3.0)

esden:
Thank you for your contribution @Garag The patch that you proposed was attempting to attempting to change the [BlackPill V2](https://stm32-base.org/boards/STM32F411CEU6-WeAct-Black-Pill-V2.0) board. Not V1.2 or V3.0 that you are referring to. So I rejected that patch.

In general the policy is to not tack on platform variants onto each other, unless the firmware will be guaranteed to always work and take advantage of the hardware at hand. Each platform has to stay self contained with a specific pin-out and chip in mind. If changes are needed, you either have to add another target platform sub-directory or implement some runtime system for auto-detection of the host chip you have at hand. The latter would be the most powerful but also most difficult solution to implement. You are welcome to work on that and a patch that consolidates blackpill boards into one single firmware is something we can consider merging. Otherwise, if you want to support your specific blackpill hardware version, please make it a separate platform with the correct linker script, pin out and readme.

perigoso:
Of note it seems the smaller flash version is marked as discontinued, so the motivation to support it is smaller

perigoso:
They also seem to have another board they call a blackpill, with lower flash and a st "compatible" part

https://github.com/WeActTC/WeActStudio.BlackPill

Garag:
The native BMP have only 128KB Flash, so 256K Flash should be no problem.

djix123:
[...]

PR #1052, opened 2022-06-13
Changed blackpillv2 linker script to use only 256k flash and 64k RAM …

Garag:
…to be compatible with stm32f401cc

See also Issue https://github.com/blackmagic-debug/blackmagic/issues/1048

### DIFF
--- a/src/platforms/stm32/blackpillv2.ld
+++ b/src/platforms/stm32/blackpillv2.ld
@@ -20,8 +20,8 @@
 /* Define memory regions. */
 MEMORY
 {
-	rom (rx) : ORIGIN = 0x08000000, LENGTH = 512K
-	ram (rwx) : ORIGIN = 0x20000000, LENGTH = 96K
+	rom (rx) : ORIGIN = 0x08000000, LENGTH = 256K
+	ram (rwx) : ORIGIN = 0x20000000, LENGTH = 64K
 }
 
 /* Include the common ld script from libopenstm32. */


perigoso:
I would create a new linker `blackpillv1` and use that by default, in the future it might become an option, and it's better documented

esden:
Thank you for your contribution @Garag The board you are attempting to change is specifically [BlackPill V2](https://stm32-base.org/boards/STM32F411CEU6-WeAct-Black-Pill-V2.0) (as the name of the target indicates). Not V1.2 or V3.0. So I will not be able to merge this patch as it is intended for other versions of the board not V2.

In general the policy is to not tack on platform variants onto each other, unless the firmware will be guaranteed to always work and take advantage of the hardware at hand. Each platform has to stay self contained with a specific pin-out and chip in mind. If changes are needed, you either have to add another target platform sub-directory or implement some runtime system for auto-detection of the host chip you have at hand. The latter would be the most powerful but also most difficult solution to implement. You are welcome to work on that and a patch that consolidates blackpill boards into one single firmware is something we can consider merging. Otherwise, if you want to support your specific blackpill hardware version, please make it a separate platform with the correct linker script, pin out and readme.

You are welcome to continue the discussion about this in the issue itself. I will be posting a copy of the above there too.

djix123:
I do wonder if its better to accept the patch - then the blackpill platform will work with all variants of the [blackpill](https://github.com/WeActTC/MiniSTM32F4x1). Seems odd to have arbitrary support for V2 (that happens to work for V3 and not V1) when supported all three versions is straightforward.

perigoso:
In this case it might make sense to rename the platform `blackpill` removing the arbitrary version, if they really are all compatible, which is likely, and have a build variable with the version, that in this case should only affect the linker

perigoso:
I have some ideas to make handling these annoyances easier in the future, if the project switches to meson

Garag:
The changed linker script will work with all Blackpill borads with a STM32F4xx. But i thing there is a Blackpill with a STM32F10x. Maybe a name like blackpill-f4 is better.
As the only change is the linker script, it is not possible to decide about the variant at runtime.

perigoso:
Hm, if that is the case i start to align more with esdens opinion

djix123:
I think what we are discussing is the WeAct TC BlackPill based on the F4. There is a BlackPill (RobotDyn, for example made them) based on the STM32F103 ... but these will work with PROBE_HOST=stlink for example. Better name maybe 'weact-blackpill' or 'weact-blackpill-f4' then the original patch will work for all variations of that board.

Garag:
WeAct seems to have only the F4 BlackPill. From RobotDyn i know a BlackPill like board with a STM32F303CC on it.

PR #1053, opened 2022-06-13
Change from tty.usb* to cu.usb* for more consistent behavior on OSX

djix123:
When using BMPA on OSX, BMPA will hang on the second invocation if the device is not specified as /dev/cu* rather than /dev/tty*. Change default BMPA behavior to use /dev/cu* on OSX.


esden (approved):
Thank you for the contribution and for fixing this. It was a known problem (#714) but it turns out an issue was never opened for this, and a patch promised but never submitted. Sorry for the inconvenience.

PR #1055, opened 2022-06-15
Enable power control in BMPA and add required functions

sidprice:
This PR enables target power control in BMPA and adds the support functions required.

Note, this has only been tested on Windows, however, given the nature of the changes it should work across all platforms.

### DIFF
--- a/src/platforms/hosted/platform.c
+++ b/src/platforms/hosted/platform.c
@@ -392,6 +392,38 @@ void platform_target_set_power(bool power)
 	}
 }
 
+bool platform_target_get_power(void)
+{
+	switch (info.bmp_type) {
+	case BMP_TYPE_BMP:
+	return remote_target_get_power() ;
+
+	default:
+		return false ;
+	}
+}
+
+uint32_t platform_target_voltage_sense(void)
+{
+	uint32_t	targetVoltage = 0 ;
+	switch (info.bmp_type) {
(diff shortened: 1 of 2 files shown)


src/platforms/hosted/platform.c
```diff
@@ -392,6 +392,38 @@ void platform_target_set_power(bool power)
...
+uint32_t platform_target_voltage_sense(void)
+{
+	uint32_t	targetVoltage = 0 ;
+	switch (info.bmp_type) {
+	case BMP_TYPE_BMP: {
+		const char * result ;
+		uint32_t units = 0, tenths = 0 ;
+		result = remote_target_voltage() ;
+		if (result != NULL) {
+			sscanf(result,"%u.%u", (unsigned int *) &units, (unsigned int *) &tenths) ;
```
dragonmux, line 415: Rather than casting here, use PRIu32 - as in `"%" PRIu32 ".%" PRIu32` - it's undefined behaviour and quite illegal to do the cast as, depending on platform, you'll end up with only some of the bytes in units and tenths written and which ones and what that means will depend on endianess.

dragonmux (changes requested):
This generally looks good, overlooking formatting. The formatting of this code will get fixed when the feature/clang-tidy-format branch is merged and the big reformatting done to the repository.

dragonmux (approved):
LGTM, merging.

PR #1056, opened 2022-06-16
Improve reset terminology usage

dragonmux:
This PR attempts to do a couple of things, but one thing is does not is change anything terribly fundamental in the code base (despite how severe the change set looks!)

In short, this PR attempts to do the following:

* Bring consistency between board physical pin-out naming and their platform headers and implementations
* Bring consistency into our use of 'TRST' so it's always used for the same thing in all places and with the same consistent name
* Bring consistency into our use of 'SRST' vs 'nRST' as the former only appears in ADIv5 to mean 'System Reset', but this conflicts with the terminology used in architectures being introduced in firmware version 2.0

What this PR guarantees in the change:

* Absolute consistency between platform README's, their physical pinouts, and how we talk about and use them in code
* A lack of ABI break as while we have renamed a lot of stuff, we kept the remote ABI unchanged

The other complication this addresses is our 'platform_srst_{get,set}_val' which, to say it lightly, causes some intense cognitive dissonance in the AVR and RISC-V support code as on these architectures there is Software Reset which is done over JTAG, JTAG reset done through nTRST, and hardware reset done through nRST. By fixing this bit of naming, it solves a lot of readability problems for these future architectures.

A couple of bug fixes have crept in due to reformatting we performed during the refactor, if it's wished that these be in a separate PR, we will move them into their own branch. Likewise with the more major reformatting items.


esden:
Thank you @dragonmux! That is a great patch that will make it easier to talk about the new platforms in the near future. Great step forward for maintainability.

That said, maybe we should add to this patch a document that describes the conventions you used here so that they can be easily referenced? Maybe in a `target/HACKING.md` document? This can reference the fact that ADIv5 uses a different terminology, and describe what our terminology is and how it is supposed to be used moving forward.

esden (approved):
Looks great! Thank you very much for smoothing this wrinkle out. Now up up to the new architecture skies and beyond. :D

PR #1057, opened 2022-06-16
Feature: auto scan

dragonmux:
This PR implements the `auto_scan` monitor command and BMDA command line option.

This makes it so we have JTAG scanning, SWD scanning (and may the two never meet when we can't find devices), and now automatic scanning which first tries JTAG and if that fails, tries SWD.

The way it has been written provides an easy way to extend this when the time comes to PDI scanning and Flash scanning as the approach is chosen for extensibility over a certain allure of simplicity.

Tested on BMP. Both the monitor command and the CLI option.


esden (approved):
LGTM

PR #1058, opened 2022-06-16
RP2040: Better detection of flash size

jamesturton:
Following on from the discussions in #1045, this PR aims to provide a better way to identify the size of the SPI flash connected to the RP2040. It does this in two steps:

First, look for repeating sectors in the flash region. This was the original approach that was used and should work if a valid program is loaded in memory, but it will fail if the flash memory is blank, or written with other uniform data. The idea is that if you attempt to read an address outside of the valid flash sector, you will get the data from the start of the sector. 

The second attempt is a bit more aggressive: trying to reading the JEDEC ID of the flash chip. This will interrupt the flash XIP and revert back to SPI mode, but we assume that there is no valid program stored in the memory, otherwise we should have been able to detect it in the previous method. Once we have the JEDEC ID, we can get the log base 2 size of the flash chip from it's second byte.

### DIFF
--- a/src/target/rp.c
+++ b/src/target/rp.c
@@ -33,7 +33,7 @@
  */
 
 /* This file implements Raspberry Pico (RP2040) target specific functions
- * for detecting the device, providing the XML memory map but not yet
+ * for detecting the device, providing the XML memory map and
  * Flash memory programming.
  */
 
@@ -61,7 +61,7 @@
  * original Pico board from Raspberry Pi.
  * https://www.winbond.com/resource-files/w25q16jv%20spi%20revd%2008122016.pdf
  * All dev boards supported by Pico SDK V1.3.1 use SPI flash chips which support
- * these commands. Other customs boards using different SPI flash chips might
+ * these commands. Other custom boards using different SPI flash chips might
  * not support these commands
  */ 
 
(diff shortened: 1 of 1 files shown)


src/target/rp.c
```diff
@@ -385,10 +389,117 @@ static void rp_add_flash(target *t, uint32_t addr, size_t length)
...
+}
+
+uint32_t rp_read_flash_chip(target *t, uint32_t cmd)
+{
+	uint32_t value = 0;
+
+	rp_ssel_active(t, true);
+
+	/* write command into SPI peripheral's FIFO */
+	for (int count = 0; (count < 4); count++)
```
dragonmux, line 414: no need for the extra () here as the comparison binds tighter than the ;'s do. Likewise size_t rather than int - no point having this be signed.

src/target/rp.c
```diff
@@ -385,10 +389,117 @@ static void rp_add_flash(target *t, uint32_t addr, size_t length)
...
+	rp_ssel_active(t, false);
+
+	return value;
+}
+
+uint32_t rp_get_flash_length(target *t)
+{
+	uint32_t size = MAX_FLASH;
+	uint32_t bootsec[16];
+	int i;
```
dragonmux, line 433: size_t as no point having this be signed

dragonmux (changes requested):
This is looking great - just a couple of things highlighted in the review. Once those are addressed and we've tested this on real hardware, we'll merge this ASAP. Good job!

jamesturton:
Thanks for the review @dragonmux! Hopefully I've addressed everything you mentioned above

jamesturton (commented):
Resolved in 50e43e7

dragonmux (approved):
LGTM, we'll build and test this on our local RP2040 hardware and merge if it all works properly

dragonmux:
We can now confirm this code works. Merging

PR #1061, opened 2022-06-22
Feature: CH32F1 cleanup and correctness improvements

dragonmux:
A user reported that all their STM32 targets were no longer found when doing SWD scan (reported via Discord). This lead us to looking at the CH32F1 code which we spotted was breaking `t->idcode` state and needed some improving.

This PR fixes this regression in v1.8 caused by the CH32F1 probe code.

This should be merged into both master and the v1.8 branch.


esden (approved):
LGTM

PR #1066, opened 2022-06-25
Fix a formatting bug that breaks, among other things, breakpoints

### LINKED ISSUE #1054: Breakpoints no longer work on ESP32 host
When using an ESP32 host, breakpoints no longer work. Single-stepping and code inspection work, however breakpoints in general never get hit.

Reverting 488df7e6f73c8cf53483a4e856c4aad51fda6330 fixes the issue.

dragonmux:
This PR fixes a small regression that made it into v1.8 which breaks the ability to set and hit breakpoints. This fixes #1054.

This should be merged into both master and v1.8

### DIFF
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -602,7 +602,7 @@ handle_z_packet(char *packet, int plen)
 	uint32_t addr;
 	int ret;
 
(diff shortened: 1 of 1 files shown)


esden (approved):
LGTM

PR #1068, opened 2022-06-25
Serial number regression fix

dragonmux:
Fixed a small regression that made it in with #1041 resulting in a transposition of alpha characters in serial numbers. This should restore the serial number generation behaviour we had.

Many thanks to @Qyriad for pointing it out to us with a clear before and after example to lead the way

### DIFF
--- a/src/platforms/stm32/serialno.c
+++ b/src/platforms/stm32/serialno.c
@@ -47,13 +47,14 @@ char *serial_no_read(char *s)
 	for (size_t i = 0; i < 24U; ++i) {
 		const size_t chunk = i >> 3U;
 		const size_t nibble = i & 7U;
+		const size_t idx = (chunk << 3U) + (7U - nibble);
 		if (nibble == 0)
 			unique_id = unique_id_p[chunk];
-		s[chunk + (7U - nibble)] = ((unique_id >> (i * 4)) & 0x0F) + '0';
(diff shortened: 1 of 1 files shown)


Qyriad (dismissed):
Seems to fix the issue on our hardware

dragonmux:
Uwe, simply undoing the code changes like that isn't acceptable. It was changed for a good reason and improved Flash usage. Please fix the actual bug and don't just revert a change like that.

UweBonnes:
Well, the original code worked but does not compile on gcc11. Several attempt to rewrite it have been done, but as written in the commit message result in a serial all 0:
 usb 2-1.1.3: Product: Black Magic Probe (STLINK/V2) v1.8.0-85-ge97fc5e5
 usb 2-1.1.3: Manufacturer: Black Magic Debug
 usb 2-1.1.3: SerialNumber: 0000000000
This working code is 20 bytes larger than the no-working code before. Are we really in a need to sink much time to save eventually  20 bytes?

dragonmux:
With regards to space savings, yes.. yes we are as the RISC-V and AVR code pushes us right to the limits of what we can fit in Flash, so we need to claw back space

esden (approved):
LGTM, I have compile tested it on GCC7, 10 and 12. It all compiles nicely (including the 24char serial platforms). And fixes the serial string instability.

PR #1069, opened 2022-06-26
CMSIS-DAP support type clairty

dragonmux:
Fixed the naming of the cmsis_type_e enum as it was inconsistent with the naming scheme we use.


esden (approved):
LGTM

PR #1070, opened 2022-06-26
Fix a couple of bugs in Cortex-M watchpoint and register handling

aksdfauytv:
I fixed some terrible bugs. You're welcome.

### DIFF
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -506,7 +506,7 @@ bool cortexm_attach(target *t)
 	priv->flash_patch_revision = (r >> 28);
 	priv->hw_watchpoint_max = CORTEXM_MAX_WATCHPOINTS;
 	r = target_mem_read32(t, CORTEXM_DWT_CTRL);
-	if ((r >> 28) > priv->hw_watchpoint_max)
+	if ((r >> 28) < priv->hw_watchpoint_max)
 		priv->hw_watchpoint_max = r >> 28;
 
 	/* Clear any stale breakpoints */
@@ -679,7 +679,7 @@ static int dcrsr_regnum(target *t, unsigned reg)
 		return regnum_cortex_m[reg];
 	} else if ((t->target_options & TOPT_FLAVOUR_V7MF) &&
 	           (reg < (sizeof(regnum_cortex_m) +
-	                   sizeof(regnum_cortex_mf) / 4))) {
+	                   sizeof(regnum_cortex_mf)) / 4)) {
 		return regnum_cortex_mf[reg - sizeof(regnum_cortex_m)/4];
 	} else {
 		return -1;


src/target/adiv5.c
```diff
@@ -351,8 +351,10 @@ static uint32_t cortexm_initial_halt(ADIv5_AP_t *ap)
 			} else {
 				trncnt = 0xfff;
 			}
-			dhcsr = adiv5_dp_low_access(
+			adiv5_dp_low_access(
```
UweBonnes, line 354: It is no a bug, it is a feature. When running the while loop for the second time, we get the real data. If the device is halted in the first loop, your approach with 4 transactions is better. If the device is hard to catch, like a STM32F7 mostly in WFI, the original loop with 3 transactiond gives less dead time where we do not hammer on DHCSR. The shrt loop is good tested annd allows me to hot plug to a STM32F7 mostly in sleep where most other debugger do not catch the device and require cold plug(connect under reset)

UweBonnes (commented):
LGTM. Probably we have never seen a device with less than 4 watchpoints and so we did not hit that bug.

I propose to add a warning if we ever meet a device with more watchpoints, like

if ((r >> 28) > priv->hw_watchpoint_max) {
DEBUG_WARN("Device with %d watchpoints found. Extend CORTEXM_MAX_BREAKPOINTS\n"), r >> 28);
}

UweBonnes (dismissed):
Good catch. LGTM.

aksdfauytv:
> LGTM. Probably we have never seen a device with less than 4 watchpoints and so we did not hit that bug.
> 
> I propose to add a warning if we ever meet a device with more watchpoints, like
> 
> if ((r >> 28) > priv->hw_watchpoint_max) { DEBUG_WARN("Device with %d watchpoints found. Extend CORTEXM_MAX_BREAKPOINTS\n"), r >> 28); }

Actually, the bigger problem is if the target has more than 4 breakpoints, then memory gets overwritten.

dragonmux (changes requested):
Thank you for your contribution. However, please edit the description properly explain what you've fixed and why (there may be extremely good reasons for the code being how it was - we can't tell that with so little information for review).

Please revert the change in the ADIv5 that Uwe highlighted as that wasn't a bug.

We will test the rest of the changes before approving for merge as they deeply change how important features work.

src/target/adiv5.c
dragonmux: This makes the STM32 detection problems we've been having even worse because of the rapid turn-arounds. Not checking the result of advi5_dp_low_access the first time is a bug and additional problems waiting to happen.

UweBonnes: Where did you see problems? When? How to reproduce?
Rapid turnaround is no problem here. It is needed to have a high probability to catch the target in a small moment of awareness. We repeat the loop as long as the result makes no sense. We check that it is not all ones or zeros. This code exposes other problem,  see #1071, but that is no fault of this code.

dragonmux: In response to a user on Discord who cannot see any of their STM32 targets after going from v1.7.1 -> v1.8 and with our own Nucleo hardware.

UweBonnes: I have tested with BMP native, StlinkV2/3, stlinkv2/3 firmware and f072. No problem seen here. Please show "blackmagic -t" on an affected board!

src/target/cortexm.c
```diff
@@ -679,7 +679,7 @@ static int dcrsr_regnum(target *t, unsigned reg)
 		return regnum_cortex_m[reg];
 	} else if ((t->target_options & TOPT_FLAVOUR_V7MF) &&
 	           (reg < (sizeof(regnum_cortex_m) +
-	                   sizeof(regnum_cortex_mf) / 4))) {
+	                   sizeof(regnum_cortex_mf)) / 4)) {
```
dragonmux, line 682: From what this looks like, we were previously generating the wrong number of registers in this expression because only the last sizeof() was being divided by 4 but they all should have been - is this a correct assessment?

src/target/cortexm.c
```diff
@@ -506,7 +506,7 @@ bool cortexm_attach(target *t)
 	priv->flash_patch_revision = (r >> 28);
 	priv->hw_watchpoint_max = CORTEXM_MAX_WATCHPOINTS;
 	r = target_mem_read32(t, CORTEXM_DWT_CTRL);
-	if ((r >> 28) > priv->hw_watchpoint_max)
+	if ((r >> 28) < priv->hw_watchpoint_max)
```
dragonmux, line 509: Please describe the impacts of this change as that seems pretty radical - a before-and-after of what `priv->hw_watchpoint_max` looks like would be very helpful just to get a feel for this.

dragonmux (commented):
Looks good - but to emphasise, we need some explanation here as to why what's been changed has been changed before we approve.

src/target/adiv5.c
dragonmux: Now there's part of the rub.. BMDA does not reproduce the issues because its turnarounds are slowed down sufficiently by the USB stack of the host. It only repros on real hardware and at that only on certain choices and combinations. This isn't the place to talk about this as it's muddying the contributor's PR and it's something Esden and us are already tracking and working on.

src/target/cortexm.c
aksdfauytv: This condition is just a sanity check, but yes, it had a typo that made it too relaxed.

src/target/cortexm.c
aksdfauytv: There were 2 issues with this bug. For targets with less than 4 watchpoints, it made GDB think there were actually 4 watchpoints available. For targets with more than 4 watchpoints, it caused memory corruption in BMP.
hw_watchpoints is a static array with size 4. This condition was meant to clamp the number of watchpoints to that limit, but due to the typo it didn't do that.

dragonmux: Ok, that makes this seem much more reasonable. Thank you for the explanation!

dragonmux (commented):
This looks basically ready for merge now - our only final request is, could you please merge the two commits down into one, and provide a more meaningful commit message such as `cortexm: fixed an issue with watchpoint handling and a register sanity check`

src/target/cortexm.c
UweBonnes: As all CortexM has 4 watchpoints, we never hit the issue. But your finding is right. For the FPU registers, i guess few users looked at the FPU registers in the debugger and so noone noticed, but your fix is right too.

dragonmux (approved):
LGTM! Thank you again for the contribution. Merging.

dragonmux:
@esden it would be useful for these fixes to be merged into v1.8 too

PR #1072, opened 2022-06-26
stlinkv2: Treat a fault after wait as an error

### LINKED ISSUE #1071: Random STLINKV3 problems when halting the device
> ./blackmagic -t -C
Black Magic Debug App v1.8.0-84-g80c98df2
 for Black Magic Probe, ST-Link v2 and v3, CMSIS-DAP, JLink and libftdi/MPSSE
Using 0483:374e 003800134741500120383733 STMicroelectronics
 STLINK-V3
STLink firmware version: V3J7M2B0S0
...
AP   0: IDR=24770011 CFG=00000000 BASE=e00ff003 CSW=23000040 (AHB-AP var1 rev2
write_retry failed. stlink_read_dp_register error -2

This is caused by unexpected STLINK_SWD_AP_FAULT results from the STLink where STLINK_ERROR_WAIT is expected.
Can be fixed handling STLINK_SWD_AP_FAULT  as  STLINK_ERROR_WAIT in that case.

UweBonnes:
cortexm_initial_halt() repeats the DHCSR write with high values for TRNCNT in
CSW. This is needed to catch a STM32F7 mostly in WFI. While the repeated write
is running, STLINKV3 on a Nucleo-WL55 (V3J7M2B0S0) answers first with
STLINK_SWD_AP_WAIT and on more read if write command is still running with
STLINK_SWD_AP_FAULT. At some point when the last command is finished, normal
STLINK_ERROR_OK indicates finally successful read. Treat STLINK_SWD_AP_FAULT
after STLINK_ERROR_WAIT as STLINK_ERROR_WAIT in that case. STLINK_SWD_AP_FAULT
may still be issued on other invalid accesses and should still be treated as
error in the other possible cases. Fixes #1071.

### DIFF
--- a/src/platforms/hosted/stlinkv2.c
+++ b/src/platforms/hosted/stlinkv2.c
@@ -211,6 +211,7 @@ static int stlink_usb_get_rw_status(bool verbose);
 int debug_level = 0;
 
 #define STLINK_ERROR_DP_FAULT -2
+#define STLINK_ERROR_AP_FAULT -3
 
 /**
     Converts an STLINK status code held in the first byte of a response to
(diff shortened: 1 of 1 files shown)


dragonmux (approved):
What this changes and how looks reasonable, even if the explanation of what's wrong and how this fixes it could really use some work.

Merging as CI succeeded and we don't have hardware to test on but see you used a STM32F7

PR #1073, opened 2022-06-26
Add monitor option to skip device probe

UweBonnes:
Some vendors do not provide a valid hint for their device in the ROM table.
To detect those devices, random values of the device are probed. This may
result in hickup on unrelated devices and defeat later debugging.
Provide a valid empty RAM region to suppress GDB warning.

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -53,6 +53,7 @@ static bool cmd_morse(target *t, int argc, char **argv);
 static bool cmd_halt_timeout(target *t, int argc, const char **argv);
 static bool cmd_connect_srst(target *t, int argc, const char **argv);
 static bool cmd_hard_srst(target *t, int argc, const char **argv);
+static bool cmd_probe(target *t, int argc, const char **argv);
 #ifdef PLATFORM_HAS_POWER_SWITCH
 static bool cmd_target_power(target *t, int argc, const char **argv);
 #endif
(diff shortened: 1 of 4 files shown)


src/command.c
```diff
@@ -72,6 +73,7 @@ const struct command_s cmd_list[] = {
 	{"help", (cmd_handler)cmd_help, "Display help for monitor commands"},
 	{"jtag_scan", (cmd_handler)cmd_jtag_scan, "Scan JTAG chain for devices" },
 	{"swdp_scan", (cmd_handler)cmd_swdp_scan, "Scan SW-DP for devices" },
+	{"probe", (cmd_handler)cmd_probe, "Probe for devices (enable|disable)" },
```
dragonmux, line 76: This command - does it actually turn probing on and off? It looks like no..?
Also, it appears to only be for Cortex-M, so please word it as such.

Confusing wording for what this actually does

src/include/command.h
```diff
@@ -25,6 +25,8 @@
 
 #include "target.h"
 
+extern bool do_probe;
```
dragonmux, line 28: Please name this more along the lines of `no_skip_cortexm_probe` - we know this is longer but it makes it clearer what this actually does.

src/platforms/pc/cl_utils.c
```diff
@@ -167,6 +167,7 @@ static void cl_help(char **argv)
 		"\t-R[h]           Reset the device. If followed by 'h', this will be done using\n"
 		"\t                  the hardware reset line instead of over the debug link\n"
 		"\t-H              Do not use the high level command API (bmp-remote)\n"
+		"\t-N              Do not probe. Usefull if probe causes hickup\n"
```
dragonmux, line 170: Same wording problem as the monitor command

src/target/cortexm.c
```diff
@@ -476,6 +480,7 @@ bool cortexm_probe(ADIv5_AP_t *ap)
 		PROBE(lpc17xx_probe);
 	}
 #undef PROBE
+	target_add_ram(t, 0x00000000, 0); /* Provide some memory map to gdb*/
```
dragonmux, line 483: This is unguarded - that feels very bad and we can foresee this causing problems

dragonmux (changes requested):

PR #1076, opened 2022-06-26
Fix uninitialized variables in target sam3x

sidprice:
Exposed while building hosted with -Og option


dragonmux (approved):
A quick dry-run through that code indicates this is a good fix to those warnings (and that they're very correct). Merging once we see the CI build complete.

PR #1078, opened 2022-06-27
Feature: clang-tidy and clang-format support

dragonmux:
This adds:

* support for clang-tidy, a script for parallel running it on the code base, and a Makefile target to trigger it
* support for clang-format, and a Makefile target to trigger it

For IDEs like VSCode, the clang-tidy extension by notskm should automatically pick the new configuration up and display linting inline. Likewise, the clang-format extension by xaver can be used to automatically enforce the formatting requirements this provides by enabling format-on-save (`"editor.formatOnSave": true` in settings.json).

In a future PR we will introduce a PR-time CI job that will validate code meets the formatting requirements and add a CI job that runs clang-tidy on merge to main to help with tracking progress on linting cleanup

PR #1079, opened 2022-06-27
VSCode extension recomendations

dragonmux:
This is a follow-up PR to #1078 which comes about from a suggestion by Esden on Discord that to help with clang-tidy and clang-format use and enforcement, the project should come with extension recommendations as appropriate to reduce the barriers to entry.

This adds recommendations for VSCode extensions to be used with the project


esden (approved):
LGTM

PR #1080, opened 2022-06-27
target/sam3x: saner uninitialized variable prevention

perigoso:
in #1076 we initialized variables to 0, but in the event that a new driver is added and not properly handled, we would try to do GPNVM bit operations on a flash peripheral with address 0x0, I think this is a saner approach, where if a driver is not known the command immediately fails

### DIFF
--- a/src/target/sam3x.c
+++ b/src/target/sam3x.c
@@ -33,7 +33,7 @@ static int sam3_flash_erase(struct target_flash *f, target_addr addr, size_t len
 static int sam_flash_write(struct target_flash *f, target_addr dest,
                              const void *src, size_t len);
 
-static uint32_t sam_gpnvm_get(target *t, uint32_t base);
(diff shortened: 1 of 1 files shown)


dragonmux (approved):
LGTM. We like the approach though having the numbers initialised to something all the same would make us more comfortable as a preventative measure to ever accessing them uninitalised.

We'll merge this as is, but perhaps we can find values to set base and gpnvm_mask to that are clearly bad/wrong but which don't make this code do whacky stuff if ever triggered. Thinking about debug as much as anything.

perigoso:
wait on the merge then, let me take another look

perigoso:
added an additional check in the flash cmd for a non 0 eefc addr

dragonmux (approved):
Looks much better - happy to merge this once the build finishes as you've probably tested it too.

perigoso:
ah bummer, i made one commit with the work email by mistake, oh well

PR #1081, opened 2022-06-27
STM32F1 general cleanup and GD32E230 option write special case

### LINKED ISSUE #921: GD32E230 Option byte write needs to be 32bit sized
I've stumbled across an issue that prevented me from resetting the flash protection option bytes. I've traced the issue, and it appears that target_mem_write16 does not work on GD32E230 to write option bytes.
Quick fix below, but not sure how it would affect other targets.

```
static bool stm32f1_option_write_erased(target *t, uint32_t addr, uint16_t value)
{
	if (value == 0xffff)
		return true;
	/* Erase option bytes instruction */
	target_mem_write32(t, FLASH_CR, FLASH_CR_OPTPG | FLASH_CR_OPTWRE);
	target_mem_write32(t, addr, 0xFFFF0000 | value); // <-- change here
	/* Read FLASH_SR to poll for BSY bit */
	while (target_mem_read32(t, FLASH_SR) & FLASH_SR_BSY)
		if(target_check_error(t))
			return false;

	return true;
}
```

perigoso:
trying to do something about #921 
I am not able to physically test this as I don't have any GD32E230 devices

@walmis do you think you could do some testing for me?

fixes #921

### DIFF
--- a/src/target/stm32f1.c
+++ b/src/target/stm32f1.c
@@ -277,7 +277,7 @@ static int stm32f1_flash_erase(struct target_flash *f, target_addr addr, size_t 
 	target_addr end = addr + len - 1;
 	target_addr start = addr;
 
-	if ((t->idcode == 0x430) && (end >= FLASH_BANK_SPLIT))
+	if (t->idcode == 0x430 && end >= FLASH_BANK_SPLIT)
 		if (stm32f1_flash_unlock(t, FLASH_BANK2_OFFSET))
 			return -1;
 
@@ -321,7 +321,7 @@ static int stm32f1_flash_erase(struct target_flash *f, target_addr addr, size_t 
 			return -1;
 		}
 	}
-	if ((t->idcode == 0x430) && (end >= FLASH_BANK_SPLIT)) {
+	if (t->idcode == 0x430 && end >= FLASH_BANK_SPLIT) {
 		uint32_t sr = target_mem_read32(t, FLASH_SR + FLASH_BANK2_OFFSET);
 		if ((sr & SR_ERROR_MASK) || !(sr & SR_EOP)) {
 			DEBUG_INFO("stm32f1 bank 2 flash erase error 0x%" PRIx32 "\n", sr);
@@ -366,7 +366,7 @@ static int stm32f1_flash_write(struct target_flash *f, target_addr dest, const v
 	}
 
 	length = len - length;
-	if ((t->idcode == 0x430) && length) { /* Write on bank 2 */
+	if (t->idcode == 0x430 && length) { /* Write on bank 2 */
 		target_mem_write32(t, FLASH_CR + FLASH_BANK2_OFFSET, FLASH_CR_PG);
 		cortexm_mem_write_sized(t, dest, src, length, ALIGN_HALFWORD);
 		/* Read FLASH_SR to poll for BSY bit */
@@ -465,28 +465,34 @@ static bool stm32f1_option_write_erased(target *t, uint32_t addr, uint16_t value
 static bool stm32f1_option_write(target *t, uint32_t addr, uint16_t value)
 {
 	uint16_t opt_val[8];
-	int i, index;
+	int index;
 
 	index = (addr - FLASH_OBP_RDP) / 2;
-	if ((index < 0) || (index > 7))
+	if (index < 0 || index > 7)
 		return false;
+
 	/* Retrieve old values */
-	for (i = 0; i < 16; i = i + 4) {
+	for (size_t i = 0; i < 16; i += 4) {
 		uint32_t val = target_mem_read32(t, FLASH_OBP_RDP + i);
 		opt_val[i / 2] = val & 0xffff;
 		opt_val[i / 2 + 1] = val >> 16;
 	}
+
 	if (opt_val[index] == value)
 		return true;
+
 	/* Check for erased value */
 	if (opt_val[index] != 0xffff)
-		if (!(stm32f1_option_erase(t)))
+		if (!stm32f1_option_erase(t))
 			return false;
+
 	opt_val[index] = value;
+
 	/* Write changed values*/
-	for (i = 0; i < 8; i++)
-		if (!(stm32f1_option_write_erased(t, FLASH_OBP_RDP + i * 2, opt_val[i])))
+	for (size_t i = 0; i < 8; i++)
+		if (!stm32f1_option_write_erased(t, FLASH_OBP_RDP + i * 2, opt_val[i]))
 			return false;
+
 	return true;
 }
 
@@ -519,7 +525,7 @@ static bool stm32f1_cmd_option(target *t, int argc, const char **argv)
 	target_mem_write32(t, FLASH_OPTKEYR, KEY1);
 	target_mem_write32(t, FLASH_OPTKEYR, KEY2);
 
-	if ((argc == 2) && !strcmp(argv[1], "erase")) {
+	if (argc == 2 && !strcmp(argv[1], "erase")) {
 		stm32f1_option_erase(t);
 		stm32f1_option_write_erased(t, FLASH_OBP_RDP, flash_obp_rdp_key);
 	} else if (rdprt) {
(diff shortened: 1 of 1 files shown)


src/target/stm32f1.c
```diff
@@ -486,9 +499,14 @@ static bool stm32f1_cmd_option(target *t, int argc, const char **argv)
 	target_mem_write32(t, FLASH_OPTKEYR, KEY1);
 	target_mem_write32(t, FLASH_OPTKEYR, KEY2);
 
-	if ((argc == 2) && !strcmp(argv[1], "erase")) {
+	if ((argc == 2) && strcmp(argv[1], "erase") == 0) {
```
dragonmux, line 502: the ()'s around the argc check are redundant and suppress warnings about accidental assignment, please remove them as part of this change

src/target/stm32f1.c
```diff
@@ -486,9 +499,14 @@ static bool stm32f1_cmd_option(target *t, int argc, const char **argv)
...
 		stm32f1_option_erase(t);
-		stm32f1_option_write_erased(t, FLASH_OBP_RDP, flash_obp_rdp_key);
+		bool width_word = false;
+		if (strcmp(t->driver, "GD32E230") == 0) {
```
dragonmux, line 505: If possible, can we detect this based on, eg, t->idcode? This works but we'd like a cheaper option that uses a bit less Flash and time

src/target/stm32f1.c
```diff
@@ -448,16 +453,24 @@ static bool stm32f1_option_write(target *t, uint32_t addr, uint16_t value)
...
+	bool width_word = false;
+	if (strcmp(t->driver, "GD32E230") == 0) {
+		/* GD32E230 special case, target_mem_write16 does not work */
+		width_word = true;
+	}
+
 	for (i = 0; i < 8; i++)
-		if (!(stm32f1_option_write_erased
-			(t, FLASH_OBP_RDP + i*2,opt_val[i])))
+		if (!(stm32f1_option_write_erased(t, FLASH_OBP_RDP + i * 2, opt_val[i], width_word)))
```
dragonmux, line 471: There's an extra pair of redundant brackets here, please remove them

src/target/stm32f1.c
```diff
@@ -448,16 +453,24 @@ static bool stm32f1_option_write(target *t, uint32_t addr, uint16_t value)
...
+
 	/* Check for erased value */
 	if (opt_val[index] != 0xffff)
 		if (!(stm32f1_option_erase(t)))
 			return false;
 	opt_val[index] = value;
+
 	/* Write changed values*/
+	bool width_word = false;
+	if (strcmp(t->driver, "GD32E230") == 0) {
```
dragonmux, line 465: This is covered by the note below about strcmp() vs testing t->idcode or similar

src/target/stm32f1.c
```diff
@@ -418,13 +418,17 @@ static bool stm32f1_option_erase(target *t)
...
 
-static bool stm32f1_option_write_erased(target *t, uint32_t addr, uint16_t value)
+static bool stm32f1_option_write_erased(target *t, uint32_t addr, uint16_t value, bool width_word)
 {
 	if (value == 0xffff)
 		return true;
 	/* Erase option bytes instruction */
 	target_mem_write32(t, FLASH_CR, FLASH_CR_OPTPG | FLASH_CR_OPTWRE);
-	target_mem_write16(t, addr, value);
+	if (width_word) {
```
dragonmux, line 427: the {}'s in this are redundant and not part of the code style. Please remove them

src/target/stm32f1.c
```diff
@@ -440,6 +444,7 @@ static bool stm32f1_option_write(target *t, uint32_t addr, uint16_t value)
 	index = (addr - FLASH_OBP_RDP) / 2;
 	if ((index < 0) || (index > 7))
```
dragonmux, line 445: the ()'s around the individual sub-clauses here are redundant and suppress warnings, please remove them

src/target/stm32f1.c
```diff
@@ -440,6 +444,7 @@ static bool stm32f1_option_write(target *t, uint32_t addr, uint16_t value)
...
 		 return false;
+
 	/* Retrieve old values */
 	for (i = 0; i < 16; i = i +4) {
```
dragonmux, line 449: This feels like i is defined elsewhere and whoever wrote this doesn't know about += - please rewrite this using size_t for i and `i += 4`

src/target/stm32f1.c
```diff
@@ -448,16 +453,24 @@ static bool stm32f1_option_write(target *t, uint32_t addr, uint16_t value)
...
 	opt_val[index] = value;
+
 	/* Write changed values*/
+	bool width_word = false;
+	if (strcmp(t->driver, "GD32E230") == 0) {
+		/* GD32E230 special case, target_mem_write16 does not work */
+		width_word = true;
+	}
+
 	for (i = 0; i < 8; i++)
```
dragonmux, line 470: same comment as above about i defined elsewhere and size_t

dragonmux (changes requested):
This is looking good, though we'd like to see if a cheaper alternative to the strcmp() of `t->driver` is available and, as you're touching this code, clean it up to make it more conformant to the code base style, removing warning-suppressing brackets, etc. (Cleanup should probably be a second commit)

If you don't wish to clean up the code, then no worries, it can be done as a separate pass now we know it needs work.

perigoso:
@dragonmux  theres a code snippet weirding me out 

```
if (0 && flash_obp_rdp_key == FLASH_OBP_RDP_KEY_F3) {
    /* Reload option bytes on F0 and F3*/
    val = adiv5_ap_mem_read(ap, FLASH_CR);
    val |= FLASH_CR_OBL_LAUNCH;
    stm32f1_option_write(t, FLASH_CR, val);
    val &= ~FLASH_CR_OBL_LAUNCH;
    stm32f1_option_write(t, FLASH_CR, val);
}
```
this code block is never executed, I can only assume it's intentional, but i would like to know why, it was added by @UweBonnes  in b1a942aca3e80aaaa8405d313424f1d6f9273bd9

dragonmux:
That's a really fantastic question.. given how old that commit is.. we're going to say just nix it. If we find we need the code we can always pull it back out the Git history.

perigoso:
this should be close to or mergeable, waiting on feedback
there's a couple of commits by now, I can squash some, but I think they're reasonably organized

perigoso:
ready for review

src/target/stm32f1.c
```diff
@@ -253,83 +265,91 @@ static int stm32f1_flash_unlock(target *t, uint32_t bank_offset)
...
 			len = 0;
+
 		addr += f->blocksize;
 	}
 
 	/* Check for error */
 	if (start < FLASH_BANK_SPLIT) {
 		uint32_t sr = target_mem_read32(t, FLASH_SR);
-		if ((sr & SR_ERROR_MASK) || !(sr & SR_EOP)) {
+		if (sr & SR_ERROR_MASK || !(sr & SR_EOP)) {
```
dragonmux, line 317: This pair of ()'s is required as & binds looser than ||

src/target/stm32f1.c
```diff
@@ -253,83 +265,91 @@ static int stm32f1_flash_unlock(target *t, uint32_t bank_offset)
...
 			DEBUG_INFO("stm32f1 flash erase error 0x%" PRIx32 "\n", sr);
 			return -1;
 		}
 	}
-	if ((t->idcode == 0x430) && (end >= FLASH_BANK_SPLIT)) {
+	if (t->idcode == 0x430 && end >= FLASH_BANK_SPLIT) {
 		uint32_t sr = target_mem_read32(t, FLASH_SR + FLASH_BANK2_OFFSET);
-		if ((sr & SR_ERROR_MASK) || !(sr & SR_EOP)) {
+		if (sr & SR_ERROR_MASK || !(sr & SR_EOP)) {
```
dragonmux, line 324: This pair of ()'s is required as & binds looser than ||

dragonmux (changes requested):
This introduces a couple of accidental bugs outlined below, but otherwise looks ready for merge. Good work perigoso!

PR #407, opened 2018-11-14
Better ST-Link support

mbrunnen:
Add LED pin and a RX pin for USB-UART on embedded ST-Links.

Tested on Nucleo-144.

PR #1083, opened 2022-06-28
ST-Link: Fix the LED pinout on the Nucleo boards

dragonmux:
This PR is a rehash of #407 which has unfortunately bitrot despite being a valuable contribution.

While the original PR's RX pin addition is no longer needed (it was done in another PR), the LED pinout fix it offers is still valid and useful.


esden (approved):
LGTM

PR #1084, opened 2022-06-28
Fix 24 digit serial numbers

dragonmux:
There was still a small issue with the current 24 digit serial number code, which this fixes.

We also ran clang-format on the file, bringing it fully in line with the required style and formatting.

Before: 0670FF530000000000000000
 After: 0670FF535567494867085055


esden (approved):
LGTM

PR #1086, opened 2022-06-28
Implement semihosting debug output redirection to usb-uart port.

dragonmux:
This PR replaces #326 which appears to have become abandoned. This aims to implement stdout redirection when semihosting is in use. As given in the original PR:

Some IDEs such as Eclipse do not show the semihosting output via GDB. In addition, it is sometimes more comfortable to use a serial port terminal program (like when debugging to UART directly but for the boards where UART is not available). For such use cases, it would be useful to redirect the semihost stdout and stderr streams to the already existing USB UART (which is probably unused when semihosting is used).

This patch adds a command - `redirect_stdout`. By running `monitor redirect_stdout enable`, the semihosting stdout and stderr are sent via the USB UART.


esden (approved):
LGTM, also tested on BMP23+1Bitsy.

PR #396, opened 2018-10-01
Add option to turn SWIM/RST into UART RX/TX

silkeh:
This PR allows building for `stlink` with `SWIM_AS_UART=1`, which enables the use of the
SWIM and RST ports as RX and TX on cheap ST-Link V2 clones.

This is done by using USART1 (on the alternative port) instead of USART2.

PR #1087, opened 2022-06-29
Feature: ST-Link SWIM-UART switchability

dragonmux:
This PR replaces #396 which appears to have become abandoned.

This enables using the stlink platform with ST-Link clones that have no UART interface, only SWIM. By specifying `SWIM_AS_UART=1` when building, it repurposes the SWIM_RST and SWIM pins as UART TX and RX respectively.

### DIFF
--- a/src/platforms/stlink/Makefile.inc
+++ b/src/platforms/stlink/Makefile.inc
@@ -25,6 +25,10 @@ else
 LDFLAGS += --specs=nosys.specs
 endif
 
+ifeq ($(SWIM_AS_UART), 1)
+CFLAGS += -DSWIM_AS_UART=1
+endif
+
 VPATH += platforms/stm32
 
 SRC += 	cdcacm.c	\
@@ -47,4 +51,3 @@ blackmagic_dfu.elf: usbdfu.o dfucore.o dfu_f1.o stlink_common.o serialno.o
 
 host_clean:
 	-$(Q)$(RM) *.bin *elf *hex
-
--- a/src/platforms/stlink/README.md
+++ b/src/platforms/stlink/README.md
@@ -1,6 +1,7 @@
 # Blackmagic for ST-Link Adapters
 
-ST-Link v3, ST-Link v2-1 and v2 with original, recent ST firmware can can use the hosted branch, running the GDB server on PC.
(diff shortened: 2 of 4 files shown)


esden (approved):
LGTM

positron96:
Hi all and thanks for this particular PR. 
I'm trying to figure out why UART TX does not work on my stlink clone (just flashed 1.10.2, compiled as SWIM_AS_UART and STLINK_FORCE_CLONE). The device only has a SWIM pin on the header, no SWIM_RST, and UART RX works well on SWIM pin. However, I see that this PR remaps USART1_TX to PB6, but PB6 is also defined as NRST pin for clones. Is TX even expected to work if the pin is also used as NRST in the code? 
I recompiled bmp with a different NRST pin, and TX started working for me fine, which suggests that NRST messes with UART remapping.

dragonmux:
We were unable to test this as we don't have any of the relevant hardware and this was part of an effort to clear the PR backlog - so it's entirely likely there are subtle things broken like this. We believe you've found a real bug and that given the pin usage issue mentioned, the nRST pin should be remapped when building SWIM_AS_UART. We'd welcome a PR that fixes this if you can work out what pin nRST is intended to be on your dongle, and remap it to that in this build mode.

positron96:
Hi @dragonmux, let me clarify. My device has only 4 gpio pins on the header, SWDIO, SWDCLK, RST and SWIM (like on the photo) 
image

Since I don't really use RST pin, I am quite happy to repurpose RST and SWIM as UART pins (that's what I did in code by changing NRST to a pin that's not on the header). I am not sure that is acceptable for other users. Or is this PR aimed at achieving exactly that? It is for stlink clones, and mentions SWIM_RST pin, but I cannot find it on any of the photos of other ST link clones (metal usb sticks), is it the same as RST pin? If it is, I think it would be nice to rename the compiler switch to SWIM_AND_RST_AS_UART to reduce confusion and to make it clear that RST functionality will be lost.

In any way, AFAIK this PR remaps UART TX to PB6, which is also a NRST pin for stlink clones, so pin configurations seem to clash. I can make a PR that moves RST pin to a pin not on the header of my hardware (e.g. to PB0, same as original stlink v2) in presence of this PR's build flag.

PR #1088, opened 2022-06-30
blackpillv2: fix PROBE_HOST in Readme.md

mniestroj:
After renaming directory with s/blackpill/blackpillv2/, Readme.md still
contained the old 'blackpill' name in PROBE_HOST. Fix that by replacing
it with 'blackpillv2'.

Fixes: 3ccb0af21c02 ("blackpillv2: Renamed blackpill to blackpillv2 to
  prevent ambiguity.")


dragonmux (approved):
Thank you for the contribution! Indeed, this documentation was wrong as a result of that change. Merging.

PR #1089, opened 2022-06-30
blackpillv2: clarify how to enter to bootloader mode

mniestroj:
According to [1] all PCB revisions have BOOT0 and NRST labels near
buttons. Reference those in README for ease of use.

PB2/BOOT1 has a 10k pull-down on PCB (in all revisions, according to
[1]), so there is no need to do anything on the board (like "jumpering
to GND") unless the board was modified by user. Hence don't reference
PB2/BOOT1 in the README, as it was just confusing to the user.

[1] https://github.com/WeActTC/MiniSTM32F4x1/tree/master/HDK

### DIFF
--- a/src/platforms/blackpillv2/Readme.md
+++ b/src/platforms/blackpillv2/Readme.md
@@ -36,7 +36,7 @@ make PROBE_HOST=blackpillv2
   * 2) Force the F4 into system bootloader mode by keeping BOOT0 button pressed while pressing and releasing NRST button. System bootloader should appear.
   * 3) `dfu-util -a 0 --dfuse-address 0x08000000 -D blackmagic.bin`
 
-To exit from dfu mode press a "key" and "reset", release reset. BMP firmware should appear
+To exit from dfu mode just press and release NRST. The newly Flashed BMP firmware should boot and enumerate.
 
 ## 10 pin male from pins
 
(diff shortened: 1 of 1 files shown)


src/platforms/blackpillv2/Readme.md
```diff
@@ -33,10 +33,10 @@ make PROBE_HOST=blackpill
 
 * After build:
   * 1) `apt install dfu-util`
-  * 2) Force the F4 into system bootloader mode by jumpering "BOOT0" to "3V3" and "PB2/BOOT1" to "GND" and reset (RESET button). System bootloader should appear.
+  * 2) Force the F4 into system bootloader mode by keeping BOOT0 button pressed while pressing and releasing NRST button. System bootloader should appear.
   * 3) `dfu-util -a 0 --dfuse-address 0x08000000 -D blackmagic.bin`
 
-To exit from dfu mode press a "key" and "reset", release reset. BMP firmware should appear
+To exit from dfu mode just press and release NRST. BMP firmware should appear.
```
dragonmux, line 39: This looks like it could benefit from a slight additional rewording - "BMP firmware should appear" => "The newly Flashed BMP firmware should boot and enumerate"

dragonmux (changes requested):
Thank you for the contribution! This looks good - we have one small thing we suggest additionally fixing, but this otherwise looks ready for merge.

src/platforms/blackpillv2/Readme.md
mniestroj: Updated, as suggested! Thanks for review!

dragonmux (approved):
Excellent. Merging

PR #423, opened 2018-12-28
Traceswo enhanced w fifo

anyn99:
I was finally able to improve the traceswo (Manchester encoding).
I had to add a fifo to allow async buffering, maybe it would be also useful in other parts of the project?

The output works very reliable now up to 115Khz SWO frequency here.

PR #1092, opened 2022-07-04
Feature: Improved Trace/SWO reliability

dragonmux:
This PR replaces #423 which appears to have become abandoned. This aims to improve Trace/SWO reliability and introduces the use of a FIFO to help prevent trace data getting dropped.

Per the original PR, the result is reliability up to a SWO frequency of 115kHz.

### DIFF
--- a/src/platforms/common/Makefile.inc
+++ b/src/platforms/common/Makefile.inc
@@ -24,6 +24,7 @@ CFLAGS += -Iplatforms/common
 SRC +=             \
 	jtagtap.c      \
 	swdptap.c      \
+	fifo.c         \
 	usb.c          \
 	usb_serial.c   \
 	usb_dfu_stub.c \
--- a/src/platforms/common/fifo.c
+++ b/src/platforms/common/fifo.c
@@ -1,0 +1,130 @@
+/*
+ * This file is part of the Black Magic Debug project.
+ *
+ * Copyright (C) 2022 1BitSquared <info@1bitsquared.com>
+ * Written by anyn99 <blnkdnl@googlemail.com>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <stdlib.h>
+#include <string.h>
+#include "fifo.h"
+
+bool fifo_push(fifo_t *const f, const uint8_t byte)
+{
+	if (f->is_full)
+		return false;
+
+	f->buffer[f->head++] = byte;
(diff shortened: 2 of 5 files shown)


mubes:
I have tested this patch. It reliably works at 100Khz;

```
$ time orbcat -c 0,"%c" | sort | uniq -c
      1 ABCDEFGHIJKLMNOPQRSTUV
2294987 ABCDEFGHIJKLMNOPQRSTUVWXYZ
      1 HIJKLMNOPQRSTUVWXYZ
$
real	177m19.772s
user	1m23.122s
sys	8m6.931s
```

This represents a user data rate of 11.65KBytes/sec.

However, If the speed is increased significantly beyond 100KHz then the comms link fails to report any data and it does not recover when the speed is dialled back down.  The only way to get SWO output again is by power cycling the probe. This will need to be fixed before the patch can be incorporated as otherwise it will lead to user confusion. An error should also be returned if the user tries to set the SWO speed higher than that which the probe can support.

Testing was against Orbuculum v2.00InProgress tag d3925fc. 

DAVE

dragonmux:
Ohey, thanks for testing this! What was the speed you could achieve before applying this patch, if you know?

Agreed regarding needing to properly handle trying to deal with >100kHz and automatically recovering. We'll look into that once we have our own test setup for Trace/SWO working

dragonmux:
As was noted in a review done yesterday, currently there is a problematic lack of license to the FIFO implementation added to src/platforms/common by this PR.

In order to take this forward from a draft, we need to find a resolution for this. @anyn99, as the original contributor, could you please advise us under what license you'd like this contribution made so we can make some progress and get this merged if possible?

anyn99:
HI there!

Sorry i left my code so unattended ;-)
I saw the request to comment and/or help on merging but didn't have time for it.
Happy to hear my code will make it into next release!

I think i would contribute this under GPLv3, if it is suitable for you, too?
What do you need from me, i am not so used to license stuff :-)

dragonmux:
Heya anyn99, thank you for the fast response. Now we know how you'd like it licensed, we can edit that into the commits in question as we've got your written agreement here.

Our only other question is how you'd like the attribution line to read - would `Written by anyn99 <your@git.email>` be sufficient for you?

anyn99:
Please add my attribution line like this:
`Written by anyn99 <blnkdnl@googlemail.com>`
And yes, i hereby release the copyright to the project.

Good luck with the blackmagic development and thanks for your effort to get my code in!
I hope i will contribute more one day in the future, see you then!

mubes:
100KHz is the maximum 'safe' limit to be claimed. It is working up to
102KHz but I wouldn't like to state that in formal documentation. The nice
thing about Manchester encoding is that you don't need to configure the
limit in the script ... It's all automatic.

Biggest issue is the lockup if you try and send SWO too quickly since it
doesn't recover. You could get fast intervals due to noise or
misconfiguration so this problem does need to be fixed.

Reading my original notes I suggested that an error should be returned, but
there's no explicit speed set for Manchester swo so that would be
difficult...ignore that!

Dave

dragonmux:
Because this has stagnated and with the difficulty in trying to actually test it that we're facing + with being so close to the release window, we're going to bump this into v2.0 - if we don't wind up just rewriting a large part of the Trace/SWO code anyway.

dragonmux:
Whether adding a FIFO still benefits the SWO implementation does need investigating, but this PR has now been superseded by #1934 which fully rewrites the Manchester SWO subsystem in particular. Closing as deprecated.

@mubes when you get a chance, if you could repeat your speed test on that other PR (merged to `main`), it would be greatly appreciated.

PR #1093, opened 2022-07-05
Fix: target Flash buffer lifetime bounding

### LINKED ISSUE #960: Blackmagic FW gets stuck in blocking_handler() on flashing an NRF52
In our project, we want to use the black magic probe as a less pricey alternative to a jlink, because we need quite a lot of probes for our automatic testing.
We wrote gdb scripts to automate some stuff, like doing a mass erase followed by writing some bytes into the UICR and then flashing a fw hex file via file/load.

When running those scripts several times, the probe eventually gets stuck within the fw file flashing just after the output of it loading the first section (Loading section .sec1, size 0x10000 lma 0x0)

This happens on three different probe platforms:

- 1bitsquared Black Magic Probe v2.1e
- ST-LINK/V2
- 'baite' stlink clone

So I guessed this is somehow bmp fw related.

I soldered some wires to the 1bitsquared probe to debug it via a jlink using the blackmagic.elf file and tried to reproduce it, which gave me following call stack when the error happens:

blocking_handler@0x0801124a (/home/mikedk/src/ng/blackmagic/libopencm3/lib/cm3/vector.c:104)
<signal handler called>@0xfffffff9 (Unbekannte Quelle:0)
_malloc_r@0x08011a0e (/_malloc_r.cdasm:39)
target_flash_write_buffered@0x0800ecf0 (/home/mikedk/src/ng/blackmagic/src/target/target.c:293)
target_flash_write@0x0800f014 (/home/mikedk/src/ng/blackmagic/src/target/target.c:263)
handle_v_packet@0x0800614a (/home/mikedk/src/ng/blackmagic/src/gdb_main.c:504)
gdb_main_loop@0x080068a8 (/home/mikedk/src/ng/blackmagic/src/gdb_main.c:314)
gdb_main@0x080068ea (/home/mikedk/src/ng/blackmagic/src/gdb_main.c:553)
main@0x08008a48 (/home/mikedk/src/ng/blackmagic/src/main.c:47)

So somehow the malloc fails and triggers a hardfault which then gets stuck within blocking_handler().

The only way to get the probe up and running again is to replug the USB connection.

For us, this is a blocking issue because the effort needed to automatically disconnect/reconnect the usb connections of a lot of probes would exceed the price of a stlink/v2 by quite a lot ;)

So please tell me if I can collect some more symptoms to help get this bug fixed.

Regards,
MikeDK

dragonmux:
Make the buffers used to program a target's Flash better bounded in lifetime and memory usage.

In the old setup, if the user requested the programming of, say, 3 unique Flash regions, we would allocate 3 buffers, work away programming the regions, and only at the very end run `target_flash_done_buffered` on each region to clean up - making the final writes possibly out of order, and meaning we gradually grow memory usage with quite large buffers till we finally de-allocate them all in one shot. This clearly causes issues on some targets as we're memory constrained, and they can have large buffer requirements and many Flash regions.

In this new regime, as we get to the end of each Flash region when running `target_flash_write`, we run `target_flash_done_buffered` to clean up. This won't catch all possible combinations of writes and allocations, but it should bound writes that fill Flash regions and step into new regions.

This partially fixes #960. If we can figure out a way to catch partial regions which won't be filled further, we will amend the PR so we can fully bound all temporary Flash buffers and the resulting memory usage.

### DIFF
--- a/src/target/target.c
+++ b/src/target/target.c
@@ -280,6 +280,11 @@ int target_flash_write(target *t,
 		dest += tmplen;
 		src += tmplen;
 		len -= tmplen;
+		/* If the current chunk of Flash is now full from this operation
+		 * then finish operations on the Flash chunk and free the internal buffer.
+		 */
+		if (dest == f->start + f->length)
+			ret |= target_flash_done_buffered(f);
 	}
 	return ret;
 }


esden (approved):
LGTM

Issue #228, opened 2017-05-08
Backdoor key not supported on Kinetis K22F


mondalaci:
Usually, `swdp_scan` produces the following result:

```
(gdb) monitor swdp_scan
Target voltage: 3.3V
Available Targets:
No. Att Driver
 1      K22F
 2      Kinetis Recovery (MDM-AP)
```

But when the backdoor key is set, the K22F target doesn't show up. We use the backdoor key to secure the bootloader, so that no matter what, the bootloader cannot overwite itself.

Here's some help regarding the location of the backdoor key: 

![Backdoor key](https://raw.githubusercontent.com/UltimateHackingKeyboard/bootloader/master/littleEndian_keyStorage.png)

Also, probably the following chapters of the K22FN512 reference manual are relevant:
* Chapter 29.1 - Flash configuration Field Description
* Chapter 29.3 - Flash registers
* Chapter 29.4.12 - Flash security (Functional description)

I hope that you reconsider supporting the backdoor key because it's a critical feature for us. Without that, we cannot migrate to Black Magic Probe, which we'd very much love to.

PR #1094, opened 2022-07-05
Kinetis handling implementation cleanup

dragonmux:
In preparation for fixing the Kinetis part-killer bug we did some cleanup, reformatting and rationalising of the Kinetis code so it should be easier to add new supported parts and work on in the future.

In light that Perigoso is working on #228 and depends on these changes, we have elected to bring this PR forward before the part-killer bugfix. As the bugfix is based on this we recommend merging to both main and v1.8


esden (approved):
LGTM

PR #1098, opened 2022-07-11
Fix: kinetis Flash control correctness

### LINKED ISSUE #1029: Blackmagic will write 0s on S32K devices, bricking them
BMP has broken something between 8def28d and f6edb54395d3d4cb2bb186f079cb3e3181eff518 that has caused it to brick S32K devices. This is because the second word written is now `0x00000000` regardless of the actual value. Crucially, address `0x400` is the Flash Configuration Field that locks flash. If this field is anything other than `0xffffffff`, the chip will become locked and cannot be unlocked.

```
(gdb) load
Loading section .interrupts, size 0x400 lma 0x0
Loading section .flash_config, size 0x10 lma 0x400
Loading section .text, size 0x1cb90 lma 0x410
Loading section .ARM, size 0x8 lma 0x1cfa0
Loading section .data, size 0x4b8 lma 0x1cfa8
Start address 0x00000410, load size 119904
Transfer rate: 9 KB/sec, 944 bytes/write.
(gdb) x 0
0x0 <__isr_vector>:     0x2001f000
(gdb) x 4
0x4 <__isr_vector+4>:   0x00000000
(gdb) x 0x400
0x400:  0xffffffff
(gdb) x 0x404
0x404:  0x00000000
(gdb) x 0x408
0x408:  0xffffffff
(gdb)
```

Going back to an older version of BMP, notably 8def28d, resolves the issue for me, meaning something happened between these two versions that caused this failure.

## Recovering a device

If you discover that `0x404` has been overwritten, **do not power off the target**. Instead, issue a `erase_mass`:

```
(gdb) att 2
A program is being debugged already.  Kill it? (y or n) y
Attaching to program: /opt/Ion/gridania-bringup-dvt1/148/_build/gridania-dvt1-148.elf, Remote target
warning: while parsing target description: no element found
warning: Could not load XML target description; ignoring

Program stopped.
warning: while parsing target memory map (at line 1): Required element <memory> is missing
0xc22ea4c2 in ?? ()
(gdb) mon erase_mass
Requesting mass erase (status = 0x1003e)
Mass erase acknowledged
Mass erase complete
(gdb) att 1
A program is being debugged already.  Kill it? (y or n) y
Attaching to program: /opt/Ion/gridania-bringup-dvt1/148/_build/gridania-dvt1-148.elf, Remote target

Program stopped.
0x00000000 in __isr_vector ()
(gdb) x 0
0x0 <__isr_vector>:     0xffffffff
(gdb) x 4
0x4 <__isr_vector+4>:   0xffffffff
(gdb)
```

Because this field is latched on power up, the board will continue to operate with debugging enabled until you power cycle the board.

dragonmux:
This PR fixes a part-killing regression found by xobs. It does this by reworking the Kinetis Flash command generation and issuing routines so more closely match the behaviour of the v1.7 ones while retaining the benefits of the revamped approach.

This closes #1029 and should be merged both to main and to v1.8.

### DIFF
--- a/src/target/kinetis.c
+++ b/src/target/kinetis.c
@@ -466,7 +466,7 @@ static int kinetis_flash_cmd_write(struct target_flash *f, target_addr dest, con
 		write_cmd = FTFx_CMD_PROGRAM_LONGWORD;
 
 	while (len) {
-		if (kinetis_fccob_cmd(f->t, write_cmd, dest, src, 1)) {
+		if (kinetis_fccob_cmd(f->t, write_cmd, dest, src, kf->write_len >> 2U)) {
 			if (len > kf->write_len)
 				len -= kf->write_len;
 			else
--- a/src/target/kinetis.c
+++ b/src/target/kinetis.c
@@ -493,16 +493,16 @@ static int kinetis_flash_done(struct target_flash *const f)
 	 * vs 4 byte phrases).
 	 */
 	if (kf->write_len == K64_WRITE_LEN) {
-		uint32_t vals[2];
-		vals[0] = target_mem_read32(f->t, FLASH_SECURITY_BYTE_ADDRESS - 4);
-		vals[1] = target_mem_read32(f->t, FLASH_SECURITY_BYTE_ADDRESS);
-		vals[1] = (vals[1] & 0xffffff00) | FLASH_SECURITY_BYTE_UNSECURED;
+		uint32_t vals[2] = {
+			target_mem_read32(f->t, FLASH_SECURITY_BYTE_ADDRESS - 4),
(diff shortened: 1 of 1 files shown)


esden (approved):
Thank you very much. The patch looks correct to me and it addresses the 8byte write issue as far as I can tell. I was only able to test if this patch introduces regression for 4byte writes (with a KL03Z target) because I don't have a target that is using 8byte writes yet. (a S32K144EVB-Q100 is ordered and on the way to be added to the test target collection) 

I also rely on the fact that @xobs tested this patch and says it fixes the issue on his S32 target. All things considered I am ready to merge this into main as well as the V1.8 branch preparing for the V1.8.1 release.

PR #1099, opened 2022-07-12
Feature: mass erase progress indication

dragonmux:
Targets which support an `erase_mass` monitor command and which take a while to complete the operation cause GDB issues by making it hit internal state timeouts for the GDB server likely having gone away ("Ignoring packet error, continuing...") but it trying to keep going regardless.

This makes for a poor user experience and gives no indication whether BMP has locked up or if the operation is just taking a while. This PR addresses that by refactoring the mass erase command into the target layer, and introducing new timeout-based reporting that lets the user know BMP is still working on it and hasn't crashed.

This has the benefit of letting GDB know we're alive too and resetting the state timeout that causes the error messages otherwise.

In a separate issue, we will bring up the issue of the command's naming and whether to change it in v2. Internally as of this PR, it's called mass_erase in the target structure and related places, but erase_mass to the user.

### DIFF
--- a/src/target/efm32.c
+++ b/src/target/efm32.c
@@ -42,7 +42,6 @@
 #include "target_internal.h"
 #include "cortexm.h"
 #include "adiv5.h"
-#include "gdb_packet.h"
 
 #define SRAM_BASE		0x20000000
 #define STUB_BUFFER_BASE	ALIGN(SRAM_BASE + sizeof(efm32_flash_write_stub), 4)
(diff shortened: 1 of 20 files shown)


dragonmux:
Even though we didn't yet fix the lack of mass erase for STM32L0/L1 (#1100), going to call this enough for this one PR as it's already giant and touches almost every target supported in the project. It has been tested on the small set of devices we have available and works there.

This is now ready for review and merge.

esden (approved):
LGTM, tested with STM32F415 and KL03Z targets.

PR #1102, opened 2022-07-12
efm32 cleanup

perigoso:
couple cleanups, and implement missing methods in mass erase, devices with >= 512 KiB flash support read-while-write and have two separate flash regions with independent mass erase commands

addendum, fix bootloader bit handling, current method only allows erasing of the bit

this is untested, both before and after, solely based on documentation, I have access to some efm32 dev boards at work I might be able to borrow, but unsure if any would be affected by these changes

emlib mass erase function:
```
MSC_Status_TypeDef MSC_MassErase(void)
{
  bool wasLocked;
  wasLocked = MSC_IS_LOCKED();
  MSC->LOCK = MSC_LOCK_LOCKKEY_UNLOCK;

  /* Enable writing to the MSC module. */
  MSC->WRITECTRL |= MSC_WRITECTRL_WREN;

  /* Unlock the device mass erase. */
  MSC->MASSLOCK = MSC_MASSLOCK_LOCKKEY_UNLOCK;

  /* Erase the first 512 K block. */
  MSC->WRITECMD = MSC_WRITECMD_ERASEMAIN0;

  /* Waiting for erase to complete. */
  while ((MSC->STATUS & MSC_STATUS_BUSY) != 0U) {
  }

#if ((FLASH_SIZE >= (512 * 1024)) && defined(_MSC_WRITECMD_ERASEMAIN1_MASK))
  /* Erase the second 512 K block. */
  MSC->WRITECMD = MSC_WRITECMD_ERASEMAIN1;

  /* Waiting for erase to complete. */
  while ((MSC->STATUS & MSC_STATUS_BUSY) != 0U) {
  }
#endif

  /* Restore the mass erase lock. */
  MSC->MASSLOCK = MSC_MASSLOCK_LOCKKEY_LOCK;

  /* Disable writing to the MSC module. */
  MSC->WRITECTRL &= ~MSC_WRITECTRL_WREN;

  if (wasLocked) {
    MSC->LOCK = MSC_LOCK_LOCKKEY_LOCK;
  }

  /* This will only successfully return if calling function is also in SRAM. */
  return mscReturnOk;
}
```

extract from one datasheet (efm32gg11)
```
9 ERASEMAIN1 0 W1 Mass Erase Region 1
Initiate mass erase of region 1. For devices supporting read-while-write, this is the upper half of the flash. Before use
MSC_MASSLOCK must be unlocked. To completely prevent access from software, clear bit 1 in the mass erase lock-word
(MLW)

8 ERASEMAIN0 0 W1 Mass Erase Region 0
Initiate mass erase of region 0. For devices supporting read-while-write, this is the lower half of the flash. For other devices
it is the entire flash. Before use MSC_MASSLOCK must be unlocked. To completely prevent access from software, clear bit
0 in the mass erase lock-word (MLW)
```

### DIFF
--- a/src/target/efm32.c
+++ b/src/target/efm32.c
@@ -87,6 +87,7 @@ const struct command_s efm32_cmd_list[] = {
 #define EFM32_MSC_WRITECMD_WRITETRIG  (1 << 4)
 #define EFM32_MSC_WRITECMD_ERASEABORT (1 << 5)
 #define EFM32_MSC_WRITECMD_ERASEMAIN0 (1 << 8)
+#define EFM32_MSC_WRITECMD_ERASEMAIN1 (1 << 9)
 
 #define EFM32_MSC_STATUS_BUSY       (1 << 0)
 #define EFM32_MSC_STATUS_LOCKED     (1 << 1)
@@ -688,7 +689,17 @@ static bool efm32_mass_erase(target *t)
 	if (device == NULL) {
 		return true;
 	}
+
+	if (device->family_id == 71 || device->family_id == 73) {
+		/* original Gecko and Tiny Gecko families don't support mass erase */
+		tc_printf(t, "This device does not support mass erase through MSC.\n");
+		return false;
+	}
+
 	uint32_t msc = device->msc_addr;
+
+	uint8_t di_version = t->driver[0] - 48; /* di version hidden in driver str */
+	uint16_t flash_kib = efm32_read_flash_size(t, di_version);
 
 	/* Set WREN bit to enable MSC write and erase functionality */
 	target_mem_write32(t, EFM32_MSC_WRITECTRL(msc), 1);
@@ -706,6 +717,19 @@ static bool efm32_mass_erase(target *t)
 		if (target_check_error(t))
 			return false;
 		target_print_progress(&timeout);
+	}
+
+	/* Parts with >= 512 kiB flash have 2 mass erase regions */
+	if (flash_kib >= 512) {
+		/* Erase operation */
+		target_mem_write32(t, EFM32_MSC_WRITECMD(msc), EFM32_MSC_WRITECMD_ERASEMAIN1);
+
+		/* Poll MSC Busy */
+		while ((target_mem_read32(t, EFM32_MSC_STATUS(msc)) & EFM32_MSC_STATUS_BUSY)) {
+			if (target_check_error(t))
+				return false;
+			target_print_progress(&timeout);
+		}
 	}
 
 	/* Relock mass erase */
--- a/src/target/efm32.c
+++ b/src/target/efm32.c
@@ -860,17 +860,18 @@ static bool efm32_cmd_bootloader(target *t, int argc, const char **argv)
 	}
 
 	uint32_t clw0 = target_mem_read32(t, EFM32_LOCK_BITS_CLW0);
-	bool bootloader_status = (clw0 & EFM32_CLW0_BOOTLOADER_ENABLE) ? 1 : 0;
 
 	if (argc == 1) {
+		const bool bootloader_status = clw0 & EFM32_CLW0_BOOTLOADER_ENABLE;
 		tc_printf(t, "Bootloader %s\n", bootloader_status ? "enabled" : "disabled");
 		return true;
-	} else {
-		bootloader_status = (argv[1][0] == 'e');
 	}
 
 	/* Modify bootloader enable bit */
-	clw0 &= bootloader_status ? ~0 : ~EFM32_CLW0_BOOTLOADER_ENABLE;
+	if (argv[1][0] == 'e')
+		clw0 |= EFM32_CLW0_BOOTLOADER_ENABLE;
+	else
+		clw0 &= ~EFM32_CLW0_BOOTLOADER_ENABLE;
 
 	/* Unlock */
 	target_mem_write32(t, EFM32_MSC_LOCK(msc), EFM32_MSC_LOCK_LOCKKEY);
--- a/src/target/efm32.c
+++ b/src/target/efm32.c
@@ -512,32 +512,26 @@ static void efm32_add_flash(target *t, target_addr addr, size_t length, size_t p
 }
 
 /* Lookup device */
-static size_t efm32_lookup_device_index(target *t, uint8_t di_version)
+static efm32_device_t const *efm32_get_device(target *t, uint8_t di_version)
 {
 	uint8_t part_family = efm32_read_part_family(t, di_version);
 
 	/* Search for family */
 	for (size_t i = 0; i < (sizeof(efm32_devices) / sizeof(efm32_device_t)); i++) {
 		if (efm32_devices[i].family_id == part_family) {
-			return i;
+			return &efm32_devices[i];
 		}
 	}
 
 	/* Unknown family */
-	return UINT32_MAX;
-}
-
-static efm32_device_t const *efm32_get_device(size_t index)
-{
-	if (index >= (sizeof(efm32_devices) / sizeof(efm32_device_t))) {
-		return NULL;
-	}
-	return &efm32_devices[index];
+	return NULL;
 }
 
 /* Probe */
 struct efm32_priv_s {
 	char efm32_variant_string[60];
+	uint8_t di_version;
+	efm32_device_t const *device;
 };
 
 bool efm32_probe(target *t)
@@ -572,15 +566,9 @@ bool efm32_probe(target *t)
 	}
 
 	/* Read the part family, and reject if unknown */
-	size_t device_index = efm32_lookup_device_index(t, di_version);
-	if (device_index >= (sizeof(efm32_devices) / sizeof(efm32_device_t))) {
-		/* unknown device family */
+	efm32_device_t const *device = efm32_get_device(t, di_version);
+	if (!device)
 		return false;
-	}
-	efm32_device_t const *device = &efm32_devices[device_index];
-	if (device == NULL) {
-		return false;
-	}
 
 	t->mass_erase = efm32_mass_erase;
 	uint16_t part_number = efm32_read_part_number(t, di_version);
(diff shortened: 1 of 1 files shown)


dragonmux:
unfortunately going to be a pain and ask when feature/mass-erase-progress-indication lands as a PR if you could rebase against that - in the mean time, we'll review with that in mind

src/target/efm32.c
```diff
@@ -349,79 +344,59 @@ typedef struct efm32_v2_di_miscchip_t {
...
- * Reads the Unique Number (DI V2 only)
- */
+/* Reads the Unique Number (DI V2 only) */
 static uint64_t efm32_v2_read_unique(target *t, uint8_t di_version)
 {
 	uint64_t unique;
 	switch (di_version) {
 	case 2:
-		unique  = (uint64_t)target_mem_read32(t, EFM32_V2_DI_UNIQUEH) << 32;
+		unique = (uint64_t)target_mem_read32(t, EFM32_V2_DI_UNIQUEH) << 32;
```
dragonmux, line 391: Given how this code is shaped and written, it makes sense to move the definition of `unique` into this block and drop the `<< 0` on the next line.

Additionally, the cast for the second line is not required - it's only required on this definition line.

perigoso:
I was aware of that, no worries

src/target/efm32.c
```diff
@@ -349,79 +344,59 @@ typedef struct efm32_v2_di_miscchip_t {
...
- * Reads the EFM32 Extended Unique Identifier EUI64 (V1)
- */
+/* Reads the EFM32 Extended Unique Identifier EUI64 (V1) */
 static uint64_t efm32_v1_read_eui64(target *t)
 {
 	uint64_t eui;
 
-	eui  = (uint64_t)target_mem_read32(t, EFM32_V1_DI_EUI64_1) << 32;
-	eui |= (uint64_t)target_mem_read32(t, EFM32_V1_DI_EUI64_0) <<  0;
+	eui = (uint64_t)target_mem_read32(t, EFM32_V1_DI_EUI64_1) << 32;
```
dragonmux, line 379: Given how this code is shaped and written, it makes sense to move the definition of `eui` here and drop the `<< 0` on the next line.

Additionally, the cast for the second line is not required - it's only required on this definition line.

src/target/efm32.c
```diff
@@ -620,57 +583,55 @@ bool efm32_probe(target *t)
 	}
 
 	/* Read the part family, and reject if unknown */
-	size_t device_index  = efm32_lookup_device_index(t, di_version);
-	if (device_index > (127-32)) {
+	size_t device_index = efm32_lookup_device_index(t, di_version);
+	if (device_index > (127 - 32)) {
```
dragonmux, line 587: the ()'s aren't required and both constants should be unsigned - perhaps manually simplify this with a comment on how such a magic value has been arrived at?

src/target/efm32.c
```diff
@@ -620,57 +583,55 @@ bool efm32_probe(target *t)
...
 	if (device == NULL) {
 		return false;
 	}
 
 	uint16_t part_number = efm32_read_part_number(t, di_version);
 
 	/* Read memory sizes, convert to bytes */
-	uint16_t flash_kib  = efm32_read_flash_size(t, di_version);
+	uint16_t flash_kib = efm32_read_flash_size(t, di_version);
 	uint32_t flash_size = flash_kib * 0x400;
```
dragonmux, line 600: U on the end of constants in an unsigned context - don't want to accidentally invoke integer promotion conversions here.

src/target/efm32.c
```diff
@@ -1000,19 +967,19 @@ void efm32_aap_probe(ADIv5_AP_t *ap)
 
 	/* New target */
 	target *t = target_new();
+	if (!t) {
+		return;
+	}
+
 	adiv5_ap_ref(ap);
```
dragonmux, line 974: It looks like we might have forgotten to assign t->mass_erase here - please check and verify the mass erase behaviour of the AAP

src/target/efm32.c
```diff
@@ -1000,19 +967,19 @@ void efm32_aap_probe(ADIv5_AP_t *ap)
...
 
 	/* Read status */
-	DEBUG_INFO("EFM32: AAP STATUS=%08"PRIx32"\n", adiv5_ap_read(ap, AAP_STATUS));
+	DEBUG_INFO("EFM32: AAP STATUS=%08" PRIx32 "\n", adiv5_ap_read(ap, AAP_STATUS));
 
 	struct efm32_aap_priv_s *priv_storage = calloc(1, sizeof(*priv_storage));
-	sprintf(priv_storage->aap_driver_string,
-			"EFM32 Authentication Access Port rev.%d",
-			aap_revision);
+	sprintf(priv_storage->aap_driver_string, "EFM32 Authentication Access Port rev.%d", aap_revision);
```
dragonmux, line 982: `%d` used for a uint16_t which should be `%hu`

dragonmux (changes requested):
Generally looks good, there are just a couple of small cleanup items we spotted

src/target/efm32.c
perigoso: you're thinking ahead, this is not yet rebased on the mass erase unification

dragonmux: Ah, that'd do it! Resolving for now with a note to check again when it has been.

perigoso:
suggestions applied, and preemptive rebase on `mass-erase-progress-indication`

src/target/efm32.c
perigoso: I could not understand how this was arrived at, my best guess is it came from some sort of char conversion or so, either way, it didn't guard for an out of bound index for that table, so I changed it for its size

perigoso: ah! it comes from here

`	snprintf(priv_storage->efm32_variant_string, sizeof(priv_storage->efm32_variant_string), "%c\b%c\b%s %hu F%hu %s",
		di_version + 48, (uint8_t)device_index + 32, device->name, part_number, flash_kib, device->description);`

`(uint8_t)device_index + 32`

it is indeed some int to char conversion, but it looks like it was just checking if it was a valid char and not if it was out of bounds, this is used later in the cmd's to get the device with `efm32_get_device(t->driver[2] - 32);`

perigoso: the more I look at it the less sense it makes, I get that it's hiding the index in the driver string, and 32 (' ') is the first char that can be safely printed out the terminal, but surely there was a better way, even back when this was written

perigoso:
further addressing https://github.com/blackmagic-debug/blackmagic/pull/1102#discussion_r919504784

Ready for re-review.

dragonmux:
Excellent! We'll re-review once we get the feature branch this is on top of PR'd in and merged so we can get a clean review and hopefully immediately merge

dragonmux:
#1099 has now been merged, so could you please rebase this?

perigoso:
@dragonmux done

src/target/efm32.c
```diff
@@ -349,79 +343,49 @@ typedef struct efm32_v2_di_miscchip_t {
...
+/* Reads the EFM32 Extended Unique Identifier EUI64 (V1) */
 static uint64_t efm32_v1_read_eui64(target *t)
 {
-	uint64_t eui;
-
-	eui  = (uint64_t)target_mem_read32(t, EFM32_V1_DI_EUI64_1) << 32;
-	eui |= (uint64_t)target_mem_read32(t, EFM32_V1_DI_EUI64_0) <<  0;
-
-	return eui;
+	return (uint64_t)target_mem_read32(t, EFM32_V1_DI_EUI64_1) << 32 | target_mem_read32(t, EFM32_V1_DI_EUI64_0);
```
dragonmux, line 376: combining this is fine but that makes the ()'s around the shift expression required due to operator precedence.

src/target/efm32.c
```diff
@@ -349,79 +343,49 @@ typedef struct efm32_v2_di_miscchip_t {
...
-	case 2:
-		unique  = (uint64_t)target_mem_read32(t, EFM32_V2_DI_UNIQUEH) << 32;
-		unique |= (uint64_t)target_mem_read32(t, EFM32_V2_DI_UNIQUEL) << 0;
-		return unique;
-	default:
+	if (di_version != 2)
 		return 0;
-	}
+
+	return (uint64_t)target_mem_read32(t, EFM32_V2_DI_UNIQUEH) << 32 | target_mem_read32(t, EFM32_V2_DI_UNIQUEL);
```
dragonmux, line 385: Same issue as the EUI reader

src/target/efm32.c
```diff
@@ -766,107 +711,127 @@ static bool efm32_mass_erase(target *t)
...
+	/* Parts with >= 512 kiB flash have 2 mass erase regions */
+	if (flash_kib >= 512) {
+		/* Erase operation */
+		target_mem_write32(t, EFM32_MSC_WRITECMD(msc), EFM32_MSC_WRITECMD_ERASEMAIN1);
+
+		/* Poll MSC Busy */
+		while ((target_mem_read32(t, EFM32_MSC_STATUS(msc)) & EFM32_MSC_STATUS_BUSY)) {
+			if (target_check_error(t))
+				return false;
+		}
```
dragonmux, line 723: This needs a call to `target_print_progress(&timeout);` inside the loop

dragonmux (changes requested):
This is looking much better. It'll be ready for merge one these 5 points are addressed.

src/target/efm32.c
dragonmux: oh, right.. it does take precedence over `|`, just not `||` - still would ask to have them added as it's then consistent with the rest of the code base and allays these kinds of confusion

src/target/efm32.c
perigoso: oops, just blindly rebased

src/target/efm32.c
perigoso: yep, that is fine

dragonmux (approved):

PR #1108, opened 2022-07-14
Fix: stm32l0/l1 is missing a mass erase implementation

### LINKED ISSUE #1100: Support for STM32L100
I am trying to attach to the STM32L100 on a Wink Hub v1. swdp_scan and jtag_scan both show the device, but neither will allow me to attach to it.
Screenshot from 2022-06-21 17-35-53
It would be great to be able to erase this chip and upload my own firmware to it.

dragonmux:
Following on from https://github.com/blackmagic-debug/blackmagic/pull/1099, this PR implements mass erase and protected chip detection for STM32L0 and STM32L1 series devices.

There are 3 possible protection states to consider:

* Level 2 protection is unrecoverable and likely results in no-scan when attempting JTAG or SWD scan.
* When L1 protected, most memory operations will fail and our only real recourse is to switch back into L0 protection - this is by nature mass erase (and performs a target reset on completion).
* L0 protection is anything goes - but, importantly, attempting to switch into this mode does nothing to the erase state of the device. So, for this protection mode we must implement sending an actual mass erase command. This does not result in communications reset.

This fixes https://github.com/blackmagic-debug/blackmagic/pull/1100.

### DIFF
--- a/src/target/stm32l0.c
+++ b/src/target/stm32l0.c
@@ -135,7 +135,7 @@ static bool stm32lx_protected_mass_erase(target_s *target);
 static bool stm32lx_mass_erase(target_s *target);
 
 static bool stm32lx_cmd_option(target_s *target, int argc, const char **argv);
-static bool stm32lx_cmd_eeprom(target_s *t, int argc, const char **argv);
+static bool stm32lx_cmd_eeprom(target_s *target, int argc, const char **argv);
 
 static const command_s stm32lx_cmd_list[] = {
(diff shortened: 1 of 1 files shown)


dragonmux:
Unfortunately we don't own any STM32L0 or L1 test devices, however this PR cleans up that code sufficiently that along with implementing the bones of mass erase, we're going to mark this as ready for review. We don't expect it's quite merge ready yet though

esden (approved):
LGTM, tested on 32L0538DISCOVERY board.

PR #1109, opened 2022-07-14
Fix: nRF52 mass erase

dragonmux:
Under certain conditions outlined below, mass erase on nRF51/nRF52 targets fails. This PR fixes that issue.

For the bug to be reproduced, one must meet the following conditions:

* Be attached
* Have run the target (using the GDB run command)
* Interrupt the target (Ctrl + C)

If one tries mass erasing the target after meeting these conditions, it will appear to work but do nothing.

Many thanks to Discord user giobauermeister for reporting this


giobauermeister:
Thanks! This fix is working for me.

esden (approved):
LGTM

PR #1110, opened 2022-07-14
Add homebrew include so that libusb headers found.

djix123:
When building BMPA with HOSTED_BMP_ONLY=0 build will fail on the Mac (ARM64, at least) since the compilation does not find the libusb headers. This patch adds appropriate include. Maybe a better place to add this to the file. If there are strong feels let me know and I can move it as directed.

### DIFF
--- a/src/platforms/hosted/Makefile.inc
+++ b/src/platforms/hosted/Makefile.inc
@@ -45,6 +45,7 @@ else ifneq (, $(findstring cygwin, $(SYS)))
 else ifneq (filter, macosx darwin, $(SYS))
 	SRC += serial_unix.c
 	LDFLAGS += -framework CoreFoundation
+	CFLAGS += -I /opt/homebrew/include -I /opt/homebrew/include/libusb-1.0
 endif
 
 ifneq ($(HOSTED_BMP_ONLY), 1)


zyp:
> I do have a question though. Why is your homebrew installed in /opt? That is not the default way of how homebrew is usually installed.

For ARM macOS, Homebrew switched from `/usr/local` to `/opt/homebrew` as the default.

esden:
Thanks @zyp ... thanks homebrew 🤦 

I still think pkg-config approach would be better and more robust... :/

djix123:
Yes, on ARM64 homebrew is installed in /opt/homebrew.

If I use pkg-config then I would have to deal with this:

```
diff --git a/src/platforms/hosted/bmp_hosted.h b/src/platforms/hosted/bmp_hosted.h
index 4783f2c..dee86cc 100644
--- a/src/platforms/hosted/bmp_hosted.h
+++ b/src/platforms/hosted/bmp_hosted.h
@@ -4,7 +4,7 @@
 #include "cli.h"
 
 #if HOSTED_BMP_ONLY != 1
-# include <libusb-1.0/libusb.h>
+# include <libusb.h>
 struct trans_ctx {
 #define TRANS_FLAGS_IS_DONE (1 << 0)
 #define TRANS_FLAGS_HAS_ERROR (1 << 1)
diff --git a/src/platforms/hosted/bmp_libusb.c b/src/platforms/hosted/bmp_libusb.c
index 11faa16..afd7f73 100644
--- a/src/platforms/hosted/bmp_libusb.c
+++ b/src/platforms/hosted/bmp_libusb.c
@@ -19,7 +19,7 @@
 
 /* Find all known usb connected debuggers */
 #include "general.h"
-#include "libusb-1.0/libusb.h"
+#include "libusb.h"
 #include "cli.h"
 #include "ftdi_bmp.h"
 #include "version.h"
```

since ```pkg-config --cflags libusb-1.0``` returns ```-I/opt/homebrew/Cellar/libusb/1.0.26/include/libusb-1.0``` already including the libusb-1.0 directory.

I can deal with this via #define ... would like to confirm preference before submitting an update.

dragonmux:
Given this will be an attempt at dealing with how to find libusb1 on Windows.. and is the wrong code even on Linux.. please submit the fix without any additional `#ifdef` logic as that needs sorting out on all platforms.

djix123:
@dragonmux - just to confirm - only add the CFLAGS += -I/opt/homebrew/install between lines 46 & 47?

dragonmux:
The suggested no-`#ifdef`'s diff above to fix the libusb include paths, and addition of the ARM Homebrew path to the Mac-specific block.

The internal conversation this caused has suggested we check if `pkg-config` exists on Mac (as it doesn't by default in fresh Homebrew installations), if it does then use that to find the required packages, otherwise insert the normal homebrew path into the include flags. We won't ask you to do the extra work this approach entails, but if you're feeling keen then it would be good.

djix123:
I can certainly have a good at doing that *but* when ```pkg-config``` does exist on the Mac it returns a path that includes ```libusb-1.0``` making the includes in ```bmp_hosted.h``` and ```bmp_libusb.c```.

dragonmux:
We'd suggest this means the no-pkg-config fix should be `CFLAGS+=-I /opt/homebrew/include -I /opt/homebrew/include/libusb-1.0` so it works the same either way

djix123:
ok, I can make that change.

dragonmux:
The revised change looks good - could you rebase the two commits together into one (apply the newer as a fixup to the older, adjusting the commit message accordingly) please?

We'll leave this open to let you work on the pkg-config part of this unless you want us to merge this now and open a new PR for the other part.

djix123:
Thanks @dragonmux - I've rebased the two commits into one. Could you merge this now, and I'll work on the ```pkg-config``` version separately. How would you deal with the fact that ```pkg-config``` returns an include path that is incompatible with the current code?

dragonmux (approved):
LGTM, merging.

PR #1111, opened 2022-07-14
target: Fixed several clang-tidy warnings which improve readability of the code

dragonmux:
As part of ongoing work to clean up the code and improve readability, this addresses some clang-tidy warnings from the target API and layer implementation.

Tested on all our available hardware without regression.


esden (approved):
LGTM

PR #1112, opened 2022-07-14
Fix: DAP SWDP init

dragonmux:
With the help of @mubes, this addresses a CMSIS-DAP protocol compliance issue which was breaking use of BMDA with some CMSIS-DAP debuggers but was otherwise shielded/hidden by the immediate link reset that occurs following the removed sequence test code.

This improves our compatibility with DragonProbe.


esden (approved):
LGTM

PR #1113, opened 2022-07-15
target: Implemented a command to allow erasing Flash ranges on any target

dragonmux:
This PR is based on a branch that does a similar thing on Uwe's repo (thank you for the idea Uwe!)

This implements a command - `monitor erase_range <address> <count>` - for erasing selected ranges of Flash memory on a target (eg, to erase a wrongly Flashed firmware image, or a bad bootloader).

This will be the first in a series of PRs which pull useful fixes and improvements in off forks to enhance the code base and close outstanding issues. We will in all likelihood follow this PR up with another that guts per-target equivalents of this command.

### DIFF
--- a/src/target/target.c
+++ b/src/target/target.c
@@ -29,13 +29,15 @@ target *target_list = NULL;
 
 #define STDOUT_READ_BUF_SIZE	64
 
-static int target_flash_write_buffered(struct target_flash *f,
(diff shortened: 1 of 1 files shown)


esden (approved):
Wonderful! Thanks for the addition. :)

PR #1114, opened 2022-07-15
Fix: Remove the BMDA `-d`/`--device` deprecation warning

### LINKED ISSUE #1096: BMDA "-d" option deprecated?
Curious as to why this has been marked as deprecated. Seems like a useful option when you have multiple BMDPs connected. 
Sid

dragonmux:
This option is still useful to many users and there isn't a particular reason for it to be marked deprecated.

This fixes #1096


esden (approved):
LGTM

PR #1115, opened 2022-07-15
Fix: BMDA lib detect

esden:
This patch set fixes some issues with the BMDA makefile.

Outside of targets we should use space indentation as tabs can be problematic. In this case when a library is not found the `$(error ...)` function is called. This fails as it is indented with tabs and make does not like that. So it is better to just generally indent everything outside of the target with spaces.

Older versions of make do not support the ` .SHELLSTATUS` variable. So we need to do some acrobatics to implement that with just the `$(shell ...)` function.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -32,7 +32,7 @@ jobs:
         run: make clean
 
       - name: Install BMDA dependencies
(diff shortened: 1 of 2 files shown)


dragonmux (approved):
LGTM, mering

PR #1116, opened 2022-07-15
target/nrf51: Add command to enable flash protection

giobauermeister:
When chip is in protected state, it is possible to disable read/write protection doing a mass_erase. This new command `mon protect_flash` enables flash read/write protection. This is useful when using BMP in production programming.

The following command sequence was used to test this branch:

```
(gdb) target extended-remote /dev/ttyBmpGdb
A program is being debugged already.  Kill it? (y or n) y
Remote debugging using /dev/ttyBmpGdb
(gdb) monitor swdp_scan
Target voltage: 3.21V
Available Targets:
No. Att Driver
 1      Nordic nRF52 M4
(gdb) att 1
Attaching to program: firmware.hex, Remote target
0xfffffffe in ?? ()
(gdb) mon protect_flash
protect flash..
(gdb) target extended-remote /dev/ttyBmpGdb
A program is being debugged already.  Kill it? (y or n) y
Remote debugging using /dev/ttyBmpGdb
0x05fa0004 in ?? ()
(gdb) monitor swdp_scan
Target voltage: 3.23V
Available Targets:
No. Att Driver
 1      Nordic nRF52 Access Port (protected) 
(gdb) 
```

Hardware used: STM32 Bluepill as BMP and target nRF52840

### DIFF
--- a/src/target/nrf51.c
+++ b/src/target/nrf51.c
@@ -33,6 +33,7 @@ static int nrf51_flash_write(struct target_flash *f,
 static bool nrf51_mass_erase(target *t);
 
 static bool nrf51_cmd_erase_uicr(target *t, int argc, const char **argv);
+static bool nrf51_cmd_protect_flash(target *t, int argc, const char **argv);
 static bool nrf51_cmd_read_hwid(target *t, int argc, const char **argv);
 static bool nrf51_cmd_read_fwid(target *t, int argc, const char **argv);
 static bool nrf51_cmd_read_deviceid(target *t, int argc, const char **argv);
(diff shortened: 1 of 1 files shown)


src/target/nrf51.c
```diff
@@ -273,6 +278,30 @@ static bool nrf51_cmd_erase_uicr(target *t, int argc, const char **argv)
...
+{
+	(void)argc;
+	(void)argv;
+	tc_printf(t, "protect flash..\n");
+
+	/* Enable write */
+	target_mem_write32(t, NRF51_NVMC_CONFIG, NRF51_NVMC_CONFIG_WEN);
+
+	/* Poll for NVMC_READY */
+	while (target_mem_read32(t, NRF51_NVMC_READY) == 0)
```
dragonmux, line 291: While this code is correct, please include the outer {}'s for the while loop for readability. As written this is bug-prone and confusing to newcomers.

dragonmux (changes requested):
Thank you for the contribution! This looks great. Please see the one thing we picked up in review (we've only highlighted one of two instances) which is a small nitpick.

With that nitpick fixed, this will be ready for merge.

src/target/nrf51.c
giobauermeister: Thanks @dragonmux! Added the braces for the two while loops.

PR #1118, opened 2022-07-18
Fix: SWDP and JTAG TAP routine types and understandability

dragonmux:
In this PR we address a few aspects of the SWDP and JTAG TAP routines which were less than ideal. This resulted in over 400 bytes of Flash saved and what should be much more understandable routines to maintain moving forward.

The first major thing we addressed was the type confusion (`uint8_t` instead of `bool`) and inappropriate use of `int` for values that must not be negative. The second was refactoring out the delayed and non-delayed paths that can be taken. The third was creating a much higher degree of re-use and re-usability where possible with clarified names for parameters to better indicate what they do.

This also applies clang-format to many of the files involved, bringing the code base closer to formatting unity.

This has been tested on a couple of flavors of STM32 target (many thanks to @mubes), and Tiva-C (JTAG and SWD) without any sign of regressions caused by it.

### DIFF
--- a/src/include/jtagtap.h
+++ b/src/include/jtagtap.h
@@ -46,7 +46,7 @@ typedef struct jtag_proc_s {
 	 * - DO may be point to the same address as DI.
 	 */
 	void (*jtagtap_tdi_tdo_seq)(uint8_t *data_out, const uint8_t final_tms, const uint8_t *data_in, int clock_cycles);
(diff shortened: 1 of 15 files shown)


esden (approved):
Did some testing too. Seems to be working. 🤞 😸  Merging!

esden:
Oops, missed the fact that it has conflicts due to the other PRs that I merged. When you have a moment can you take a look at it @dragonmux ? :)

esden (approved):
LGTM

PR #1121, opened 2022-07-19
Feature: Switch to C11

### LINKED ISSUE #1120: code style: asm inconsistency
Right now in the code base there are multiple instances of inline assembly, but with inconsistent keywords

some are using `asm` while others `__asm__`

afaik `__asm__` is more widely supported (as asm is an extension and is not allowed by flags like std=c99 or ansi) and should be safer for naming conflicts, so that should be adopted

dragonmux:
This PR splits PR #1092 to stage and merge the switch to C11 component, which then also fixes #1120.

Fairly straight-forwardly, this fixes up standards violations that prevent the clean compilation of the code in C11 mode without changing what anything is doing. This includes upgrading libopencm3 and bringing consistency to how we write inline assembly statements.

The other major thing that had to be changed was including the proper header for alloca and marking all presently un-fixable uses as a bug due to the dangers of accidentally overflowing or smashing the stack with alloca.

Sorting out a setup to test the code in #1092 is proving time-consuming and difficult, so this gets the biggest pre-requisite part of that PR out the way.


esden (approved):
LGTM

Issue #1105, opened 2022-07-13
No recovery from SWD parity error over CMSIS-DAP


mubes:
When a parity error is received over CMSIS-DAP communication with the target is instantly dropped. There does not appear to be any retry. This is polling after a 'continue' with the '09' representing a correctly received message with a parity error;
```
cmd :   05.00.05.08.00.00.00.00.01.52.00.00.23.05.f0.ed.00.e0.0f.0e.
cmd res:05.05.01.01.00.01.01.01.00.01.01.
memread @ e000edf0 len 4, align 2 , start: 
cmd :   05.00.05.08.00.00.00.00.01.52.00.00.23.05.f0.ed.00.e0.0f.0e.
cmd res:05.05.01.01.00.01.01.01.00.01.01.
memread @ e000edf0 len 4, align 2 , start: 
cmd :   05.00.05.08.00.00.00.00.01.52.00.00.23.05.f0.ed.00.e0.0f.0e.
cmd res:05.05.01.01.00.01.01.01.00.01.01.
memread @ e000edf0 len 4, align 2 , start: 
cmd :   05.00.05.08.00.00.00.00.01.52.00.00.23.05.f0.ed.00.e0.0f.0e.
cmd res:05.05.01.01.00.01.01.01.00.01.01.
memread @ e000edf0 len 4, align 2 , start: 
cmd :   05.00.05.08.00.00.00.00.01.52.00.00.23.05.f0.ed.00.e0.0f.0e.
cmd res:05.05.01.01.00.01.01.01.00.01.01.
memread @ e000edf0 len 4, align 2 , start: 
cmd :   05.00.05.08.00.00.00.00.01.52.00.00.23.05.f0.ed.00.e0.0f.0e.
cmd res:05.05.01.01.00.01.01.01.00.01.01.
memread @ e000edf0 len 4, align 2 , start: 
cmd :   05.00.05.08.00.00.00.00.01.52.00.00.23.05.f0.ed.00.e0.0f.0e.
cmd res:05.05.09.
TARGET LOST.
```

Platform was ORBTrace with STM32L151.

UweBonnes:
For a short test for BMDA CMSIS error recovery, can you please try to change in dap.c/wait_word() from
	if(buf[1] == SWDP_ACK_FAULT) {
to
      if(buf[1] >  SWDP_ACK_WAIT) {
Thanks

dragonmux:
[...]

PR #1122, opened 2022-07-19
dap: Allow recovery from protocol errors

UweBonnes:
Tested with picoprobe against another Pico and code that casual injected
errors, 09 as in #1105 and 2 for wait. Exception for NO_TARGET is needed to
detect the recovery AP of the RP2040.

### DIFF
--- a/src/platforms/hosted/dap.c
+++ b/src/platforms/hosted/dap.c
@@ -333,7 +333,7 @@ static uint32_t wait_word(uint8_t *buf, int size, int len, uint8_t *dp_fault)
 			break;
 	} while (buf[1] == DAP_TRANSFER_WAIT);
 
-	if(buf[1] == SWDP_ACK_FAULT) {
+	if (buf[1] != DAP_TRANSFER_OK && buf[1] != DAP_TRANSFER_NO_TARGET) {
 		*dp_fault = 1;
 		return 0;
 	}


src/platforms/hosted/dap.c
```diff
@@ -333,7 +333,7 @@ static uint32_t wait_word(uint8_t *buf, int size, int len, uint8_t *dp_fault)
 			break;
 	} while (buf[1] == DAP_TRANSFER_WAIT);
 
-	if(buf[1] == SWDP_ACK_FAULT) {
+	if ((buf[1] > DAP_TRANSFER_OK) && (buf[1] != DAP_TRANSFER_NO_TARGET)) {
```
dragonmux, line 336: Remove the extra ()'s - they're not necessary and suppress GDB warnings about assignments.

dragonmux (changes requested):
Before we even entertain merging this, noting that mubes made it clear he has a robust fix in mind for #1105 when he gets a chance to write it up, please fix the patch to match the code style for the repo.

Additionally, please clearly describe what this PR does, why it does it and how - the description given is unintelligible.

dragonmux:
As per the related PR and as this code has been rewritten since, we are closing this PR as the issue it tries to solve has been fixed, but with the understanding we may re-open the PR if needed

PR #1124, opened 2022-07-20
general: check _DEFAULT_SOURCE and _GNU_SOURCE

xobs:
The _DEFAULT_SOURCE and _GNU_SOURCE defines must exist in order to
enable certian glibc features. Therefore, it is common to define these
either in header files or on the command line.

Some platforms define these in their system headers. For example,
ESP-IDF defines this in their `features.h` file. This results in
a clash when blackmagic is compiled.

Check that these defines do not already exist prior to defining them.

### DIFF
--- a/src/include/general.h
+++ b/src/include/general.h
@@ -21,8 +21,12 @@
 #ifndef __GENERAL_H
 #define __GENERAL_H
 
+#ifndef _GNU_SOURCE
 #define _GNU_SOURCE
+#endif
+#ifndef _DEFAULT_SOURCE
 #define _DEFAULT_SOURCE
+#endif
 #if !defined(__USE_MINGW_ANSI_STDIO)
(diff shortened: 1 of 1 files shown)


dragonmux (approved):

LGTM, though we think this might just a patch over a deeper problem - that being we depend on these regardless of platform, so should probably just define them in the build system and remove all defines from the source as we're doing the work of the build system in code here.

Qyriad:
> we think this might just a patch over a deeper problem

iirc -std=gnuXX automatically defined _GNU_SOURCE and friends, so it might be that #1121 didn't fully remove all dependency on GNU C

dragonmux:
glibc and newlibc are structured in a way where, in order to use the sisprintf(), etc (which are the versions that BMP uses) one has to define they want the GNU-specific library functions, which is what _GNU_SOURCE is doing in this case. That stands separate from the standards compliance side of things.

PR #1127, opened 2022-07-21
gdb_main: Fixed a small DEBUG_GDB regression for vFlashErase and vFlashWrite debugging

dragonmux:
A small amount of fallout from #1125 - we accidentally regressed the ability to debug flash erase and write because of a nano newlibc bug whereby they don't support the size_t size modifier in their implementation of sscanf() or sprintf(). Our change to use size_t more and print with %zX therefore makes for problems.

This PR fixes that regression. This needs merging to both main and v1.8


esden (approved):
LGTM

PR #1129, opened 2022-07-22
black pill with stm32f401cc

koendv:
Linker script for black pill with stm32f401cc processor (256kbyte flash, 64kbyte ram).
Build with:
```
make PROBE_HOST=blackpillv2 BLACKPILL_F401CC=1
```

### DIFF
--- a/src/platforms/stm32/blackpillv2.ld
+++ b/src/platforms/stm32/blackpillv2.ld
@@ -20,8 +20,8 @@
 /* Define memory regions. */
 MEMORY
 {
-	rom (rx) : ORIGIN = 0x08000000, LENGTH = 512K
-	ram (rwx) : ORIGIN = 0x20000000, LENGTH = 96K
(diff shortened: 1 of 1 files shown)


esden:
Hi @koendv, thank you for the contribution.

Unfortunately this is not the correct solution. Making define based variations on target devices is not the way to do this. You either completely copy the blackpillv2 and give it a distinct name or make your private local variation for the device. Defines make it impossible to automatically test if the target still compiles cleanly.

At this point I would almost be ok with the approach from #1048 where we use the least common denominator. Also take a look at the changes planned for V2 of BMD.

Cheers,
Piotr

koendv:
OK. Is this admissible: lowering flash to 256k, and ram to 64k?
Ought to work in both f401cc and f411.

koen

koendv:
OK, forget it.
koen

PR #1131, opened 2022-07-25
Feature: JTAG TAP RISC-V prep

dragonmux:
This PR finishes the cleanup of the JTAG TAP code and adds a couple of new things needed for supporting buggy targets and RISC-V debugging.

The two new items this introduces are:

* Support for running the JTAG clock for a number of cycles with fixed TDI and TMS pin values
* Support for skipping or dwelling in the JTAG Idle state as required

This should mark the last of the RISC-V support setup.


esden (approved):
LGTM

PR #1132, opened 2022-07-25
target: fix probing for lpc1343

### LINKED ISSUE #912: Add support for LPC1343
Please report Designer   0 and Partno   0 and the probed device
Available Targets:
No. Att Driver
*** 1 Unknown ARM Cortex-M Designer   0 Partno   0 M3

perigoso:
I have no means to test this, but looking at the symptoms, and the git history, I believe this should fix it

fix #912

### DIFF
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -464,6 +464,7 @@ bool cortexm_probe(ADIv5_AP_t *ap)
 			PROBE(ch32f1_probe);
 			PROBE(stm32f1_probe); /* Care for other STM32F1 clones (?) */
 			PROBE(lpc15xx_probe); /* Thanks to JojoS for testing */
+			PROBE(lpc11xx_probe); /* LPC1343 */
 		} else if (ap->ap_partno == 0x471)  { /* Cortex-M0 ROM */
 			PROBE(lpc11xx_probe); /* LPC24C11 */
 			PROBE(lpc43xx_probe);


dragonmux (approved):
LGTM, it's such a trivial patch and put last in the Cortex-M3 ROM list that.. if it's wrong we can just pull it back out so it's no big deal. If it's right (which seems more likely), we fix a bug/regression :)

Merging.

oleg-login:
Hello is problem LPC1343,

(gdb) monitor swdp_scan
Target voltage: 3.3V
Please report probed device with Designer code 0x8000 and Partno 0x  0
Available Targets:
No. Att Driver
*** 1 Unknown ARM Cortex-M Designer 0x8000 Partno 0x000 M3
(gdb)

dragonmux:
Thank you for letting us know, could you please open a new issue ticket with that information and we'll copy the designer code out and deeper dive both this patch and what's going on with detection.

perigoso:
this might have to do with the designer code changes I made, before this device was showing designer code 0x0, but now its 0x8000, which would mark it as a legacy ascii code, I bet that blocks some code down the line

perigoso:
neverming, that is not checked in the default case, what's weird is theres no partno, and that is what is checked for M3 ROM where I placed the probe

oleg-login:
> Hello is problem LPC1343,
> 
> (gdb) monitor swdp_scan Target voltage: 3.3V Please report probed device with Designer code 0x8000 and Partno 0x 0 Available Targets: No. Att Driver *** 1 Unknown ARM Cortex-M Designer 0x8000 Partno 0x000 M3 (gdb)
> 
> Earlier, when there was no check ap->designer_code and ap->ap_partno, everything worked well.

ok

PR #1133, opened 2022-07-26
Feature: renesas ra support

perigoso:
This adds support for Renesas RA family of microcontrollers

What is not implemented:
- debug unlock, these devices have a feature very similar to "back access key" on Kinetis parts, this is done on a secondary AP (I think that code needs some work before I try to implement it)
- MF3 flash type, there are multiple types of flash peripheral on this family, only the "RV40" types have been implemented, it's also the only I have the means to test
- IDCODE for most devices is unknown, a method to "brute force" detection is in place

~This is not quite ready, as right now, although completely functional (for RV40), writing to flash is extremely(unusably so) slow, and needs optimization~

debugging looks to work, but testing was superficial

~I'm opening the PR as I've run out of spoons to work on the missing bits~

### DIFF
--- a/src/target/renesas.c
+++ b/src/target/renesas.c
@@ -1,0 +1,583 @@
+/*
+ * This file is part of the Black Magic Debug project.
+ *
+ * Copyright (C) 2022 1BitSquared <info@1bitsquared.com>
+ * Written by Rafael Silva <perigoso@riseup.net>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its
+ *    contributors may be used to endorse or promote products derived from
+ *    this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/* Support for Renesas RA family of microcontrollers (Arm Core) */
+
+#include "general.h"
+#include "target.h"
+#include "target_internal.h"
+#include "cortexm.h"
+#include "adiv5.h"
+
+#define RENESAS_PARTID_RA6M2 0x0150
+#define RENESAS_PARTID_RA4M3 0x0310
+
+/*
+ * Part numbering scheme
+ *
+ *  R7   F   A   xx   x   x   x   x   x   xx
+ * \__/ \_/ \_/ \__/ \_/ \_/ \_/ \_/ \_/ \__/
+ *  |    |   |   |    |   |   |   |   |   |
+ *  |    |   |   |    |   |   |   |   |   \_ Package type
+ *  |    |   |   |    |   |   |   |   \_____ Quality Grade
+ *  |    |   |   |    |   |   |   \_________ Operating temperature
+ *  |    |   |   |    |   |   \_____________ Code flash memory size
+ *  |    |   |   |    |   \_________________ Feature set
+ *  |    |   |   |    \_____________________ Group number
+ *  |    |   |   \__________________________ Series name
+ *  |    |   \______________________________ family (A: RA)
+ *  |    \__________________________________ Flash memory
+ *  \_______________________________________ Renesas microcontroller (always 'R7')
+ *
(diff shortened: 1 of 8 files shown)


perigoso:
I think I added way too many checks in the flash routines, checking for alignment, as is the bmp code already writes and erases aligned correct? @dragonmux

dragonmux:
So, the range erase command does do aligned erase, but it's /possible/ for GDB to issue an unaligned erase and write which is never corrected for in the GDB protocol machinery.. however, given how target_flash_erase and target_flash_write work, they should be changed to ensure alignment (we'll do this in a separate PR) which then means targets never need worry they'll get unaligned addresses

perigoso:
Build on top of #1205

perigoso:
I should test this once more and port the logs, but it should be good for review

src/target/renesas.c
```diff
@@ -0,0 +1,938 @@
...
+
+#define RV40_FCPSR         (RV40_BASE + 0xE0U)
+#define RV40_FCPSR_ESUSPMD 1U
+
+#define RENESAS_CF_END (0x00200000UL) /* End of Flash (maximum possible accross families) */
+
+static bool renesas_uid(target *t, int argc, const char **argv);
+
+const struct command_s renesas_cmd_list[] = {
+	{"uid", (cmd_handler)renesas_uid, "Prints unique number"},
```
dragonmux, line 281: This cast should be removed - the function signature should match the cmd_handler type

src/target/renesas.c
```diff
@@ -0,0 +1,938 @@
...
+	{NULL, NULL, NULL},
+};
+
+struct renesas_priv_s {
```
dragonmux, line 285: Could this be turned into a typedef using the _s suffix for the typedef please?

src/target/renesas.c
```diff
@@ -0,0 +1,938 @@
...
+}
+
+static void renesas_uid_read(target *t, const uint32_t base, uint8_t *uid)
+{
+	uint32_t uidr[4];
+	for (size_t i = 0U; i < 4U; i++)
+		uidr[i] = target_mem_read32(t, base + i * 4U);
+
+	for (size_t i = 0U; i < 16U; i++)
+		uid[i] = uidr[i / 4U] >> (i % 4U) * 8U;
```
dragonmux, line 303: To ensure this stays cheap, could the `% 4U` be replaced with `& 3U`? We're aware this makes the intention a little less obvious but it avoids the possibility of accidentally invoking the rather expensive remainder polyfill routine

src/target/renesas.c
```diff
@@ -0,0 +1,938 @@
...
+{
+	uint32_t pnrr[4];
+	for (size_t i = 0U; i < 4U; i++)
+		pnrr[i] = target_mem_read32(t, base + i * 4U);
+
+	if (base == RENESAS_FIXED1_PNR) {
+		/* Renesas... look what you made me do...  */
+		/* see 'Part numbering scheme' note for context */
+		for (size_t i = 0U; i < 13U; i++)
+			pnr[i] = pnrr[3U - (i + 3U) / 4U] >> (24U - ((i + 3U) % 4U) * 8U);
```
dragonmux, line 316: Same query as above

src/target/renesas.c
```diff
@@ -0,0 +1,938 @@
...
+		series = (series << 8) | pnr[PNR_FAMILY_INDEX + i];
+
+	return (pnr_series_t)series;
+}
+
+static uint32_t renesas_flash_size(const uint8_t *pnr)
+{
+	switch (pnr[PNR_MEMSIZE_INDEX]) {
+	case PNR_MEMSIZE_16KB:
+		return 16UL * 1024UL;
```
dragonmux, line 340: Rather than suffixing the individual constant values, this looks like a good candidate for expressing as `UINT32_C(16 * 1024)`

src/target/renesas.c
```diff
@@ -0,0 +1,938 @@
...
+	target_mem_write8(t, SYSC_FWEPROR, SYSC_FWEPROR_PERMIT);
+
+	return 0;
+}
+
+typedef enum pe_mode_e {
+	PE_MODE_READ,
+	PE_MODE_CF,
+	PE_MODE_DF,
+} pe_mode_t;
```
dragonmux, line 380: Please use `pe_mode_e` here and remove the suffix from the base enum name

src/target/renesas.c
```diff
@@ -0,0 +1,938 @@
...
+		platform_timeout_set(&timeout, 10);
+
+		/* Wait until the operation has completed or timeout */
+		/* Read FRDY bit until it has been set to 1 indicating that the current  operation is complete.*/
+		while (!(target_mem_read32(t, RV40_FSTATR) & RV40_FSTATR_RDY)) {
+			if (target_check_error(t) || platform_timeout_is_expired(&timeout))
+				return error;
+		}
+
+		if (target_mem_read8(t, RV40_FASTAT) & RV40_FASTAT_CMDLK) {
```
dragonmux, line 470: The braces here are extra and should be removed

src/target/renesas.c
```diff
@@ -0,0 +1,938 @@
...
+
+		return false;
+	}
+
+	/* code flash or data flash operation */
+	const bool code_flash = f->start < RENESAS_CF_END;
+
+	/* Transition to PE mode */
+	const pe_mode_t pe_mode = code_flash ? PE_MODE_CF : PE_MODE_DF;
+	if (!renesas_rv40_pe_mode(t, pe_mode) || renesas_rv40_error_check(t, RV40_FSTATR_ILGLERR))
```
dragonmux, line 494: This looks like it can be return elided? `return renesas_rv40_pe_mode(t, pe_mode) && !renesas_rv40_error_check(t, RV40_FSTATR_ILGLERR);`

src/target/renesas.c
```diff
@@ -0,0 +1,938 @@
...
+
+	return true;
+}
+
+static bool renesas_rv40_done(target_flash_s *f)
+{
+	target *t = f->t;
+
+	/* return to read mode */
+	if (!renesas_rv40_pe_mode(t, PE_MODE_READ))
```
dragonmux, line 505: This also looks like a prime target for return elision

src/target/renesas.c
```diff
@@ -0,0 +1,938 @@
...
+
+		/* wait until the operation has completed or timeout */
+		/* read FRDY bit until it has been set to 1 indicating that the current operation is complete.*/
+		while (!(target_mem_read32(t, RV40_FSTATR) & RV40_FSTATR_RDY)) {
+			if (target_check_error(t) || platform_timeout_is_expired(&timeout))
+				return -1;
+		}
+	}
+
+	if (renesas_rv40_error_check(t, RV40_FSTATR_PRGERR | RV40_FSTATR_ILGLERR))
```
dragonmux, line 609: This looks like another good target for return elision if we first merge the Flash return type PR before this one

dragonmux (changes requested):
This is looking fantastic, quite excited to get to merging this! There are a few fairly small feedback items found and we're going to suggest merging #1211 before this so you can benefit from the cleaner Flash function return type interface here and not have to re-do work or have work re-done with a follow-up in that PR or another.

src/target/renesas.c
dragonmux: This one got missed

dragonmux (commented):
Looking very good. There was one item from the previous review that got missed (highlighted above). Once that's been fixed, we'll merge this.

src/target/renesas.c
perigoso: oops

dragonmux (approved):
LGTM, merging.

Issue #1134, opened 2022-07-26
ADIv5: APs badly enumerated


ylm:
BMP bails out if the first AP is not supported. All 8 possible APs should be checked instead.
[This line](https://github.com/blackmagic-debug/blackmagic/blob/6f81cb6a68a79efb3d29b1e0ea463557d51fd23b/src/target/adiv5.c#L793) should check if void_aps equals 8 instead.

dragonmux:
Thank you for reporting this issue. This is indeed a regression and we have now opened a PR that fixes this buggy behaviour

PR #1135, opened 2022-07-26
Fix: ADIv5 AP scan early bailout

dragonmux:
the ADIv5 AP scan bails out too early (on the first encountered invalid AP) rather than after 8 invalid APs.

This was a regression introduced by cda83d30 because that commit checks the wrong loop variable, and then checks the variable against the wrong value. We also fixed the nomenclature up as the code was a litte hard to understand.

This PR fixes this behaviour, fixing #1134.


perigoso (approved):
LGTM

esden (approved):
LGTM

PR #1136, opened 2022-07-26
Feature: LPC82x low power reset

### LINKED ISSUE #1126: Manual tdi control for lpc82x bootloader
lpc82x (and probably other LPC series) have an embedded bootloader that is enabled when a GPIO is pulled low on boot. This happens to be the same GPIO pin mapped to TDI.

If using a connector wired for full JTAG but connecting via SWD, the BMP will by default pull TDI low, which will make the lpc82x always execute its bootloader. However, running a jtag_scan first will leave the TDI pin in a high state, allowing normal program execution.

This is actually sort of handy, because entering the bootloader this way seems to be the only way to recover a device that enters power-down mode where clocks have stopped. It would be nice to have it be an intentional monitor command though, rather than happening by chance if jtag_scan is run.

dragonmux:
This PR implements a new command for doing resets while holding pins in the JTAG interface in specific state. The reason this is necessary is given by LPC82x processors where if in a deep low-power mode do not have running clocks but if reset with TDI held low, will wind up in their bootloaders, allowing BMP to scan for them and attach.

This fixes #1126

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -33,6 +33,7 @@
 #include "morse.h"
 #include "version.h"
 #include "serialno.h"
+#include "jtagtap.h"
 
 #ifdef ENABLE_RTT
 #include "rtt.h"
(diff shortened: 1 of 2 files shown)


esden (approved):
LGTM, Merging. 

With one thing to keep in mind going forward. We should be cautious of overwhelming the users with too many monitor commands. We should start thinking of ways to combine those commands where reasonable and splitting the help output when appropriate. :)

PR #1137, opened 2022-07-26
Feature: monitor command reformatting and cleanup

dragonmux:
In this PR we address the state of command.c with its various styles, formatting choices, type badness and general inconsistencies.

This PR cleans up the commands and improves their output to GDB to make for a nicer user experience while doing what it can to streamline the code to claw back a bit of space.

Tested working on both v2.1e and v2.3a hardware, and against TM4C and LPC43xx targets.

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -587,19 +587,17 @@ static bool cmd_traceswo(target *t, int argc, const char **argv)
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


esden (approved):
All LGTM.

One note, as already mentioned through "sideband". We should change the way we declare struct types so that it is bit more ergonomic using `_s` suffix. That would then apply to enum `_e`, union `_u` and so on. But it can be addressed with a separate spring pass cleanup PR. :)

PR #1143, opened 2022-07-28
raspberry magic number begone

perigoso:
a second round of cleanup on adi code, this fixes the misleading idcode naming, we were actually reading the DP DPIDR register, it also fixes the JEP106 code for raspberry, and the handling of designer code in a couple registers (the reason for the wrong jep code)

### DIFF
--- a/src/target/adiv5.h
+++ b/src/target/adiv5.h
@@ -46,7 +46,7 @@
 #define ADIV5_DP_RDBUFF    ADIV5_DP_REG(0xCU)
 #define ADIV5_DP_TARGETSEL ADIV5_DP_REG(0xCU)
 
-/* AP DPIDR */
+/* DP DPIDR */
 #define ADIV5_DP_DPIDR_REVISION_OFFSET 28U
 #define ADIV5_DP_DPIDR_REVISION_MASK   (0xfU << ADIV5_DP_DPIDR_VERSION_OFFSET)
 #define ADIV5_DP_DPIDR_PARTNO_OFFSET   20U
@@ -59,6 +59,21 @@
 #define ADIV5_DP_DPIDR_VERSION_DPv2    (2U << ADIV5_DP_DPIDR_VERSION_OFFSET)
 #define ADIV5_DP_DPIDR_DESIGNER_OFFSET 1U
 #define ADIV5_DP_DPIDR_DESIGNER_MASK   (0x7ffU << ADIV5_DP_DPIDR_DESIGNER_OFFSET)
+
+/* DP TARGETID */
+#define ADIV5_DP_TARGETID_TREVISION_OFFSET 28U
+#define ADIV5_DP_TARGETID_TREVISION_MASK   (0xfU << ADIV5_DP_TARGETID_TREVISION_OFFSET)
+#define ADIV5_DP_TARGETID_TPARTNO_OFFSET   12U
+#define ADIV5_DP_TARGETID_TPARTNO_MASK     (0xffffU << ADIV5_DP_TARGETID_TPARTNO_OFFSET)
+#define ADIV5_DP_TARGETID_TDESIGNER_OFFSET 1U
+#define ADIV5_DP_TARGETID_TDESIGNER_MASK   (0x7ffU << ADIV5_DP_TARGETID_TDESIGNER_OFFSET)
+
+/* DP DPIDR/TARGETID DESIGNER */
+/* Bits 10:7 - JEP-106 Continuation code */
+/* Bits 6:0 - JEP-106 Identity code */
+#define ADIV5_DP_DESIGNER_JEP106_CONT_OFFSET 7U
+#define ADIV5_DP_DESIGNER_JEP106_CONT_MASK   (0xfU << ADIV5_DP_DESIGNER_JEP106_CONT_OFFSET)
+#define ADIV5_DP_DESIGNER_JEP106_CODE_MASK   (0x7fU)
 
 /* AP Abort Register (ABORT) */
 /* Bits 31:5 - Reserved */
@@ -167,7 +182,7 @@
 #define JEP106_MANUFACTURER_SPECULAR     0x501U /* LPC845 with code 501. Strange!? Specular Networks */
 #define JEP106_MANUFACTURER_ENERGY_MICRO 0x673U /* Energy Micro */
 #define JEP106_MANUFACTURER_GIGADEVICE   0x751U /* GigaDevice */
-#define JEP106_MANUFACTURER_RASPBERRY    0x927U /* Raspberry Pi */
+#define JEP106_MANUFACTURER_RASPBERRY    0x913U /* Raspberry Pi */
 
 /*
  * This code is not listed in the JEP106 standard, but is used by some stm32f1 clones
@@ -205,6 +220,10 @@ typedef struct ADIv5_DP_s {
 	int refcnt;
 
 	uint32_t debug_port_id;
+
+	uint16_t designer_code;
+	uint16_t partno;
+
 	uint32_t targetid; /* Contains IDCODE for DPv2 devices.*/
 
 	void (*seq_out)(uint32_t tms_states, size_t clock_cycles);
--- a/src/target/adiv5.h
+++ b/src/target/adiv5.h
@@ -38,7 +38,7 @@
 #define ADIV5_DP_BANK4 0x40U
 
 /* ADIv5 DP Register addresses */
-#define ADIV5_DP_IDCODE    ADIV5_DP_REG(0x0U)
+#define ADIV5_DP_DPIDR     ADIV5_DP_REG(0x0U)
 #define ADIV5_DP_ABORT     ADIV5_DP_REG(0x0U)
 #define ADIV5_DP_CTRLSTAT  ADIV5_DP_REG(0x4U)
 #define ADIV5_DP_TARGETID  (ADIV5_DP_BANK2 | ADIV5_DP_REG(0x4U))
@@ -204,7 +204,7 @@ typedef struct ADIv5_AP_s ADIv5_AP_t;
 typedef struct ADIv5_DP_s {
 	int refcnt;
 
-	uint32_t idcode;
+	uint32_t debug_port_id;
 	uint32_t targetid; /* Contains IDCODE for DPv2 devices.*/
 
 	void (*seq_out)(uint32_t tms_states, size_t clock_cycles);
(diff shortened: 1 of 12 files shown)


dragonmux:
We'll let you figure out why the build broke before we review this, and we'll make sure to test against our RP2040 to prevent any possible regressions we might expect

perigoso:
should be good, added a couple commits that changed/removed some superfluous checks

I also want to rename t->idcode to maybe t->partno, as it actually inherits the ap->partno, nothing to do with idcode, but I'll leave that for another day

src/target/cortexm.c
```diff
@@ -638,13 +640,14 @@ static void cortexm_regs_write(target *t, const void *data)
 			regs++;
 		}
 		if (t->target_options & TOPT_FLAVOUR_V7MF)
-			for(size_t z = 0; z < sizeof(regnum_cortex_mf) / 4; z++) {
+			for (size_t z = 0; z < sizeof(regnum_cortex_mf) / 4; z++) {
 				ap->dp->ap_reg_write(ap, regnum_cortex_mf[z], *regs);
 				regs++;
 			}
 	}
 #else
```
dragonmux, line 648: This `#else` branch should be eliminated.

src/target/cortexm.c
```diff
@@ -638,13 +640,14 @@ static void cortexm_regs_write(target *t, const void *data)
...
-	if (0) {}
+	if (0) {
+	}
 #endif
 	else {
```
dragonmux, line 652: this `else` should be in the `#if` block. The {}'s are valid regardless

src/target/cortexm.c
```diff
@@ -589,42 +594,39 @@ static void cortexm_regs_read(target *t, void *data)
...
 		ap->dp->ap_regs_read(ap, base_regs);
-		for(i = 0; i < sizeof(regnum_cortex_m) / 4; i++)
+		for (i = 0; i < sizeof(regnum_cortex_m) / 4; i++)
 			*regs++ = base_regs[regnum_cortex_m[i]];
 		if (t->target_options & TOPT_FLAVOUR_V7MF)
-			for(i = 0; i < sizeof(regnum_cortex_mf) / 4; i++)
+			for (i = 0; i < sizeof(regnum_cortex_mf) / 4; i++)
 				*regs++ = ap->dp->ap_reg_read(ap, regnum_cortex_mf[i]);
 	}
 #else
```
dragonmux, line 603: This `#else` block should be eliminated

src/target/cortexm.c
```diff
@@ -589,42 +594,39 @@ static void cortexm_regs_read(target *t, void *data)
...
-	if (0) {}
+	if (0) {
+	}
 #endif
 	else {
```
dragonmux, line 607: This `else` should be in the above `#if` block

src/target/cortexm.c
```diff
@@ -333,14 +329,12 @@ bool cortexm_probe(ADIv5_AP_t *ap)
 		t->core = "M0";
 		break;
 	default:
-		if (ap->designer_code != JEP106_MANUFACTURER_ATMEL) /* Protected Atmel device?*/{
+		if (ap->designer_code != JEP106_MANUFACTURER_ATMEL) /* Protected Atmel device?*/ {
```
dragonmux, line 332: The braces around the body of this if block should be eliminated

src/target/cortexm.c
```diff
@@ -118,103 +117,101 @@ struct cortexm_priv {
...
-	"    <reg name=\"d9\" bitsize=\"64\" type=\"float\"/>"
-	"    <reg name=\"d10\" bitsize=\"64\" type=\"float\"/>"
-	"    <reg name=\"d11\" bitsize=\"64\" type=\"float\"/>"
-	"    <reg name=\"d12\" bitsize=\"64\" type=\"float\"/>"
-	"    <reg name=\"d13\" bitsize=\"64\" type=\"float\"/>"
-	"    <reg name=\"d14\" bitsize=\"64\" type=\"float\"/>"
-	"    <reg name=\"d15\" bitsize=\"64\" type=\"float\"/>"
-	"  </feature>"
-	"</target>";
+static const char tdesc_cortex_m[] = "<?xml version=\"1.0\"?>"
```
dragonmux, line 135: Suggest putting this in a `/* clang-format: off */` block and formatting it so the indentation is in the source but not inside the strings. This will save us some space int the firmware and make it easier to read

src/target/cortexm.c
```diff
@@ -68,12 +68,11 @@ const struct command_s cortexm_cmd_list[] = {
...
 	{"redirect_stdout", (cmd_handler)cortexm_redirect_stdout, "Redirect semihosting stdout to USB UART"},
 #endif
-	{NULL, NULL, NULL}
-};
+	{NULL, NULL, NULL}};
 
 /* target options recognised by the Cortex-M target */
-#define	TOPT_FLAVOUR_V6M	(1<<0)	/* if not set, target is assumed to be v7m */
-#define	TOPT_FLAVOUR_V7MF	(1<<1)	/* if set, floating-point enabled. */
+#define TOPT_FLAVOUR_V6M  (1 << 0) /* if not set, target is assumed to be v7m */
```
dragonmux, line 74: `U` on these constants please

dragonmux (changes requested):
This is already looking really good - unfortunate that in reformatting `cortexm.c`, it makes our review look really harsh, but most of the points raised should be simple fixes.

We're happy with how this does what it does and will test and merge once the major review points are addressed.

dragonmux:
Going to suggest that renaming idcode -> target_id should be a separate PR based on this one. We'd like to get this merged and see how it plays out with the other big 6, keeping concerns separated as best as possible (it's a good change though and lends some much needed clarity to the code)

src/target/adiv5.h
```diff
@@ -204,7 +219,9 @@ typedef struct ADIv5_AP_s ADIv5_AP_t;
 typedef struct ADIv5_DP_s {
 	int refcnt;
 
-	uint32_t idcode;
+	uint32_t dpidr;
```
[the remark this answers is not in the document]
perigoso: DPIDR is actually part of the adi spec, `DPIDR, Debug Port Identification Register`, but i agree it's not immediately clear

src/target/cortexm.c
```diff
@@ -1587,46 +1636,51 @@ static int cortexm_hostio_request(target *t)
...
-		}
+	}
 
 	case SYS_READC: { /* readc */
-		uint8_t ch='?';
+		uint8_t ch = '?';
 		//DEBUG("SYS_READC ch addr %p\n", &ch);
-		void (*saved_mem_read)(target *t, void *dest, target_addr src, size_t len);
-		void (*saved_mem_write)(target *t, target_addr dest, const void *src, size_t len);
+		void (*saved_mem_read)(target * t, void *dest, target_addr src, size_t len);
```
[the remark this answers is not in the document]
perigoso: uh, yeah that doesnt feel right

src/target/adiv5.h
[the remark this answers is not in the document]
dragonmux: Ah, sorry, should have made ourself clearer.. it's not in the ARMv7-M TRM or the Cortex-M4 TRM (for example), so you'd have to be looking at just the right spec to find the term, which is probably not the spec you're reading if you're trying to find out how values are encoded and at what addresses they can be found (that's in the ARMv7 TRM debugging section, and core-specific TRMs)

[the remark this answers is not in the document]
perigoso: right, I was looking at the ADIv5 and ADIv6 specs, debu_port_id sounds good enough, as long as its doesnt imply something like the idcode did

src/target/cortexm.c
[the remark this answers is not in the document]
dragonmux: The other instance of clang-format doing things we don't want was in `command.c` where it tried to put the command array initialiser closing bracket on the same line as the sentinel, and then also put the first item of the array on the same line as the array's defined and opened.. need to figure out why and fix it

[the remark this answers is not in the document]
perigoso: I manually reverted back this format, but future formats may try to do it again

src/target/cortexm.c
```diff
@@ -1179,30 +1163,30 @@ static bool cortexm_vector_catch(target *t, int argc, char *argv[])
...
-#define SYS_RENAME	0x0F
-#define SYS_SEEK	0x0A
-#define SYS_SYSTEM	0x12
-#define SYS_TICKFREQ	0x31
-#define SYS_TIME	0x11
-#define SYS_TMPNAM	0x0D
-#define SYS_WRITE	0x05
-#define SYS_WRITEC	0x03
-#define SYS_WRITE0	0x04
+#define SYS_CLOCK         0x10
```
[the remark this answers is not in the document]
perigoso: I added an extra commit moving these, it made sense

src/target/cortexm.c
```diff
@@ -1162,7 +1146,7 @@ static bool cortexm_vector_catch(target *t, int argc, char *argv[])
 
 /* Windows defines this with some other meaning... */
 #ifdef SYS_OPEN
```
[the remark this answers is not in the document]
perigoso: renamed them to SEMIHOSTING_SYS_etc to avoid conflicts

src/target/cortexm.c
perigoso: fyi `/* clang-format: off */` doesnt work, must be `/* clang-format off */` TIL too

src/target/cortexm.c
```diff
@@ -1244,31 +1186,33 @@ static int cortexm_hostio_request(target *t)
...
+		/* code that runs in pc-hosted process. use linux system calls. */
 
-	case SYS_OPEN:{	/* open */
-		target_addr fnam_taddr =  params[0];
+	case SEMIHOSTING_SYS_OPEN: { /* open */
+		target_addr fnam_taddr = params[0];
 		uint32_t fnam_len = params[2];
 		ret = -1;
-		if ((fnam_taddr == TARGET_NULL) || (fnam_len == 0)) break;
+		if ((fnam_taddr == TARGET_NULL) || (fnam_len == 0))
```
dragonmux, line 1204: The extra perens should be removed

dragonmux (changes requested):
Spotted just one last thing that needs fixing but otherwise this LGTM and is ready for merge. We'll test this against our RPi Pico and approve for merge once the one item is fixed.

dragonmux (dismissed):
LGTM, we'll merge this once the 6 PRs that are ahead of this are in and we've been able to properly test this

perigoso:
Tested probe (hosted and native) with success on:
- Nordic nRF52
- SAMS70J19A
- Atmel SAM3U
- 1 Unknown ARM Cortex-M Designer 0x423 Part ID 0x150 (Renesas RA6M2)
- RP2040

dragonmux:
Please rebase this then we'll merge it.

dragonmux (approved):
LGTM! Testing against Tiva-C then merging.

PR #1139, opened 2022-07-27
Fix: Tiva-C/Stelaris device probing issues

dragonmux:
In this PR we seek to address issues probing for TI Tiva-C and Stelaris devices which, because of where the probe routine is in the sequence, get put into a bad state and fail to scan under certain circumstances by the time the routine fires.

In addressing this we have inadvertantly ended up diving into the LM3S datasheets and adding support for a couple of new micros + fixing the Flash and RAM regions for another.

This fixes #958

PR #1144, opened 2022-07-29
fix: Use standard macros for debug message formating

ylm:


dragonmux:
Thank you for the contribution! Unfortunately, this is a duplicate of the ADIv5 cleanup commit in #1139 - certainly can't blame you for not knowing that though.

For reasons to do with the way the constant is presently defined, formatting it actually has to be done with `"%lX"` rather than `"%" PRIx32`- If you alter this PR to make the definition of the constant use the `UINT32_C` macro to eliminate this problem, we'll take this change over our own.

ylm:
Would it make sense for me to cherry-pick your ADIv5 commit and include it as part of this PR for thoroughness as it's not a complete overlap?

dragonmux:
yes, that does make sense.

ylm:
That should do it

ylm:
Thanks a lot for the code review

src/target/adiv5.c
```diff
@@ -64,9 +64,9 @@
  *                                 V
  *                           Component ID
  */
-#define CID_PREAMBLE    0xB105000DUL
-#define CID_CLASS_MASK  0x0000F000UL
-#define CID_CLASS_SHIFT 12UL
+#define CID_PREAMBLE    UINT32_C(0xB105000D)
+#define CID_CLASS_MASK  UINT32_C(0x0000F000)
+#define CID_CLASS_SHIFT UINT32_C(12)
```
dragonmux, line 69: This one can and should just be `12U`

dragonmux (approved):
LGTM, there's one advisory item but this is otherwise ready for merge

dragonmux (approved):
LGTM, merging.

PR #1141, opened 2022-07-28
Fix/lpc1143 probe

perigoso:
new attempt at fixing #912 / #1140

PR #1146, opened 2022-07-29
Rework/cleanup TARGETID handling

perigoso:
built on top of #1143 mind the extra commits

This tries to improve readabilty and correctness in the initialization of the dp and the reading and usage of targetid

This needs some more thorough testing, but so far it looks good, on some embedded jlink targets target id is now being read where it was not before

changed the logic in the target probe "selection" including the change mentioned on discord

~~note: i think we can, and should, safely move the target id reading inside the dp_init method, but I'll leave that for a future cleanup~~

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -359,8 +359,8 @@ static void display_target(int i, target *t, void *context)
 	const char attached = target_attached(t) ? '*' : ' ';
 	const char *const core_name = target_core_name(t);
 	if (!strcmp(target_driver_name(t), "ARM Cortex-M"))
-		gdb_outf("***%2d %c Unknown %s Designer 0x%03x Partno 0x%03x %s\n", i, attached, target_driver_name(t),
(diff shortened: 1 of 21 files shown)


perigoso:
This PR also supersedes #1141, so that may be closed if this is to be merged

perigoso:
Tested probe (hosted and native) with success on:
- Nordic nRF52
- SAMS70J19A
- Atmel SAM3U
- 1 Unknown ARM Cortex-M Designer 0x423 Part ID 0x150 (Renesas RA6M2)
- RP2040

dragonmux:
Please rebase this then we'll review and merge it.

perigoso:
ready.

perigoso:
should be good to go right after 1143 merge

dragonmux (approved):
LGTM! We'll merge once we've tested against our Tiva-C target too

PR #1147, opened 2022-07-30
fix: Add nop instruction for timing purpose

ylm:
Adding a nop instruction in the no_delay TMS sequence routine fixes timing on the f4discovery platform


dragonmux (approved):
LGTM, merging.

PR #1148, opened 2022-07-30
Fixed compatibility issues with CH32F1 driver and other STM32F103 clones

grumat:
I receive a new Bluepill+ board from WeAct Studios which has a APM32F103CBT6 µC part number (APEX/Geehy). But had an issue with BMP V1.8.0. Writing flash simply froze (GDB load command).
My board was identified as `CH32F1 medium density (stm32f1 clone) M3` which caused the issue, since this clone does not support the Fast Flash feature. Reading documentation, I saw that the Fast Flash Flag has a default value of **0** in a regular STM32 and also on the APM32Fxx.
Reading the source code, I noticed that this condition is not tested during the `probe` function.
The PR that I am posting solves the issue for me, but I have no CH32F part to test if the patch breaks something for this family.
Now my board is identified as `STM32F1 (clone) medium density M3` and works as expected.

### DIFF
--- a/src/target/ch32f1.c
+++ b/src/target/ch32f1.c
@@ -150,6 +150,12 @@ static int ch32f1_flash_lock(target *t)
 {
 	DEBUG_INFO("CH32: flash lock \n");
 	SET_CR(FLASH_CR_LOCK);
+	uint32_t cr = target_mem_read32(t, FLASH_CR);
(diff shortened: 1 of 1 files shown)


dragonmux (approved):
LGTM. The register read back should be read into a const value, and the flash lock function should return a bool, but those are cleanup items we can do separately. We'll merge this now.

PR #1149, opened 2022-07-30
Fix: ch32f1 cleanup

dragonmux:
This PR cleans up some type abuse and missing `static` in the CH32F1 code.

The cleanup was prompted by #1148 which fixes a bug in how CH32F1 parts are probed for vs other clones.


esden (approved):
LGTM

PR #1150, opened 2022-07-30
Supports for BlackPill AT32F403A

grumat:
While Chip crisis is ongoing one more common alternative developed by WeActStudio, the [BlackPill AT32F403A](https://github.com/WeActStudio/WeActStudio.BlackPill).
This PR adds support for the entire family of chips. This family may offer an _external Flash_, which requires custom flash code to be used. With this PR the interface common to STM32F103 allows for 256KB usage, which is quite a good start.
Tested with the referred board.

### DIFF
--- a/src/target/target_probe.h
+++ b/src/target/target_probe.h
@@ -37,6 +37,7 @@ bool efm32_aap_probe(ADIv5_AP_t *ap);
 bool rp_rescue_probe(ADIv5_AP_t *ap);
 
 bool ch32f1_probe(target *t); // will catch all the clones
+bool at32fxx_probe(target *t); // STM32 clones from Artery
 bool gd32f1_probe(target *t);
 bool stm32f1_probe(target *t);
 bool stm32f4_probe(target *t);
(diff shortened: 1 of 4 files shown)


dragonmux:
Thank you for the contribution! Before we dive into the code and do a full review, could we please ask you to update your default branch against main and rebase this PR to remove the merge commit?

Unfortunately we don't have any of the AT32F devices for now (though Esden has some AT32F103 clones on the way iwrc), so if you could provide some GDB output showing this working that would be helpful as a reference and to show it works.

src/target/target_probe.h
```diff
@@ -37,6 +37,7 @@ bool efm32_aap_probe(ADIv5_AP_t *ap);
 bool rp_rescue_probe(ADIv5_AP_t *ap);
 
 bool ch32f1_probe(target *t); // will catch all the clones
+bool at32fxx_probe(target *t); // STM32 clones from Artery
```
perigoso, line 40: Pleas add a probe alias on `target_probe.c`

grumat:
Sorry I am not used to git. Will try to figure out what this "rebase" stuff means. I'll update as soon as this is solved.

dragonmux:
The following two parts of the Git reference manual might help in explaining how rebasing works and what it is: https://git-scm.com/book/en/v2/Git-Branching-Rebasing and https://git-scm.com/docs/git-rebase

PR #1153, opened 2022-07-31
Fix: rp2040 code cleanup

dragonmux:
In this PR we re-order and clean up the RP2040 support code to make it easier to read and ordered/structured more like other existing targets.

This does include some clang-tidy warnings fixes and reduces code size by a few bytes, but is otherwise purely reorganisation and renaming work.

Tested working against our RP2040.


esden (approved):
LGTM

esden:
I tested this on my RP2040 board and all seems to be good as far as I can tell.

PR #1160, opened 2022-08-04
Feature: LPC43x0 external flash support

dragonmux:
In this PR we overhaul the LPC43xx support code and provide support for the external SPI Flash on processors such as the LPC4370.

This started out as just trying to implement LPC43x0 Flashless part support, but as we dove into the problem we found the implementation for LPC43xx processors was mostly a stub that only supported the LPC4337 with proper mappings. Due to a mix of errata and other factors, we've ended up mostly rewriting the LP43xx support so this should be considered the addition of a brand new family of targets.

We don't have any parts with on-chip Flash to test so we can't be sure the IAP-driven code is correct but it should be ok based on a read of the user manual.

We intend to land a follow-up PR for this in the future that makes use of the SFDP data where possible on the target's attached SPI Flash to automatically configure block and page size information, erase opcodes, and other information that depends on the exact Flash that's been attached.


dragonmux:
This PR now provides support for all LPC43x2/3/5/7 parts and both SPIFI and SSP0 boot modes for all LPC43x0 parts. Stubs are provided for eventual EMC 8-/16-/32-bit support, but in the interests of seeing this land any time soon, this mode has been omitted for now

dragonmux:
This PR is ready for review when you get a chance, @esden

esden (approved):
As the "support" was essentially just a stub I am optimistically merging this, even though we don't have user feedback for the parts in question. The reason is that, it is significantly more fleshed out than what we had before. And merging it will increase chances of users testing this code for the parts in question.

PR #1164, opened 2022-08-05
Fix: BMDA uses the wrong libusb

dragonmux:
This PR cleans up the CFLAGS and LDFLAGS building to use pkg-config to find libusb1 and corrects the pkg-config presence check


esden (approved):
LGTM

Issue #1159, opened 2022-08-04
stm32h750 target wrong part_id read as 0x4500 not 0x450


klenSA:
(gdb) tar ext /dev/kgp/bmp
Remote debugging using /dev/kgp/bmp
(gdb) mon auto
Target voltage: 0.0V
JTAG scan found no devices, trying SWD!
Please report unknown device with Designer 20 Part ID 450
Available Targets:
No. Att Driver
*** 1   Unknown ARM Cortex-M Designer 0x20 Part ID 0x4500 M7
(gdb)

dragonmux:
Thank you for reporting this. As shown in the message in the middle, the part ID is read out correctly (need to look into why it's formatted wrong for the driver string but that's a separate problem).

What it looks like is happening here is that the stm32h7_probe routine failed in some manner.
It's also interesting that your probe is reporting a target voltage of 0V

Can we confirm that this is with latest v1.8.2 firmware or main?

If you are able to do a debug-enabled build of the firmware, that would be very helpful in figuring out what's really going wrong here. To make such a build requires building with ENABLE_DEBUG=1 and then once Flashed, running `mon debug en` just before invoking auto-scan while monitoring the debug serial port with minicom or similar (with the project udev rules that looks like: `minicom -D /dev/ttyBmpTarg -8`).

perigoso:
This is very likely a regression of #1146

The different Part IDs can be explained by the fact that I missed a change from ´ap->partno' -> `t->part_id` (will PR a fixup in sec)
The Id actually being used to probe is `t->part_id` which is as of the new change 0x4500, and 0x450 would be the old ID (`ap->partno`)

Without more information we can't know for certain what is happening, but I can make an assumption I think is likely:
The `stm32h750` has a valid TARGETID register and as such the PARTNO present in it is being used to run the probe routine, but this PARTNO has one extra nibble over the one in the ap, I was hoping they were the same as to not break any probe routines, but it seems they're not, the first 3 nibbles match, it just has the extra nibble that shows as an extra 0 at the end

[...]

PR #1167, opened 2022-08-05
target/cortexm: print t->part_id not ap->partno in unknown device warning

perigoso:
this is the id used to probe, which may or may not be the same as the one on the ap

fixup related to #1159


dragonmux (approved):
LGTM, merging.

PR #1169, opened 2022-08-06
Fix: unify dpidr targetid

perigoso:
This Pr is an attempt at unifying the DPIDR and TARGETID register handling, which was being read on the lower level protocol before `adiv5_dp_init` this should hopefully make behavior more consistent in particular for JTAG-DP vs SWD-DP for the same target

This makes the assumption that all DPs will have architecture version 1 or higher, which may not be the case on JTAG scan routines, there's is a naive check that may catch dpv0 but is not confirmed to work in reality

I would be more comfortable merging this after some more testing, in particular with the probes I was unable to test with (STLink)

hosted JLINK:
```
DP DPIDR 0x2ba01477 (v1 rev0) designer 0x43b partno 0xba
  1      Nordic nRF52 M4
 ```
 ```
 DP DPIDR 0x2ba01477 (v1 rev0) designer 0x43b partno 0xba
  1      Atmel SAM3U M3
  ```
  
 hosted BMP
 ```
 DP DPIDR 0x2ba01477 (v1 rev0) designer 0x43b partno 0xba
  1      Nordic nRF52 M4
 ```
 
 ```
 DP DPIDR 0x2ba01477 (v1 rev0) designer 0x43b partno 0xba
  1      Atmel SAM3U M3
``` 

```
DP DPIDR 0x0bc12477 (v2 MINDP rev0) designer 0x43b partno 0xbc
TARGETID 0x01002927designer 0x913 partno 0x1002
DP DPIDR 0x0bc12477 (v2 MINDP rev0) designer 0x43b partno 0xbc
TARGETID 0x01002927designer 0x913 partno 0x1002
DP DPIDR 0x10212927 (v2 MINDP rev0) designer 0x913 partno 0x2
TARGETID 0x01002927designer 0x913 partno 0x1002
 1      Raspberry RP2040 M0+
 2      Raspberry RP2040 M0+
 3      Raspberry RP2040 Rescue (Attach to reset!) 
 ```
 
 native BMP
```
 No. Att Driver
 1      Raspberry RP2040 M0+
 2      Raspberry RP2040 M0+
 3      Raspberry RP2040 Rescue (Attach to reset!) 
```

### CODE REVIEW

**Comment by perigoso** [comment]
Multidrop scan may be able to be moved to a higher level too, but I left that mostly as is

File: src/target/adiv5_jtagdp.c [hunk 1]
```diff
@@ -29,76 +29,72 @@
...
 	jtag_dev_write_ir(&jtag_proc, dp->dp_jd_index, APnDP ? IR_APACC : IR_DPACC);
 
 	platform_timeout_set(&timeout, 250);
 	do {
-		jtag_dev_shift_dr(&jtag_proc, dp->dp_jd_index, (uint8_t*)&response,
-						  (uint8_t*)&request, 35);
+		jtag_dev_shift_dr(&jtag_proc, dp->dp_jd_index, (uint8_t *)&response, (uint8_t *)&request, 35);
 		ack = response & 0x07;
-	} while(!platform_timeout_is_expired(&timeout) && (ack == JTAGDP_ACK_WAIT));
+	} while (!platform_timeout_is_expired(&timeout) && (ack == JTAGDP_ACK_WAIT));
```
dragonmux, line 90: The extra parens should go

src/target/adiv5.c
```diff
@@ -682,44 +683,72 @@ static void rp_rescue_setup(ADIv5_DP_t *dp)
...
-	if (dp->version > 0) {
-		/* debug_port_id is valid DPIDR */
-		dp->partno = (dp->debug_port_id & ADIV5_DP_DPIDR_PARTNO_MASK) >> ADIV5_DP_DPIDR_PARTNO_OFFSET;
-		dp->mindp = !!(dp->debug_port_id & ADIV5_DP_DPIDR_MINDP);
-	} else {
-		/* called from JTAG scan, debug_port_id is JTAG TAP IDCODE */
-		dp->partno = (dp->debug_port_id & JTAG_IDCODE_PARTNO_MASK) >> JTAG_IDCODE_PARTNO_OFFSET;
-		dp->mindp = false;
-	}
+	dp->version = (dpidr & ADIV5_DP_DPIDR_VERSION_MASK) >> ADIV5_DP_DPIDR_VERSION_OFFSET;
```
dragonmux, line 719: We should probably leave a note on how to do the JTAG DPv0 ID code stuff

src/target/adiv5.c
```diff
@@ -743,8 +772,7 @@ void adiv5_dp_init(ADIv5_DP_t *dp)
 	dp->mem_write_sized = firmware_mem_write_sized;
 #endif
 
-	volatile uint32_t ctrlstat = 0;
-	volatile struct exception e;
+	volatile uint32_t ctrlstat;
```
dragonmux, line 775: Two points - why is this volatile, and why not = 0?

src/platforms/hosted/platform.c
```diff
@@ -241,65 +260,82 @@ int platform_jtag_dp_init(ADIv5_DP_t *dp)
...
+	case BMP_TYPE_NONE:
 		return "NONE";
-	  case BMP_TYPE_BMP:
+
+	case BMP_TYPE_BMP:
 		return "BMP";
-	  case BMP_TYPE_STLINKV2:
+
+	case BMP_TYPE_STLINKV2:
 		return "STLINKV2";
```
dragonmux, line 285: It might be nice to make these strings nicer to read - for example "ST-Link v2"

src/platforms/hosted/platform.c
```diff
@@ -310,71 +346,82 @@ bool platform_nrst_get_val(void)
...
 	case BMP_TYPE_STLINKV2:
 		return stlink_max_frequency_get(&info);
+
 	case BMP_TYPE_JLINK:
 		return jlink_max_frequency_get(&info);
+
 	default:
 		DEBUG_WARN("Reading max SWJ frequency not yet implemented\n");
-		break;
+		return false;
```
dragonmux, line 423: This should return 0, not false - inappropriate value for the return type

src/platforms/hosted/platform.c
```diff
@@ -310,71 +346,82 @@ bool platform_nrst_get_val(void)
...
 	}
-	return false;
 }
 
 void platform_target_set_power(bool power)
```
dragonmux, line 427: Can the parameter be marked `const`?

src/platforms/hosted/platform.c
```diff
@@ -395,39 +443,42 @@ bool platform_target_get_power(void)
...
 {
-	uint32_t	targetVoltage = 0 ;
+	uint32_t targetVoltage = 0;
+
 	switch (info.bmp_type) {
 	case BMP_TYPE_BMP: {
-		const char * result ;
-		uint32_t units = 0, tenths = 0 ;
-		result = remote_target_voltage() ;
+		const char *result;
```
dragonmux, line 459: Please move the definition to where the initialisation happens. That'll allow the pointer to be marked const as well as what it points to

src/platforms/hosted/platform.c
```diff
@@ -395,39 +443,42 @@ bool platform_target_get_power(void)
...
+		uint32_t units = 0, tenths = 0;
```
dragonmux, line 460: These should be moved inside the if, and broken onto separate declaration lines.

src/platforms/hosted/platform.c
```diff
@@ -509,17 +590,16 @@ uint32_t adiv5_dp_read(ADIv5_DP_t *dp, uint16_t addr)
...
 
-uint32_t adiv5_dp_low_access(struct ADIv5_DP_s *dp, uint8_t RnW,
-							 uint16_t addr, uint32_t value)
+uint32_t adiv5_dp_low_access(struct ADIv5_DP_s *dp, uint8_t RnW, uint16_t addr, uint32_t value)
 {
 	uint32_t ret = dp->low_access(dp, RnW, addr, value);
 	if (cl_debuglevel & BMP_DEBUG_TARGET) {
 		ap_decode_access(addr, RnW);
-		fprintf(stderr, " 0x%08" PRIx32 "\n", (RnW)? ret : value);
+		fprintf(stderr, " 0x%08" PRIx32 "\n", (RnW) ? ret : value);
```
dragonmux, line 602: Please remove the extra parens

dragonmux (changes requested):
This is looking good, and we'll test it on a ST-Link v2.1 and STM32 targets, but there are a few points we'd like addressed before merge

src/platforms/hosted/jlink_adiv5_swdp.c
```diff
@@ -50,137 +49,86 @@ enum {
...
+	uint8_t cmd[44];
+	memset(cmd, 0, sizeof(cmd));
+
+	cmd[0] = CMD_HW_JTAG3;
+	/* write 19 Bytes.*/
+	cmd[2] = 19 * 8;
+	uint8_t *direction = cmd + 4;
+	memset(direction, 0, 5);
+	memset(direction + 5, 0xffU, 9);
+	memset(direction + 14, 0xffU, 4);
```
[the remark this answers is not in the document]
perigoso: memset to 0 at the start at the function, next to the cmd declaration

src/platforms/hosted/jlink_adiv5_swdp.c
```diff
@@ -50,137 +49,86 @@ enum {
...
+	memset(data, 0xffU, 7);
 	data[7] = 0x9e;
 	data[8] = 0xe7;
-	data[9] = 0xff;
-	data[10] = 0xff;
-	data[11] = 0xff;
-	data[12] = 0xff;
-	data[13] = 0xff;
-	data[14] = 0xff;
-	data[15] = 0;
```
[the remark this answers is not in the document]
perigoso: likewise it is all initialized to zero next to declaration

src/target/adiv5.c
perigoso: on dpv0 we dont really have an analogue, we just dont have DPIDR, unless we use the IDCODE, but that is also present on DPv1/2 so its not quite comparable. the handling of dpv0 is theres nothing to handle

dragonmux: We're thinking along the lines that, we don't want to loose easy access to the version = 0 branch here for when we get to figuring out how to do DPv0

src/target/adiv5.c
perigoso: actually don't know why I touched that and it should remain initialized to 0, as for volatile, I don't know

src/target/adiv5_swdp.c
```diff
@@ -93,93 +97,98 @@ int adiv5_swdp_scan(uint32_t targetid)
...
+	if (!initial_dp->dp_low_write) {
+		DEBUG_WARN("CMSIS_DAP < V1.2 can not handle multi-drop!\n");
+		/* E.g. CMSIS_DAP < V1.2 can not handle multi-drop!*/
+		scan_multidrop = false;
+	}
+
+	DEBUG_WARN("scan_multidrop: %s\n", scan_multidrop ? "true" : "false");
+
+	const volatile size_t max_dp = (scan_multidrop) ? 16U : 1U;
+	for (volatile size_t i = 0; i < max_dp; i++) {
```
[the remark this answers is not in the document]
perigoso: imo it should not be required, but I was weary to remove previous volatiles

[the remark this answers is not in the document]
dragonmux: To be honest.. if volatile is the only think making that loop work, we've got much bigger problems - there's nothing about it that should need volatile. It looks like it was added "in case" the compiler tried optimising the loop out, but the compiler won't as it makes calls and does work

src/target/adiv5.c
dragonmux: We can't see any reason it should be marked volatile

dragonmux:
This all appears to work on the STM32F411 after rebasing onto #1170. Still need to test ST-Link v2.1 though

src/target/adiv5.c
perigoso: heres the reason
`error: variable ‘ctrlstat’ might be clobbered by ‘longjmp’ or ‘vfork’ [-Werror=clobbered]`

src/target/adiv5_swdp.c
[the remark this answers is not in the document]
perigoso: like `ctrlstat`, removing volatile triggers `error: variable ‘i’ might be clobbered by ‘longjmp’ or ‘vfork’ [-Werror=clobbered]`

src/target/adiv5.c
dragonmux: Ok, that's a good explanation.. it indicates an underlying bigger problem, but we'll take it for now. Please add a note about this to the definition.

src/target/adiv5_swdp.c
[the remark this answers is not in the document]
dragonmux: Ok, that's a good explanation.. it indicates an underlying bigger problem, but we'll take it for now. Please add a note about this to the definition.

perigoso:
I also want to split the multidrop handling, but that may come as another PR

src/include/platform_support.h
```diff
@@ -45,7 +45,7 @@ int platform_hwversion(void);
 void platform_nrst_set_val(bool assert);
 bool platform_nrst_get_val(void);
 bool platform_target_get_power(void);
-void platform_target_set_power(bool power);
+void platform_target_set_power(const bool power);
```
dragonmux, line 48: clang-tidy complains if you put const here in the declaration. It's only required on the definition.

src/target/jtag_scan.h
```diff
@@ -43,10 +43,10 @@ typedef struct jtag_dev_s {
 } jtag_dev_t;
 
 extern struct jtag_dev_s jtag_devs[JTAG_MAX_DEVS + 1];
-extern uint32_t jtag_dev_count;
+extern size_t jtag_dev_count;
```
dragonmux, line 46: We'd made this a uint32_t because of the remote.c issues bought up by size_t, otherwise we'd have done the same. Likewise for jtag_add_device()

dragonmux (changes requested):
Looking much better. It seems we goofed when we said change the scan routines to size_t rather than uint32_t as this actually causes some problems in the remote protocol layer, but we'll own that. Apologies!

src/include/platform_support.h
perigoso: Ah, so that's why it didn't have it, TIL

src/target/jtag_scan.h
perigoso: Ah, my bad, i went a little ham with that

dragonmux: We failed to document why we'd made the type choice we had, so there was nothing to indicate why it was a bad idea.. to be fair

perigoso:
I won't be able to touch this until Monday, so if you're keen on merging feel free to fixup those, otherwise I will when I can

dragonmux:
Okay, we'll run a round of fixes shortly then so we can get this merged

dragonmux (approved):
LGTM, merging.

perigoso:
Thank you for dealing with the last couple fixups

PR #1170, opened 2022-08-06
target/stm32h7: fix ids

perigoso:
this is a naive attempt at fixing #1159, following the assumptions I made there

Other targets that were probably affected by the same problem:
stm32f4
stm32g0
stm32lx

### DIFF
--- a/src/target/stm32h7.c
+++ b/src/target/stm32h7.c
@@ -146,9 +146,9 @@ enum stm32h7_regs {
 #define FLASH_SECTOR_SIZE 	0x20000
 #define BANK2_START         0x08100000
 enum ID_STM32H7 {
-	ID_STM32H74x  = 0x450,      /* RM0433, RM0399 */
-	ID_STM32H7Bx  = 0x480,      /* RM0455 */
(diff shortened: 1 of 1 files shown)


dragonmux (approved):
LGTM, we don't have hardware to test on but this does look like it fixes the immediate problem.

dragonmux:
Tested on the STM32F411 Nucleo which was also possibly effected but this bug, and it's working. Merging so we have the fix

perigoso:
> Tested on the STM32F411 Nucleo which was also possibly effected but this bug, and it's working. Merging so we have the fix

could you confirm what DP version it is? I assume v1, this would affect v2s, maybe its not as widespread a problem as i thought

dragonmux:
`DP DPIDR 0x2ba01477 (v1 rev0) designer 0x43b partno 0xba` so.. that looks like DPv1?

perigoso:
Yep, so it retained the old behavior thus not breaking
Documenting this somewhere is probably not a bad idea, afair this information is not usually publicly documented by the implementers

PR #1172, opened 2022-08-06
Correct BMDA build error for BMP only

sidprice:
Windows does not have alloca.h, conditionally use malloc.h on Windows


dragonmux (approved):
LGTM, merging.

PR #1163, opened 2022-08-05
Fix: ADIv5 format string

dragonmux:
This PR addresses a couple of small defects found while working on #1161.

The one in adiv5_jtagdp.c is purely cosmetic. The one in adiv5.c prevents ENABLE_DEBUG=1 builds from succeeding due to the format string mismatching the arguments provided

PR #1175, opened 2022-08-07
adiv5: use PRIxNN specifiers for debug printfs

xobs:
This fixes a few printf-style warnings that slipped through due to the
fact that they're only enabled on debug builds. This enables building
this file on an Xtensa platform where -Werror and -Wformat is enabled by
default.

### DIFF
--- a/src/target/adiv5.c
+++ b/src/target/adiv5.c
@@ -721,7 +721,7 @@ void adiv5_dp_init(ADIv5_DP_t *dp)
 
 		/* Check for a valid DPIDR / designer */
 		if (dp->designer_code != 0) {
-			DEBUG_INFO("DP DPIDR 0x%08" PRIx32 " (v%d %srev%d) designer 0x%" PRIx32 " partno 0x%" PRIx32 "\n", dpidr,
+			DEBUG_INFO("DP DPIDR 0x%08" PRIx32 " (v%x %srev%"PRId32") designer 0x%x partno 0x%x\n", dpidr,
 				dp->version, dp->mindp ? "MINDP " : "",
 				(dpidr & ADIV5_DP_DPIDR_REVISION_MASK) >> ADIV5_DP_DPIDR_REVISION_OFFSET, dp->designer_code,
 				dp->partno);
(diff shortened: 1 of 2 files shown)


src/target/adiv5.c
```diff
@@ -721,7 +721,7 @@ void adiv5_dp_init(ADIv5_DP_t *dp)
 
 		/* Check for a valid DPIDR / designer */
 		if (dp->designer_code != 0) {
-			DEBUG_INFO("DP DPIDR 0x%08" PRIx32 " (v%d %srev%d) designer 0x%" PRIx32 " partno 0x%" PRIx32 "\n", dpidr,
+			DEBUG_INFO("DP DPIDR 0x%08" PRIx32 " (v%"PRIx8" %srev%"PRId32") designer 0x%" PRIx16 " partno 0x%" PRIx16 "\n", dpidr,
```
dragonmux, line 724: PRIx16 and PRIx8 doesn't work on nano newlibc last we tried. These should use %x

src/target/adiv5.c
```diff
@@ -753,7 +753,7 @@ void adiv5_dp_init(ADIv5_DP_t *dp)
 
 		dp->target_partno = (targetid & ADIV5_DP_TARGETID_TPARTNO_MASK) >> ADIV5_DP_TARGETID_TPARTNO_OFFSET;
 
-		DEBUG_INFO("TARGETID 0x%08" PRIx32 " designer 0x%" PRIx32 " partno 0x%" PRIx32 "\n", targetid,
+		DEBUG_INFO("TARGETID 0x%08" PRIx32 " designer 0x%" PRIx16 " partno 0x%" PRIx16 "\n", targetid,
```
dragonmux, line 756: This duplicates #1163 but suffers from the same problem as the above format string.

dragonmux (changes requested):
Thank you for fixing some extra instances of format specifiers being wrong! Much appreciated. We're in dire need for Esden reviewing and merging our PRs right now as this PR partially duplicates one of ours (not your fault) - if you wish to fully consume the changes we made in that PR, we'll close it in favour of this one.

src/target/adiv5.c
xobs: I've switched it to `%x`, which still compiles on ESP32.

src/target/adiv5.c
xobs: I've also switched this to `%x`, and confirmed that it still builds.

xobs:
I've updated changes and force-pushed the branch. I've verified that the changes still build on ESP32. Additionally, I've added the same fix to `adiv5_jtagdp.c`.

dragonmux (approved):
LGTM, considering this as replaces #1163 and merging in favour of that PR.

PR #1178, opened 2022-08-08
native: Enable uart on HW >=V3 when debugging.

### LINKED ISSUE #1103: Enable UART on newer BMP versions when debugged (inception)
We are not enabling the AUX UART when a debugger is connected to the BMP hardware for all hardware versions except 0.

https://github.com/blackmagic-debug/blackmagic/blob/main/src/platforms/native/platform.c#L230

The comment is not exactly true as the debug connector was disconnected from the AUX UART pins, and has it's dedicated connector, since quite a while now. We can do debugging and use the UART at the same time, which can be very useful.

esden:

## Detailed description

We were disabling the UART on all Hardware V1 and newer, as V1 and V2
had the SWD pins connected to the UART pins. Since V3 this is not the
case any more so we can keep the UART enabled when an SWD debugger is
connected to the BMP.

## Closing issues

Closes #1103

### DIFF
--- a/src/platforms/native/platform.c
+++ b/src/platforms/native/platform.c
@@ -225,9 +225,11 @@ void platform_init(void)
 	platform_timing_init();
 	cdcacm_init();
 
-	/* On mini hardware, UART and SWD share connector pins.
+	/* On hardware version 1 and 2, UART and SWD share connector pins.
 	 * Don't enable UART if we're being debugged. */
-	if ((platform_hwversion() == 0) || !(SCS_DEMCR & SCS_DEMCR_TRCENA))
+	if (platform_hwversion() == 0 ||
+	    platform_hwversion() >= 3 ||
+	    !(SCS_DEMCR & SCS_DEMCR_TRCENA))
 		usbuart_init();
 
 	setup_vbus_irq();


src/platforms/native/platform.c
```diff
@@ -225,9 +225,11 @@ void platform_init(void)
 	platform_timing_init();
 	cdcacm_init();
 
-	/* On mini hardware, UART and SWD share connector pins.
+	/* On hardware version 1 and 2, UART and SWD share connector pins.
 	 * Don't enable UART if we're being debugged. */
-	if ((platform_hwversion() == 0) || !(SCS_DEMCR & SCS_DEMCR_TRCENA))
+	if ((platform_hwversion() == 0) ||
```
dragonmux, line 230: the outer pair of parens around each equality suppresses a GCC warning so should be removed. (note: this doesn't apply to the bit mask in the final check which binds looser than || and !)

dragonmux (changes requested):
Looks good, one small thing to fix that's been a slow-moving change in the code base but otherwise looks ready to merge

src/platforms/native/platform.c
esden: Sounds good. I amended the patch.

dragonmux (approved):
LGTM, merging.

PR #1181, opened 2022-08-08
Fix: ATSAMD attach errata

### LINKED ISSUE #578: ATSAMD11 can fail to probe if bugged firmware triggers errata 9905
While writing a firmware for the ATSAMD11D14A-SS, I failed to implement a workaround for a silicon bug that Atmel describes as "Errata 9905." Various sources online summarize this errata as "the DFLL clock must be requested before being configured otherwise a write access to a DFLL register can freeze the device." If bugged firmware attempts to program the DFLL in an incorrect way too soon after reset, the BMP will fail to scan it and give "Timeout during scan. Is target stuck in WFI?" This will make it impossible to erase the firmware and try to flash a working one.

As far as I can tell, what is happening is that the `samd_probe` function releases the ARM core from Atmel's "extended reset" state (https://github.com/blacksphere/blackmagic/blob/master/src/target/samd.c#L545). However, the BMP does not halt the core at this point, and the core can start running the bugged firmware. By the time GDB tries to attach to the target, the firmware has triggered the errata and frozen the core. The BMP then gets very confused.

As a workaround, I wrote the following patch that was able to work around the problem for me by moving the logic that releases the extended reset state to the attach function. I'm not entirely sure why it works since it seems like the code should still have a race window with the firmware, but it worked when I tried it. I don't expect this patch to get merged, but I'm not sure how this should be properly fixed. I am filing this issue for now in order to open a discussion for a proper solution to the problem.

```patch
diff --git a/src/target/samd.c b/src/target/samd.c
index 3507d02..7116ce6 100644
--- a/src/target/samd.c
+++ b/src/target/samd.c
@@ -37,6 +37,7 @@
 #include "target.h"
 #include "target_internal.h"
 #include "cortexm.h"
+#include "gdb_packet.h"
 
 static int samd_flash_erase(struct target_flash *t, target_addr addr, size_t len);
 static int samd_flash_write(struct target_flash *f,
@@ -363,6 +364,21 @@ bool samd_protected_attach(target *t)
 	return true;
 }
 
+bool samd_hax_attach(target *t) {
+	t->attach = cortexm_attach;
+
+	if (target_mem_read32(t, SAMD_DSU_CTRLSTAT) &
+	    SAMD_STATUSA_CRSTEXT) {
+		gdb_outf("in hooked attach\n");
+
+		/* Write bit to clear from extended reset */
+		target_mem_write32(t, SAMD_DSU_CTRLSTAT,
+		                   SAMD_STATUSA_CRSTEXT);
+	}
+
+	return cortexm_attach(t);
+}
+
 /**
  * Use the DSU Device Indentification Register to populate a struct
  * describing the SAM D device.
@@ -474,6 +490,8 @@ static void samd_add_flash(target *t, uint32_t addr, size_t length)
 char variant_string[60];
 bool samd_probe(target *t)
 {
+	gdb_outf("in samd_probe\n");
+
 	uint32_t cid = samd_read_cid(t);
 	uint32_t pid = samd_read_pid(t);
 
@@ -540,18 +558,28 @@ bool samd_probe(target *t)
 	samd_add_flash(t, 0x00000000, 0x40000);
 	target_add_commands(t, samd_cmd_list, "SAMD");
 
+	gdb_outf("about to poke reset bullshit\n");
+
+	t->attach = samd_hax_attach;
+
+#if 0
 	/* If we're not in reset here */
 	if (!platform_srst_get_val()) {
+		gdb_outf("in first if statement\n");
 		/* We'll have to release the target from
 		 * extended reset to make attach possible */
 		if (target_mem_read32(t, SAMD_DSU_CTRLSTAT) &
 		    SAMD_STATUSA_CRSTEXT) {
+			gdb_outf("in second if statement\n");
 
 			/* Write bit to clear from extended reset */
 			target_mem_write32(t, SAMD_DSU_CTRLSTAT,
 			                   SAMD_STATUSA_CRSTEXT);
 		}
 	}
+#endif
+
+	gdb_outf("leaving samd_probe\n");
 
 	return true;
 }
```

dragonmux:
## Detailed description

This PR provides a workaround fix for SAMD11 errata which can be triggered by the normal cortexm_attach routine.

If the DFLL48M is left unrequested and then a configuration attempt is made, it locks up the chip till reset. This stops attach working in GDB when trying to work with bugged firmware that triggers this errata.

We work around this by releasing the target from reset before attach, which corrects this problem state.

Tested working against the ATSAMD11 on LUNA.

## Closing issues

Fixes #578

### DIFF
--- a/src/target/samd.c
+++ b/src/target/samd.c
@@ -217,13 +217,13 @@ static const struct samd_part samd_l22_parts[] = {
 	{0xFF, 0, 0, 0}
 };
 
-/**
+/*
  * Overloads the default cortexm reset function with a version that
  * removes the target from extended reset where required.
  */
 void samd_reset(target *t)
 {
-	/**
+	/*
 	 * nRST is not asserted here as it appears to reset the adiv5
 	 * logic, meaning that subsequent adiv5_* calls PLATFORM_FATAL_ERROR.
 	 *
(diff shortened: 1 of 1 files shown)


esden (approved):
LGTM

PR #1184, opened 2022-08-08
start seperating swd multi drop

perigoso:
## Detailed description

I initially meant  to properly abstract away the multidrop handling and swd scan in general, but left it rather early, this PR doesnt have much going for it, just a place to document and later continue work from

as is the jlink scan (and I believe stlink too, but i have not tested or looked at it) does not handle multi-drop, I was meaning to get rp2040 working on jlink as means to develop and test this work but it is unable to comunicate with its DP, the reason is I would guess the Lack of "selection alert sequence for SW-DP-V2" that is present on the regular adiv5_scan

The jlink probe and api should be able to do it, but it needs to be implemented, ideally as a `seq_out` method, and the cmds in jlink in general could do with some work

since the jlink usb spec is not that easy to find I will leave it here
[RM08001_JLinkUSBProtocol_Rev.7.pdf](https://github.com/blackmagic-debug/blackmagic/files/9283562/RM08001_JLinkUSBProtocol_Rev.7.pdf)

### DIFF
--- a/src/target/adiv5_swd.c
+++ b/src/target/adiv5_swd.c
@@ -79,10 +79,6 @@ static uint32_t firmware_dp_low_read(const uint16_t addr)
 	return res == SWDP_ACK_OK ? data : 0;
 }
 
-/* Try first the dormant to SWD procedure.
- * If target id given, scan DPs 0 .. 15 on that device and return.
- * Otherwise
- */
 uint32_t adiv5_swdp_scan(const uint32_t targetid)
 {
 	/* Free the device list if any */
@@ -127,6 +123,19 @@ uint32_t adiv5_swdp_scan(const uint32_t targetid)
 
 	if (!dp_targetid) {
 		/* No targetID given on the command line Try to read ID */
+		/*
+		 * ARM Debug Interface Architecture Specification, ADIv5.0 to ADIv5.2. ARM IHI 0031C
+		 *
+		 * §4.2.6 Limitations of multi-drop
+		 * 
+		 * It is not possible to interrogate a multi-drop Serial Wire Debug system that includes multiple devices to establish
+		 * which devices are connected. Because all devices are selected on coming out of a line reset, no communication with
+		 * a device is possible without prior selection of that target using its target ID. Therefore, connection to a multi-drop
+		 * Serial Wire Debug system that includes multiple devices requires that either:
+		 * - The host has prior knowledge of the devices in the system and is configured before target connection.
+		 * - The host attempts auto-detection by issuing a target select command for each of the devices it has been
+		 * configured to support.
+	 	 */
 
 		swd_line_reset();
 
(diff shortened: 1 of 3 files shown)


dragonmux:
It would be useful please if this PR could be rebased on main, especially now that the J-Link implementation implements those low-level SWD functions mentioned as being a blocker for you.

Having proper support for multi-drop and it being refactored like this is of great interest to us

perigoso:
> It would be useful please if this PR could be rebased on main, especially now that the J-Link implementation implements those low-level SWD functions mentioned as being a blocker for you.
> 
> Having proper support for multi-drop and it being refactored like this is of great interest to us

Thank you for letting me know, this was just the start of the implementation there's work to do still, but I was looking for something to pick up!

perigoso:
Note to self, possible problem to investigate, amidst the rebases `adiv5_dp_read_dpidr` lost the fault check, verify if it's need

perigoso:
After taking a new look, this is ready for review @dragonmux it doesn't achieve much worthy of note, but, stepping stones...

I started to poke the targetsel handling, but I only got higher not lower flash usage and the code didn't improve significantly, so I'm leaving that for another time.

src/target/adiv5_swd.c
```diff
@@ -53,10 +53,66 @@ uint8_t make_packet_request(uint8_t RnW, uint16_t addr)
...
+	 * Note: in some non-conformant devices (STM32) at least 51 HIGH cycles and/or 3/4 idle cycles are required
+	 *
+	 * for robustness, we use 60 HIGH cycles and 4 idle cycles
+	 */
+	swd_proc.seq_out(0xffffffffU, 32U);                     /* 32 cycles HIGH */
+	swd_proc.seq_out(0x0fffffffU, idle_cycles ? 32U : 28U); /* 28 cycles HIGH + 4 idle cycles if idle is requested */
+}
+
+/* Switch out of dormant state into SWD */
+static void dormant_state_to_swd_sequence()
```
dragonmux, line 69: We know this isn't technically required because of that `static`, but to be in keeping with enforcing proper prefixing, we suggest making this `swd_switch_from_dormant_sequence()` - either that or drop the word state, making this `dormant_to_swd_sequence()`

src/target/adiv5_swd.c
```diff
@@ -53,10 +53,66 @@ uint8_t make_packet_request(uint8_t RnW, uint16_t addr)
...
+	swd_line_reset_sequence(false);
+	/* Send the 128-bit Selection Alert sequence on SWDIOTMS */
+	swd_proc.seq_out(ADIV5_SELECTION_ALERT_SEQUENCE_0, 32U);
+	swd_proc.seq_out(ADIV5_SELECTION_ALERT_SEQUENCE_1, 32U);
+	swd_proc.seq_out(ADIV5_SELECTION_ALERT_SEQUENCE_2, 32U);
+	swd_proc.seq_out(ADIV5_SELECTION_ALERT_SEQUENCE_3, 32U);
+	/* Send 4 SWCLKTCK cycles with SWDIOTMS LOW */
+	/* Send the required activation code sequence on SWDIOTMS */
+	/* We combine the last two actions in a single seq_out as an optimization */
+	swd_proc.seq_out(ADIV5_ACTIVATION_CODE_ARM_SWD_DP << 4U, 8U + 4U);
```
dragonmux, line 88: Please make this a proper block comment, and note how many cycles the activation code requires in the comment so the number passed can be simplified to just `12U` rather than relying on constant folding

src/target/adiv5_swd.c
```diff
@@ -79,133 +135,200 @@ static uint32_t firmware_dp_low_read(const uint16_t addr)
...
+			dpidr = adiv5_dp_read_dpidr(dp);
+			if (dpidr != 0)
+				/* Successfully read the DPIDR */
+				break;
+
+			if (!tried_jtag_to_swd) {
+				jtag_to_swd_sequence();
+
+				dp->fault = 0;
+
```
dragonmux, line 200: Please drop the extra newline here. These statements feel naturally grouped

src/target/adiv5_swd.c
```diff
@@ -79,133 +135,200 @@ static uint32_t firmware_dp_low_read(const uint16_t addr)
...
+		DEBUG_WARN("Discovered multi-drop enabled target but CMSIS_DAP < v1.2 cannot handle multi-drop\n");
 		scan_multidrop = false;
 	}
 #endif
 
-	DEBUG_WARN("scan_multidrop: %s\n", scan_multidrop ? "true" : "false");
+	if (scan_multidrop)
+		adiv5_swdp_multidrop_scan(dp, dp_targetid);
+	else {
+		adiv5_dp_error(dp);
```
dragonmux, line 241: This should be `adiv5_dp_abort(dp, ADIV5_DP_ABORT_STKERRCLR);` still from what we can see, to preserve the existing behaviour

src/target/adiv5_swd.c
```diff
@@ -79,133 +135,200 @@ static uint32_t firmware_dp_low_read(const uint16_t addr)
 	return res == SWDP_ACK_OK ? data : 0;
 }
 
-/* Try first the dormant to SWD procedure.
- * If target id given, scan DPs 0 .. 15 on that device and return.
- * Otherwise
- */
-uint32_t adiv5_swdp_scan(uint32_t targetid)
+uint32_t adiv5_swdp_scan(const uint32_t targetid)
```
dragonmux, line 138: Given this appears to return 0 for failure and 1 for success, does it make sense to change this to return `bool` and clean up the returns accordingly?

If we do that, lets also rename this `adiv5_swd_scan()` as the 'p' in the name is unhelpful - it's scanning the SWD bus. That puts the nomenclature in line with the JTAG scan engine

src/target/adiv5_swd.c
```diff
@@ -79,133 +135,200 @@ static uint32_t firmware_dp_low_read(const uint16_t addr)
...
+ * configured to support.
+ * 
+ * This means that debug tools cannot connect seamlessly to targets in a multi-drop Serial Wire Debug system that they
+ * have never seen before. However, if the debug tools can be provided with the target ID of such targets by the user
+ * then the contents of the target can be auto-detected as normal.
+ * To protect against multiple selected devices all driving the line simultaneously SWD protocol version 2 requires:
+ * - For multi-drop SWJ-DP, the JTAG connection is selected out of powerup reset. JTAG does not drive the line.
+ * - For multi-drop SW-DP, the DP is in the dormant state out of powerup reset.
+ */
+void adiv5_swdp_multidrop_scan(adiv5_debug_port_s *target_dp, const uint32_t targetid)
```
dragonmux, line 273: `target_dp` here can just be `dp` as we're not sure it adds anything to include the `target_` nomenclature in this instance. Think it can be marked `const` too as in `adiv5_debug_port_s *const dp`

Likewise, please either drop the `p` from the end of `swdp` or make it `swd_dp` - we've been working to slowly eliminate this confusing naming, so not adding another instance to later fix is a good idea.

src/target/adiv5_swd.c
```diff
@@ -79,133 +135,200 @@ static uint32_t firmware_dp_low_read(const uint16_t addr)
...
+		/* Select the instance */
+		target_dp->dp_low_write(ADIV5_DP_TARGETSEL,
+			instance << ADIV5_DP_TARGETSEL_TINSTANCE_OFFSET |
+				(targetid & (ADIV5_DP_TARGETID_TDESIGNER_MASK | ADIV5_DP_TARGETID_TPARTNO_MASK)) | 1U);
 
-		adiv5_debug_port_s *dp = calloc(1, sizeof(*dp));
-		if (!dp) { /* calloc failed: heap exhaustion */
-			DEBUG_ERROR("calloc: failed in %s\n", __func__);
+		/* Read DPIDR */
+		if (adiv5_dp_read_dpidr(target_dp) == 0)
```
dragonmux, line 298: This should still check `dp->fault`, or `adiv5_dp_read_dpidr()` should turn fault being set into a 0 return like it does an exception being thrown. Fault being set because of getting a no-response response is one of the failure modes for this call.

src/target/adiv5_swd.c
```diff
@@ -79,133 +135,200 @@ static uint32_t firmware_dp_low_read(const uint16_t addr)
...
+			/* No DP here, next instance */
 			continue;
-		}
 
-		memcpy(dp, initial_dp, sizeof(adiv5_debug_port_s));
-		dp->instance = i;
+		/* Store the current instance DP, we will yield it to adiv5_dp_init and continue with a new one */
+		adiv5_debug_port_s *const instance_dp = target_dp;
```
dragonmux, line 303: With the DP variable fed in renamed, this can become `target_dp = dp;` which will help the below code make more sense. We'd suggest keeping the original calloc-and-memcpy behaviour though as that's less confusing (less if logic and nesting) and allows the final action taken in this function to be `free(dp);` to keep things neat.

We know what you're doing here would result in less memory fragmentation, but the convolution of the resulting logic doesn't feel worth it.

dragonmux (changes requested):
This is looking great and provides some much needed clean-up and magic number naming - thank you for the contribution!

There are a few items that we've spotted in review that could do with addressing before we merge to try to ensure this is the best possible version of a SWD bus scanner, but with them addressed, we'll be happy to merge this.

src/target/adiv5_swd.c
perigoso: It does make sense, from what I gathered reading the code, I did it, but the changes started to cascade a bit so I think I'll submit that later, in an effort to get this merged faster

With that said I took the name note and renamed multidrop scan to align with the clearer nomenclature

src/target/adiv5_swd.c
perigoso: naming suggestions taken, regarding the const, it cant as that same pointer was being used to track the new dp allocations, but if I use the dp allocation strategy suggested in the other comment I might be able to

src/target/adiv5_swd.c
perigoso: from reading the `low_access` implementation I gathered that when `dp->fault` is non zero, 0 is also always returned, I suppose an additional check doesn't hurt, but I don't think it's *required*, did I miss something?

src/target/adiv5_swd.c
perigoso: I was trying to be smart, but you're right it's a bit convoluted

src/target/adiv5_swd.c
perigoso: The reason they were separated was because I wanted to clearly show they were individual steps (as they are listed on the spec), that we combine into a single sequence. That said, it wasn't really that obvious or useful.

src/target/adiv5_swd.c
dragonmux: Ah, you're right. Fair point!

src/target/adiv5_swd.c
dragonmux: Ah, sure. We can address that in a follow-up cleanup PR then. You're right that it probably does have some far reaching consequences and so will cause a lot of churn.

dragonmux (approved):

We'd run our own tests but we know you're careful to thoroughly test these kinds of thing.

PR #1189, opened 2022-08-10
Add some basic LPC55xx flashing support, tested only on LPC5504

TomCrypto:
## Detailed description

This pull request adds some support for flashing LPC55xx devices. Because I only have an LPC5504 device, this is the only device I have tested. It _should_ work for LPC5502 and LPC5506 and the corresponding S-variants as well. Based on user manuals and other online sources, I have tried to write the code such that it will also work for LPC551x, LPC552x and LPC556x, but I can provide no guarantees as to what this code will do to them.

Based on my research, it is possible that there may be two different silicon revisions in the wild with (very) different IAP interfaces. My device has "ROM API version 1", so that is what I have tested. The code in this PR has support for "ROM API version 0" but it is untested.

The IAP interface for this device family is very poorly documented with a lot of contradicting information online. I recommend gathering all of the devices if possible and running tests on them to see what each device actually needs us to do.

Overall, with this change I am able to flash and execute programs fine with no issues, and everything seems stable. I've also implemented the readuid monitor command which makes an IAP call and it appears to behave sensibly.

There's some limited debug logging, on probing the LPC55xx it will detect flash and print something like this:

    LPC55xx: Detected flash with 131072 bytes, 512-byte pages

so that can be a quick way to smoke-test part of the IAP code on a device without flashing anything.

### DIFF
--- a/src/target/lpc55xx.c
+++ b/src/target/lpc55xx.c
@@ -37,11 +37,509 @@
 #include "target_internal.h"
 #include "cortexm.h"
 
-#define LPC55xx_CHIPID 0x50000ff8U
-#define LPC55_DMAP_IDR 0x002a0000U
-
+#define LPC55_DMAP_IDR                 0x002a0000U
 #define LPC55_DMAP_BULK_ERASE          0x02U
 #define LPC55_DMAP_START_DEBUG_SESSION 0x07U
+
+#define LPC55xx_FLASH_BASE     0x00000000U
+#define LPC55xx_ERASE_KEY      0x6b65666cU
+#define LPC55xx_CHIPID_ADDRESS 0x50000ff8U
+
+#define LPC55xx_WRITE_SIZE 4096
+#define LPC55xx_ERASE_SIZE 4096
+
+/* Target memory layout for IAP calls, we will scribble over SRAM */
+
+#define LPC55xx_FLASH_CONFIG_ADDRESS 0x04000000U
+#define LPC55xx_CODE_PATCH_ADDRESS   0x0400003cU
+#define LPC55xx_UUID_ADDRESS         0x04000040U
+#define LPC55xx_WRITE_BUFFER_ADDRESS 0x20000000U
+#define LPC55xx_SCRATCH_MEMORY_LEN   0x40
+#define LPC55xx_UUID_LEN             0x10
+
+/* The ROM code seems to default to an MSP in the SRAM-X region, which
+ * is code-only. This seems fairly safe, so do the same for IAP calls.
+ *
+ * This may not matter as supposedly IAP doesn't use the stack (?)
+ */
+
+#define LPC55xx_IAP_MSP_ADDRESS 0x04003000U
+#define LPC55xx_IAP_FREQ_IN_MHZ 96
+
+/* Device chip IDs */
+
+#define LPC5502_CHIPID  0xa1003702U // UM11424
+#define LPC5504_CHIPID  0xa1001504U // UM11424
+#define LPC5506_CHIPID  0xa1000406U // UM11424
+#define LPC5512_CHIPID              // (unknown)
+#define LPC5514_CHIPID              // (unknown)
+#define LPC5516_CHIPID              // (unknown)
+#define LPC5524_CHIPID              // (unknown)
+#define LPC5526_CHIPID  0xa010221aU // NXP forum
+#define LPC5528_CHIPID  0xa010111cU // NXP forum
+#define LPC55S04_CHIPID 0x51001584U // UM11424
+#define LPC55S06_CHIPID 0x51000486U // UM11424
+#define LPC55S14_CHIPID             // (unknown)
+#define LPC55S16_CHIPID             // (unknown)
+#define LPC55S26_CHIPID 0xa010229aU // NXP forum
+#define LPC55S28_CHIPID 0xa010119cU // NXP forum
+#define LPC55S66_CHIPID             // (unknown)
+#define LPC55S69_CHIPID             // (unknown)
+
+/* The available IAP commands that we support, mostly flash access */
+
+typedef enum lpc55xx_iap_cmd {
+	IAP_CMD_FLASH_INIT,
+	IAP_CMD_FLASH_ERASE,
+	IAP_CMD_FLASH_PROGRAM,
+	IAP_CMD_FFR_INIT,
+	IAP_CMD_FFR_GET_UUID,
+} lpc55xx_iap_cmd_e;
+
+/* The possible IAP errors are documented here for easy reference */
+
+typedef enum lpc55xx_iap_status {
+	IAP_STATUS_FLASH_SUCCESS = 0,
+	IAP_STATUS_FLASH_INVALID_ARGUMENT = 4,
+	IAP_STATUS_FLASH_SIZE_ERROR = 100,
+	IAP_STATUS_FLASH_ALIGNMENT_ERROR = 101,
+	IAP_STATUS_FLASH_ADDRESS_ERROR = 102,
+	IAP_STATUS_FLASH_ACCESS_ERROR = 103,
+	IAP_STATUS_FLASH_COMMAND_FAILURE = 105,
+	IAP_STATUS_FLASH_UNKNOWN_PROPERTY = 106,
+	IAP_STATUS_FLASH_ERASE_KEY_ERROR = 107,
+	IAP_STATUS_FLASH_COMMAND_NOT_SUPPORTED = 111,
+	IAP_STATUS_FLASH_ECC_ERROR = 116,
+	IAP_STATUS_FLASH_COMPARE_ERROR = 117,
+	IAP_STATUS_FLASH_REGULATION_LOSS = 118,
+	IAP_STATUS_FLASH_INVALID_WAIT_STATE_CYCLES = 119,
+	IAP_STATUS_FLASH_OUT_OF_DATE_CFPA_PAGE = 132,
+	IAP_STATUS_FLASH_BLANK_IFR_PAGE_DATA = 133,
+	IAP_STATUS_FLASH_ENCRYPTED_REGIONS_ERASE_NOT_DONE_AT_ONCE = 134,
+	IAP_STATUS_FLASH_PROGRAM_VERIFICATION_NOT_ALLOWED = 135,
+	IAP_STATUS_FLASH_HASH_CHECK_ERROR = 136,
+	IAP_STATUS_FLASH_SEALED_FFR_REGION = 137,
+	IAP_STATUS_FLASH_FFR_REGION_WRITE_BROKEN = 138,
+	IAP_STATUS_FLASH_NMPA_ACCESS_NOT_ALLOWED = 139,
+	IAP_STATUS_FLASH_CMPA_CFG_DIRECT_ERASE_NOT_ALLOWED = 140,
+	IAP_STATUS_FLASH_FFR_BANK_IS_LOCKED = 141,
+} lpc55xx_iap_status_e;
+
+static target_addr_t lpc55xx_get_bootloader_tree_address(target_s *target)
+{
+	switch (target_mem_read32(target, LPC55xx_CHIPID_ADDRESS)) {
+	//case LPC5512_CHIPID:
+	//case LPC5514_CHIPID:
+	//case LPC55S14_CHIPID:
+	//case LPC5516_CHIPID:
+	//case LPC55S16_CHIPID:
+	//case LPC5524_CHIPID:
+	case LPC5502_CHIPID:
+	case LPC5504_CHIPID:
+	case LPC5506_CHIPID:
+	case LPC55S04_CHIPID:
+	case LPC55S06_CHIPID:
+		return 0x1301fe00U;
+
+	//case LPC55S69_CHIPID:
+	case LPC5526_CHIPID:
+	case LPC55S26_CHIPID:
+	case LPC5528_CHIPID:
+	case LPC55S28_CHIPID:
+		//case LPC55S66_CHIPID:
+		return 0x130010f0U;
+
+	default:
+		return 0;
+	}
+}
+
+static const char *lpc55xx_get_device_name(uint32_t chipid)
+{
+	switch (chipid) {
+	case LPC5502_CHIPID:
+		return "LPC5502";
+	case LPC5504_CHIPID:
+		return "LPC5504";
+	case LPC5506_CHIPID:
+		return "LPC5506";
+	case LPC55S04_CHIPID:
+		return "LPC55S04";
+	case LPC55S06_CHIPID:
+		return "LPC55S06";
+	case LPC5526_CHIPID:
+		return "LPC5526";
+	case LPC55S26_CHIPID:
+		return "LPC55S26";
+	case LPC5528_CHIPID:
+		return "LPC5528";
+	case LPC55S28_CHIPID:
+		return "LPC55S28";
+
+	default:
+		return "unknown";
+	}
+}
+
+static int lpc55xx_get_rom_api_version(target_s *target, target_addr_t bootloader_tree_address)
+{
+	return ((target_mem_read32(target, bootloader_tree_address + 0x4) >> 16) & 0xff) == 3 ? 1 : 0;
+}
+
+static target_addr_t lpc55xx_get_flash_table_address(target_s *target, target_addr_t bootloader_tree_address)
+{
+	return target_mem_read32(target, bootloader_tree_address + 0x10);
+}
+
+static target_addr_t lpc55xx_get_flash_init_address(target_s *target)
+{
+	target_addr_t bootloader_tree_address = lpc55xx_get_bootloader_tree_address(target);
+
+	target_addr_t flash_table_address = lpc55xx_get_flash_table_address(target, bootloader_tree_address);
+	return target_mem_read32(target, flash_table_address + sizeof(uint32_t));
+}
+
+static target_addr_t lpc55xx_get_flash_erase_address(target_s *target)
+{
+	target_addr_t bootloader_tree_address = lpc55xx_get_bootloader_tree_address(target);
+
+	if (lpc55xx_get_rom_api_version(target, bootloader_tree_address) == 0)
+		return 0x1300413bU; // UNTESTED: found in SDK, not referenced in UM
+
+	target_addr_t flash_table_address = lpc55xx_get_flash_table_address(target, bootloader_tree_address);
+	return target_mem_read32(target, flash_table_address + 2 * sizeof(uint32_t));
+}
+
+static target_addr_t lpc55xx_get_flash_program_address(target_s *target)
+{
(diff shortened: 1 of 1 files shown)


src/target/adiv5.c
```diff
@@ -833,7 +833,7 @@ void adiv5_dp_init(ADIv5_DP_t *dp, const uint32_t idcode)
 		}
 	}
 
-	if (dp->designer_code == JEP106_MANUFACTURER_NXP)
```
TomCrypto, line 836: Recommended by Perigoso on Discord

src/target/cortexm.c
```diff
@@ -678,10 +678,7 @@ bool cortexm_probe(ADIv5_AP_t *ap)
 		PROBE(renesas_probe);
 		break;
 	case JEP106_MANUFACTURER_NXP:
-		if (t->part_id == 0x4c9U) /* Cortex-M33 ROM */
-			PROBE(lpc55xx_probe);
-		else
-			DEBUG_WARN("Unhandled NXP device\n");
```
TomCrypto, line 684: Recommended by Perigoso on Discord

src/target/lpc55xx.c
```diff
@@ -37,12 +37,485 @@
...
+	}
+
+	return true;
+}
+
+static bool lpc55xx_enter_flash_mode(target *t)
+{
+	// TODO: do we need to reset the target here?
+	// (reset here doesn't work for some reason)
+	// target_reset(t);
```
TomCrypto, line 443: I really need help with this one. Clearly _some_ target resets work, but resetting it here breaks everything. It hasn't been needed in my tests but apparently is a good idea if the target is in an IRQ (?)

src/target/lpc55xx.c
```diff
@@ -37,12 +37,478 @@
...
+
+	// Initialize the IAP flash context once in a predefined location
+	// of SRAM, the flash erase/write functions assume it is present.
+
+	status = iap_call_raw(t, IAP_CMD_FLASH_INIT, 0, 0, 0);
+
+	if (status != IAP_STATUS_FLASH_SUCCESS) {
+		DEBUG_WARN("LPC55xx: IAP error: FLASH_INIT (%d)\n", status);
+		return false;
+	}
```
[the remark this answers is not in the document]
TomCrypto: In some ways it kind of fits in both, because it initializes the flash controller AND writes a structure in RAM. But I agree I think `prepare` would be the better place for it, I will try to move it.

src/target/lpc55xx.c
```diff
@@ -37,12 +37,478 @@
 #include "target_internal.h"
 #include "cortexm.h"
 
-#define LPC55xx_CHIPID 0x50000ff8U
-#define LPC55_DMAP_IDR 0x002A0000U
+#define IAP_FREQ_MHZ 96
+#define ARM_BKPT_PATTERN 0xBEBE
 
+#define LPC55_DMAP_IDR                 0x002A0000
```
dragonmux, line 43: U on the end of the constant please and run clang-format

src/target/lpc55xx.c
```diff
@@ -37,12 +37,478 @@
...
 #define LPC55_DMAP_BULK_ERASE          0x02U
 #define LPC55_DMAP_START_DEBUG_SESSION 0x07U
 
+#define LPC55XX_FLASH_BASE     0x00000000
```
dragonmux, line 47: U on the ends of all these constants please, clang-format, and use `LPC55xx_` in the macro naming scheme so it's more obvious the X's refer to missing part number digits

src/target/lpc55xx.c
```diff
@@ -37,12 +37,478 @@
...
+	//case LPC55S14_CHIPID:
+	//case LPC5516_CHIPID:
+	//case LPC55S16_CHIPID:
+	//case LPC5524_CHIPID:
+	case LPC5502_CHIPID:
+	case LPC5504_CHIPID:
+	case LPC5506_CHIPID:
+	case LPC55S04_CHIPID:
+	case LPC55S06_CHIPID:
+		return 0x1301fe00;
```
dragonmux, line 142: This should be suffixed with U, same with the address below

src/target/lpc55xx.c
```diff
@@ -37,12 +37,478 @@
...
+		return "LPC55S28";
+
+	default:
+		return "unknown";
+	}
+}
+
+static int lpc55xx_get_rom_api_version(target* t, target_addr_t bootloader_tree_address)
+{
+	return (((target_mem_read32(t, bootloader_tree_address + 0x4) >> 16) & 0xff) == 3) ? 1 : 0;
```
dragonmux, line 186: The outer parens are not required so please remove them. In a similar vein, please suffix the bit manipulation constants with U

src/target/lpc55xx.c
```diff
@@ -37,12 +37,478 @@
...
+{
+	return target_mem_read32(t, bootloader_tree_address + 0x10);
+}
+
+static target_addr_t lpc55xx_get_flash_init_address(target* t)
+{
+	target_addr_t bootloader_tree_address = lpc55xx_get_bootloader_tree_address(t);
+
+	target_addr_t flash_table_address = lpc55xx_get_flash_table_address(t, bootloader_tree_address);
+	return target_mem_read32(t, flash_table_address + 1 * sizeof(uint32_t));
```
dragonmux, line 199: Is this supposed to be `(flash_table_addres + 1) * sizeof(uint32_t)`, or is that `1 *` redundant?

src/target/lpc55xx.c
```diff
@@ -37,12 +37,478 @@
...
+
+	target_mem_write16(t, LPC55XX_BKPT_ADDRESS, ARM_BKPT_PATTERN);
+	regs[REG_LR] = LPC55XX_BKPT_ADDRESS | 1; // set arm thumb bit
+
+	/* Write the registers to the target and perform the IAP call */
+
+	target_regs_write(t, regs);
+
+	target_halt_resume(t, false);
+	while (!target_halt_poll(t, NULL)) {}
```
dragonmux, line 299: Please don't do this. Please use the following pattern:
```c
while (!target_halt_poll(t, NULL))
    continue;
```

src/target/lpc55xx.c
```diff
@@ -37,12 +37,478 @@
...
+	} else {
+		return target_mem_read32(t, flash_table_address + 13 * sizeof(uint32_t));
+	}
+}
+
+static enum iap_status iap_call_raw(target *t, enum iap_cmd cmd, uint32_t r1, uint32_t r2, uint32_t r3)
+{
+	/* Prepare the registers for the IAP call. R0 is always flash_config */
+
+	uint32_t regs[t->regs_size / sizeof(uint32_t)];
```
dragonmux, line 254: If t->regs_size is a constant size for this processor series, please use that constant here instead of this. This constructs a VLA and we're trying to remove all variable length arrays from the code base.

src/target/lpc55xx.c
```diff
@@ -37,12 +37,478 @@
...
+};
+
+static void lpc55xx_prepare_flash_config(target *t, target_addr_t address)
+{
+	/* The flash_config structure is 60 bytes in size, zero it out as that
+	 * is what the SDK does. For some reason you have to fill in the clock
+	 * speed field ("sysFreqInMHz") before flash_init. Set it to 96MHz (?)
+	 */
+
+	struct flash_config config = { 0 };
```
dragonmux, line 327: This only 0's the first member, and leaves the rest in an undefined state. Better here to do:
```c
struct flash_config config;
memset(&config, 0, sizeof(config));
```

src/target/lpc55xx.c
```diff
@@ -37,12 +37,478 @@
...
+	config.sysFreqInMHz = IAP_FREQ_MHZ;
+
+	target_mem_write(t, address, &config, sizeof(config));
+}
+
+static bool lpc55xx_flash_init(target *t, struct flash_config *config)
+{
+	uint8_t backup_memory[LPC55XX_SCRATCH_MEMORY_LEN];
+	uint32_t regs[t->regs_size / sizeof(uint32_t)];
```
dragonmux, line 336: Same comment as above about VLAs

src/target/lpc55xx.c
```diff
@@ -37,12 +37,478 @@
...
+	target_mem_write(t, LPC55XX_FLASH_CONFIG_ADDRESS, backup_memory, sizeof(backup_memory));
+	target_regs_write(t, regs);
+
+	return success;
+}
+
+static bool lpc55xx_get_uuid(target *t, uint8_t *uuid)
+{
+	uint8_t backup_memory[LPC55XX_SCRATCH_MEMORY_LEN + LPC55XX_UUID_LEN];
+	uint32_t regs[t->regs_size / sizeof(uint32_t)];
```
dragonmux, line 369: Same comment about VLAs

src/target/lpc55xx.c
```diff
@@ -37,12 +37,478 @@
...
+	target_flash_s *f = calloc(1, sizeof(*f));
+
+	if (!f) {			/* calloc failed: heap exhaustion */
+		DEBUG_WARN("calloc: failed in %s\n", __func__);
+		return NULL;
+	}
+
+	struct flash_config config;
+
+	if (!lpc55xx_flash_init(t, &config)) {
```
dragonmux, line 473: Please remove the extra {}'s, they're not required

src/target/lpc55xx.c
```diff
@@ -37,12 +37,478 @@
...
+	(void)argc;
+	(void)argv;
+
+	uint8_t uuid[LPC55XX_UUID_LEN];
+
+	if (!lpc55xx_get_uuid(t, uuid))
+		return false;
+
+	tc_printf(t, "UID: 0x");
+	for (uint32_t i = 0; i < sizeof(uuid); ++i)
```
dragonmux, line 505: Please use size_t here - sizeof() returns a size_t value and you're using i for array indexing.

src/target/lpc55xx.c
```diff
@@ -75,26 +541,54 @@ void lpc55_dp_prepare(ADIv5_DP_t *dp)
 	/* At this point we assume that we've got access to the debug mailbox and can continue normally. */
 }
 
+const struct command_s lpc55xx_cmd_list[] = {
+	{"readuid", lpc55xx_read_uid, "Read out the 16-byte UID."},
+	{NULL, NULL, NULL}
```
dragonmux, line 546: Please append a `,` after the closing }

src/target/lpc55xx.c
```diff
@@ -75,26 +541,54 @@ void lpc55_dp_prepare(ADIv5_DP_t *dp)
...
 bool lpc55xx_probe(target *const t)
 {
 	ADIv5_AP_t *const ap = cortexm_ap(t);
 	if (ap->apsel == 1)
 		return false;
-#ifdef ENABLE_DEBUG
-	const uint32_t chipid = target_mem_read32(t, LPC55xx_CHIPID);
-	DEBUG_WARN("Chip ID: %08" PRIx32 "\n", chipid);
-#endif
+	uint32_t chipid = target_mem_read32(t, LPC55XX_CHIPID_ADDRESS);
```
dragonmux, line 554: chipid looks like it can be const, please make it const if you can

src/target/lpc55xx.c
```diff
@@ -75,26 +541,54 @@ void lpc55_dp_prepare(ADIv5_DP_t *dp)
...
-	target_add_ram(t, 0x20000000, 0x44000);
 	t->target_options |= CORTEXM_TOPT_INHIBIT_NRST;
-	switch (t->cpuid & CPUID_PATCH_MASK) {
-	case 3:
-		t->driver = "LPC55";
+	t->driver = lpc55xx_get_device_name(chipid);
+
+	switch (chipid) {
+	case LPC5502_CHIPID:
+		target_add_ram(t, 0x04000000, 0x4000); // SRAM_X
```
dragonmux, line 562: U on the ends of all these constants please

dragonmux (changes requested):
This is shaping up very nicely. We've done a normal review and found a few things we'd like addressed, but we're very impressed at your work on this

src/target/lpc55xx.c
TomCrypto: Just for consistency, will remove it

src/target/lpc55xx.c
TomCrypto: Agreed, I actually just copied this from the lpc11xx readuid

src/target/lpc55xx.c
TomCrypto: What's the right way to do this? It looks like they are all M33 with FPU, so I think it would be `sizeof(regnum_cortex_m) + sizeof(regnum_cortex_mf)` bytes, but these are defined in `cortexm.c`. Should I just hardcode it to 20 + 33 = 53? I couldn't find any other example of register saving that doesn't use VLAs.

TomCrypto: I am unable to complete this item because the macros discussed in the discord (to retrieve the number of Cortex-M33 registers) are not available yet.

dragonmux: We were suggesting this PR would include a patch to add those macros - but if you wish us to do that first, then no worries, we can do that.

TomCrypto: Oh sorry I didn't understand that you intended this patch to do it. I figured it was unrelated work that could be separate to this work. My bad for this miscommunication; I'll let you do it and will update the PR afterwards

dragonmux: ok, works for us. We'll add the macros either to the base branch or to main once v1.9 is out (probably the base branch for this for now so you can finish this PR)

TomCrypto:
Rebased on latest `feature/lpc55-support`. Outstanding issues:

* the debug mailbox attach got broken due to incorrect handling of `target_regs_read/write` when `regs_size` is not zero and no read/write handlers are provided (as is the case for the debug mailbox), @dragonmux as discussed
* the target reset before/after the flash operations is still not working for unknown reasons, would be good to look at it as otherwise it requires annoying manual workarounds to make sure the target won't be servicing interrupts during the flash IAP calls

Otherwise the target detection and flashing still works as before on my board.

src/target/lpc55xx.c
```diff
@@ -37,12 +37,515 @@
...
+#define LPC55S14_CHIPID             // (unknown)
+#define LPC55S16_CHIPID             // (unknown)
+#define LPC55S26_CHIPID 0xa010229aU // NXP forum
+#define LPC55S28_CHIPID 0xa010119cU // NXP forum
+#define LPC55S66_CHIPID             // (unknown)
+#define LPC55S69_CHIPID             // (unknown)
+
+/* The available IAP commands that we support, mostly flash access */
+
+enum iap_cmd {
```
dragonmux, line 91: this could be typedef'd to `iap_cmd_e` as in
```c
typedef enum iap_cmd {
    [...]
} iap_cmd_e;
```
which would then simplify uses below (same for the iap_status enum)

src/target/lpc55xx.c
```diff
@@ -37,12 +37,515 @@
...
+
+	target_addr_t flash_table_address = lpc55xx_get_flash_table_address(target, bootloader_tree_address);
+	return target_mem_read32(target, flash_table_address + sizeof(uint32_t));
+}
+
+static target_addr_t lpc55xx_get_flash_erase_address(target_s *target)
+{
+	target_addr_t bootloader_tree_address = lpc55xx_get_bootloader_tree_address(target);
+
+	if (lpc55xx_get_rom_api_version(target, bootloader_tree_address) == 0) {
```
dragonmux, line 206: Excess braces, please remove them

src/target/lpc55xx.c
```diff
@@ -37,12 +37,515 @@
...
+	target_addr_t flash_table_address = lpc55xx_get_flash_table_address(target, bootloader_tree_address);
+	return target_mem_read32(target, flash_table_address + 3 * sizeof(uint32_t));
+}
+
+static target_addr_t lpc55xx_get_ffr_init_address(target_s *target)
+{
+	target_addr_t bootloader_tree_address = lpc55xx_get_bootloader_tree_address(target);
+	target_addr_t flash_table_address = lpc55xx_get_flash_table_address(target, bootloader_tree_address);
+
+	if (lpc55xx_get_rom_api_version(target, bootloader_tree_address) == 0) {
```
dragonmux, line 231: Excess braces, please remove them

src/target/lpc55xx.c
```diff
@@ -37,12 +37,515 @@
...
+		continue;
+
+	/* Read back the status code from r0 and return */
+
+	target_regs_read(target, regs);
+
+	return (enum iap_status)regs[0];
+}
+
+struct flash_config {
```
dragonmux, line 310: Same comment about typedef'ing:
```c
typedef struct lpc55xx_flash_config {
    [...]
} lpc55xx_flash_config_s;
```

src/target/lpc55xx.c
```diff
@@ -37,12 +37,515 @@
...
+};
+
+static void lpc55xx_prepare_flash_config(target_s *target, target_addr_t address)
+{
+	/* The flash_config structure is 60 bytes in size, zero it out as that
+	 * is what the SDK does. For some reason you have to fill in the clock
+	 * speed field ("sysFreqInMHz") before flash_init. Set it to 96MHz (?)
+	 */
+
+	struct flash_config config;
```
dragonmux, line 329: You can make use of named initialisation syntax here (example assumes the typedef from above):
```c
lpc55xx_flash_config config = {
    .sysFreqInMHz = LPC55xx_IAP_FREQ_IN_MHZ,
};
```
Also, please use lower_snake_case for all the member naming

dragonmux (changes requested):
There are a couple of items that need addressing and a couple of items of clean-up to optionally address, but this is looking great. With those handled, this looks ready for merge

src/target/lpc55xx.c
TomCrypto: Yeah sorry about that, I think it has this case because I copied the name from the NXP SDK and never noticed it was out of place. Will fix!

src/target/lpc55xx.c
```diff
@@ -37,12 +37,511 @@
...
+	else
+		return target_mem_read32(target, flash_table_address + 10 * sizeof(uint32_t));
+}
+
+static target_addr_t lpc55xx_get_ffr_get_uuid_address(target_s *target)
+{
+	target_addr_t bootloader_tree_address = lpc55xx_get_bootloader_tree_address(target);
+	target_addr_t flash_table_address = lpc55xx_get_flash_table_address(target, bootloader_tree_address);
+
+	if (lpc55xx_get_rom_api_version(target, bootloader_tree_address) == 0) {
```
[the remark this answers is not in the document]
TomCrypto: Done @dragonmux I didn't find any other excess braces in the file

src/target/lpc55xx.c
dragonmux: We finally got back around to fixing this and are introducing a new Cortex header that has stuff common across the CPU family, including these much awaited register count macros

PR #1192, opened 2022-08-14
Feature: high-z TCK/SWCLK when idle

### LINKED ISSUE #945: Set SWCLK and SWDIO to Hi-Z on detatch
There are many cases where the SWD interface pins are re-purposed in the target application - eg those pins are remapped for other GPIO / peripheral interfaces when the embedded firmware is running. In these cases the SWD interface is only available immediately after power-up while the NRST is asserted. After programming and subsequent reset there will be contention on the SWD pins if the BMP continues to drive these signals, so it's desirable for the BMP to set these pins to Hi-Z after completing the programming cycle.

Current native BMP hardware uses external buffers between the STM32F103 MCU running BMP firmware and the target and will not be able to completely disable the SWD lines, but non-native hardware (and much older versions of BMP) will still benefit if these pins are disabled on the STM32F103.

Note that simply disabling the SWD buffers by means of TPWR is not sufficient since it's still necessary to power the target.

dragonmux:
## Detailed description

Certain low pin-count parts very heavily overload the device pins with functionality. This can mean it's inappropriate to drive TCK/SWCLK before initiating a scan, after scan but before attach, and after detach.

This PR introduces tristating support for these devices when running native hardware revision 6 and newer, allowing BMP to work around this problem. This also means BMP will play nicer with other devices on the roadmap.

This has been tested working against both the SWD and JTAG modes of the Tiva-C TM4C123GH6PM.

## Closing issues

Fixes #945

### DIFF
--- a/COPYING-MIT
+++ b/COPYING-MIT
@@ -1,0 +1,21 @@
+ MIT License
+
+ Copyright (C) 2022 1BitSquared <info@1bitsquared.com>
+
+ Permission is hereby granted, free of charge, to any person obtaining a copy
+ of this software and associated documentation files (the "Software"), to deal
+ in the Software without restriction, including without limitation the rights
+ to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ copies of the Software, and to permit persons to whom the Software is
+ furnished to do so, subject to the following conditions:
+
+ The above copyright notice and this permission notice shall be included in all
+ copies or substantial portions of the Software.
+
+ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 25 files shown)


emeb:
I've tested this out with recent BMP hardware on an STM32G030J6 that needs to share its SWCLK pin with a UART TX signal. The Hi-Z feature on the BMP SWCLK signal allows the G030J6 to safely share the pin for both programing and functional operation without danger of contention which is confirmed by measuring the Voh/Vol levels prior / during / post flashing. Levels were observed to be fully 3.3V / 0V with no evidence of excessive current flow.

esden (approved):
LGTM, tested on 1bitsy (STM32F415) and BMP V2.3 (GD32F103) all seems good so far. :crossed_fingers:

PR #1195, opened 2022-08-18
Fix: JTAG ADIv5 DPv0 detection

dragonmux:
## Detailed description

As part of the ongoing work to improve our ADIv5 compliance correctness, Perigoso had to "brick" support for ADIv5 JTAG for version 0 Debug Ports as the standard provides no reasonable way to detect these. As it happens, unfortunately, both the Tiva-C and the LPC43xx devices have version 0 DPs in JTAG mode (but not SWD!).

This PR restores JTAG on these parts to working by looking for a "magic" JTAG ID Code which Perigoso and us have jointly determined a probable safe way to detect at least some of these DPv0 parts.

This is tested on and works to detect both the aforementioned processor series.

### DIFF
--- a/src/platforms/hosted/stlinkv2.c
+++ b/src/platforms/hosted/stlinkv2.c
@@ -1099,7 +1099,7 @@ uint32_t stlink_swdp_scan(bmp_info_t *info)
 
 	stlink_dp_error(dp);
 
-	adiv5_dp_init(dp);
+	adiv5_dp_init(dp, 0);
 
 	return target_list ? 1U : 0U;
 }
--- a/src/target/adiv5.c
+++ b/src/target/adiv5.c
@@ -681,7 +681,7 @@ static void rp_rescue_setup(ADIv5_DP_t *dp)
 	return;
 }
 
-void adiv5_dp_init(ADIv5_DP_t *dp)
+void adiv5_dp_init(ADIv5_DP_t *dp, const uint32_t idcode)
 {
 	/*
 	 * Assume DP v1 or later.
@@ -691,10 +691,11 @@ void adiv5_dp_init(ADIv5_DP_t *dp)
 	 *
 	 * for SWD-DP, we are guaranteed to be DP v1 or later.
 	 */
-	volatile uint32_t dpidr;
+	volatile uint32_t dpidr = 0;
 	volatile struct exception e;
 	TRY_CATCH (e, EXCEPTION_ALL) {
-		dpidr = adiv5_dp_read(dp, ADIV5_DP_DPIDR);
+		if (idcode != JTAG_IDCODE_ARM_DPv0)
+			dpidr = adiv5_dp_read(dp, ADIV5_DP_DPIDR);
 	}
 	if (e.type) {
 		DEBUG_WARN("DP not responding!...\n");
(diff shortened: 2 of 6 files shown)


src/target/adiv5.c
```diff
@@ -691,10 +691,11 @@ void adiv5_dp_init(ADIv5_DP_t *dp)
 	 *
 	 * for SWD-DP, we are guaranteed to be DP v1 or later.
 	 */
-	volatile uint32_t dpidr;
+	volatile uint32_t dpidr = idcode;
```
perigoso, line 694: I think we should not attribute idcode to dpidr, they are explicitly "different", and it was the kind of thing that initialy confused me leadind to the rework on this code,  since in the end we opted to look for a particular match idcode we dont need to either, but we should initialize it to 0

perigoso, line 694: I think we should not attribute idcode to dpidr, they are explicitly "different", and it was the kind of thing that initialy confused me leadind to the rework on this code, since in the end we opted to look for a particular match idcode we dont need to either, but we should initialize it to 0

src/platforms/hosted/stlinkv2.c
```diff
@@ -1099,7 +1099,7 @@ uint32_t stlink_swdp_scan(bmp_info_t *info)
 
 	stlink_dp_error(dp);
 
-	adiv5_dp_init(dp);
+	adiv5_dp_init(dp, UINT32_MAX);
```
perigoso, line 1102: we can pass 0 here, and the other `adiv5_dp_init` calls alike, since in the end we went for looking for an exact match and not parsing the version from it, that said i dont see anything wrong with passing this either

src/target/adiv5.c
```diff
@@ -731,13 +732,9 @@ void adiv5_dp_init(ADIv5_DP_t *dp)
...
 	}
-	if (dp->version == 0) {
+	if (dp->version == 0)
 		/* DP v0 */
-		DEBUG_WARN("DPv0 detected, no designer code available\n");
-		dp->designer_code = 0;
-		dp->partno = 0;
-		dp->mindp = 0;
-	}
+		DEBUG_WARN("DPv0 detected, designer detected from JTAG IDCode\n");
```
perigoso, line 737: we dont actually parse the designer from the idcode, so the comment isnt right, but we could

perigoso (dismissed):
LGTM, in general

we could parse the idcode, if we have a dp v0 and an idcode was passed here, but since that data will not be needed elsewhere we can leave this as is, resulting in 0's for dp's v0 (that is if you go with my suggestion)

src/target/adiv5.c
```diff
@@ -731,13 +732,9 @@ void adiv5_dp_init(ADIv5_DP_t *dp)
 		}
```
perigoso, line 732: this is now missing the cleanup
```
dp->designer_code = 0;
dp->partno = 0;
dp->mindp = 0;
```
realistically this doesnt make a difference, and its still unknown if this check actually catches anything

perigoso (approved):
LGTM

esden (approved):
LGTM, thanks for the hard work sorting this out @dragonmux and @perigoso :D

PR #1196, opened 2022-08-19
target/cortex: dynamically generate target description strings to save code size

Qyriad:
## Detailed description

This PR size-optimizes the target description XML strings used for Cortex-M and Cortex-A targets, by removing the string literals which contained the verbatim XML and replacing them with functions `create_tdesc_cortex_m()`, `create_tdesc_cortexmf()`, and `create_tdesc_cortex_a()`. These functions generate the target description strings at runtime with sequences of string formatting instead of monolithic string literals.

The size-optimized code is, unfortunately, much less readable than the verbatim string literals, but the XML they generate has been left in as comments for clarity, and the string deduplication has overall saved ~3.2 KiB of flash space.

### Left to Do

As you may have noticed, this PR is marked as a draft. We have *not* tested this code on any Cortex-A targets (or non floating point Cortex-M targets, though `create_tdesc_cortex_mf()` uses `create_tdesc_cortex_m()` internally) as we don't have any on hand. We compiled and ran these functions as host code to compare their output with the string literals that were being used before, and did find them identical (post-format), but this code should still probably be tested on actual hardware before being merged.

### DIFF
--- a/src/include/general.h
+++ b/src/include/general.h
@@ -42,6 +42,8 @@
 #include "platform.h"
 #include "platform_support.h"
 
+#define ARRAY_SIZE(arr) (sizeof(arr) / sizeof(arr[0]))
+
 extern uint32_t delay_cnt;
 
 enum BMP_DEBUG {
--- a/src/include/general.h
+++ b/src/include/general.h
@@ -42,7 +42,9 @@
 #include "platform.h"
 #include "platform_support.h"
 
-#define ARRAY_SIZE(arr) (sizeof(arr) / sizeof(arr[0]))
+#ifndef ARRAY_LENGTH
+#define ARRAY_LENGTH(arr) (sizeof(arr) / sizeof(arr[0]))
+#endif
 
 extern uint32_t delay_cnt;
 
--- a/src/target/cortexa.c
+++ b/src/target/cortexa.c
@@ -36,8 +36,6 @@
 
 #include <assert.h>
 
-#define ARRAY_SIZE(arr) (sizeof(arr) / sizeof(arr[0]))
-
 static const char cortexa_driver_str[] = "ARM Cortex-A";
 
 static bool cortexa_attach(target *t);
--- a/src/target/cortexa.c
+++ b/src/target/cortexa.c
@@ -174,7 +174,7 @@ static const gdb_reg_type_e cortex_a_spr_types[] = {
 	GDB_TYPE_UNSPECIFIED // cpsr
 };
 
-static_assert(ARRAY_SIZE(cortex_a_spr_types) == ARRAY_SIZE(cortex_a_spr_names), "SPR array length mixmatch! SPR type "
+static_assert(ARRAY_LENGTH(cortex_a_spr_types) == ARRAY_LENGTH(cortex_a_spr_names), "SPR array length mixmatch! SPR type "
                                                                                 "array should have the same length as "
                                                                                 "SPR name array.");
 
@@ -270,7 +270,7 @@ static size_t create_tdesc_cortex_a(char *buffer, size_t max_len)
 	// all of the Cortex-A target description SPRs have the same bitsize, and none of them
 	// have a specified save-restore value. So we only need one "associative array" here.
 	// NOTE: unlike the other loops, this loop uses a size_t for its counter, as it's used to index into arrays.
-	for (size_t i = 0; i < ARRAY_SIZE(cortex_a_spr_names); ++i) {
+	for (size_t i = 0; i < ARRAY_LENGTH(cortex_a_spr_names); ++i) {
 		gdb_reg_type_e type = cortex_a_spr_types[i];
 
 		if (max_len != 0)
--- a/src/target/cortexa.c
+++ b/src/target/cortexa.c
@@ -31,7 +31,12 @@
 #include "exception.h"
 #include "adiv5.h"
 #include "target.h"
+#include "gdb_reg.h"
 #include "target_internal.h"
+
+#include <assert.h>
+
+#define ARRAY_SIZE(arr) (sizeof(arr) / sizeof(arr[0]))
 
 static const char cortexa_driver_str[] = "ARM Cortex-A";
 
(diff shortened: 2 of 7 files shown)


src/target/cortexa.c
```diff
@@ -199,6 +240,90 @@ static const char tdesc_cortex_a[] =
...
+	// follow it with the GDB ARM Core feature tag.
+	total += snprintf(buffer, printsz,
+		"%s feature %s "
+		"<feature name=\"org.gnu.gdb.arm.core\">",
+		gdb_arm_preamble_first,
+		gdb_arm_preamble_second
+	);
+
+	// Then the general purpose registers, which have names of r0 to r12.
+	for (uint8_t i = 0; i <= 12; ++i) {
```
dragonmux, line 272: we recommend using size_t here as the codegen won't be more expensive but it avoids integer promotion interactions

src/target/cortexa.c
```diff
@@ -199,6 +240,90 @@ static const char tdesc_cortex_a[] =
...
+
+		if (max_len != 0)
+			printsz = max_len - total;
+
+		total += snprintf(buffer + total, printsz, "<reg name=\"r%d\" bitsize=\"32\"/>", i);
```
dragonmux, line 277: Please use %u, %d is for signed integers

src/target/cortexa.c
```diff
@@ -199,6 +240,90 @@ static const char tdesc_cortex_a[] =
...
+	// Now onto the floating point registers.
+	// The first register is unique; the rest all follow the same format.
+	total += snprintf(buffer + total, printsz,
+		"</feature>"
+		"<feature name=\"org.gnu.gdb.arm.vfp\">"
+		"<reg name=\"fpscr\" bitsize=\"32\"/>"
+	);
+
+	// Now onto the simple ones.
+	for (uint8_t i = 0; i <= 15; ++i) {
```
dragonmux, line 309: Same comment as the loop above

src/target/cortexa.c
```diff
@@ -199,6 +240,90 @@ static const char tdesc_cortex_a[] =
...
+		if (max_len != 0)
+			printsz = max_len - total;
+
+		total += snprintf(buffer + total, printsz,
+			"<reg name=\"d%d\" bitsize=\"64\" type=\"float\"/>",
```
dragonmux, line 314: Same comment as the snprintf above

src/target/cortexm.c
```diff
@@ -165,9 +261,80 @@ static const char tdesc_cortex_m[] =
...
+	total += snprintf(buffer, printsz,
+		"%s target %s "
+		"<feature name=\"org.gnu.gdb.arm.m-profile\">",
+		gdb_arm_preamble_first,
+		gdb_arm_preamble_second
+	);
+
+	// Then the general purpose registers, which have names of r0 to r12,
+	// and all the same bitsize.
+	for (uint8_t i = 0; i <= 12; ++i) {
```
dragonmux, line 292: This loop has the same nitpick as the one above in cortexa.c

src/target/cortexm.c
```diff
@@ -216,8 +383,52 @@ static const char tdesc_cortex_mf[] =
...
+	}
+
+
+	total += snprintf(buffer + total, printsz,
+		"<feature name=\"org.gnu.gdb.arm.vfp\">"
+		"<reg name=\"fpscr\" bitsize=\"32\"/>"
+	);
+
+	// After fpscr, the rest of the vfp registers follow a regular format: d0-d15, bitsize 64, type float.
+	for (uint8_t i = 0; i <= 15; ++i) {
```
dragonmux, line 416: This loop has the same nitpick as the one above in cortexa.c

src/target/cortexm.c
```diff
@@ -355,8 +566,32 @@ bool cortexm_probe(ADIv5_AP_t *ap)
...
+
+
 	/* Should probe here to make sure it's Cortex-M3 */
-	t->tdesc = tdesc_cortex_m;
+
+	// Find the buffer size needed for the target description string we need to send to GDB,
+	// and then compute the string itself.
+	size_t size_needed;
+	char *buffer;
+	if (!is_cortexmf) {
```
dragonmux, line 584: Does it make sense, as these allocate, to move building the tdesc buffers to cortexm_attach() (w/ cortexm_detach() free()'ing the buffers)? Thinking about allocation lifetimes and trying to minimise the RAM overheads here

dragonmux (changes requested):
This is looking really good and we look forward to getting to merge it!

Our only other general nitpick than the review items is that the patch appears to be using two blank lines between functions and blocks which is counter to the style the rest of the codebase uses - please only use one blank line between functions and blocks. (clang-format should be collapsing them to single line for that matter)

src/target/cortexa.c
Qyriad: `size_t` won't really avoid integer promotion interactions as variadic arguments are promoted to (signed) `int`. If we want to avoid any actual problems due to integer promotion interactions, then we should either use an unsigned "small" type like `unsigned short` or `uint8_t` to avoid any actual consequences of integer promotion; or we should use a (signed) `int` to prevent promotion in the first place.

That being said, we also already know there cannot be any consequences of integer promotion interactions in this case as the value of `i` is very clearly and strictly bounded.

We're fine with changing it from `uint8_t` as a matter of style, but we don't think that `size_t` specifically makes much sense here.

src/target/cortexm.c
Qyriad: That might make sense. We were treating these strings as "lazily allocated" semi-static, but you're right that that multiple attaches and detaches can happen throughout the firmware program's lifetime. But in that case it might actually make the most sense to store each target description as a static global pointer, which is checked upon attach and allocated and initialized if it hasn't been already, that way each kind of tdesc is only allocated and initialized once.

Would that be acceptable or would you prefer we `malloc()` and `free()` on attach and detach respectively for each target, as you suggested?

src/target/cortexa.c
dragonmux: We were referring to promotion shenanigans within the loop, not specifically the sprintf(). Trying to keep the code base additions `-Wconversion` and friends clean and this may ping a warning when we enable that - size_t or uint32_t won't (they're also the same underlying type on ARM which is why we suggested size_t here).

We've also found the transform we're requesting consistently reduces code size by a couple of instructions (strb/ldrb vs str/ldr).

src/target/cortexm.c
dragonmux: We'd prefer an alloc/free on every attach/detach as for processors that have multiple cores or long scan-chains, we're liable to run out of memory if users are also running flashing commands as those allocate some fairly large buffers too. Basically, concerned with getting a repeat of #960 (#1093 being the partial fix)

src/target/cortexa.c
Qyriad: So, `strb`/`ldrb` have the same size as `str`/`ldr`, so that's not going to reduce code size. We checked `-Wconversion` and it doesn't raise any warnings there, because `-Wconversion` is for cases where integer promotion changes the value/result, and in the loop itself there aren't any interactions where it changes the result (since it's just simple compares and additions with statically bounded values to literals). Checking `-Wconversion` *did* point out that `snprintf()` returns an `int` and thus adding its return to a `size_t` changes the sign, so for correctness most of the `size_t`s that are in these functions should actually be `int`s.

We could also use `uint32_t` like you suggested (or probably `int32_t`, due to the aforementioned sign issue), but from a correctness standpoint that's equally wrong to `size_t`, and all of these should simply be `int`. `uint32_t` does happen to be typedef'd to `unsigned long` on our platform, which also happens to be the same width as `unsigned int` on our platform, but if we're looking from the standpoint of what Actually Happens instead of from a Correctness standpoint, then none of this matters in the first place — we checked the disassembly and confirmed that none of these changes make any difference.

src/include/general.h
```diff
@@ -42,6 +42,8 @@
 #include "platform.h"
 #include "platform_support.h"
 
+#define ARRAY_SIZE(arr) (sizeof(arr) / sizeof(arr[0]))
```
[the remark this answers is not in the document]
Qyriad: If libopencm3 is going to define the same macro, perhaps we should leave it without a guard so the warning pops up when libopencm3 gets that macro, and then we can remove it from our codebase? Having potentially two versions of the same macro showing up in the codebase could get a bit confusing.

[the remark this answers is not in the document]
dragonmux: We figured that, as long as BMP's version of the macro matches locm3's and behaves the same, having it in general.h for things that don't include locm3's USB headers but with a definition guard where both are included, provides the best possible coverage for the code base and as they'd be named the same and operate the same, no confusion should result.

src/target/cortexa.c
dragonmux: We've talked about this over on Discord and decided a way forward, marking this as resolved pending the update discussed

Qyriad:
Formatted, guarded the ARRAY_LENGTH macro, and a little other cleanup. This does still needs testing on a Cortex-A device, though.

dragonmux (approved):
LGTM, we'll merge this once the build finishes

PR #1199, opened 2022-08-20
Fix: newlib file IO hooks

dragonmux:
## Detailed description

While trying to figure out how printf() worked for the umpteenth time in the code base, we finally stumbled across the RDI file IO halo code which serves `_write()`, `isatty()` and other parts of newlib's `FILE *` support via the SVC interrupt.

Having stared at the code long and hard, we figured out that we can make use of the weak linkage of the two aforementioned functions to avoid having to do this interrupt vector halo, saving around 200 bytes of Flash space, and then write out the manual assembly to insert the RDI vector routine by careful manipulation of state in a bare function.

We've then added documentation to make this easier to find in future, and removed a private function from the public platform headers. This PR is the result.

### DIFF
--- a/src/platforms/f072/platform.h
+++ b/src/platforms/f072/platform.h
@@ -145,7 +145,6 @@
 
 #ifdef ENABLE_DEBUG
 extern bool debug_bmp;
-int usbuart_debug_write(const char *buf, size_t len);
 # define DEBUG printf
 #else
 # define DEBUG(...)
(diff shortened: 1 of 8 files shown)


esden (approved):
LGTM

For context to others that might read this. We agreed with @dragonmux that we will bring back some kind of output through the "inception" debugger for BMP. This used to be done through semihosting when a debugger was connected. We now have more options than just semihosting and there will be some solution brought back to make debug output work in "inception" mode again.

For now this code is a cleanup and improvement especially from the flash usage perspective which we are trying to focus on.

PR #1202, opened 2022-08-22
Fix: serial code reoganisation

dragonmux:
## Detailed description

After the initial work done in #1199, we found the code in usbuart.{c,h} to be very confused in purpose - most of it being to do with the AUX serial port and not USB.

Due to this, we've ended up refactoring this code into usb_serial.{c,h} and a new aux_serial.{c,h}, unifying AUX serial handling across all platforms.

In doing this, we inadvertently found and fixed an issue with debug serial newline handling due to '\r' injection not being quite right, and found and fixed that two platforms were no-build with ENABLE_DEBUG=1 (f3, f072).

This has been tested on native hardware and works.


src/platforms/common/usb_serial.c
```diff
@@ -206,14 +213,12 @@ size_t debug_uart_write(const char *buf, const size_t len)
 
 	for (size_t i = 0; i < len && (usb_dbg_in + 1) % AUX_UART_BUFFER_SIZE != usb_dbg_out; ++i) {
 		if (buf[i] == '\n') {
-			usb_dbg_buf[usb_dbg_in++] = '\r';
-			usb_dbg_in %= AUX_UART_BUFFER_SIZE;
+			debug_serial_append_char('r');
```
esden: Should this not be `\r`? Or is it fixed in a later commit?

esden (commented):
As far as I can tell I found only one thing that is potentially an issue. Otherwise it is ready to merge.

src/platforms/common/usb_serial.c
dragonmux: It should and while it is fixed in a later commit, we've opted to edit this one and rebase so that the bug's not introduced in the first place.

esden (approved):
LGTM

PR #1214, opened 2022-08-26
sam4l: fix debug formatting

xobs:
SAM4L added a DEBUG_INFO command that uses insufficiently-specific formatting.
This breaks compiles on ESP32.

Use specific formatters. Additionally, ensure that FLASHCALW_BASE is
defined as a long int, which allows it to be formatted with "%x".

## Detailed description

This fixes the build for ESP32 by adding more formatting information to sam4l

### DIFF
--- a/src/target/sam4l.c
+++ b/src/target/sam4l.c
@@ -32,7 +32,7 @@
 /*
  * Flash Controller defines
  */
-#define FLASHCALW_BASE				0x400A0000
+#define FLASHCALW_BASE				UINT32_C(0x400A0000)
 
 /* Flash Control Register */
 #define	FLASHCALW_FCR				(FLASHCALW_BASE + 0x00)
@@ -295,7 +295,8 @@ sam4l_extended_reset(target *t)
 static bool sam4l_flash_command(target *t, uint32_t page, uint32_t cmd)
 {
 	DEBUG_INFO(
-		"\nSAM4L: sam4l_flash_command: FSR: 0x%08" PRIx32 ", page = %lu, command = %lu\n", FLASHCALW_FSR, page, cmd);
+		"\nSAM4L: sam4l_flash_command: FSR: 0x%08" PRIx32
+		", page = %" PRIu32 ", command = %" PRIu32 "\n", FLASHCALW_FSR, page, cmd);
 
 	/* wait for Flash controller ready */
 	platform_timeout timeout;


src/target/sam4l.c
```diff
@@ -32,7 +32,7 @@
 /*
  * Flash Controller defines
  */
-#define FLASHCALW_BASE				0x400A0000
+#define FLASHCALW_BASE				0x400A0000L
```
dragonmux, line 35: Rather than suffixing L here, could you use U please (this is used in an unsigned context, so should be marked as such)

dragonmux (changes requested):
Ah, excellent spot. There's one small thing we noticed in review, but with that addressed we'll merge this ASAP.

src/target/sam4l.c
xobs: Unfortunately, U doesn't work:

```
C:/Users/Sean/esp/esp-idf/components/log/include/esp_log.h:270:27: note: in expansion of macro 'LOG_COLOR'
  270 | #define LOG_COLOR_I       LOG_COLOR(LOG_COLOR_GREEN)
      |                           ^~~~~~~~~
C:/Users/Sean/esp/esp-idf/components/log/include/esp_log.h:282:37: note: in expansion of macro 'LOG_COLOR_I'
  282 | #define LOG_FORMAT(letter, format)  LOG_COLOR_ ## letter #letter " (%" PRIu32 ") %s: " format LOG_RESET_COLOR "\n"
      |                                     ^~~~~~~~~~
C:/Users/Sean/esp/esp-idf/components/log/include/esp_log.h:414:86: note: in expansion of macro 'LOG_FORMAT'
  414 |         else                                { esp_log_write(ESP_LOG_INFO,       tag, LOG_FORMAT(I, format), esp_log_timestamp(), tag, ##__VA_ARGS__); } \
      |                                                                                      ^~~~~~~~~~
C:/Users/Sean/esp/esp-idf/components/log/include/esp_log.h:432:41: note: in expansion of macro 'ESP_LOG_LEVEL'
  432 |         if ( LOG_LOCAL_LEVEL >= level ) ESP_LOG_LEVEL(level, tag, format, ##__VA_ARGS__); \
      |                                         ^~~~~~~~~~~~~
C:/Users/Sean/esp/esp-idf/components/log/include/esp_log.h:340:38: note: in expansion of macro 'ESP_LOG_LEVEL_LOCAL'
  340 | #define ESP_LOGE( tag, format, ... ) ESP_LOG_LEVEL_LOCAL(ESP_LOG_ERROR,   tag, format, ##__VA_ARGS__)
      |                                      ^~~~~~~~~~~~~~~~~~~
E:/Code/Farpatch/format-demo/main/main.c:63:5: note: in expansion of macro 'ESP_LOGE'
   63 |     ESP_LOGE("foo", "Some value: %"PRIx32, 0x90800000U);
      |     ^~~~~~~~
cc1.exe: some warnings being treated as errors
```

Is it alright if I use `UL`?

dragonmux: in that case, could you use `UINT32_C()` around the constant - that'll make it unsigned and suffix it according to platform requirements which is more correct than 'UL'

xobs: Alright, I removed the `U` suffix and wrapped the debug statement argument in `UINT32_C()`.

dragonmux (approved):
LGTM, merging.

src/target/sam4l.c
xobs: It seems that `UINT32_C()` doesn't work on hosted mode. Hmm...

xobs:
One fix (that's used elsewhere in that file) is to add this patch, which feels terribly wrong but seems to work?

```patch
diff --git a/src/target/sam4l.c b/src/target/sam4l.c
index 9fadd990..4fe283f1 100644
--- a/src/target/sam4l.c
+++ b/src/target/sam4l.c
@@ -296,7 +296,7 @@ static bool sam4l_flash_command(target *t, uint32_t page, uint32_t cmd)
 {
        DEBUG_INFO(
                "\nSAM4L: sam4l_flash_command: FSR: 0x%08" PRIx32
-               ", page = %" PRIu32 ", command = %" PRIu32 "\n", UINT32_C(FLASHCALW_FSR), page, cmd);
+               ", page = %" PRIu32 ", command = %" PRIu32 "\n", (unsigned long)FLASHCALW_FSR, page, cmd);

        /* wait for Flash controller ready */
        platform_timeout timeout;
```

perigoso:
uh, funny I just removed that, hence the breakage, it does indeed feel terrible

dragonmux:
That would be because, based on our current understanding of the parts of the standard at play here, it is wrong as it only covers up the warning and its implications rather than fixing the problem which is that integer constants default to being 'int' (and therefore signed) when the formatting string calls for unsigned and the standard makes no bones that the constant you wrote may in some cases not end up as the value you see formatted out if you do that cast to unsigned long

dragonmux (approved):

PR #1216, opened 2022-08-28
Fix AppleClang compatibility

### LINKED ISSUE #1204: BMPA build broken on OSX: aliases are not supported on darwin
Building the latest HEAD revision is broken on OSX. Build fails with the following messages:

```
target/target_probe.c:50:58: error: aliases are not supported on darwin
bool cortexa_probe(ADIv5_AP_t *apb, uint32_t debug_base) CORTEXA_PROBE_WEAK_NOP;
                                                         ^
target/target_probe.c:22:53: note: expanded from macro 'CORTEXA_PROBE_WEAK_NOP'
#define CORTEXA_PROBE_WEAK_NOP __attribute__((weak, alias("cortexa_probe_nop")))
```

amyspark:
## Detailed description

👋 

This PR fixes the usage of weak aliasing for probe targets under macOS. 

AppleClang has several incompatibilities with `__attribute__((weak,
alias))` that prevent sane GCC-like usage on macOS:

- it does not support `__attribute__((alias))` at all, only the `#pragma
  weak symbol1 = symbol2` syntax (1)

- the compiler removes unused static functions prior to aliasing, which renders the above [into an error](https://bugs.llvm.org/show_bug.cgi?id=17775).

To fix this, I implemented weak aliasing using the `#pragma` directive and [the recommended indirection macros from the GCC docs](https://gcc.gnu.org/onlinedocs/cpp/Pragmas.html) and removing the static inline specifiers from the functions if building under macOS.

[1] : See, for instance, https://sourceforge.net/p/openocd/tickets/304/ and Google "error: aliases are not supported on darwin".

  * I believe AppleClang isn't compatible with hardware native BMP, so I've not tested it

## Closing issues

Fixes #1204

### DIFF
--- a/src/target/target_probe.c
+++ b/src/target/target_probe.c
@@ -19,24 +19,39 @@
 
 #include "target_probe.h"
 
-#define CORTEXA_PROBE_WEAK_NOP __attribute__((weak, alias("cortexa_probe_nop")))
-#define CORTEXM_PROBE_WEAK_NOP __attribute__((weak, alias("cortexm_probe_nop")))
-#define TARGET_PROBE_WEAK_NOP __attribute__((weak, alias("target_probe_nop")))
+#ifdef __APPLE__
+// https://bugs.llvm.org/show_bug.cgi?id=17775
+#define APPLE_STATIC
+// https://gcc.gnu.org/onlinedocs/cpp/Pragmas.html
+#define DO_PRAGMA_(x) _Pragma(#x)
+#define DO_PRAGMA(x)  DO_PRAGMA_(x)
+// __attribute__((alias)) is not supported in AppleClang.
+#define weak_alias(name, aliasname)  DO_PRAGMA(weak name = aliasname)
+#define CORTEXA_PROBE_WEAK_NOP(name) weak_alias(name, cortexa_probe_nop)
+#define CORTEXM_PROBE_WEAK_NOP(name) weak_alias(name, cortexm_probe_nop)
+#define TARGET_PROBE_WEAK_NOP(name)  weak_alias(name, target_probe_nop)
+#else
+#define APPLE_STATIC static inline
+#define CORTEXA_PROBE_WEAK_NOP(name) \
+	extern bool name(ADIv5_AP_t *, uint32_t) __attribute__((weak, alias("cortexa_probe_nop")));
+#define CORTEXM_PROBE_WEAK_NOP(name) extern bool name(ADIv5_AP_t *) __attribute__((weak, alias("cortexm_probe_nop")));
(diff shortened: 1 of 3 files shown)


src/target/target_probe.c
```diff
@@ -19,24 +19,39 @@
...
+#define weak_alias(name, aliasname)  DO_PRAGMA(weak name = aliasname)
+#define CORTEXA_PROBE_WEAK_NOP(name) weak_alias(name, cortexa_probe_nop)
+#define CORTEXM_PROBE_WEAK_NOP(name) weak_alias(name, cortexm_probe_nop)
+#define TARGET_PROBE_WEAK_NOP(name)  weak_alias(name, target_probe_nop)
+#else
+#define APPLE_STATIC static inline
+#define CORTEXA_PROBE_WEAK_NOP(name) \
+	extern bool name(ADIv5_AP_t *, uint32_t) __attribute__((weak, alias("cortexa_probe_nop")))
+#define CORTEXM_PROBE_WEAK_NOP(name) extern bool name(ADIv5_AP_t *) __attribute__((weak, alias("cortexm_probe_nop")))
+#define TARGET_PROBE_WEAK_NOP        extern bool name(target *) __attribute__((weak, alias("target_probe_nop")))
```
dragonmux, line 38: This appears to be missing `(name)` in the base macro definition?

dragonmux (changes requested):
Thank you for the contribution! There's one small thing we spotted on review. With that fixed, we're happy to merge this.

src/target/target_probe.c
```diff
@@ -19,24 +19,39 @@
...
+#define DO_PRAGMA(x)  DO_PRAGMA_(x)
+// __attribute__((alias)) is not supported in AppleClang.
+#define weak_alias(name, aliasname)  DO_PRAGMA(weak name = aliasname)
+#define CORTEXA_PROBE_WEAK_NOP(name) weak_alias(name, cortexa_probe_nop)
+#define CORTEXM_PROBE_WEAK_NOP(name) weak_alias(name, cortexm_probe_nop)
+#define TARGET_PROBE_WEAK_NOP(name)  weak_alias(name, target_probe_nop)
+#else
+#define APPLE_STATIC static inline
+#define CORTEXA_PROBE_WEAK_NOP(name) \
+	extern bool name(ADIv5_AP_t *, uint32_t) __attribute__((weak, alias("cortexa_probe_nop")))
```
dragonmux, line 36: Looks like these macros need ';' added to the end to mark the ends of the decls properly

dragonmux (approved):
LGTM, merging once the build completes.

PR #1224, opened 2022-09-01
rp: invert success logic in rp_flash_write

xobs:
## Detailed description

Target write operations return boolean values: `true` indicates a write
was successful, while `false` indicates there was a failure.

Inside the rpi target, there is a delegate function `rp_rom_call` that
performs the heavy lifting. It also returns `true` on success and
`false` on failure.

The rp_flash_write() function was inverting this logic, which caused it
to return `false` on success and `true` on failure. This behaviour was
exhibited as only the first 0x89c bytes successfully write to the
device.

Invert this logic in order to get rpi working again.


dragonmux (approved):
LGTM, glad this was a fairly simple fix - thank you for the patch. Merging after testing on our local RP2040.

dragonmux:
Finally got around to testing this. Confirmed working, merging. Thank you again for the contribution!

PR #1227, opened 2022-09-01
rp and adiv5 fixes 

jamesturton:
## Detailed description

Fixes a regression caused by 733d8ddc19a9fbefd17495b51359fc5ae509693b which does not resume a halted target.
Also fixes some bugs where we prematurely de-assert  CDBGRSTREQ and CDBGPWRUPREQ


src/target/adiv5.c
```diff
@@ -848,8 +846,7 @@ void adiv5_dp_init(ADIv5_DP_t *dp, const uint32_t idcode)
 				dp->ap_cleanup(i);
 #endif
 			if (++invalid_aps == 8) {
-				adiv5_dp_unref(dp);
-				return;
+				break;
```
dragonmux, line 849: It would be helpful to include a comment here with where this jumps to (for loop on line 874) and why. Additionally, remove the now defunct {}'s please.

src/target/adiv5.c
```diff
@@ -810,16 +810,12 @@ void adiv5_dp_init(ADIv5_DP_t *dp, const uint32_t idcode)
 	/* This AP reset logic is described in ADIv5, but fails to work
 	 * correctly on STM32.	CDBGRSTACK is never asserted, and we
 	 * just wait forever.  This scenario is described in B2.4.1
-	 * so we have a timeout mechanism in addition to the sensing one.
-	 *
-	 * Write request for debug reset */
+	 * so we have a timeout mechanism in addition to the sensing one. */
+	platform_timeout_set(&timeout, 201);
```
dragonmux, line 814: Did you mean to use 201 as the timeout in ms here? Would have expected 200 as the value for 10 loops of the loop below

dragonmux (changes requested):
Thank you for the contribution! it's generally looking very promising, with just two things we picked up in review that seemed a little off or in need of extra explanation.

Obviously, we'll need to test this before merging with at least 3 of the targets we have available here. However, with the items found addressed and that testing complete, we're happy to then merge.

src/target/adiv5.c
jamesturton: Comment added!

src/target/adiv5.c
jamesturton: Changed value to 200 😄

dragonmux (approved):
LGTM, we'll test this then merge.

jamesturton:
@dragonmux is there anything else holding up this being merged?

PR #1229, opened 2022-09-01
rp: Don't abort program when attaching

### LINKED ISSUE #1228: rp: Target can no longer continue after attaching
Firstly, nice work @dragonmux for implementing the reading of the SFDP table in the rp driver!
9681475ab5e9472443ebe9b8231cb8fa758843f2 caused a regression where by the spi peripheral will always be modified causing the running program to be aborted. Of course it is really nice to know exactly the page and block sizes of the flash chip, but maybe this could wait to be read immediately before erasing the flash, not when attaching?

jamesturton:
## Detailed description

Reading the sfdp parameter table in the flash causes the spi peripheral to be reconfigure, and, while the previous configuration is restored, this causes the running program to be aboarded. Instead read the sfdp table only if there is no valid program in the flash, or we are going to be modifying the flash anyway (e.g. on erase).

This also includes some other enhancements for accelerating the flash loading.

## Closing issues

Fixes #1228

### DIFF
--- a/src/target/rp.c
+++ b/src/target/rp.c
@@ -157,8 +157,8 @@ static bool rp_flash_write(target_flash_s *f, target_addr_t dest, const void *sr
 
 static bool rp_read_rom_func_table(target *t);
 static bool rp_attach(target *t);
-static void rp_flash_prepare(target *t);
-static void rp_flash_resume(target *t);
+static bool rp_flash_prepare(target *t);
+static bool rp_flash_resume(target *t);
 static void rp_spi_read(target *t, uint16_t command, target_addr_t address, void *buffer, size_t length);
 static uint32_t rp_get_flash_length(target *t, rp_flash_s *flash);
 static bool rp_mass_erase(target *t);
@@ -221,6 +221,8 @@ bool rp_probe(target *t)
 	t->driver = RP_ID;
 	t->target_options |= CORTEXM_TOPT_INHIBIT_NRST;
 	t->attach = rp_attach;
+	t->enter_flash_mode = rp_flash_prepare;
+	t->exit_flash_mode = rp_flash_resume;
 	target_add_commands(t, rp_cmd_list, RP_ID);
 	return true;
 }
@@ -345,30 +347,34 @@ static bool rp_rom_call(target *t, uint32_t *regs, uint32_t cmd, uint32_t timeou
 	return ret;
 }
 
-static void rp_flash_prepare(target *t)
+static bool rp_flash_prepare(target *t)
 {
 	rp_priv_s *ps = (rp_priv_s *)t->target_storage;
+	bool ret = true; /* catch false returns with &= */
 	if (!ps->is_prepared) {
 		DEBUG_INFO("rp_flash_prepare\n");
 		/* connect*/
-		rp_rom_call(t, ps->regs, ps->rom_connect_internal_flash, 100);
+		ret &= rp_rom_call(t, ps->regs, ps->rom_connect_internal_flash, 100);
 		/* exit_xip */
-		rp_rom_call(t, ps->regs, ps->rom_flash_exit_xip, 100);
+		ret &= rp_rom_call(t, ps->regs, ps->rom_flash_exit_xip, 100);
 		ps->is_prepared = true;
 	}
+	return ret;
 }
 
-static void rp_flash_resume(target *t)
+static bool rp_flash_resume(target *t)
 {
 	rp_priv_s *ps = (rp_priv_s *)t->target_storage;
+	bool ret = true; /* catch false returns with &= */
 	if (ps->is_prepared) {
 		DEBUG_INFO("rp_flash_resume\n");
 		/* flush */
-		rp_rom_call(t, ps->regs, ps->rom_flash_flush_cache, 100);
+		ret &= rp_rom_call(t, ps->regs, ps->rom_flash_flush_cache, 100);
 		/* enter_cmd_xip */
-		rp_rom_call(t, ps->regs, ps->rom_flash_enter_xip, 100);
+		ret &= rp_rom_call(t, ps->regs, ps->rom_flash_enter_xip, 100);
 		ps->is_prepared = false;
 	}
+	return ret;
 }
 
 /*
@@ -384,7 +390,6 @@ static bool rp_flash_erase(target_flash_s *f, target_addr_t addr, size_t len)
 	target *t = f->t;
 
 	/* Update our assumptions with the SFDP params */
-	rp_flash_prepare(t);
 	spi_parameters_s spi_parameters;
 	rp_flash_s *flash = (rp_flash_s *)f;
 	if (!(flash->table_has_been_read) && sfdp_read_parameters(t, &spi_parameters, rp_spi_read_sfdp)) {
@@ -396,12 +401,10 @@ static bool rp_flash_erase(target_flash_s *f, target_addr_t addr, size_t len)
 
 	if (addr & (f->blocksize - 1)) {
 		DEBUG_WARN("Unaligned erase\n");
-		rp_flash_resume(t);
 		return false;
 	}
 	if ((addr < f->start) || (addr >= f->start + f->length)) {
 		DEBUG_WARN("Address is invalid\n");
-		rp_flash_resume(t);
 		return false;
 	}
 	addr -= f->start;
@@ -451,7 +454,6 @@ static bool rp_flash_erase(target_flash_s *f, target_addr_t addr, size_t len)
 		if (full_erase)
 			target_print_progress(&timeout);
 	}
-	rp_flash_resume(t);
 	DEBUG_INFO("Erase done!\n");
 	return ret;
 }
@@ -467,7 +469,6 @@ static bool rp_flash_write(target_flash_s *f, target_addr_t dest, const void *sr
 	dest -= f->start;
 	rp_priv_s *ps = (rp_priv_s *)t->target_storage;
 	/* Write payload to target ram */
-	rp_flash_prepare(t);
 	bool ret = true;
 	while (len) {
 		uint32_t chunksize = (len <= MAX_WRITE_CHUNK) ? len : MAX_WRITE_CHUNK;
@@ -489,7 +490,6 @@ static bool rp_flash_write(target_flash_s *f, target_addr_t dest, const void *sr
 		src += chunksize;
 		dest += chunksize;
 	}
-	rp_flash_resume(t);
 	DEBUG_INFO("Write done!\n");
 	return ret;
 }
@@ -498,9 +498,12 @@ static bool rp_mass_erase(target *t)
 {
 	rp_priv_s *ps = (rp_priv_s *)t->target_storage;
 	ps->is_monitor = true;
-	const bool result = rp_flash_erase(t->flash, t->flash->start, t->flash->length);
+	bool ret = true; /* catch false returns with &= */
+	ret &= rp_flash_prepare(t);
+	ret &= rp_flash_erase(t->flash, t->flash->start, t->flash->length);
+	ret &= rp_flash_resume(t);
 	ps->is_monitor = false;
-	return result;
+	return ret;
 }
 
 static void rp_spi_chip_select(target *const t, const bool active)
@@ -638,13 +641,16 @@ static bool rp_cmd_erase_sector(target *t, int argc, const char **argv)
 	} else if (argc == 2)
 		length = strtoul(argv[1], NULL, 0);
 	else
-		return -1;
+		return false;
 
 	rp_priv_s *ps = (rp_priv_s *)t->target_storage;
 	ps->is_monitor = true;
-	const bool result = rp_flash_erase(t->flash, start, length);
+	bool ret = true; /* catch false returns with &= */
+	ret &= rp_flash_prepare(t);
+	ret &= rp_flash_erase(t->flash, start, length);
+	ret &= rp_flash_resume(t);
 	ps->is_monitor = false;
-	return result;
+	return ret;
 }
 
 static bool rp_cmd_reset_usb_boot(target *t, int argc, const char **argv)
--- a/src/target/rp.c
+++ b/src/target/rp.c
@@ -174,31 +174,21 @@ static void rp_add_flash(target *t)
 		return;
 	}
 
-	rp_flash_prepare(t);
-	spi_parameters_s spi_parameters;
-	if (!sfdp_read_parameters(t, &spi_parameters, rp_spi_read_sfdp)) {
-		/* SFDP readout failed, so make some assumptions and hope for the best. */
-		spi_parameters.page_size = 256U;
-		spi_parameters.sector_size = 4096U;
-		spi_parameters.capacity = rp_get_flash_length(t);
-		spi_parameters.sector_erase_opcode = SPI_FLASH_CMD_SECTOR_ERASE;
-	}
-	rp_flash_resume(t);
-
-	DEBUG_INFO("Flash size: %zu MB\n", spi_parameters.capacity / (1024U * 1024U));
-
+	/* Make some assumptions and hope for the best. */
 	target_flash_s *const f = &flash->f;
 	f->start = RP_XIP_FLASH_BASE;
-	f->length = spi_parameters.capacity;
-	f->blocksize = spi_parameters.sector_size;
+	f->length = rp_get_flash_length(t);
+	f->blocksize = 4096U;
 	f->erase = rp_flash_erase;
 	f->write = rp_flash_write;
 	f->writesize = 2048; /* Max buffer size used otherwise */
 	f->erased = 0xffU;
 	target_add_flash(t, f);
 
-	flash->page_size = spi_parameters.page_size;
-	flash->sector_erase_opcode = spi_parameters.sector_erase_opcode;
+	DEBUG_INFO("Flash size: %d MB\n", f->length / (1024U * 1024U));
+
+	flash->page_size = 256U;
+	flash->sector_erase_opcode = SPI_FLASH_CMD_SECTOR_ERASE;
 }
 
 bool rp_probe(target *t)
@@ -220,6 +210,8 @@ bool rp_probe(target *t)
 		DEBUG_WARN("calloc: failed in %s\n", __func__);
 		return false;
 	}
+	priv_storage->is_prepared = false;
+	priv_storage->is_monitor = false;
 	t->target_storage = (void *)priv_storage;
 
 	t->mass_erase = rp_mass_erase;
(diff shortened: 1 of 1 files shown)


src/target/rp.c
```diff
@@ -387,6 +388,17 @@ static bool rp_flash_erase(target_flash_s *f, target_addr_t addr, size_t len)
 {
 	DEBUG_INFO("Erase addr 0x%08" PRIx32 " len 0x%" PRIx32 "\n", addr, (uint32_t)len);
 	target *t = f->t;
+
+	/* Update our assumptions with the SFDP params */
+	spi_parameters_s spi_parameters;
+	rp_flash_s *flash = (rp_flash_s *)f;
+	if (!(flash->table_has_been_read) && sfdp_read_parameters(t, &spi_parameters, rp_spi_read_sfdp)) {
```
dragonmux, line 395: The brackets here around table_has_been_read are not required and suppress a GCC warning we want enabled. Please remove them

src/target/rp.c
```diff
@@ -595,6 +603,22 @@ static uint32_t rp_get_flash_length(target *t)
 	size = MAX_FLASH;
 
 	rp_flash_prepare(t);
+
+	spi_parameters_s spi_parameters;
+	if (!(flash->table_has_been_read) && sfdp_read_parameters(t, &spi_parameters, rp_spi_read_sfdp)) {
```
dragonmux, line 608: Same comment as above.

src/target/rp.c
```diff
@@ -617,26 +641,27 @@ static bool rp_cmd_erase_sector(target *t, int argc, const char **argv)
 	} else if (argc == 2)
 		length = strtoul(argv[1], NULL, 0);
 	else
-		return -1;
+		return false;
 
 	rp_priv_s *ps = (rp_priv_s *)t->target_storage;
 	ps->is_monitor = true;
-	const bool result = rp_flash_erase(t->flash, start, length);
+	bool ret = true; /* catch false returns with &= */
```
dragonmux, line 648: keeping the "result" naming here would be good to be honest, 'ret' in other places in the code is something we've let in but we're not especially happy with

dragonmux (changes requested):
Thank you for the contribution! We're going to need to give this very careful consideration as there's a major problem with this approach being GDB is unable to be told about the changes to the target Flash object which creates a mismatch, which is part of why we write the change we did how we did.

That said, we understand the desire to be able to break into an existing program to inspect state (for example, finding out why you ended up in a hardfault handler).

It might be that, rather than running the ROM prepare/resume routines, we actually need to manually manipulate the SPI Flash controller and do some register preservation either side of operations to preserve device state - further thought and investigation is required to find the best approach here weighing up the various problem domains, aspects, benefits and drawbacks.

src/target/rp.c
jamesturton: Fixed!

src/target/rp.c
jamesturton: Fixed!

src/target/rp.c
jamesturton: Fixed!

jamesturton:
> It might be that, rather than running the ROM prepare/resume routines, we actually need to manually manipulate the SPI Flash controller and do some register preservation either side of operations to preserve device state - further thought and investigation is required to find the best approach here weighing up the various problem domains, aspects, benefits and drawbacks.

I have now implemented our own flash access functions so it is now possible to read the SFDP table and JEDEC ID without aborting the program execution. Now I can't see any reason to keep the old method of scanning for repeating sections in the addressable flash space - instead reading the SFDP table as the primary approach!

Thoughts and comments would be greatly appreciated!

src/target/rp.c
```diff
@@ -53,16 +56,30 @@
 #define RP_SRAM_BASE          0x20000000U
 #define RP_SRAM_SIZE          0x42000U
 
-#define RP_GPIO_QSPI_BASE_ADDR     0x40018000U
-#define RP_GPIO_QSPI_CS_CTRL       (RP_GPIO_QSPI_BASE_ADDR + 0x0cU)
-#define RP_GPIO_QSPI_CS_DRIVE_LOW  (2U << 8U)
-#define RP_GPIO_QSPI_CS_DRIVE_HIGH (3U << 8U)
-#define RP_GPIO_QSPI_CS_DRIVE_MASK 0x00000300U
+#define RP_GPIO_QSPI_BASE_ADDR                 0x40018000U
```
dragonmux, line 59: clang-format would force this block to be aligned to the longest definition in the block + 1 space - please let it do so

src/target/rp.c
```diff
@@ -85,6 +102,31 @@
 #define RP_SSI_XIP_SPI_CTRL0_ADDRESS_LENGTH(x) (((x) * 2U) << 2U)
 #define RP_SSI_XIP_SPI_CTRL0_INSTR_LENGTH_8b   (2U << 8U)
 #define RP_SSI_XIP_SPI_CTRL0_WAIT_CYCLES(x)    (((x) * 8U) << 11U)
+#define RP_SSI_XIP_SPI_CTRL0_XIP_CMD_SHIFT     24U
+#define RP_SSI_XIP_SPI_CTRL0_XIP_CMD(x)        ((x) << RP_SSI_XIP_SPI_CTRL0_XIP_CMD_SHIFT)
+#define RP_SSI_XIP_SPI_CTRL0_TRANS_1C1A        (0U << 0U)
+#define RP_SSI_XIP_SPI_CTRL0_TRANS_1C2A        (1U << 0U)
+#define RP_SSI_XIP_SPI_CTRL0_TRANS_2C2A        (2U << 0U)
+
+#define RP_PADS_QSPI_BASE_ADDR                 0x40020000U
```
dragonmux, line 111: These and the two additional blocks below are over-aligned same as the first block - please let clang-format set the alignment on these.

(for what it's worth, we agree on what you've done here, we just haven't figured out how to make clang-format do the same as it'd clean up a pile of stuff and give greater visual consistency, so for now let's do the thing that keeps this code clang-format clean)

src/target/rp.c
```diff
@@ -155,12 +198,18 @@ static bool rp_flash_write(target_flash_s *f, target_addr_t dest, const void *sr
...
-static void rp_flash_prepare(target *t);
-static void rp_flash_resume(target *t);
+static bool rp_flash_prepare(target *t);
+static bool rp_flash_resume(target *t);
 static void rp_spi_read(target *t, uint16_t command, target_addr_t address, void *buffer, size_t length);
 static uint32_t rp_get_flash_length(target *t);
 static bool rp_mass_erase(target *t);
 
+// Our own implementation of bootloader functions for handling flash chip
+static void __attribute__((unused))rp_flash_connect_internal(target *t);
```
dragonmux, line 208: There's a missing space after the attribute here, and two lines down. Additionally, as they're both actually unused, it might make sense to either comment them out for now (if you think they have value "just in case"), or remove them entirely. Just seems off to be adding intentionally unused code here.

src/target/rp.c
```diff
@@ -183,17 +233,18 @@ static void rp_add_flash(target *t)
 		spi_parameters.capacity = rp_get_flash_length(t);
 		spi_parameters.sector_erase_opcode = SPI_FLASH_CMD_SECTOR_ERASE;
 	}
-	rp_flash_resume(t);
 
-	DEBUG_INFO("Flash size: %zu MB\n", spi_parameters.capacity / (1024U * 1024U));
+	rp_flash_enter_xip(t);
+
+	DEBUG_INFO("Flash size: %" PRIu16 " MB\n", spi_parameters.capacity / (1024U * 1024U));
```
dragonmux, line 239: PRIu16 doesn't work on ARM newlib, please use `%u` here. Also, we own this one but `MB` => `MiB` as this is using IEC not SI units, so we can't use the SI unit bases as that's wrong. Might as well also remove the extra space between the format string and units so it doesn't read so weirdly.

UweBonnes:
Not using the ROM routine seems as one good approach. If you have the knowledge of the RP internal, that is fine. Another poor mans approach to debugger warm plug could be to skip probe for ROM routine approach on attach. I do not think it is important to have the exact size of Flash send to gdb. The largest possible value is fine. It is the task of the linker script to care for the right flash size. Then only when reset for flashing is needed, the ROM routines are called.

src/target/rp.c
```diff
@@ -554,56 +611,216 @@ static void rp_spi_read(
...
+	);
+	// Slave selected when transfers in progress
+	target_mem_write32(t, RP_SSI_SER, 1);
+	// Re-enable
+	target_mem_write32(t, RP_SSI_ENABLE, 1);
+}
+
+// Also allow any unbounded loops to check whether the above abort condition
+// was asserted, and terminate early
+static int rp_flash_was_aborted(target *t) {
```
dragonmux, line 667: The { here belongs on a new line per the codebase style

src/target/rp.c
```diff
@@ -554,56 +611,216 @@ static void rp_spi_read(
...
-	size = MAX_FLASH;
+// Put bytes from one buffer, and get bytes into another buffer.
+// These can be the same buffer.
+// If tx is NULL then send zeroes.
+// If rx is NULL then all read data will be dropped.
+//
+// If rx_skip is nonzero, this many bytes will first be consumed from the FIFO,
+// before reading a further count bytes into *rx.
+// E.g. if you have written a command+address just before calling this function.
+static void rp_flash_put_get(target *t, const uint8_t *tx, uint8_t *rx, size_t count, size_t rx_skip) {
```
dragonmux, line 679: The { here belongs on a new line per the codebase style. Also, count appears to be constant, so please mark it such - same for any other parameters that aren't modified (t for example -> `target *const t`).

src/target/rp.c
```diff
@@ -554,56 +611,216 @@ static void rp_spi_read(
...
+    // Make sure there is never more data in flight than the depth of the RX
+    // FIFO. Otherwise, when we are interrupted for long periods, hardware
+    // will overflow the RX FIFO.
+    const uint max_in_flight = 16 - 2; // account for data internal to SSI
+    size_t tx_count = count;
+    size_t rx_count = count;
+    while (tx_count || rx_skip || rx_count) {
```
dragonmux, line 686: As much as this is a reasonably complex condition, is it possible to turn this into a for loop, allowing tx and rx to be marked as const pointers? (`const uint8_t *const tx, uint8_t *const rx`)

It's our experience that this kind of while loop that's really a for loop both hurts codegen and so also Flash usage while being harder to reason about.

src/target/rp.c
```diff
@@ -554,56 +611,216 @@ static void rp_spi_read(
...
+	uint32_t padctrl_save = target_mem_read32(t, RP_PADS_QSPI_GPIO_SD0);
+    uint32_t padctrl_tmp = (padctrl_save
+                            & ~(RP_PADS_QSPI_GPIO_SD0_OD_BITS | RP_PADS_QSPI_GPIO_SD0_PUE_BITS |
+                                RP_PADS_QSPI_GPIO_SD0_PDE_BITS)
+                           ) | RP_PADS_QSPI_GPIO_SD0_OD_BITS | RP_PADS_QSPI_GPIO_SD0_PDE_BITS;
+
+    // First two 32-clock sequences
+    // CSn is held high for the first 32 clocks, then asserted low for next 32
+    rp_spi_chip_select(t, RP_GPIO_QSPI_CS_DRIVE_HIGH);
+    for (int i = 0; i < 2; ++i) {
```
dragonmux, line 739: Please use an unsigned type (size_t would be suitable) for this loop

src/target/rp.c
```diff
@@ -554,56 +611,216 @@ static void rp_spi_read(
...
-	for (i = 0; i < 16; i++) {
-		if ((bootsec[i] != 0x00) && (bootsec[i] != 0xff))
-			break;
-	}
+    // Use hard reset to force IO and pad controls to known state (don't touch
+    // IO_BANK0 as that does not affect XIP signals)
+	uint32_t reset = target_mem_read32(t, RP_RESETS_RESET);
+    target_mem_write32(t, RP_RESETS_RESET, reset | RP_RESETS_RESET_IO_QSPI_BITS | RP_RESETS_RESET_PADS_QSPI_BITS);
+    target_mem_write32(t, RP_RESETS_RESET, reset);
+	while (~target_mem_read32(t, RP_RESETS_RESET_DONE) & (RP_RESETS_RESET_IO_QSPI_BITS | RP_RESETS_RESET_PADS_QSPI_BITS));
```
dragonmux, line 630: ~ is the wrong operation here - should be ! or an explicit comparison. Also, please use `continue` on a new line as the loop body so the loop body is obvious and explicit. A ';' on the end is just too easy to miss.

src/target/rp.c
```diff
@@ -554,56 +611,216 @@ static void rp_spi_read(
...
+        bool did_something = false; // Expect this to be folded into control flow, not register
+        if (tx_count && tx_level + rx_level < max_in_flight) {
+            target_mem_write32(t, RP_SSI_DR0, (uint32_t) (tx ? *tx++ : 0));
+            --tx_count;
+            did_something = true;
+        }
+        if (rx_level) {
+            uint8_t rxbyte = target_mem_read32(t, RP_SSI_DR0);
+            did_something = true;
+            if (rx_skip) {
```
dragonmux, line 699: The {}'s around this if body should be removed.

dragonmux (changes requested):
This is looking significantly better - thank you for addressing our concern/issue with the original approach. There are a few stylistic things we spotted on review, however this is very close to mergable.

We will run a test on our own RP2040 just to convince ourself that it works and then once the stylistic points are addressed, we will merge this.

dragonmux:
> Not using the ROM routine seems as one good approach. If you have the knowledge of the RP internal, that is fine. Another poor mans approach to debugger warm plug could be to skip probe for ROM routine approach on attach. I do not think it is important to have the exact size of Flash send to gdb. The largest possible value is fine. It is the task of the linker script to care for the right flash size. Then only when reset for flashing is needed, the ROM routines are called.

Actually, it turns out to be extremely important for several reasons to do with how the Flash code works and to get GDB to behave properly with BMP. It wasn't for nothing that we went to the effort to pull the SFDP data and use it.

UweBonnes:
Please explain the reasons. 

We already fake the Flash size for devices like STM32F103x8 and STM32F103xB. Those devices are the same chip with a different flash size indicated by ST. It should be up to the user what to do with this flash.

src/target/rp.c
jamesturton: Fixed!

src/target/rp.c
jamesturton: Fixed!

src/target/rp.c
jamesturton: I agree, I left it in the code because I was testing to see how calling this function effected the target, but it seems that it is not needed to be called from our code. The linker will cleanup these unused functions so having them preserved in the code for future reference will not effect the size of the binary, but I can hide them so that the compiler never sees them if you prefer.

src/target/rp.c
jamesturton: I agree with `MiB` :smile: 

As for `PRIu16`, I'm testing compiling for `native` and `PRIu16` is working exactly as expected when compiling with `arm-none-eabi-gcc-10.2.1`. `arm-none-eabi-gcc-10.2.1` is compiled with `newlib-3.3.0`:
https://sourceware.org/git/?p=newlib-cygwin.git;a=blob;f=newlib/libc/include/inttypes.h;h=417ff224cb1f347ab6438a4ba75421bbe1ca85d7;hb=26894de91d296af502b53c243e71f88c06ea1a3d#l119
So I'm not too sure what you mean by:
> PRIu16 doesn't work on ARM newlib

src/target/rp.c
jamesturton: It is the correct operation... We are bitwise inverting the value in register `RP_RESETS_RESET_DONE` and then bitwise and to the `IO_QSPI_BITS` and `PADS_QSPI_BITS`. I will re-write this to help improve clarity.

src/target/rp.c
jamesturton: Fixed!

src/target/rp.c
jamesturton: Fixed!

src/target/rp.c
jamesturton: Fixed!

src/target/rp.c
jamesturton: Fixed!

src/target/rp.c
jamesturton: Fixed!

src/target/rp.c
jamesturton: Ok, fixed!

src/target/rp.c
jamesturton: It might be possible, but I feel that there is a large amount of clarity gained by using a while loop, over using a for loop.
The number of times the loop should be run is not possible to be determined before hand as it depends on the levels read from `RP_SSI_TXFLR` and `RP_SSI_RXFLR`.
But I would be happy to accept suggestions for changes!

src/target/rp.c
```diff
@@ -554,56 +613,220 @@ static void rp_spi_read(
...
+// 2. CSn = 0, IO = 4'hf (via pullup to avoid contention), x32 clocks
+// 3. CSn = 1 (brief deassertion)
+// 4. CSn = 0, MOSI = 1'b1 driven, x16 clocks
+//
+// Part 4 is the sequence suggested in W25X10CL datasheet.
+// Parts 1 and 2 are to improve compatibility with Micron parts
+static void rp_flash_exit_xip(target *const t)
+{
+	uint8_t buf[2];
+	buf[0] = 0xff;
```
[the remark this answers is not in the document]
jamesturton: Fixed!

src/target/rp.c
dragonmux: So.. this does make the result 8 bytes larger in Flash (not entirely sure why).. but.. this is what we meant about finding this complex and hard to follow so could it be turned into a for loop:

```c
static void rp_flash_put_get(target *const t, const uint8_t *const tx, uint8_t *const rx,
	const size_t count, const size_t rx_skip)
{
	// Make sure there is never more data in flight than the depth of the RX
	// FIFO. Otherwise, when we are interrupted for long periods, hardware
	// will overflow the RX FIFO.
	static const size_t max_in_flight = 16 - 2; // account for data internal to SSI
	size_t rx_offset = 0;
	for (size_t tx_count = 0, rx_count = 0; rx_count < count + rx_skip || tx_count < count; ) {
		// NB order of reads, for pessimism rather than optimism
		const uint32_t tx_level = target_mem_read32(t, RP_SSI_TXFLR);
		const uint32_t rx_level = target_mem_read32(t, RP_SSI_RXFLR);
		bool idle = true; // Expect this to be folded into control flow, not register

		if (tx_count < count && tx_level + rx_level < max_in_flight) {
			target_mem_write32(t, RP_SSI_DR0, tx ? tx[tx_count] : 0U);
			++tx_count;
			idle = false;
		}

		if (rx_level) {
			const uint8_t data = target_mem_read32(t, RP_SSI_DR0) & 0xFFU;
			if (rx_count >= rx_skip && rx)
				rx[rx_offset++] = data;
			++rx_count;
			idle = false;
		}

		if (idle && rp_flash_was_aborted(t))
			break;
	}
	rp_spi_chip_select(t, RP_GPIO_QSPI_CS_DRIVE_HIGH);
}
```

It's still a bit messy to follow, but we find both because it's doing less shenanigans with count-down parameters, and because of the reduced indentation levels, the trade off is worth it. We'll let you have final say but this would be our suggestion.

jamesturton: I think for clarity it should be left as a while loop as the step size for the loop is not constant and therefore does not align with the for loop idiom. But I have updated the function to make the `tx` and `rx` const pointers, and performed a little bit more clean-up, so I'm happy now!

dragonmux:
Uwe, both of the devices you quoted use on-chip Flash with a dedicated Flash controller. Both of them have dedicated hardware logic for going past the end of Flash and handling that. Both of them will refuse to program past the end. That's the problem with using those as examples. 

This part is Flashless, it uses SPI Flash and that external Flash device is the one that determines behaviour more than the RP2040 - and SPI Flash does not have a consistent past-the-end behaviour, some chips will fail, some will wrap (screwing up your start of Flash programming), and some will hard lock. It depends on what SPI Flash part is attached. It's not safe to assume that it will safely, detectably fail. GDB can also behave in some pretty strange ways with an over-big mapping on these parts when reading too, so it's ill-advised on that front.

In other words, Flashless parts require an accurate map to be returned to GDB because BMP, not the target, is the Flash controller.

dragonmux (approved):
With that revised buffer handling logic making the code much easier to reason about and follow, we're now happy to merge this. Thank you again for the contribution James!

PR #1231, opened 2022-09-02
L0 prog

UweBonnes:
## Detailed description
Fix bug introduced in 085dcb7bc942a9dc32245b13708425040fe28a2b
Additional return immediate when erasing a chunk fails and report failure location.

* [ ] It builds as BMDA (`make PROBE_HOST=hosted`)


esden:
Hi @UweBonnes, thank you for pointing out this issue. Good catch.

But first things first. If you want us to merge any of your PR going forward, you will have to read the checklist items in the PR template, make sure you have done them and set the check marks accordingly. These are the most basic things every contributor has to remember to do. Those are an effort to decrease maintenance effort and probability of regressions.

As for the bugfix itself. The way it is done here is not exactly right. I talked to @dragonmux and they will comment on the way the fix needs to be implemented.

Cheers,
Piotr

src/target/stm32l0.c
```diff
@@ -326,10 +326,9 @@ static bool stm32lx_nvm_busy_wait(target *t, uint32_t nvm)
 	uint32_t sr;
 	do {
 		sr = target_mem_read32(t, STM32Lx_NVM_SR(nvm));
-		if ((sr & STM32Lx_NVM_SR_ERR_M) || !(sr & STM32Lx_NVM_SR_EOP) || target_check_error(t))
-			return false;
 	} while (sr & STM32Lx_NVM_SR_BSY);
-
+	if ((sr & STM32Lx_NVM_SR_ERR_M) || !(sr & STM32Lx_NVM_SR_EOP) || target_check_error(t))
+		return false;
```
dragonmux, line 331: There are a couple of problems with this change that we see - the first is that `target_check_error(t)` must be called every loop otherwise the loop might never terminate, and the second is to do with the if-return logic (clang-tidy warns not do this!)

We would suggest this should either be:
```c
uint32_t sr;
do {
    sr = target_mem_read32(t, STM32Lx_NVM_SR(nvm));
    if ((sr & STM32Lx_NVM_SR_ERR_M) || target_check_error(t))
        return false;
} while ((sr & STM32Lx_NVM_SR_BSY) && !(sr & STM32Lx_NVM_SR_EOP));
```
with the original post-while return, or.. to keep with the theme here:
```c
while (target_mem_read32(t, STM32Lx_NVM_SR(nvm)) & STM32Lx_NVM_SR_BSY) {
    if (target_check_error(t))
        return false;
}
const uint32_t sr = target_mem_read32(t, STM32Lx_NVM_SR(nvm));
return !target_check_error(t) && !(sr & STM32Lx_NVM_SR_ERR_M) && (sr & STM32Lx_NVM_SR_EOP);
```

The former method keeps the original flow and intent of this code in tact, the latter might actually be slightly more correct and definitely solves your issue still.

src/target/target_flash.c
```diff
@@ -143,7 +143,10 @@ bool target_flash_erase(target *t, target_addr_t addr, size_t len)
 		if (!flash_prepare(f))
 			return false;
 
-		ret &= f->erase(f, local_start_addr, f->blocksize);
+		if (!f->erase(f, local_start_addr, f->blocksize)) {
+			DEBUG_WARN("Erase failed at %" PRIx32 "\n", local_start_addr);
+			return false;
```
dragonmux, line 148: There is major problems here - calling `flash_done()` prior to exiting from this routine is not optional. It must be done to create the correct semantics for a variety of devices and perform buffer allocation cleanup that is required to avoid heap exhaustion.

The other reason we need `flash_done()` called is that in several cases, the correct functioning of the device hinges on it as entering Flash mode inhibits operations like reading the Flash mapping region. As mentioned, not an optional call.

There are two solutions we see here - the first is to insert a call to `flash_done()` just prior to this return (ignoring the return value) - which is not.. great.. the second is to have the old line, then an if check for ret - if ret is false, issue the debug warn and then break, moving the flash_done at the bottom of the loop one line down outside the loop. We favour the latter option.

dragonmux (changes requested):
As Esden mentioned, while we appreciate that STM32L0 programming is broken at current, this patch only solves your immediate problem while causing a pile more. Please see below for our review comments about what we see wrong and suggestions for ways to do it differently that should still solve the problem you're facing without breaking other things.

src/target/stm32l0.c
UweBonnes: I think that other SR bits are only valid after BSY is released. Look at my proposal in the updated PR.

UweBonnes: Up to 91d1ef8bf6abc5f5ffff1 it was:
        do {
                sr = target_mem_read32(t, STM32Lx_NVM_SR(nvm));
        } while (sr & STM32Lx_NVM_SR_BSY);

        if ((sr & STM32Lx_NVM_SR_ERR_M) || !(sr & STM32Lx_NVM_SR_EOP) ||
            target_check_error(t))
                        return -1;

UweBonnes:
I removed the early check for failing flash erase as #1236 seem to target the same problem. However L0 flashing is still broken and so this PR show move on.

src/target/stm32l0.c
dragonmux: This looks better now, but it's still doing a pessimistic do-while loop which is what we're trying to suggest you avoid in the original review above. Otherwise it looks ok. That leaves one last review point before we can merge:

This is still doing `if (cond) return false; return true;` which clang-tidy will tell you not to do (so don't!) - the condition of the if statement should be turned into the return value directly with a single return - `return !(cond);`, so in this case `return !(sr & STM32Lx_NVM_SR_ERR_M) && (sr & STM32Lx_NVM_SR_EOP);`

UweBonnes:
'make  clang-tidy' show zillons of warnings, but not in stm32lx_nvm_busy_wait(). Having the old setup i.m.h.o. helps debugging . When single stepping, you step on the return false and you still have access to sr. With the clang-tidy  approach, you only realize thing got wrong outside that function and sr is gone. Isn't gcc clever  enough to optimize good enough with -O2 and allow the single step debugging when compiled with -Og?

dragonmux:
We know for a fact that it should be warning on that (it appears it isn't because it's failing to recognise it, but the diagnostic is [readability-simplify-boolean-expr](https://clang.llvm.org/extra/clang-tidy/checks/readability/simplify-boolean-expr.html#readability-simplify-boolean-expr)) also, when you get to the return statement you have access to 'sr', it doesn't improve or hinder debugging, but it does pessimise codegen, it is against the policies in place on new code, and it's extra verbiage to get wrong.

UweBonnes:
If I am not wrong, capability to set a breakpoint on the failing case while still access to SR is possible is lost with that code style. And that does not hinder debugging? But if politics require...

dragonmux:
You can still set a breakpoint on the return statement, further you can use GDB to print what the function would return, or you can use a conditional breakpoint to only break if the function would return false. There is no information loss there.

You also still always have the option of inspecting the value of the register on the target which does not change after the return as it's not read-to-clear.

dragonmux (approved):

dragonmux:
Close enough.. it introduces a needless extra pair of parentheses, and doesn't do the DeMorgan's simplification on the return suggested in the review, but we'll merge this as is as we want to get the fix in.

PR #1241, opened 2022-09-15
Feature: HC32L110 support

vesim987:
## Detailed description
In this PR I've implemented support for detecting and flashing HC32L110B(A variant should work too, but i don't have it to test). And also introduce flash_operation in `flash.prepare` and `flash.done`


dragonmux:
Please can you rebase this against main, or let us know if you're happy for us to do so for you, so we can review and merge this?

dragonmux:
Given the time between asking for this to be rebased and now, we're going to get on and do so, addressing Perigoso's review points. Getting this sorted out and merged for v1.10 would be really good.

src/target/target_flash.c
perigoso, line 1: The current_flash_operation state/variable is implemented on the wrong scope, it's on the target scope when it should belong to the flash structs. It's also incorrectly implemented at the moment as it is working under the assumption that the target is reset into "flash mode" every time the type of operation on the flash changes, which is incorrect. I addressed this in my last review but it was never addressed.

src/target/hc32l110.c
```diff
@@ -0,0 +1,236 @@
...
+static bool hc32l110_enter_flash_mode(target_s *target);
+static bool hc32l110_flash_prepare(target_flash_s *flash, flash_operation_e op);
+static bool hc32l110_flash_done(target_flash_s *flash, flash_operation_e op);
+static bool hc32l110_flash_erase(target_flash_s *flash, target_addr_t addr, size_t length);
+static bool hc32l110_flash_write(target_flash_s *flash, target_addr_t dest, const void *src, size_t length);
+static bool hc32l110_mass_erase(target_s *target);
+
+static void hc32l110_add_flash(target_s *target, const uint32_t flash_size)
+{
+	target_flash_s *flash = calloc(1, sizeof(*flash));
```
perigoso, line 85: This is missing the null/heap exhaust check

src/target/hc32l110.c
```diff
@@ -0,0 +1,236 @@
...
+
+static bool hc32l110_flash_erase(target_flash_s *const flash, const target_addr_t addr, const size_t length)
+{
+	(void)length;
+	/* The Flash controller automatically erases the whole sector after one write operation */
+	target_mem_write32(flash->t, addr, 0);
+
+	const bool result = hc32l110_check_flash_completion(flash->t, 1000);
+	if (!result)
+		DEBUG_ERROR("Failed to erase %" PRIx32 " bytes at 0x%08" PRIx32 "\n", (uint32_t)length, addr);
```
perigoso, line 203: Isn't this debug message redundant? `target_flash_erase` does a `DEBUG_ERROR("Erase failed at %" PRIx32 "\n", local_start_addr);` on erase failure

src/target/hc32l110.c
```diff
@@ -0,0 +1,236 @@
...
+
+static bool hc32l110_flash_write(
+	target_flash_s *const flash, const target_addr_t dest, const void *const src, const size_t length)
+{
+	(void)length;
+	target_mem_write32(flash->t, dest, *(const uint32_t *)src);
+
+	const bool result = hc32l110_check_flash_completion(flash->t, 1000);
+	if (!result)
+		DEBUG_ERROR("Failed to write %" PRIu32 " bytes at 0x%08" PRIx32 "\n", (uint32_t)length, dest);
```
perigoso, line 215: likewise for the read, Isn't this debug message redundant? `target_flash_write` does a `DEBUG_ERROR("Write failed at %" PRIx32 "\n", dest);` on write failure, which unlike the erase case will be less granular, but I still question the usefulness of this debug message

perigoso (changes requested):
One of the issues I noted in a previous review was not yet addressed, that is, the new flash operation status is implemented on the wrong scope, it's kept on the target scope when it should belong to the flash structs.

Instead of thoroughly explaining what needed to change in review I found it easier to just re-write the affected code, you can find the changes in [perigoso/fix/hc32l110](https://github.com/perigoso/blackmagic/tree/fix/hc32l110), 8c353a475bd1a17c560fecac98564e0e0e54ce7a and 08c57862a879b7498ca35c6c7141873b250a232d replace a9ae7c458ed8211b8947608879123b39772ab3d6 and there's an additional fix-up because of function signature changes 8d213d9c6d793d4b290461da79ea6c9177e2b565

src/target/target_flash.c
dragonmux: We've been thinking about how to pull that out into its own PR anyway because it really should be its own entirely separate change-set anyway and we aren't convinced of the utility of it

src/target/hc32l110.c
dragonmux: You're right, we weren't thinking about the target layer, only what was in front of us from the original contributor.

src/target/hc32l110.c
dragonmux: You're right, we weren't thinking about the target layer, only what was in front of us from the original contributor.

src/target/target_flash.c
dragonmux: Now that #1538 is merged, this has been addressed, with this PR now only implementing support for the new target.

esden (approved):
LGTM

PR #1242, opened 2022-09-16
stm32f4: fix wrong idcode when probing

andrew1x000:
STM32F4 has bug in stm32f4_probe() function

## Detailed description

When program reach stm32f4_probe() function, target structure had wrongly fulfilled `t->idcode` field (with value `1041`). Which causes wrong chip selection ahead in the program. This issue doesn't present in [BMP ver 1.7.1](https://github.com/blackmagic-debug/blackmagic/tree/v1.7.1).
My fix works for me with the STM32F446 board. But maybe similar bugs are present in other targets as well. 

### DIFF
--- a/src/target/stm32f4.c
+++ b/src/target/stm32f4.c
@@ -207,16 +207,17 @@ static void stm32f4_detach(target *t)
 
 bool stm32f4_probe(target *t)
 {
-	const uint16_t mcu_idcode = target_mem_read32(t, DBGMCU_IDCODE) & 0xfffU;
-	switch (mcu_idcode) {
-	/* F405 revision A has the wrong IDCODE, use ARM_CPUID to make the
-	 * distinction vs the F205. Revision is also wrong (0x2000 instead
-	 * of 0x1000). See F40x/F41x errata. */
-	case ID_STM32F20X:
-		if ((t->cpuid & 0xfff0U) != CORTEX_M4)
-			break;
-		mcu_idcode = ID_STM32F40X;
+	uint16_t mcu_idcode = target_mem_read32(t, DBGMCU_IDCODE) & 0xfffU;
 
+	if (mcu_idcode == ID_STM32F20X) {
+		/* F405 revision A have a wrong IDCODE, use ARM_CPUID to make the
+		 * distinction with F205. Revision is also wrong (0x2000 instead
+		 * of 0x1000). See F40x/F41x errata. */
+		if ((t->cpuid & 0xFFF0) == CORTEX_M4)
+			mcu_idcode = ID_STM32F40X;
+	}
+
+	switch (mcu_idcode) {
 	case ID_STM32F74X: /* F74x RM0385 Rev.4 */
 	case ID_STM32F76X: /* F76x F77x RM0410 */
 	case ID_STM32F72X: /* F72x F73x RM0431 */
--- a/src/target/stm32f4.c
+++ b/src/target/stm32f4.c
@@ -207,37 +207,35 @@ static void stm32f4_detach(target *t)
 
 bool stm32f4_probe(target *t)
 {
-	uint16_t stored_idcode = t->idcode;
 	t->idcode = target_mem_read32(t, DBGMCU_IDCODE) & 0xFFF;
 
 	if (t->idcode == ID_STM32F20X) {
 		/* F405 revision A have a wrong IDCODE, use ARM_CPUID to make the
-		* distinction with F205. Revision is also wrong (0x2000 instead
-		* of 0x1000). See F40x/F41x errata. */
+		 * distinction with F205. Revision is also wrong (0x2000 instead
+		 * of 0x1000). See F40x/F41x errata. */
 		if ((t->cpuid & 0xFFF0) == CORTEX_M4)
-		t->idcode = ID_STM32F40X;
+			t->idcode = ID_STM32F40X;
 	}
 	switch (t->idcode) {
-		case ID_STM32F74X:  /* F74x RM0385 Rev.4 */
-		case ID_STM32F76X:  /* F76x F77x RM0410 */
-		case ID_STM32F72X:  /* F72x F73x RM0431 */
-		case ID_STM32F40X:
-		case ID_STM32F42X:  /* 427/437 */
-		case ID_STM32F46X:  /* 469/479 */
-		case ID_STM32F20X:  /* F205 */
-		case ID_STM32F446:  /* F446 */
-		case ID_STM32F401C: /* F401 B/C RM0368 Rev.3 */
-		case ID_STM32F411:  /* F411     RM0383 Rev.4 */
-		case ID_STM32F412:  /* F412     RM0402 Rev.4, 256 kB Ram */
-		case ID_STM32F401E: /* F401 D/E RM0368 Rev.3 */
-		case ID_STM32F413:  /* F413     RM0430 Rev.2, 320 kB Ram, 1.5 MB flash. */
-			t->detach = stm32f4_detach;
-			t->driver = stm32f4_get_chip_name(t->idcode);
-			t->attach = stm32f4_attach;
+	case ID_STM32F74X: /* F74x RM0385 Rev.4 */
+	case ID_STM32F76X: /* F76x F77x RM0410 */
+	case ID_STM32F72X: /* F72x F73x RM0431 */
+	case ID_STM32F40X:
+	case ID_STM32F42X:  /* 427/437 */
+	case ID_STM32F46X:  /* 469/479 */
+	case ID_STM32F20X:  /* F205 */
+	case ID_STM32F446:  /* F446 */
+	case ID_STM32F401C: /* F401 B/C RM0368 Rev.3 */
+	case ID_STM32F411:  /* F411     RM0383 Rev.4 */
+	case ID_STM32F412:  /* F412     RM0402 Rev.4, 256 kB Ram */
+	case ID_STM32F401E: /* F401 D/E RM0368 Rev.3 */
+	case ID_STM32F413:  /* F413     RM0430 Rev.2, 320 kB Ram, 1.5 MB flash. */
+		t->detach = stm32f4_detach;
+		t->driver = stm32f4_get_chip_name(t->idcode);
+		t->attach = stm32f4_attach;
 		target_add_commands(t, stm32f4_cmd_list, t->driver);
 		return true;
-		default:
-			t->idcode = stored_idcode;
+	default:
 		return false;
 	}
 }
--- a/src/target/stm32f4.c
+++ b/src/target/stm32f4.c
@@ -233,6 +233,7 @@ bool stm32f4_probe(target *t)
 	case ID_STM32F413:  /* F413     RM0430 Rev.2, 320 kB Ram, 1.5 MB flash. */
 		t->attach = stm32f4_attach;
 		t->detach = stm32f4_detach;
+		t->mass_erase = stm32f4_mass_erase;
 		t->driver = stm32f4_get_chip_name(t->part_id);
 		t->part_id = mcu_idcode;
 		target_add_commands(t, stm32f4_cmd_list, t->driver);
--- a/src/target/stm32f4.c
+++ b/src/target/stm32f4.c
@@ -207,34 +207,37 @@ static void stm32f4_detach(target *t)
 
 bool stm32f4_probe(target *t)
 {
-	if (t->part_id == ID_STM32F20X) {
+	uint16_t stored_idcode = t->idcode;
+	t->idcode = target_mem_read32(t, DBGMCU_IDCODE) & 0xFFF;
+
+	if (t->idcode == ID_STM32F20X) {
 		/* F405 revision A have a wrong IDCODE, use ARM_CPUID to make the
-		 * distinction with F205. Revision is also wrong (0x2000 instead
-		 * of 0x1000). See F40x/F41x errata. */
+		* distinction with F205. Revision is also wrong (0x2000 instead
+		* of 0x1000). See F40x/F41x errata. */
 		if ((t->cpuid & 0xFFF0) == CORTEX_M4)
-			t->part_id = ID_STM32F40X;
+		t->idcode = ID_STM32F40X;
 	}
-	switch(t->part_id) {
-	case ID_STM32F74X: /* F74x RM0385 Rev.4 */
-	case ID_STM32F76X: /* F76x F77x RM0410 */
-	case ID_STM32F72X: /* F72x F73x RM0431 */
-	case ID_STM32F40X:
-	case ID_STM32F42X: /* 427/437 */
-	case ID_STM32F46X: /* 469/479 */
-	case ID_STM32F20X: /* F205 */
-	case ID_STM32F446: /* F446 */
-	case ID_STM32F401C: /* F401 B/C RM0368 Rev.3 */
-	case ID_STM32F411: /* F411     RM0383 Rev.4 */
-	case ID_STM32F412: /* F412     RM0402 Rev.4, 256 kB Ram */
-	case ID_STM32F401E: /* F401 D/E RM0368 Rev.3 */
-	case ID_STM32F413: /* F413     RM0430 Rev.2, 320 kB Ram, 1.5 MB flash. */
-		t->mass_erase = stm32f4_mass_erase;
-		t->detach = stm32f4_detach;
-		t->driver = stm32f4_get_chip_name(t->part_id);
-		t->attach = stm32f4_attach;
+	switch (t->idcode) {
+		case ID_STM32F74X:  /* F74x RM0385 Rev.4 */
+		case ID_STM32F76X:  /* F76x F77x RM0410 */
+		case ID_STM32F72X:  /* F72x F73x RM0431 */
+		case ID_STM32F40X:
+		case ID_STM32F42X:  /* 427/437 */
+		case ID_STM32F46X:  /* 469/479 */
+		case ID_STM32F20X:  /* F205 */
+		case ID_STM32F446:  /* F446 */
+		case ID_STM32F401C: /* F401 B/C RM0368 Rev.3 */
+		case ID_STM32F411:  /* F411     RM0383 Rev.4 */
+		case ID_STM32F412:  /* F412     RM0402 Rev.4, 256 kB Ram */
+		case ID_STM32F401E: /* F401 D/E RM0368 Rev.3 */
+		case ID_STM32F413:  /* F413     RM0430 Rev.2, 320 kB Ram, 1.5 MB flash. */
+			t->detach = stm32f4_detach;
+			t->driver = stm32f4_get_chip_name(t->idcode);
+			t->attach = stm32f4_attach;
 		target_add_commands(t, stm32f4_cmd_list, t->driver);
 		return true;
-	default:
+		default:
+			t->idcode = stored_idcode;
 		return false;
 	}
 }
--- a/src/target/stm32f4.c
+++ b/src/target/stm32f4.c
@@ -207,37 +207,37 @@ static void stm32f4_detach(target *t)
 
 bool stm32f4_probe(target *t)
 {
-	t->idcode = target_mem_read32(t, DBGMCU_IDCODE) & 0xFFF;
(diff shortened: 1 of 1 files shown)


src/target/stm32f4.c
```diff
@@ -207,34 +207,37 @@ static void stm32f4_detach(target *t)
...
 {
-	if (t->part_id == ID_STM32F20X) {
+	uint16_t stored_idcode = t->idcode;
+	t->idcode = target_mem_read32(t, DBGMCU_IDCODE) & 0xFFF;
+
+	if (t->idcode == ID_STM32F20X) {
 		/* F405 revision A have a wrong IDCODE, use ARM_CPUID to make the
-		 * distinction with F205. Revision is also wrong (0x2000 instead
-		 * of 0x1000). See F40x/F41x errata. */
+		* distinction with F205. Revision is also wrong (0x2000 instead
```
dragonmux, line 215: This reformatting is incorrect - those spaces weren't wrong and clang-format would not have removed them, please don't do that.

src/target/stm32f4.c
```diff
@@ -207,34 +207,37 @@ static void stm32f4_detach(target *t)
...
+		* of 0x1000). See F40x/F41x errata. */
 		if ((t->cpuid & 0xFFF0) == CORTEX_M4)
-			t->part_id = ID_STM32F40X;
+		t->idcode = ID_STM32F40X;
```
dragonmux, line 218: This is now incorrectly indented, the tab that was at the start of the line was correct, please re-add it.

src/target/stm32f4.c
```diff
@@ -207,34 +207,37 @@ static void stm32f4_detach(target *t)
...
-	case ID_STM32F401C: /* F401 B/C RM0368 Rev.3 */
-	case ID_STM32F411: /* F411     RM0383 Rev.4 */
-	case ID_STM32F412: /* F412     RM0402 Rev.4, 256 kB Ram */
-	case ID_STM32F401E: /* F401 D/E RM0368 Rev.3 */
-	case ID_STM32F413: /* F413     RM0430 Rev.2, 320 kB Ram, 1.5 MB flash. */
-		t->mass_erase = stm32f4_mass_erase;
-		t->detach = stm32f4_detach;
-		t->driver = stm32f4_get_chip_name(t->part_id);
-		t->attach = stm32f4_attach;
+	switch (t->idcode) {
```
dragonmux, line 220: We just checked with clang-format and this switch-case was already correctly formatted save for some spacing on the case comments and a missing space between `switch` and the opening `(`, please don't just rejig the formatting - use the .clang-format provided.

src/target/stm32f4.c
```diff
@@ -207,34 +207,37 @@ static void stm32f4_detach(target *t)
...
+		case ID_STM32F401E: /* F401 D/E RM0368 Rev.3 */
+		case ID_STM32F413:  /* F413     RM0430 Rev.2, 320 kB Ram, 1.5 MB flash. */
+			t->detach = stm32f4_detach;
+			t->driver = stm32f4_get_chip_name(t->idcode);
+			t->attach = stm32f4_attach;
 		target_add_commands(t, stm32f4_cmd_list, t->driver);
 		return true;
-	default:
+		default:
+			t->idcode = stored_idcode;
```
dragonmux, line 240: This is incorrectly indented.

dragonmux (changes requested):
Thank you for the contribution! So, a couple of things..

We appreciate you might not have been aware, but the change in behaviour is expected and not incorrect, as part of some ongoing work by @perigoso the ADIv5 code (and subsequently the IDs fed into the probe code) was changed to be more correct against the ADIv5 specification. It's expected that the IDs (but not the method of getting them) are what are wrong, so please target ID_STM32F446 for patch rather than the probe routine - the routine is correct, the ID code is what's wrong.

The second is that the patch changes the formatting of the probe routine in ways which clang-format would not allow. Please use the provided style and formatting configuration.

We are however grateful to know about another set of IDs which are now wrong and need correction, so would be very happy to accept a modified version of this contribution that targets the STM32F4 ID codes specifically.

UweBonnes:
I just tested with a F4 discovery board.
Black Magic Debug App v1.8.0-645-geb17a279
...
ROM: Table BASE=0xe00ff000 SYSMEM=0x00000001, Manufacturer  20 Partno 411 
...
***  1      STM32F40x M4
...

Probe succeeds, as expected. t->idcode is not used anywher in target/stm32f4.c
andrew1x000: What problem are you trying to solve?

UweBonnes:
Okay, with the F446 nucleo now I see:
...
ROM: Table BASE=0xe00ff000 SYSMEM=0x00000001, Manufacturer  20 Partno 411
...
***  1      STM32F40x M4
if (t->part_id == ID_STM32F20X)
		t->part_id = target_mem_read32(t, DBGMCU_IDCODE) & 0xFFF;
does the trick

andrew1x000:
@UweBonnes, I have the STM32F446ZE Nucleo board on hand. And in v1.8.2 after the probe, it shows for me
...
*** 1 STM32F40x M4
...

Which is wrong. 
Expected output should be
***
***  1   *  STM32F446 M4
***

You mentioned v1.8.0. [Here ](https://github.com/blackmagic-debug/blackmagic/blob/5e3cadec20a5fa3ec6eef53f23520707d0db5380/src/target/stm32f4.c#L216)is the usage of t->idcode in your version

UweBonnes:
Sorry, I did not read your original post carefully enough and missed thee F446 keyword

dragonmux:
Thank you for fixing the formatting items, it is much appreciated. As per our original review comment, the underlying bug is not in the probe routine, but in the incorrect ID codes that BMP has due to the historical way these parts were ID'd - we still need you to change [src/target/stm32f4.c:123](https://github.com/blackmagic-debug/blackmagic/blob/main/src/target/stm32f4.c#L123) as this is where the underlying issue is.

UweBonnes:
The problem is that F20x and some F4 devices like F405 have the same Romtable part ID. So using the partid read from DBGMCU_IDCODE is the only way to differentiate these devices. IDS_STM32F247 is right.

dragonmux:
In that case, rather than modifying the target structure, which has the correct ID codes for other reasons and parts.. please locally grab the DBGMCU_IDCODE and use that only where it's needed.. eg:
```c
[...]
case ID_STM32F413: { /* F413     RM0430 Rev.2, 320 kB Ram, 1.5 MB flash. */
    t->attach = stm32f4_attach;
    t->detach = stm32f4_detach;
    const uint16_t mcu_idcode = target_mem_read32(t, DBGMCU_IDCODE) & 0xFFF;
    t->driver = stm32f4_get_chip_name(mcu_idcode);
    target_add_commands(t, stm32f4_cmd_list, t->driver);
    return true;
}
```

That would be a more ideal pattern for this particular issue so we don't screw things up for later probe routines.

UweBonnes:
If done with attach, then a wrong chipname based on the wrong part_id is added with target_add_command.

dragonmux:
We weren't saying to do the readout during attach.. and target_add_commands() doesn't use the target ID at all - t->driver is set by the preceding line here, and in our suggestion that is driven by the DBGMCU_IDCODE register, so t->driver will be set correctly

UweBonnes:
If done with attach, "mon s" will report the wrong driver.

dragonmux:
`mon s` uses t->driver, t->driver is set by the suggested snippet by the DBGMCU_IDCODE, so is unable to be set to the wrong value - we fail to see how it'll report the wrong driver when the snippet sets the correct driver value.

UweBonnes:
Yes,the whole switch case can be omitted
	t->part_id = target_mem_read32(t, DBGMCU_IDCODE) & 0xFFF;
	t->driver = stm32f4_get_chip_name(t->part_id);
	if (t->driver) {
		t->mass_erase = stm32f4_mass_erase;
		t->detach = stm32f4_detach;
		t->driver = stm32f4_get_chip_name(t->part_id);
		t->attach = stm32f4_attach;
		target_add_commands(t, stm32f4_cmd_list, t->driver);
		return true;
	} else }
		return false;
	}

dragonmux:
Uwe, `t->part_id` is unused by the monitor commands that display the target information, see (src/command.c:374)[https://github.com/blackmagic-debug/blackmagic/blob/main/src/command.c#L374] - please do NOT modify t->part_id in the probe routines, use a local variable and pass it to the get chip name call. It will do the right thing and t->driver will get set to the right value.

dragonmux:
It's a thing we're trying to remove because it's possible STM create a new part and we can't probe for it because of all the messing about t->part_id gets - it should be maintained as the ADIv5 value to be quite honest

UweBonnes:
What we read from the Romtable or DP_TARGETID should be taken as a hint, not as absolute truth. E.G. STM32L55 give a valid 0 as DP_TARGETID, F405, F407 and F446 at least use the same part_id and most NXP device only use the ARM core ID. So tinkering is needed in some cases. By the way, Arm describes "DP_TARGETID" as implementation defined, so that is no defined way to decipher.

andrew1x000:
> In that case, rather than modifying the target structure, which has the correct ID codes for other reasons and parts.. please locally grab the DBGMCU_IDCODE and use that only where it's needed

But if `t->idcode` is not changed in `stm32f4_probe()` right in next call `stm32f4_attach()` program jump to the wrong case 
```
case ID_STM32F40X:
		has_ccmram = true;
		max_flashsize = 1024;
		break;
```
		
And after that, I can't erase or flash the chip. So I think we need to read `t->idcode` with DBGMCU_IDCODE. At least in the case of STM32F446

UweBonnes:
uint16_t temp_id = target_mem_read32(t, DBGMCU_IDCODE) & 0xFFF;
	t->driver = stm32f4_get_chip_name(temp_id);
	if (t->driver) {
		t->part_id = temp_id;
		t->mass_erase = stm32f4_mass_erase;
		t->detach = stm32f4_detach;
		t->driver = stm32f4_get_chip_name(t->part_id);
		t->attach = stm32f4_attach;
		target_add_commands(t, stm32f4_cmd_list, t->driver);
		return true;
	default:
		return false;
	}
}

collapses the switch case at the expense of always one added target transaction.

dragonmux:
Uwe, the problem isn't how to use `t->part_id`, it's that blazé overwriting it in the probe routines with device-specific register values, before decoding it, which then screws things up for the next probe routine which then screws things up for the next, etc. That's why we're trying to eliminate the modification of it as it's already caused a significant number of bugs that keep coming back.

@andrew1x000 then a slight modification of our suggestion, once a positive identification on the part has been made from the local, then and only then in the `return true` path of the switch-case, do `t->part_id = mcu_idcode;`, so that it's only changed once we're sure it is a STM32F4 part.

UweBonnes:
But my first approach
if (t->part_id == ID_STM32F20X)
     t->part_id = target_mem_read32(t, DBGMCU_IDCODE) & 0xFFF;
is not a "blaze"overwrite.

dragonmux:
Unless that write is valid and correct for the STM32F20x parts and any others STM makes with the same part number, it is. That's exactly why we said "a positive identification on the part" to Andrew as only after checking the value that read returns can we write it to `t->part_id` safely.

andrew1x000:
I just tested v1.8.2 with STM32F437 and it stores wrong `t->idcode` as well. Shows `1041 (0x411)` should detect as `1049 (0x419)`. 

@dragonmux, I will use your approach, but how can we be sure it is an STM32F4 part? I'm not so good in STM32 internals related to flashing.
Code now has a part that detects CORTEX_M4. Can I use that? e.g.:
```
bool stm32f4_probe(target *t)
{
	if (t->idcode == ID_STM32F20X) {
		/* F405 revision A has a wrong ID CODE, use ARM_CPUID to make the
		* distinction with F205. Revision is also wrong (0x2000 instead
		* of 0x1000). See F40x/F41x errata. */
		if ((t->cpuid & 0xFFF0) == CORTEX_M4) {
			t->idcode = target_mem_read32(t, DBGMCU_IDCODE) & 0xFFF;
	         }
	}
	switch (t->idcode) {
	***
```

Or do you suggest going with the local variable through the switch-case? e.g.:
```
bool stm32f4_probe(target *t)
{
	const uint16_t mcu_idcode = target_mem_read32(t, DBGMCU_IDCODE) & 0xFFF;
	if (mcu_idcode == ID_STM32F20X) {
		/* F405 revision A have a wrong IDCODE, use ARM_CPUID to make the
		 * distinction with F205. Revision is also wrong (0x2000 instead
		 * of 0x1000). See F40x/F41x errata. */
		if ((t->cpuid & 0xFFF0) == CORTEX_M4)
			t->idcode = ID_STM32F40X;
	}
	switch(mcu_idcode) {
	case ID_STM32F74X: /* F74x RM0385 Rev.4 */
	case ID_STM32F76X: /* F76x F77x RM0410 */
	case ID_STM32F72X: /* F72x F73x RM0431 */
	case ID_STM32F40X:
	case ID_STM32F42X: /* 427/437 */
	case ID_STM32F46X: /* 469/479 */
	case ID_STM32F20X: /* F205 */
	case ID_STM32F446: /* F446 */
	case ID_STM32F401C: /* F401 B/C RM0368 Rev.3 */
	case ID_STM32F411: /* F411     RM0383 Rev.4 */
	case ID_STM32F412: /* F412     RM0402 Rev.4, 256 kB Ram */
	case ID_STM32F401E: /* F401 D/E RM0368 Rev.3 */
	case ID_STM32F413: /* F413     RM0430 Rev.2, 320 kB Ram, 1.5 MB flash. */
		t->detach = stm32f4_detach;
		t->driver = stm32f4_get_chip_name(t->idcode);
		t->attach = stm32f4_attach;
		t->idcode = mcu_idcode;
		target_add_commands(t, stm32f4_cmd_list, t->driver);
		return true;
	default:
		return false;
	}
}
```

andrew1x000:
Also, I'm facing further problems with the STM32F4 series flashing. But the problem doesn't belong to probing. 
I wanted to write firmware to flash. When I got the right `t->idcode`, in `stm32f4_attach()` there are calls for `stm32f4_add_flash()`.
For example with STM32F446 attached it calls
```
static bool stm32f4_attach(target *t)
{
***
stm32f4_add_flash(t, 0x8000000, 0x10000,  0x4000,  0, split);
***
stm32f4_add_flash(t, 0x8020000, remains, 0x20000,  5, split);
***
```

Inside that function, each time changes start address for chip flash
```
static void stm32f4_add_flash(target *t,
                              uint32_t addr, size_t length, size_t blocksize,
                              unsigned int base_sector, int split)
{
	if (length == 0) return;
	struct stm32f4_flash *sf = calloc(1, sizeof(*sf));
	struct target_flash *f;
	if (!sf) {			/* calloc failed: heap exhaustion */
		DEBUG_WARN("calloc: failed in %s\n", __func__);
		return;
	}

	f = &sf->f;
	f->start = addr; ///<<<here
	f->length = length;
        ***
```
So after `stm32f4_attach()` BMP tries to write a file into the address stored in `t->flash->start`. In the example below it should be `0x8000000` but in fact, it writes to `0x8020000`. And firmware doesn't work.

I have such a pattern in both chips: STM32F446 and STM32F437.

Is it the right algorithm? Or I misunderstood something?

Should I create a new MR or issue for this problem?

andrew1x000:
Update on my flashing problem. I compared BMP v1.8.2 and v1.7.1. They seem to use different orders for adding flash in `stm32f4_attach()` :
		
v1.8.2:
```
		stm32f4_add_flash(t, 0x8000000, 0x10000,  0x4000,  0, split);
		if (banksize > 0x10000) {
		stm32f4_add_flash(t, 0x8010000, 0x10000, 0x10000,  4, split);
			stm32f4_add_flash(t, 0x8020000, remains, 0x20000,  5, split);
		}
```
v1.7.1 :
```
		stm32f4_add_flash(t, 0x8020000, remains, 0x20000, 5, split);
		stm32f4_add_flash(t, 0x8010000, 0x10000, 0x10000,  4, split);
		stm32f4_add_flash(t, 0x8000000, 0x10000, 0x4000, 0, split);
```

With v1.7.1 my STM32F446 flashes as expected

dragonmux:
We would prefer you use the latter approach. It's the safest available way to detect the part and not screw t->idcode up for other routines in the process.

dragonmux:
Looking at the logic, we can actually fold away the if statement and default case - we would suggest in this approach, doing this:

```c
bool stm32f4_probe(target *t)
{
	uint16_t mcu_idcode = target_mem_read32(t, DBGMCU_IDCODE) & 0xFFF;
	switch(mcu_idcode) {
	/* F405 revision A have a wrong IDCODE, use ARM_CPUID to make the
	 * distinction with F205. Revision is also wrong (0x2000 instead
	 * of 0x1000). See F40x/F41x errata. */
	case ID_STM32F20X:
		if ((t->cpuid & 0xFFF0U) != CORTEX_M4)
			break;
		mcu_idcode = ID_STM32F40X;

	case ID_STM32F74X: /* F74x RM0385 Rev.4 */
	case ID_STM32F76X: /* F76x F77x RM0410 */
	case ID_STM32F72X: /* F72x F73x RM0431 */
	case ID_STM32F42X: /* 427/437 */
	case ID_STM32F46X: /* 469/479 */
	case ID_STM32F20X: /* F205 */
	case ID_STM32F446: /* F446 */
	case ID_STM32F401C: /* F401 B/C RM0368 Rev.3 */
	case ID_STM32F411: /* F411     RM0383 Rev.4 */
	case ID_STM32F412: /* F412     RM0402 Rev.4, 256 kB Ram */
	case ID_STM32F401E: /* F401 D/E RM0368 Rev.3 */
	case ID_STM32F413: /* F413     RM0430 Rev.2, 320 kB Ram, 1.5 MB flash. */
		t->attach = stm32f4_attach;
		t->detach = stm32f4_detach;
		t->driver = stm32f4_get_chip_name(t->part_id);
		t->part_id = mcu_idcode;
		target_add_commands(t, stm32f4_cmd_list, t->driver);
		return true;
	}
	return false;
}
```

andrew1x000:
> We would prefer you use the latter approach. It's the safest available way to detect the part and not screw `t->idcode` up for other routines in the process 

Ok, will make changes in short.

> (also, isn't that supposed to be `t->part_id`? the idcode should be the part PIDR and we actually want it set to that.. `t->part_id` is what the attach routine uses and this patch doesn't change that).

Hmm. I'm looking at v1.8.2 and there is no `t->part_id`:

https://github.com/blackmagic-debug/blackmagic/blob/5cd430647ecfb6dab4a2ed858fb98567723de699/src/target/target_internal.h#L115

https://github.com/blackmagic-debug/blackmagic/blob/5cd430647ecfb6dab4a2ed858fb98567723de699/src/target/stm32f4.c#L258

Oh. I see it in the main branch. My bad

src/target/stm32f4.c
```diff
@@ -207,36 +207,37 @@ static void stm32f4_detach(target *t)
...
 {
-	if (t->part_id == ID_STM32F20X) {
-		/* F405 revision A have a wrong IDCODE, use ARM_CPUID to make the
-		 * distinction with F205. Revision is also wrong (0x2000 instead
-		 * of 0x1000). See F40x/F41x errata. */
-		if ((t->cpuid & 0xFFF0) == CORTEX_M4)
-			t->part_id = ID_STM32F40X;
-	}
-	switch(t->part_id) {
+	const uint16_t mcu_idcode = target_mem_read32(t, DBGMCU_IDCODE) & 0xfffU;
```
dragonmux, line 210: Because of the assignment in the first case, this can't be marked const (unfortunatly)

dragonmux (commented):
This is looking good to us. We'll merge it once the one thing we saw in review is fixed, which should make it build

src/target/stm32f4.c
andrew1x000: I see. But there is an error because in switch-case we have 2 identical cases. So I returned if statement before switch-case

dragonmux (dismissed):
LGTM, we'll merge this once the build completes. Thank you again for the contribution! Much appreciated too that this was your first

src/target/stm32f4.c
dragonmux: Ah, good catch. We'll have a play later with seeing if we can still somehow remove the need for the pre-switch if (think we know a way), but getting this fix in is more important right now.

dragonmux:
Ah, in the path that returns true in the probe routine, please add `t->mass_erase = stm32f4_mass_erase;` after the `t->detach` assignment to fix the build

dragonmux (approved):

Issue #1243, opened 2022-09-19
Support for Renesas RA2A1 ?


mtk11:
Trying to use it with Renesas RA2A1 and getting the following:

```
WARNING: "Debug component" !match expected "Generic IP component"
Found renesas chip (R7FA2A1AB3CNF   ) with Flash Root Table and unsupported Part ID 0x1b0 please report it
0x80000000: 0x00000004000a3004 Non ARM component ignored
```

Support for this series ?

dragonmux:
There is experimental support for Renesas parts thanks to @perigoso's #1133, we'll take their lead on how to deal with that. v1.9 is currently in a feature freeze pending release of v1.9rc1, so this will get targeted to v1.10.

mtk11:
> does it not show up as an attachable target?

```
(gdb) mon s
Available Targets:
No. Att Driver
 1      R7FA2A1AB3CNF    M23
(gdb) att 1
Attaching to program: /tmp/ra2a1-main/blinky/main.hex, Remote target
0xeffffffe in ?? ()
(gdb) load main.hex
Loading section .sec1, size 0xc0 lma 0x0
Load failed
```

dragonmux:
We have re-opened the issue and will modify the tittle accordingly.

PR #1247, opened 2022-09-22
Fix: adiv5 nuisance CID warnings

dragonmux:
## Detailed description

In f4345e4 we made adjustments to the ADIv5 component table, which were based on the SCS classes not matching up which seemed like a copy-paste mistake. As it turns out this is actually that the Cortex-M23 and Cortex-M33 SCSs are defined as debug components in the processor TRMs, but this is actually incorrect as they're generic IP. This PR fixes this by detecting those two SCS's and adjusting the class ID going into the sanity check.

It additionally adjusts the sanity check message so it's a bit more clear what's wrong. This should fix the warning seen in #1243 which raised some eyebrows.


esden (approved):
LGTM

PR #1249, opened 2022-09-24
stm32f1xx: Fix broken vFlashErase

fk0815:
## Detailed description

When trying to load the flash content via gdb the flash erase fails for a stm32f1xx target:

(gdb) load

This fixes the wrong EOP evaluation logic, and also due to errata "2.12 Flash memory BSY bit delay versus STRT bit setting", it makes sense to evaluate both status bits, EOP and BSY.

* [ x] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md)
* [ x] I've read the [guidelines for contributing](https://github.com/blackmagic-debug/blackmagic/blob/main/CONTRIBUTING.md) to this repository
* [ x] It builds for hardware native (`make PROBE_HOST=native`)
* [ x] It builds as BMDA (`make PROBE_HOST=hosted`)
* [ x] I've tested it to the best of my ability

### DIFF
--- a/src/target/stm32f1.c
+++ b/src/target/stm32f1.c
@@ -357,17 +357,25 @@ static int stm32f1_flash_unlock(target *t, uint32_t bank_offset)
 	return 0;
 }
 
-static bool stm32f1_flash_busy_wait(target *t, uint32_t bank_offset)
+static inline void stm32f1_flash_clear_eop(target * const t, const uint32_t bank_offset)
+{
+	const uint32_t status = target_mem_read32(t, FLASH_SR + bank_offset);
+	target_mem_write32(t, FLASH_SR + bank_offset, status | SR_EOP); /* EOP is W1C */
+}
+
+static bool stm32f1_flash_busy_wait(target * const t, const uint32_t bank_offset, platform_timeout * const timeout)
 {
 	/* Read FLASH_SR to poll for BSY bit */
 	uint32_t sr;
 	do {
 		sr = target_mem_read32(t, FLASH_SR + bank_offset);
-		if ((sr & SR_ERROR_MASK) || !(sr & SR_EOP) || target_check_error(t)) {
+		if ((sr & SR_ERROR_MASK) || target_check_error(t)) {
 			DEBUG_WARN("stm32f1 flash error 0x%" PRIx32 "\n", sr);
 			return false;
 		}
-	} while (sr & FLASH_SR_BSY);
+		if (timeout)
+			target_print_progress(timeout);
+	} while (!(sr & SR_EOP) && (sr & FLASH_SR_BSY));
 
 	return true;
 }
@@ -390,6 +398,8 @@ static bool stm32f1_flash_erase(target_flash_s *f, target_addr_t addr, size_t le
 		if (addr >= FLASH_BANK_SPLIT)
 			bank_offset = FLASH_BANK2_OFFSET;
 
+		stm32f1_flash_clear_eop(t, bank_offset);
+
 		/* Flash page erase instruction */
 		target_mem_write32(t, FLASH_CR + bank_offset, FLASH_CR_PER);
 		/* write address to FMA */
@@ -398,7 +408,7 @@ static bool stm32f1_flash_erase(target_flash_s *f, target_addr_t addr, size_t le
 		target_mem_write32(t, FLASH_CR + bank_offset, FLASH_CR_STRT | FLASH_CR_PER);
 
 		/* Wait for completion or an error */
-		if (!stm32f1_flash_busy_wait(t, bank_offset))
+		if (!stm32f1_flash_busy_wait(t, bank_offset, NULL))
 			return false;
 
 		if (len > f->blocksize)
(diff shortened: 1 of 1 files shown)


dragonmux:

Before we dive too deep, we might already have fixed this bug (at least partially) in 097dde2. It would be useful if you could check that commit, especially as this patch pulls in part of #1108 which we would like to avoid to avoid mergeability issues.

The fix/clang-format-cleanup branch is currently WIP but contains some of the final changes for v1.9 that were planned, so we can guarantee that changes there will make it into the final release of v.19.

src/target/stm32f1.c
```diff
@@ -452,61 +467,59 @@ static bool stm32f1_mass_erase(target *t)
...
 	/* Wait for completion or an error */
-	do {
-		sr = target_mem_read32(t, FLASH_SR);
-		if ((sr & SR_ERROR_MASK) || !(sr & SR_EOP) || target_check_error(t)) {
-			DEBUG_WARN("stm32f1 flash error 0x%" PRIx32 "\n", sr);
-			return false;
-		}
-		target_print_progress(&timeout);
-	} while (sr & FLASH_SR_BSY);
+	err = stm32f1_flash_busy_wait(t, 0, &timeout);
```
dragonmux, line 481: Rather than doing this pattern, the following would be preferred and eliminates `bool err` above:
```c
if (!stm32f1_flash_busy_wait(t, 0, &timeout))
    return false;
```

src/target/stm32f1.c
```diff
@@ -452,61 +467,59 @@ static bool stm32f1_mass_erase(target *t)
...
 		/* Wait for completion or an error */
-		do {
-			sr = target_mem_read32(t, FLASH_SR + FLASH_BANK2_OFFSET);
-			if ((sr & SR_ERROR_MASK) || !(sr & SR_EOP) || target_check_error(t)) {
-				DEBUG_WARN("stm32f1 flash error 0x%" PRIx32 "\n", sr);
-				return false;
-			}
-			target_print_progress(&timeout);
-		} while (sr & FLASH_SR_BSY);
+		err = stm32f1_flash_busy_wait(t, FLASH_BANK2_OFFSET, &timeout);
```
dragonmux, line 496: Same comment as above

src/target/stm32f1.c
```diff
@@ -357,17 +357,25 @@ static int stm32f1_flash_unlock(target *t, uint32_t bank_offset)
 	return 0;
 }
 
-static bool stm32f1_flash_busy_wait(target *t, uint32_t bank_offset)
+static void stm32f1_flash_clear_eop(target *t, uint32_t bank_offset)
+{
+	uint32_t sr = target_mem_read32(t, FLASH_SR + bank_offset);
```
dragonmux, line 362: `sr` here should be named `status`, which avoids the clang-tidy warning about too short a variable name and lends more clarity

src/target/stm32f1.c
```diff
@@ -357,17 +357,25 @@ static int stm32f1_flash_unlock(target *t, uint32_t bank_offset)
 	return 0;
 }
 
-static bool stm32f1_flash_busy_wait(target *t, uint32_t bank_offset)
+static void stm32f1_flash_clear_eop(target *t, uint32_t bank_offset)
```
dragonmux, line 360: Please add `const` to both parameters as they're not modified by the function. Given how short this is, it would be ideal to mark the function inline too as in:
`static inline void stm32f1_flash_clear_eop(target *const t, const uint32_t bank_offset)`

src/target/stm32f1.c
```diff
@@ -357,17 +357,25 @@ static int stm32f1_flash_unlock(target *t, uint32_t bank_offset)
...
+{
+	uint32_t sr = target_mem_read32(t, FLASH_SR + bank_offset);
+	target_mem_write32(t, FLASH_SR + bank_offset, sr | SR_EOP); /* EOP is W1C */
+}
+
+static bool stm32f1_flash_busy_wait(target *t, uint32_t bank_offset, platform_timeout *timeout)
```
dragonmux, line 366: Same comment as above for `const` on the parameters.

dragonmux (changes requested):
This is looking good. There are a couple of stylistic fixes we'd like to see, along side a small amount of cleanup this code was due anyway.

With the review items addressed, we'd be happy to merge this. Think it would be a good idea for us to do a review of the other STM32 code and see where else BMP is doing this anachronistic check for SR_EOP that's wrong, as this isn't the first time we've come across it (it's been a problem for at least 2 other target families) - but this is an action for @esden and ourself to worry about.

src/target/stm32f1.c
fk0815: Marking the "target *t" as const would require also to change the target_mem_read32(), which has no const marker. That would then require to change a lot more files, which should be a separate PR I think.

dragonmux: There are two kinds of const. The target being const, the pointer to the target being const. We're asking you to do the latter, no changes to target_mem_read32() needed - `target *const t`

dragonmux (approved):
This nicely improves the STM32F1 code and while it's come to our attention that stm32f1_flash_busy_wait() needs some further modification to make the STM32 clones happy, we're happy that this fixes the bug based on other users testing. LGTM, merging

PR #1253, opened 2022-10-01
Fix USART on SWIM pins for STLINK

dossalab:
Hi,
I checked if `SWIM_AS_UART` mode mentioned [here](https://github.com/blackmagic-debug/blackmagic/blob/main/src/platforms/stlink/README.md) is working as expected, and it turns out there is an issue. The flag was correctly switching USART port to 1 and setting up pin remap, however it left DMA channels and GPIO pins configured for USART2.

SWO output feature was using the same first UART controller, so I had to switch it off for this particular quirk.

Tested on random STLINK dongle from Aliexpress.

I don't like that a lot of those macros are basically duplicated twice (with the port number being the only difference), but I can't figure if there is a nice way to address it.

* [ x ] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md)
* [ x ] I've read the [guidelines for contributing](https://github.com/blackmagic-debug/blackmagic/blob/main/CONTRIBUTING.md) to this repository
* [ x ] It builds for hardware native (`make PROBE_HOST=native`)
* [ x ] It builds as BMDA (`make PROBE_HOST=hosted`)
* [ x ] I've tested it to the best of my ability
* [ x ] My commit messages provide a useful short description of what the commits do

### DIFF
--- a/src/platforms/stlink/Makefile.inc
+++ b/src/platforms/stlink/Makefile.inc
@@ -27,6 +27,8 @@ endif
 
 ifeq ($(SWIM_AS_UART), 1)
 CFLAGS += -DSWIM_AS_UART=1
+else
+SRC += traceswoasync.c
 endif
 
 VPATH += platforms/stm32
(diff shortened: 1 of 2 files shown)


dragonmux (approved):
LGTM, and you indicated in Discord that you tested it, so while we can't, we'll merge anyway. Thank you for the contribution!

PR #1254, opened 2022-10-02
support for another lpc11u6x part id

mean00:
Hi,
* Is a new feature implemented?  No
* What existing problem(s) does the pull request solve? Compatibility with some LPC11U67/LPC11U68
* How does the pull request solve these problems? MCU not identified without

Trivial patch attached to add support for another LPC11U6x part_id.
Works fine with the patch. 

Thank you

### DIFF
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -682,8 +682,10 @@ bool cortexm_probe(ADIv5_AP_t *ap)
 		PROBE(renesas_probe);
 		break;
 	case JEP106_MANUFACTURER_ARM:
-		if (t->part_id == 0x4c0 || t->part_id == 0x4c1) { /* Cortex-M0+ ROM (0x4c1 is for some newer LPC11U6xx)*/
+		if (t->part_id == 0x4c0) {        /* Cortex-M0+ ROM */
 			PROBE(lpc11xx_probe);         /* LPC8 */
+		}else if (  t->part_id == 0x4c1) { /* Cortex-M0+ ROM */
+			PROBE(lpc11xx_probe);         /* newer LPC11U6x */
 		} else if (t->part_id == 0x4c3) { /* Cortex-M3 ROM */
 			PROBE(lmi_probe);
 			PROBE(ch32f1_probe);
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -682,7 +682,7 @@ bool cortexm_probe(ADIv5_AP_t *ap)
 		PROBE(renesas_probe);
 		break;
 	case JEP106_MANUFACTURER_ARM:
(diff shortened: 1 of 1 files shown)


src/target/cortexm.c
```diff
@@ -682,7 +682,7 @@ bool cortexm_probe(ADIv5_AP_t *ap)
 		PROBE(renesas_probe);
 		break;
 	case JEP106_MANUFACTURER_ARM:
-		if (t->part_id == 0x4c0) {        /* Cortex-M0+ ROM */
+		if (t->part_id == 0x4c0 || t->part_id == 0x4c1) { /* Cortex-M0+ ROM (0x4c1 is for some newer LPC11U6xx)*/
```
dragonmux, line 685: Rather than having this in the same if block, it is preferred that the `t->part_id` is tested against `0x4c1U` in an `else if` and the probe call duplicated. Having to reorganise this later as and when new parts are added to the block will be less than idea, especially as this change don't make it clear to the reader that the block is safe to split.

dragonmux (changes requested):
Thank you for the contribution! There are a few things that stood out that we'd like addressed:

1. In the future please use our pull request template as this lets us know certain important pieces of information including that you've read the CoC, contribution guidelines, etc. If you're using a Git client like GitKraken that doesn't understand GitHub pull request templates, you can find the template at `.github/PULL_REQUEST_TEMPLATE.md` in the repository.
2. Please format your commit message according to the contribution guidelines - in this case: `cortexm: Added support for newer LPC11U6x parts`
3. While we see the value in the patch, please add a new `else if` to the set of blocks this modifies so the addition is better future proofed. This is in part because we can expect more Cortex-M0+ CPUs to need adding in the future, and because 0x4c1 is something NXP is doing quite specifically here and is not a standard ROM table value defined by ARM (at least not for the Cortex-M0+). This will keep the addition separated from the ARM-defined values and help stake it out as the vendor doing something a bit weird.

mean00:
Thank you for the review, hopefully this time it is correct.

I tried to identify where the 0x4C1 was coming from with little success. 

Missing Pull info : 
[X] I've read the Code of Conduct
[X] I've read the guidelines for contributing to this repository
[X] It builds for hardware native (make PROBE_HOST=native)
[X] It builds as BMDA (make PROBE_HOST=hosted)
[ - ] I've tested it to the best of my ability => ** Tested with the same patch on v1.8.2, main does not work with some STM chips, cannot use it **
[X] My commit messages provide a useful short description of what the commits do

dragonmux (approved):
Thank you for the speedy response in addressing the review points. I'm not going to worry about the formatting this time (clang-format will be enforced in the near future) as we're currently working on a branch to apply clang-format across the code base and the formatting will get fixed at that point anyway.

LGTM, merging.

PR #1262, opened 2022-10-09
add rtt

koendv:
## Detailed description

- Current behavior:
When compiled with ENABLE_RTT=1, characters typed at the terminal /dev/ttyBmpTarg are always sent to RTT
This is unpleasant for people who wish to use the built-in serial port.
- New behavior:
When compiled with ENABLE_RTT=1, if rtt is enabled (``mon rtt``) then characters typed at the terminal are sent to RTT.
If rtt is not enabled (default) then characters typed at the terminal are sent to the built-in serial

While testing this PR noticed the following:
``make all_platforms ENABLE_RTT=1`` fails. ``src/platforms/launchpad-icdi/rtt_if.c`` is missing.
fix:
copied  ``src/platforms/stm32/rtt_if.c`` to ``src/platforms/launchpad-icdi/rtt_if.c``

### DIFF
--- a/UsingRTT.md
+++ b/UsingRTT.md
@@ -40,6 +40,12 @@ max poll ms: 256 min poll ms: 8 max errs: 10
 The terminal emulator displays RTT output from the target, and characters typed
 in the terminal emulator are sent via RTT to the target.
 
+It is possible to use RTT and the built-in serial port at the same time. If both RTT and the built-in serial port are used at the same time, the terminal will show both RTT and serial port output, interleaved.
+
+Characters typed in at the terminal go to either RTT or serial port.
+
+-  ``monitor rtt``: If RTT is enabled, characters typed in at the terminal are sent to target RTT.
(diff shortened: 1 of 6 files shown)


dragonmux:
Hi Koen, before we review the PR, could we please ask a couple of things:

* Please could you update the PR description with a filled in copy of the issue template checklist, this is a mandatory component of PRs;
* Could you please provide a better description than "add rtt" as, well.. we already have RTT so we're unclear on what this is really doing;
* Finally, as per the contribution guidelines, please correct your commit messages to describe what part of the project they're modifying and be descriptive so we can use the short messages to understand what the commits change and why - for example, `launchpad-icdi: Coppied over rrt_if.c from stm32 base platform, enabling support on the Launchpad platform`

koendv:
You're right. I'll close this PR and try again.

dragonmux:
No need to close the PR - you can update the PR title + description (there's an edit button) and force pushing to your branch will allow you to fix the commit messages, which would be preferable to the project

koendv:
Sorry, ended up opening a new PR.

PR #1265, opened 2022-10-12
fix simultaneous use of rtt and built-in serial

koendv:
## Detailed description

- There is a problem with using rtt and built-in serial at the same time. 
On a bmp compiled with ENABLE_RTT=1, built-in serial does not work for characters from host to target. Data from target to host is fine. Cause is that, when compiled with ENABLE_RTT=1, characters typed in at the terminal always go to rtt.
After this PR,  on bmp compiled with ENABLE_RTT=1, if rtt is enabled (``mon rtt``) then characters typed in at the terminal go to rtt. If rtt is not enabled (``mon rtt dis``, default), characters typed in at the terminal go to the built-in serial.
- ``make all_platforms ENABLE_RTT=1`` fails, because src/platforms/launchpad-icdi/rtt_if.c is missing. Fix: copy src/platforms/stm32/rtt_if.c to src/platforms/launchpad-icdi/rtt_if.c
- to avoid a PR breaking rtt, have added a build with ENABLE_RTT=1 to build_pr.yml. This way, rtt is checked with every pull request. An alternative would be to make ENABLE_RTT=1 the default.  


dragonmux:
Heya, sorry this has slipped getting looked at a bit. Thank you for the replacement PR which we will do code review on later today.

One small thing that immediately hits our notice though is the commit messages - as requested in #1262: as per the contribution guidelines, please correct your commit messages to describe what part of the project they're modifying and be descriptive so we can use the short messages to understand what the commits change and why. You can force-push to your branch to update this PR.

koendv:
On Sun, 23 Oct 2022 00:42:37 -0700
Rachel Mant ***@***.***> wrote:

> please correct your commit messages

ok, done.

src/platforms/hosted/rtt_if.c
```diff
@@ -66,7 +66,8 @@ int rtt_if_exit()
 
 uint32_t rtt_write(const char *buf, uint32_t len)
 {
-	write(1, buf, len);
+	int l = write(1, buf, len);
```
dragonmux, line 69: This change suggests that the correct code here should either be `(void)write(1, buf, len);`, or `const int32_t result = write(1, buf, len); (void)result;` - iwrc the former works and feels better. At a minimum, this needs `const` and a non-single-letter variable name using.

src/platforms/launchpad-icdi/rtt_if.c
```diff
@@ -0,0 +1,134 @@
...
+/* usb uart receive buffer */
+static char recv_buf[RTT_DOWN_BUF_SIZE];
+static uint32_t recv_head = 0;
+static uint32_t recv_tail = 0;
+
+/* data from host to target: number of free bytes in usb receive buffer */
+inline static uint32_t recv_bytes_free()
+{
+	if (recv_tail <= recv_head)
+		return sizeof(recv_buf) - recv_head + recv_tail - 1;
```
dragonmux, line 50: Does this not make more sense written as `RTT_DOWN_BUF_SIZE - recv_head + recv_tail - 1U`?

dragonmux (changes requested):
This is looking very good, We're overlooking a couple of things in rtt_if.c as those are addressed already on `fix/clang-format-cleanup`, but there are a couple of things that we wanted to ask for that cleanup (included as review items) and there is an item we spotted in another file. Once those are addressed, we're happy to merge. Thank you for the contribution!

koendv:
On Sun, 23 Oct 2022 06:23:07 -0700
Rachel Mant ***@***.***> wrote:

> > @@ -66,7 +66,8 @@ int rtt_if_exit()  
>  
>  uint32_t rtt_write(const char *buf, uint32_t len)
>  {
> -	write(1, buf, len);
> +	int l = write(1, buf, len);
> 
> This change suggests that the correct code here should either be
> `(void)write(1, buf, len);`, or `const int32_t result = write(1, buf,
> len); (void)result;` - iwrc the former works and feels better. At a
> minimum, this needs `const` and a non-single-letter variable name
> using.

I agree `(void) write(1, buf, len);` is the logical thing to write, but
the compiler issues a warning on this?

koen

dragonmux:
Thought that the `(void)` on the call itself suppressed the unused result warning.. interesting - we're better versed with C++ where you can fix this with a simple `[[maybe_unused]]` attribute on the assigned variable. Well, assignment and then `(void)` casting the variable it is then.

Please either use a fixup commit, rebase and force-push or if you prefer, use a normal commit and adhere to the contribution requirements for commit messages. We're glad to have explored options here though to have the best possible patch for the problem.

koendv:
On Sun, 23 Oct 2022 09:07:23 -0700
Rachel Mant ***@***.***> wrote:

> Thought that the `(void)` on the call itself suppressed the unused
> result warning.. interesting - we're better versed with C++ where you
> can fix this with a simple `[[maybe_unused]]` attribute on the
> assigned variable. Well, assignment and then `(void)` casting the
> variable it is then.
> 
ok.
> Please either use a fixup commit, rebase and force-push or if you
> prefer, use a normal commit and adhere to the contribution
> requirements for commit messages. We're glad to have explored options
> here though to have the best possible patch for the problem.
> 
ok, done.

koen

dragonmux (approved):
LGTM! Merging

PR #1269, opened 2022-10-16
hosted: Moved function "extract_serial" from bmp_serial to probe_info to allow reuse in bmp_libusb

sidprice:
## Moved function "extract_serial" from bmp_serial to bmp_libusb

Moving this function allows it to be re-used in the USB scanning of probes.

### DIFF
--- a/src/platforms/hosted/probe_info.h
+++ b/src/platforms/hosted/probe_info.h
@@ -48,7 +48,7 @@ typedef struct probe_info {
 	struct probe_info *next;
 } probe_info_s;
 
-char *extract_serial(const char *const device, const size_t length) ;
+char *extract_serial(const char *device, size_t length);
 probe_info_s *probe_info_add(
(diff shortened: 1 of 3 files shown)


src/platforms/hosted/probe_info.h
```diff
@@ -48,6 +48,7 @@ typedef struct probe_info {
 	struct probe_info *next;
 } probe_info_s;
 
+char *extract_serial(const char *const device, const size_t length) ;
```
dragonmux, line 51: clang-tidy will complain about the excess of const here (`const char *` and `size_t` only)

dragonmux (approved):
LGTM! Only ask is for the declaration in probe_info.h to be cleaned up but we'll merge this if you don't want to and run a fixup commit to fix the clang-tidy warnings it introduces

src/platforms/hosted/probe_info.h
sidprice: Pushed update.

PR #1271, opened 2022-10-21
Implement proper support for GD32F4 microcontrollers

via:
This implements support for the GD32F450 and GD32F470.

The entire GD32F4 family uses the same flash layout, and individual chips only differ in the amount of flash.  Re-use the stm32f4 routines like the gd32f1 family.

I still need to:
* [ ] ~Implement dual bank mode for 1MB chips~ I cannot test this on my chip, it only works on the 1MB ones
* [x] Implement option bytes
* [ ] 


src/target/stm32f4.c
```diff
@@ -242,6 +245,47 @@ bool stm32f4_probe(target *t)
...
+
+	/* Third MB composed of 4 256 KB sectors, and uses sector values 12-15 */
+	stm32f4_add_flash(t, 0x8200000, 0x100000,  0x40000,  12, split);
+	return true;
+}
+
+
+bool gd32f4_probe(target *t)
+{
+	switch (t->part_id) {
```
dragonmux, line 277: It would be better if this was written as `if (t->part_id != ID_GD32F470)` with an early return false. If a switch-case is required later, we would structure it a bit differently to maintain the early return so the code is kept toward the left margin

src/target/target_probe.h
```diff
@@ -39,6 +39,7 @@ bool rp_rescue_probe(ADIv5_AP_t *ap);
 bool ch32f1_probe(target *t); // will catch all the clones
 bool at32fxx_probe(target *t); // STM32 clones from Artery
 bool gd32f1_probe(target *t);
+bool gd32f4_probe(target *t);
```
dragonmux, line 42: This change must be mirrored into target_probe.c.

src/target/stm32f4.c
```diff
@@ -242,6 +245,47 @@ bool stm32f4_probe(target *t)
...
+{
+	if (!cortexm_attach(t))
+		return false;
+
+	target_mem_map_free(t);
+	target_add_ram(t, 0x10000000, 0x10000); /* 64 k CCM Ram*/
+	target_add_ram(t, 0x20000000, 0x50000);     /* 320 k RAM */
+
+	/* TODO implement DBS mode */
+	const int split = 12;
```
dragonmux, line 258: This isn't really an int - it can't be negative, please use a more appropriate type such as `uint8_t`. This will also bring things in line with what we've done on the fix/clang-format-cleanup branch for this file.

src/target/stm32f4.c
```diff
@@ -242,6 +245,47 @@ bool stm32f4_probe(target *t)
 	return false;
 }
 
+static bool gd32f4_attach(target *t)
```
dragonmux, line 248: Given this is a fixed map, rather than creating churn on attach it might be better to construct the map once during probe and let attach be handled normally by `cortexm_attach()` directly. This isn't a blocker if you think keeping the map building code in its own attach would be better somehow.

dragonmux (changes requested):
Thank you for the contribution, this is looking really good! There are a couple of small things that need addressing and one thing that's optional to do with code layout, but otherwise we're happy with this change and once the review items are addressed, we'll merge.

dragonmux (dismissed):
LGTM! We'll merge this when the v1.10 cycle begins which should be soon as the mop-up for v1.9 is starting to draw to a close

via:
Thank you! I've got the option bytes working as another commit, do you want it pushed to this branch or left for later?  It is fundamentally identical to a stm32f427, though the registers are named differently (OBCTL0 and OBCTL1).

dragonmux:
We'll update the title of this PR to just "Implement proper support for GD32F4 microcontrollers" then - it can live on this PR.

src/target/stm32f4.c
```diff
@@ -615,7 +616,7 @@ static bool stm32f4_option_write(target *t, uint32_t *val, int count)
 	/* WRITE option bytes instruction */
 	if (((t->part_id == ID_STM32F42X) || (t->part_id == ID_STM32F46X) ||
 		 (t->part_id == ID_STM32F72X) || (t->part_id == ID_STM32F74X) ||
-		 (t->part_id == ID_STM32F76X)) && (count > 1))
+		 (t->part_id == ID_STM32F76X) || (t->part_id == ID_GD32F470)) && (count > 1))
```
dragonmux, line 619: Just a heads-up that, when fix/clang-format-cleanup lands, this change will break and will need to go in the newly introduced helper function `stm32f4_opt_bytes_for` - this seems fine otherwise.

It's not your fault, but a note for future contributions, the parentheses around the comparison are considered harmful as they suppress a useful GCC warning about accidental assignment instead of comparison. We've stripped them on the clean up branch but we can see you're following the existing style here. We've been slowly transitioning the code to not including the extra parentheses for this reason and to improve clarity.

via: That all sounds reasonable to me -- would it make sense for me to rebase this on that branch?

dragonmux: Given we've got to pull that branch apart into 4-5 for PR'ing and the code changes have yet to be reviewed, we'd hold off. All the changes in question will land in 1.9 or early 1.10 at latest, so we'll try to remember to ping you on this PR once the breaking changes for this have landed.

dragonmux: The breaking changes for this have landed so if you could please rebase on main, we will look at getting this merged ASAP.

via:
I rebased, and squashed in some other work I had to support the GD32F450 as well, since I got some of those since I made this PR originally.  I re-tested basic gdb functionality and flashing on both.

src/target/stm32f4.c
```diff
@@ -123,6 +123,7 @@ typedef struct stm32f4_priv {
 	uint32_t dbgmcu_cr;
 } stm32f4_priv_s;
 
+
```
dragonmux, line 126: Please remove the extra space here

src/target/stm32f4.c
```diff
@@ -239,6 +246,39 @@ bool stm32f4_probe(target_s *t)
 	return false;
 }
 
+bool gd32f4_probe(target_s *t)
+{
+	if ((t->part_id != ID_GD32F450) && (t->part_id != ID_GD32F470))
```
dragonmux, line 251: The inner braces are redundant, please remove them as in `if (t->part_id != ID_GD32F450 && t->part_id != ID_GD32F470)` as they suppress a GCC warning we want here

dragonmux (changes requested):
Changeset is looking good, the only things spotted are stylistic requirements for the code base (Several of them enforced by the lint checks). With them fixed, we're happy to merge this

src/target/stm32f4.c
```diff
@@ -137,6 +138,8 @@ typedef struct stm32f4_priv {
 #define ID_STM32F72X  0x452U
 #define ID_STM32F410  0x458U
 #define ID_STM32F413  0x463U
+#define ID_GD32F450   0x2B3U
```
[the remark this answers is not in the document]
dragonmux: Your fixup commit caught the second of these lines, but not this first line, hence the lint check still failing.

If you could fix that and `git rebase main --autosquash` this PR, then we can approve and merge this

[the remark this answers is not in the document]
via: Done!  Is there a way I can trigger the lint checks?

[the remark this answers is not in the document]
dragonmux: The lints are Python scripts you can clone the repo for and run, but the PR-automated run of them is something we have to trigger for first-time contributions - the next time you contribute, it'll happen automatically.

dragonmux (dismissed):
LGTM, merging when the checks complete

dragonmux:
Ah, could we get you to run clang-format on stm32f4.c please as something is still introducing formatting errors, hence the formatting lint failing

via:
Fwiw, when I run clang-format with 12.0.1 I get:
```clang-format -i src/target/stm32f4.c 
/home/via/dev/blackmagic/.clang-format:44:1: error: unknown key 'SeparateDefinitionBlocks'
SeparateDefinitionBlocks: Always
```

I commented out that line and squashed in the result of running it, I had an extra semicolon and other whitespace changes.

dragonmux (approved):

PR #1274, opened 2022-10-25
rtt_if: fix build when libopencm3 is not used

xobs:
## Detailed description

With the inclusion of a89689f3cbd32661375054f2f54669f4fcae1122, `libopencm3/usb/usbd.h` is now unconditionally included on all non-hosted builds. This causes the build to fail if libopencm3 is not present, which occurs on platforms where this library is not used.

As a separate issue, the ESP32 build uses ccache which appears to ignore the CMake `add_definitions()` directive, and thus does not get the `NO_LIBOPENCM3=1` parameter during ccache generation.

This patch makes two minor changes to `rtt_if.h`:

1. It gates the inclusion of `libopencm3/usb/usbd.h` by both `PC_HOSTED=0` and `NO_LIBOPENCM3=1`, meaning both of these must be true in order to include that file
2. `rtt_if.h` now explicitly includes `platform.h`, which fixes the ccache issue as well as ensuring that `RTT_DOWN_BUF_SIZE` and `RTT_UP_BUF_SIZE` come from `platform.h`

### DIFF
--- a/src/include/rtt_if.h
+++ b/src/include/rtt_if.h
@@ -29,9 +29,12 @@
 
 /* rtt i/o to terminal */
 
+/* Define NO_LIBOPENCM3 either in "platform.h" or on the command line
+ * in order to avoid including "libopencm3/usb/usbd.h".
+ */
 #include "platform.h"
 
-#if PC_HOSTED == 0 && NO_LIBOPENCM3 != 1
+#if PC_HOSTED == 0 && !defined(NO_LIBOPENCM3)
 #include <libopencm3/usb/usbd.h>
 
 /* usb rx callback */
--- a/src/include/rtt_if.h
+++ b/src/include/rtt_if.h
@@ -29,6 +29,8 @@
 
 /* rtt i/o to terminal */
 
+#include "platform.h"
+
 #if PC_HOSTED == 0 && NO_LIBOPENCM3 != 1
 #include <libopencm3/usb/usbd.h>
 
--- a/src/include/rtt_if.h
+++ b/src/include/rtt_if.h
@@ -29,7 +29,7 @@
 
 /* rtt i/o to terminal */
 
-#if PC_HOSTED == 0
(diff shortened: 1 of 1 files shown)


src/include/rtt_if.h
```diff
@@ -29,7 +29,9 @@
 
 /* rtt i/o to terminal */
 
-#if PC_HOSTED == 0
+#include "platform.h"
+
+#if PC_HOSTED == 0 && NO_LIBOPENCM3 != 1
```
dragonmux, line 34: If memory serves, NO_LIBOPENCM3 is not defined if we're using locm3, so could this be done as `#if PC_HOSTED == 0 && !defined(NO_LIBOPENCM3)`?

dragonmux (changes requested):
Appreciate the report. This change looks good as it is, though we'd like to explore one slight more robust option before we just merge. See the review comment below

src/include/rtt_if.h
xobs: Perhaps this is not a good approach, because it's not actually defined anywhere. It only exists in the `Makefile`, and is never passed as a build flag.

I do define a flag in Farpatch, but perhaps that approach is too platform-specific.

I'm open to suggestions for other approaches. For example, defining a dummy `struct _usbd_device;` and then having the function accept a `struct _usbd_device *` (which never actually gets defined).

dragonmux: A comment in the header would be perfectly acceptable. BMP itself does not `-DNO_LIBOPENCM3` so hence asking to use the `defined()` check rather than an integral value check.

Good question re ccache, unfortunately we don't know enough about the tool to make any suggestions on that score

xobs: Alright, I've modified it and verified that it still builds for the three targets in question.

dragonmux (approved):
LGTM! Thank you :) Merging

PR #1284, opened 2022-11-13
gdb_main: move main loop into `main()`

xobs:
## Detailed description

This is a rewrite of the main loop as suggested in https://github.com/blackmagic-debug/blackmagic/pull/1281. Rather
than constantly sitting inside `gdb_main()`, the program's main loop does its own polling of the target. This enables
a host to do its own waiting without needing to insert hooks in various functions.

This prevents WDT triggers on ESP32 where RTT blocks the IDLE thread from running. The IDLE thread is responsible for resetting the WDT. It can also be used inside the main loop to regulate hosts such as `PC_HOSTED` where constant USB polling is undesirable.

### DIFF
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -115,7 +115,7 @@ target_controller_s gdb_controller = {
 
 /* execute gdb remote command stored in 'pbuf'. returns immediately, no busy waiting. */
 
-int gdb_main_loop(target_controller_s *tc, char *pbuf, size_t size, bool in_syscall)
+int gdb_main_loop(target_controller_s *tc, char *pbuf, size_t pbuf_size, size_t size, bool in_syscall)
 {
 	bool single_step = false;
 
@@ -134,7 +134,7 @@ int gdb_main_loop(target_controller_s *tc, char *pbuf, size_t size, bool in_sysc
 			uint32_t addr, len;
 			ERROR_IF_NO_TARGET();
 			sscanf(pbuf, "m%" SCNx32 ",%" SCNx32, &addr, &len);
-			if (len > sizeof(pbuf) / 2U) {
+			if (len > pbuf_size / 2U) {
 				gdb_putpacketz("E02");
 				break;
 			}
@@ -662,9 +662,9 @@ static void handle_z_packet(char *packet, const size_t plen)
 		gdb_putpacketz("OK");
 }
 
-void gdb_main(char *pbuf, size_t size)
+void gdb_main(char *pbuf, size_t pbuf_size, size_t size)
 {
-	gdb_main_loop(&gdb_controller, pbuf, size, false);
+	gdb_main_loop(&gdb_controller, pbuf, pbuf_size, size, false);
 }
 
 /* halt target */
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -29,7 +29,6 @@
 #include "hex_utils.h"
 #include "gdb_if.h"
 #include "gdb_packet.h"
-#include "gdb_main.h"
 #include "gdb_hostio.h"
 #include "target.h"
 #include "command.h"
@@ -53,7 +52,7 @@ typedef enum gdb_signal {
 	GDB_SIGLOST = 29,
 } gdb_signal_e;
 
-#define BUF_SIZE 1024U
+#define GDB_MAX_PACKET_SIZE 1024U
 
 #define ERROR_IF_NO_TARGET()   \
 	if (!cur_target) {         \
@@ -66,10 +65,9 @@ typedef struct cmd_executer {
 	void (*func)(const char *packet, size_t len);
 } cmd_executer_s;
 
-static char pbuf[BUF_SIZE + 1U];
-
-static target_s *cur_target;
-static target_s *last_target;
+target_s *cur_target;
+target_s *last_target;
+bool gdb_target_running = false;
 static bool gdb_needs_detach_notify = false;
 
 static void handle_q_packet(char *packet, size_t len);
@@ -97,7 +95,7 @@ static void gdb_target_printf(target_controller_s *tc, const char *fmt, va_list 
 	gdb_voutf(fmt, ap);
 }
 
-static target_controller_s gdb_controller = {
+target_controller_s gdb_controller = {
 	.destroy_callback = gdb_target_destroy_callback,
 	.printf = gdb_target_printf,
 
@@ -115,18 +113,14 @@ static target_controller_s gdb_controller = {
 	.system = hostio_system,
 };
 
-int gdb_main_loop(target_controller_s *tc, bool in_syscall)
+/* execute gdb remote command stored in 'pbuf'. returns immediately, no busy waiting. */
+
+int gdb_main_loop(target_controller_s *tc, char *pbuf, size_t size, bool in_syscall)
 {
 	bool single_step = false;
 
 	/* GDB protocol main loop */
-	while (1) {
-		SET_IDLE_STATE(1);
-		size_t size = gdb_getpacket(pbuf, BUF_SIZE);
-		// If port closed and target detached, stay idle
-		if (pbuf[0] != '\x04' || cur_target) {
-			SET_IDLE_STATE(0);
-		}
+	{
 		switch (pbuf[0]) {
 		/* Implementation of these is mandatory! */
 		case 'g': { /* 'g': Read general registers */
@@ -211,52 +205,10 @@ int gdb_main_loop(target_controller_s *tc, bool in_syscall)
 			 * This packet isn't documented as being mandatory,
 			 * but GDB doesn't work without it.
 			 */
-
-			if (!cur_target) {
-				/* Report "target exited" if no target */
-				gdb_putpacketz("W00");
-				break;
-			}
-
-			/* Wait for target halt */
-			target_addr_t watch;
-			target_halt_reason_e reason;
-			while (true) {
-				reason = target_halt_poll(cur_target, &watch);
-				if (reason != TARGET_HALT_RUNNING)
-					break;
-
-				const char gdb_ch = (char)gdb_if_getchar_to(0);
-				if (gdb_ch == '\x03' || gdb_ch == '\x04')
-					target_halt_request(cur_target);
-
-				platform_pace_poll();
-
-#ifdef ENABLE_RTT
-				if (rtt_enabled)
-					poll_rtt(cur_target);
-#endif
-			}
-			SET_RUN_STATE(false);
-
-			/* Translate reason to GDB signal */
-			switch (reason) {
-			case TARGET_HALT_ERROR:
-				gdb_putpacket_f("X%02X", GDB_SIGLOST);
-				morse("TARGET LOST.", true);
-				break;
-			case TARGET_HALT_REQUEST:
-				gdb_putpacket_f("T%02X", GDB_SIGINT);
-				break;
-			case TARGET_HALT_WATCHPOINT:
-				gdb_putpacket_f("T%02Xwatch:%08X;", GDB_SIGTRAP, watch);
-				break;
-			case TARGET_HALT_FAULT:
-				gdb_putpacket_f("T%02X", GDB_SIGSEGV);
-				break;
-			default:
-				gdb_putpacket_f("T%02X", GDB_SIGTRAP);
-			}
+			if (cur_target)
+				gdb_target_running = true;
+			else
+				gdb_putpacketz("W00"); /* target exited */
 			break;
 		}
 
@@ -374,6 +326,7 @@ int gdb_main_loop(target_controller_s *tc, bool in_syscall)
 			gdb_putpacketz("");
 		}
 	}
+	return 0;
 }
 
 static bool exec_command(char *packet, const size_t length, const cmd_executer_s *exec)
@@ -408,6 +361,7 @@ static void exec_q_rcmd(const char *packet, const size_t length)
 	else {
 		const char *const response = "Failed\n";
 		const size_t length = strlen(response);
+		char pbuf[length * 2 + 1];
 		gdb_putpacket(hexify(pbuf, response, length), 2 * length);
 	}
 }
@@ -440,7 +394,7 @@ static void exec_q_supported(const char *packet, const size_t length)
 {
 	(void)packet;
 	(void)length;
-	gdb_putpacket_f("PacketSize=%X;qXfer:memory-map:read+;qXfer:features:read+", BUF_SIZE);
+	gdb_putpacket_f("PacketSize=%X;qXfer:memory-map:read+;qXfer:features:read+", GDB_MAX_PACKET_SIZE);
 }
 
 static void exec_q_memory_map(const char *packet, const size_t length)
@@ -708,7 +662,58 @@ static void handle_z_packet(char *packet, const size_t plen)
 		gdb_putpacketz("OK");
 }
 
-void gdb_main(void)
+void gdb_main(char *pbuf, size_t size)
 {
-	gdb_main_loop(&gdb_controller, false);
+	gdb_main_loop(&gdb_controller, pbuf, size, false);
+}
+
+/* halt target */
+void gdb_halt_target(void)
+{
+	if (cur_target)
+		target_halt_request(cur_target);
+	else
+		/* Report "target exited" if no target */
+		gdb_putpacketz("W00");
+}
+
+/* poll running target */
+void gdb_poll_target(void)
+{
+	target_addr_t watch;
+	enum target_halt_reason reason;
+	if (!cur_target) {
+		/* Report "target exited" if no target */
+		gdb_putpacketz("W00");
+		return;
+	}
+
+	/* poll target */
+	reason = target_halt_poll(cur_target, &watch);
+	if (!reason)
+		return;
+
+	/* switch polling off */
+	gdb_target_running = false;
+	SET_RUN_STATE(0);
+
+	/* Translate reason to GDB signal */
+	switch (reason) {
+	case TARGET_HALT_ERROR:
+		gdb_putpacket_f("X%02X", GDB_SIGLOST);
+		morse("TARGET LOST.", true);
+		break;
+	case TARGET_HALT_REQUEST:
+		gdb_putpacket_f("T%02X", GDB_SIGINT);
+		break;
+	case TARGET_HALT_WATCHPOINT:
+		gdb_putpacket_f("T%02Xwatch:%08X;", GDB_SIGTRAP, watch);
+		break;
+	case TARGET_HALT_FAULT:
+		gdb_putpacket_f("T%02X", GDB_SIGSEGV);
+		break;
+	default:
+		gdb_putpacket_f("T%02X", GDB_SIGTRAP);
+	}
+	return;
 }
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -695,7 +695,7 @@ void gdb_poll_target(void)
 
 	/* poll target */
 	target_addr_t watch;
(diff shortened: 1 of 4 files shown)


xobs:
This has not been tested -- I have only confirmed that it builds, so there may be issues.

I've created this PR to get feedback, to ensure this is the approach you were describing in https://github.com/blackmagic-debug/blackmagic/pull/1281.

koendv:
I've tried rtt on black pill v2 and hosted on arm64, seems ok. Flashing works, speed unchanged.
OK. Does it run on esp32? What's your opinion of the way the code looks now?

xobs:
It seems to work fine. I started on a refactor of Farpatch to support multiple clients attaching at once, and nearly got something working. I backed out those changes for another day, but I can confirm that this approach allows me to insert delays as necessary. Overall, I think this approach is sound.

I am, however, concerned with the fact that RP2040 debugging appears to have broken slightly. Though if you're not seeing it on your target, then perhaps it's either something on my platform or something due to flash changes recently.

koendv:
Glad to hear it works fine. 

I think this also opens up the possibility of running bmp on a real-time
os.

Another possibility is running bmp and cmsis-dap on the same probe.
bmp on usb serial together with free-dap on usb hid works; the
equivalent for esp32 would be bmp on a tcp port and free-dap on
bluetooth hid. The advantage is changing between openocd and bmp
without having to change cables.

"slightly"? Could you be more specific about what you see on rp2040?

koen

xobs:
I just checked, and it seems that RP2040 support is just weird even on `main`. For example:

```
[4:47:17 pm] E:/Code/Farpatch/farpatch> & 'E:\Software\gcc\6 2017-q2-update\bin\arm-none-eabi-gdb.exe' mdi-gateway.elf
GNU gdb (GNU Tools for ARM Embedded Processors 6-2017-q2-update) 7.12.1.20170417-git
Copyright (C) 2017 Free Software Foundation, Inc.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "--host=i686-w64-mingw32 --target=arm-none-eabi".
<http://www.gnu.org/software/gdb/bugs/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from mdi-gateway.elf...done.
(gdb) tar ext \\.\COM54
Remote debugging using \\.\COM54
(gdb) mon s
Target voltage: 3.3V
Available Targets:
No. Att Driver
 1      Raspberry RP2040 M0+
 2      Raspberry RP2040 M0+
 3      Raspberry RP2040 Rescue (Attach to reset!)
(gdb) att 1
Attaching to program: E:\Code\Ion\mdi-gateway.elf, Remote target
0x10002340 in tud_suspended () at /opt/Pico/pico-sdk/lib/tinyusb/src/device/usbd.c:372
372     /opt/Pico/pico-sdk/lib/tinyusb/src/device/usbd.c: No such file or directory.
(gdb) load
Loading section .boot2, size 0x100 lma 0x10000000
Loading section .text, size 0x61e8 lma 0x10000100
Loading section .rodata, size 0x8b8 lma 0x100062e8
Loading section .binary_info, size 0x24 lma 0x10006ba0
Loading section .data, size 0x2a8 lma 0x10006bc4
Start address 0x100001e8, load size 28268
Transfer rate: 47 KB/sec, 883 bytes/write.
(gdb) c
Continuing.

Program received signal SIGINT, Interrupt.
0x102f29d0 in ?? ()
(gdb) run
The program being debugged has been started already.
Start it from the beginning? (y or n) y
Starting program: mdi-gateway.elf

Program received signal SIGINT, Interrupt.
0x000020f0 in ?? ()
(gdb) bt
#0  0x000020f0 in ?? ()
#1  0x000002b2 in ?? ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
(gdb) run
The program being debugged has been started already.
Start it from the beginning? (y or n) y
Starting program: mdi-gateway.elf

Program received signal SIGINT, Interrupt.
0x000020f0 in ?? ()
(gdb) bt
#0  0x000020f0 in ?? ()
#1  0x000002b2 in ?? ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
(gdb) b main
Breakpoint 1 at 0x100004b0: file main.c, line 59.
(gdb) run
The program being debugged has been started already.
Start it from the beginning? (y or n) y
Starting program: mdi-gateway.elf
Note: automatically using hardware breakpoints for read-only addresses.

Program received signal SIGINT, Interrupt.
0x000020f0 in ?? ()
(gdb) x 0x10002000
0x10002000 <_hw_endpoint_buffer_control_update32+56>:   0x00000000
(gdb) disassemble main
Dump of assembler code for function main:
   0x100004b0 <+0>:     movs    r0, r0
   0x100004b2 <+2>:     movs    r0, r0
   0x100004b4 <+4>:     movs    r0, r0
   0x100004b6 <+6>:     movs    r0, r0
   0x100004b8 <+8>:     movs    r0, r0
   0x100004ba <+10>:    movs    r0, r0
   0x100004bc <+12>:    movs    r0, r0
   0x100004be <+14>:    movs    r0, r0
   0x100004c0 <+16>:    movs    r0, r0
   0x100004c2 <+18>:    movs    r0, r0
   0x100004c4 <+20>:    movs    r0, r0
   0x100004c6 <+22>:    movs    r0, r0
   0x100004c8 <+24>:    movs    r0, r0
   0x100004ca <+26>:    movs    r0, r0
   0x100004cc <+28>:    movs    r0, r0
   0x100004ce <+30>:    movs    r0, r0
   0x100004d0 <+32>:    movs    r0, r0
   0x100004d2 <+34>:    movs    r0, r0
   0x100004d4 <+36>:    movs    r0, r0
   0x100004d6 <+38>:    movs    r0, r0
   0x100004d8 <+40>:    movs    r0, r0
   0x100004da <+42>:    movs    r0, r0
   0x100004dc <+44>:    movs    r0, r0
   0x100004de <+46>:    movs    r0, r0
   0x100004e0 <+48>:    movs    r0, r0
   0x100004e2 <+50>:    movs    r0, r0
   0x100004e4 <+52>:    movs    r0, r0
   0x100004e6 <+54>:    movs    r0, r0
   0x100004e8 <+56>:    movs    r0, r0
   0x100004ea <+58>:    movs    r0, r0
   0x100004ec <+60>:    movs    r0, r0
   0x100004ee <+62>:    movs    r0, r0
   0x100004f0 <+64>:    movs    r0, r0
   0x100004f2 <+66>:    movs    r0, r0
   0x100004f4 <+68>:    movs    r0, r0
   0x100004f6 <+70>:    movs    r0, r0
   0x100004f8 <+72>:    movs    r0, r0
   0x100004fa <+74>:    movs    r0, r0
   0x100004fc <+76>:    movs    r0, r0
   0x100004fe <+78>:    movs    r0, r0
   0x10000500 <+80>:    movs    r0, r0
   0x10000502 <+82>:    movs    r0, r0
   0x10000504 <+84>:    movs    r0, r0
   0x10000506 <+86>:    movs    r0, r0
   0x10000508 <+88>:    movs    r0, r0
   0x1000050a <+90>:    movs    r0, r0
   0x1000050c <+92>:    movs    r0, r0
   0x1000050e <+94>:    movs    r0, r0
   0x10000510 <+96>:    movs    r0, r0
   0x10000512 <+98>:    movs    r0, r0
End of assembler dump.
(gdb) p main
$1 = {int (void)} 0x100004b0 <main>
(gdb) x main
0x100004b0 <main>:      0x00000000
(gdb) load
Loading section .boot2, size 0x100 lma 0x10000000
Loading section .text, size 0x61e8 lma 0x10000100
Loading section .rodata, size 0x8b8 lma 0x100062e8
Loading section .binary_info, size 0x24 lma 0x10006ba0
Loading section .data, size 0x2a8 lma 0x10006bc4
Start address 0x100001e8, load size 28268
Transfer rate: 106 KB/sec, 883 bytes/write.
(gdb) p main
$2 = {int (void)} 0x100004b0 <main>
(gdb) x 0x100004b0
0x100004b0 <main>:      0x00000000
(gdb) mon erase_mass
Erasing device Flash: done
(gdb)
```

From this point, the board has stopped responding. Examining memory returns all zeroes. Doing an `erase_mass` returns immediately rather than performing a mass erase. The board doesn't actually work unless I hard power-cycle it.

It's just coincidence that it's the board I was testing my patches with. Perhaps it's damaged. At any rate, the STM32G49 test board I've got works just fine under both JTAG and SWD, so I'd consider the rpi weirdness a separate issue.

dragonmux:
That behaviour indicates issues with getting the RP2040 to run its exit-XIP and enter-XIP sequences (rp_flash_prepare, rp_flash_resume). It's possible this is a bug in your RP2040's ROM routines, or it could indicate an issue with BMP's own implementations of those same routines..

There are two `#if 0` blocks in the RP2040 support routines guarding `rp_flash_*` routines, if you remove the preprocessor magic, and replace the prepare and resume routines' calls to the ROM table with:
```c
rp_flash_connect_internal(t);
rp_flash_exit_xip(t);
```
and
```c
rp_flash_flush_cache(t);
rp_flash_enter_xip(t);
```

for prepare and resume respectively.. please let us know how that goes in a follow-up issue report

xobs:
I think this is what you were looking for:

```patch
diff --git a/src/target/rp.c b/src/target/rp.c
index c27a0a58..06b8ec6e 100644
--- a/src/target/rp.c
+++ b/src/target/rp.c
@@ -207,7 +207,7 @@ static bool rp_mass_erase(target *t);
 // Our own implementation of bootloader functions for handling flash chip
 static void rp_flash_exit_xip(target *const t);
 static void rp_flash_enter_xip(target *const t);
-#if 0
+#if 1
 static void rp_flash_connect_internal(target *const t);
 static void rp_flash_flush_cache(target *const t);
 #endif
@@ -225,6 +225,7 @@ static void rp_add_flash(target *t)
                return;
        }

+       rp_flash_connect_internal(t);
        rp_flash_exit_xip(t);

        spi_parameters_s spi_parameters;
@@ -236,6 +237,7 @@ static void rp_add_flash(target *t)
                spi_parameters.sector_erase_opcode = SPI_FLASH_CMD_SECTOR_ERASE;
        }

+       rp_flash_flush_cache(t);
        rp_flash_enter_xip(t);

        DEBUG_INFO("Flash size: %uMiB\n", spi_parameters.capacity / (1024U * 1024U));
@@ -612,7 +614,7 @@ static void rp_spi_read(
        target_mem_write32(t, RP_SSI_ENABLE, ssi_enabled);
 }

-#if 0
+#if 1
 // Connect the XIP controller to the flash pads
 static void rp_flash_connect_internal(target *const t)
 {
@@ -769,7 +771,7 @@ static void rp_flash_exit_xip(target *const t)
        target_mem_write32(t, RP_GPIO_QSPI_CS_CTRL, 0);
 }

-#if 0
+#if 1
 // This is a hook for steps to be taken in between programming the flash and
 // doing cached XIP reads from the flash. Called by the bootrom before
 // entering flash second stage, and called by the debugger after flash
```
However, it's still behaving Oddly. Since this is orthogonal to this patch, I can move this into another issue and we can discuss it there.

Is there anything else that needs to happen in order to merge this change?

dragonmux:
That is indeed, please do move that into a new issue ticket. A trace of what you're seeing would be fantastic to help debug but otherwise we mostly need to recreate your situation to see what's going on

xobs:
I opened https://github.com/blackmagic-debug/blackmagic/issues/1295 which is at least rather reproducible

dragonmux (changes requested):
There are a couple of times we found on a cursory initial code review. We'll dive deeper once the cleanup work is all merged

src/main.c
```diff
@@ -29,22 +29,53 @@
...
+#ifdef ENABLE_RTT
+				if (rtt_enabled)
+					poll_rtt(cur_target);
+#endif
+			}
+
+			SET_IDLE_STATE(1);
+			size_t size = gdb_getpacket(pbuf, BUF_SIZE);
+			// If port closed and target detached, stay idle
+			if ((pbuf[0] != 0x04) || cur_target) {
```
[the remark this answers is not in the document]
xobs: The braces appear necessary here, because on hosted `SET_IDLE_STATE(x)` is defined as an empty macro.

If you like, I can redefine `SET_IDLE_STATE(x)` to be `do {} while(0)` on hosted, which would get around that error...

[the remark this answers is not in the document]
xobs: I went ahead and added that in c86e7964304845fc062749b94b3d3e9bc3ba0467 which fixes this on `hosted`.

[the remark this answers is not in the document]
dragonmux: ah, right.. that mess - yes, good call to turn that into a `do {} while(0)`

xobs:
Rebased to latest `main` to keep it mergable.

xobs:
Rebased to merge latest changes on `main`

xobs:
Rebased to fix conflicts with latest `main`

xobs:
Rebased again to fix conflicts with latest `main`

dragonmux:
We will try to get this reviewed today now the v1.10 merge window is open

src/gdb_main.c
```diff
@@ -115,18 +113,14 @@ static target_controller_s gdb_controller = {
...
 
 	/* GDB protocol main loop */
-	while (1) {
-		SET_IDLE_STATE(1);
-		size_t size = gdb_getpacket(pbuf, BUF_SIZE);
-		// If port closed and target detached, stay idle
-		if (pbuf[0] != '\x04' || cur_target) {
-			SET_IDLE_STATE(0);
-		}
+	{
```
dragonmux, line 123: With the removal of the loop, we'd remove the comment as it's inaccurate and de-indent the entire block

src/gdb_main.c
```diff
@@ -210,52 +204,10 @@ int gdb_main_loop(target_controller_s *tc, bool in_syscall)
...
-			case TARGET_HALT_WATCHPOINT:
-				gdb_putpacket_f("T%02Xwatch:%08X;", GDB_SIGTRAP, watch);
-				break;
-			case TARGET_HALT_FAULT:
-				gdb_putpacket_f("T%02X", GDB_SIGSEGV);
-				break;
-			default:
-				gdb_putpacket_f("T%02X", GDB_SIGTRAP);
-			}
+			if (cur_target)
```
dragonmux, line 207: Does this not need to call the new gdb_poll_target() before sending back a `W00` or no-response? If not, please add documentation as to why this otherwise disjointed-seeming approach works and what the new logic is

src/gdb_main.c
```diff
@@ -707,7 +661,58 @@ static void handle_z_packet(char *packet, const size_t plen)
...
+	else
+		/* Report "target exited" if no target */
+		gdb_putpacketz("W00");
+}
+
+/* poll running target */
+void gdb_poll_target(void)
 {
-	gdb_main_loop(&gdb_controller, false);
+	target_addr_t watch;
```
dragonmux, line 682: Please don't use C89 style declarations - having the if statement first, then declaring watch and declare-initialising reason helps the compiler and removes a bug possibility.

src/gdb_main.c
```diff
@@ -707,7 +661,58 @@ static void handle_z_packet(char *packet, const size_t plen)
...
+	case TARGET_HALT_WATCHPOINT:
+		gdb_putpacket_f("T%02Xwatch:%08X;", GDB_SIGTRAP, watch);
+		break;
+	case TARGET_HALT_FAULT:
+		gdb_putpacket_f("T%02X", GDB_SIGSEGV);
+		break;
+	default:
+		gdb_putpacket_f("T%02X", GDB_SIGTRAP);
+	}
+	return;
```
dragonmux, line 717: tautological statement - please remove per clang-tidy lints.

src/main.c
```diff
@@ -42,7 +48,31 @@ int main(int argc, char **argv)
 	while (true) {
 		volatile exception_s e;
 		TRY_CATCH (e, EXCEPTION_ALL) {
-			gdb_main();
+			SET_IDLE_STATE(false);
+			while (gdb_target_running && cur_target) {
```
dragonmux, line 52: Is it possible to flatten this out a bit by making this while loop the outer while loop?

src/main.c
```diff
@@ -29,22 +29,53 @@
...
-		TRY_CATCH(e, EXCEPTION_ALL) {
-			gdb_main();
+		TRY_CATCH (e, EXCEPTION_ALL) {
+			SET_IDLE_STATE(0);
+			while (gdb_target_running && cur_target) {
+				gdb_poll_target();
+
+				// Check again, as `gdb_poll_target()` may
+				// alter these variables.
+				if (!gdb_target_running || !cur_target) {
```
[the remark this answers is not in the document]
dragonmux: These braces (`{`, `}`) are still redundant, please remove them.

src/main.c
```diff
@@ -42,7 +48,31 @@ int main(int argc, char **argv)
...
+			SET_IDLE_STATE(false);
+			while (gdb_target_running && cur_target) {
+				gdb_poll_target();
+
+				// Check again, as `gdb_poll_target()` may
+				// alter these variables.
+				if (!gdb_target_running || !cur_target) {
+					break;
+				}
+				char c = (char)gdb_if_getchar_to(0);
```
dragonmux, line 60: This cast should be removed - it should do nothing now and was a bug that it was ever required.

dragonmux (changes requested):
There are a couple of stylistic/formatting things to resolve, and we'd like an explanation on one or two things the change set is doing so we can better understand it, but this is shaping up nicely. With the stylistic/formatting items resolved this looks ready to merge.

xobs:
I've made the requested changes, except I'm not sure what you mean by flattening out that bit.

src/gdb_main.c
```diff
@@ -707,7 +668,56 @@ static void handle_z_packet(char *packet, const size_t plen)
...
+{
+	if (!cur_target) {
+		/* Report "target exited" if no target */
+		gdb_putpacketz("W00");
+		return;
+	}
+
+	/* poll target */
+	target_addr_t watch;
+	enum target_halt_reason reason = target_halt_poll(cur_target, &watch);
```
dragonmux, line 697: `target_halt_reason_e` please - the enum is typedef'd for this reason

src/main.c
```diff
@@ -42,7 +48,30 @@ int main(int argc, char **argv)
...
+#ifdef ENABLE_RTT
+				if (rtt_enabled)
+					poll_rtt(cur_target);
+#endif
+			}
+
+			SET_IDLE_STATE(true);
+			size_t size = gdb_getpacket(pbuf, BUF_SIZE);
+			// If port closed and target detached, stay idle
+			if (pbuf[0] != 0x04 || cur_target)
```
dragonmux, line 72: `pbuf[0]` is a char, so this should compare `'\x04'` for type consistency please

dragonmux (changes requested):
Think this should be the last round of changes. Next review should be approve-to-merge.

To explain what we meant about flattening the main loop out a bit: currently there's a while loop inside which there's a "try-catch", inside which another while loop - this is a lot of levels of indentation, and especially as the outer most is `while true`.

What we were thinking is if it's possible to push the inner while condition out to the outer while loop, making the inner while body a part of the try-catch scope instead, removing a layer indentation. If not, it might be worth extracting the inner while loop into its own function to reduce the indentation level and improve readability instead?

xobs:
Alright, both of those requested changes have been made. I can't figure out how to de-indent those without adding a goto, so I opted to pull it into its own function instead.

dragonmux (approved):
LGTM, we'll give this a test locally just to make sure nothing broke on native, then merge. Many thanks xobs!

dragonmux:
Testing says this works well on both RP2040 and Tiva-C, so this is ready for merge once the branch conflict has been sorted

dragonmux (approved):

PR #1286, opened 2022-11-14
Feature: pre-commit check hex case

perigoso:
This PR adds a hook to the pre-commit file, said hooks checks for hex number declarations (i.e. `0xffff`) case (as in upper or lower case) across the committed files

~edit: an additional hook to check for accidental assignment has been added~

## Detailed description

It is set to check and fix declarations to the format `0x55aa`, i.e. all lower case

note that this depends on the merge of #1282 as it includes the addition of the pre-commit file, it also should only be commited after the code has been cleaned up (as in the hex number declarations all fixed to consistent format)

@dragonmux or anyone that wants to make use of the hook before it is configured (i.e. this has been merged), you can clone the hook repository and use it as a python script:

`git clone https://github.com/perigoso/pre-commit-hooks.git`

`${path_to_folder}/pre-commit-hooks/hooks/check_hex_case.py -i --lower-digits ${file_to_fixup}`

this will fix all hex numbers in the passed file, you can pass multiple files, that should prove useful for the cleanup

## Your checklist for this pull request

* [x] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md)
* [x] I've read the [guidelines for contributing](https://github.com/blackmagic-debug/blackmagic/blob/main/CONTRIBUTING.md) to this repository
* [x] It builds for hardware native (`make PROBE_HOST=native`)
* [x] It builds as BMDA (`make PROBE_HOST=hosted`)
* [x] I've tested it to the best of my ability
* [x] My commit messages provide a useful short description of what the commits do

PR #1302, opened 2022-11-19
Feature: pre-commit accidental assignment

perigoso:
This PR adds a hook to the pre-commit file, said hooks checks for assignments inside parentheses

## Detailed description

depends on #1286 

### DIFF
--- a/src/rtt.c
+++ b/src/rtt.c
@@ -354,28 +354,36 @@ void poll_rtt(target_s *const cur_target)
 	bool rtt_busy = false;
 
 	if (last_poll_ms + poll_ms <= now || now < last_poll_ms) {
-		target_addr_t watch;
-		target_halt_reason_e reason;
 		bool resume_target = false;
+
 		if (!rtt_found)
 			/* check if target needs to be halted during memory access */
 			rtt_halt = target_mem_access_needs_halt(cur_target);
+
+		target_addr_t watch;
 		if (rtt_halt && target_halt_poll(cur_target, &watch) == TARGET_HALT_RUNNING) {
 			/* briefly halt target during target memory access */
 			target_halt_request(cur_target);
-			while ((reason = target_halt_poll(cur_target, &watch)) == TARGET_HALT_RUNNING)
-				continue;
+
+			target_halt_reason_e reason;
+			do {
+				reason = target_halt_poll(cur_target, &watch);
+			} while (reason == TARGET_HALT_RUNNING);
+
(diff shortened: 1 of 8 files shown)


perigoso:
ah, I forgot I left this one for later

```
if ((cl_opts->opt_serial && strstr(serial_number, cl_opts->opt_serial)) ||
	(cl_opts->opt_position && cl_opts->opt_position == probes_found) ||
	/* Special case for the very first probe found. */
	(probe_identified = false, probes_found == 1)) {
	strncpy(info->serial, serial_number, sizeof info->serial);
	strncpy(info->manufacturer, "BMP", sizeof info->manufacturer);
	snprintf(info->product, sizeof info->product, "%ls", busReportedDeviceSesc);
	/* Don't bother to parse the version string. It is a part of the
	 * product description string. It seems that at the moment it
	 * is only being used to print a version string in response
	 * to the 'monitor version' command, so it doesn't really matter
	 * if the version string is printed as a part of the product string,
	 * or as a separate string, the result is pretty much the same. */
	info->version[0] = 0;
	if (probe_identified)
		return 0;
}
```

I can't change it because I don't fully understand what it's doing

src/rtt.c
```diff
@@ -354,28 +354,36 @@ void poll_rtt(target_s *const cur_target)
...
+
+		target_addr_t watch;
 		if (rtt_halt && target_halt_poll(cur_target, &watch) == TARGET_HALT_RUNNING) {
 			/* briefly halt target during target memory access */
 			target_halt_request(cur_target);
-			while ((reason = target_halt_poll(cur_target, &watch)) == TARGET_HALT_RUNNING)
-				continue;
+
+			target_halt_reason_e reason;
+			do {
```
dragonmux, line 369: would have preferred this to be a while loop w/ `reason` initialised to the truthy loop condition, but this works

dragonmux (approved):
This LGTM! Thank you for this awesome clean-up work and the pre-commit hooks.

Merging.

src/rtt.c
perigoso: Ah, yeah you're right that didn't occur to me

dragonmux: If you want to open a follow-up PR to change this, you'd be welcome to. Otherwise we'll address it when v1.10's cycle begins (assuming we remember)

perigoso: Yes i was thinking of that, I'll do one when i find a bit of dead time today

Issue #1001, opened 2022-03-24
SWD regression on scan targets(SW-DP scan failed!)


zvova7890:
Recently I have updated my bmp(on stm32f1)  board to latest revision and found bad news for me. My project board not recognized anymore. It has ~40cm SWD wire in not best noise environment. I assume that the problem is related to these nuances.
```
Target voltage: 3.22V
SW-DP scan failed!
Failed
```

I'm started bisecting commits and after long time magical behavior I found broken commits.
It is 61efe26348b8d4d3ae01c3b2ac68680a4ef68d89 and 0c6390307115b48e57deeecf0be8f5eca89604ac

61efe26348b8d4d3ae01c3b2ac68680a4ef68d89 - commit(as I'm understand) reduced count of try to detect target. 
0c6390307115b48e57deeecf0be8f5eca89604ac - commit has reduced timeout from 2000 -> 20.

Partially revert noticed commits restore my board recognition, and works fine.

```
diff --git a/src/target/adiv5_jtagdp.c b/src/target/adiv5_jtagdp.c
index cf8591b..0ea4e3d 100644
--- a/src/target/adiv5_jtagdp.c
+++ b/src/target/adiv5_jtagdp.c
@@ -85,7 +85,7 @@ uint32_t fw_adiv5_jtagdp_low_access(ADIv5_DP_t *dp, uint8_t RnW,
 
 	jtag_dev_write_ir(&jtag_proc, dp->dp_jd_index, APnDP ? IR_APACC : IR_DPACC);
 
-	platform_timeout_set(&timeout, 20);
+	platform_timeout_set(&timeout, 2000);
 	do {
 		jtag_dev_shift_dr(&jtag_proc, dp->dp_jd_index, (uint8_t*)&response,
 						  (uint8_t*)&request, 35);
diff --git a/src/target/adiv5_swdp.c b/src/target/adiv5_swdp.c
index 7797b8f..4f8682b 100644
--- a/src/target/adiv5_swdp.c
+++ b/src/target/adiv5_swdp.c
@@ -227,13 +227,15 @@ uint32_t firmware_swdp_low_access(ADIv5_DP_t *dp, uint8_t RnW,
 
 	if((addr & ADIV5_APnDP) && dp->fault) return 0;
 
-	platform_timeout_set(&timeout, 20);
+	platform_timeout_set(&timeout, 2000);
 	do {
 		dp->seq_out(request, 8);
 		ack = dp->seq_in(3);
[...]

PR #1304, opened 2022-11-19
Fix: ADIv5 SWD fault handling

dragonmux:
## Detailed description

This PR addresses a regression caused in #1001. While that PR did fix a real bug, use of dp->error(dp) in the SWD DP access function causes BMP to misbehave and behave badly. In reality this is most likely because an exception is getting throw and not getting caught, but we have no real way to find out and be sure. To solve this, we instead use adiv5_dp_write() directly so we don't go through a read cycle on the control/status DP register (and a series of traversals through the DP access routine) to write the abort DP register (which still requires traversals through the DP access routine, but which ADIv5 guarantees won't fail).

This fixes the behaviour where a user would see
```
Calling lmi_probe
Timeout while waiting for BMP response
remote_ap_mem_read error -4 around 0x400fe000
ap_memread @ 400fe000 len 4: 80 0a 01 00
Timeout while waiting for BMP response
remote_ap_mem_read error -4 around 0x400fe004
ap_memread @ 400fe004 len 4: 80 0a 01 00
Timeout while waiting for BMP response
remote_adiv5_dp_read error -4
Read  CTRL/STAT: 0x00ff001d
Write DP_ABORT : 0x00000002
Timeout while waiting for BMP response
remote_adiv5_low_access error -4
DP Error 0x00000010
```
on the console from BMDA repeatedly and be unable to do any further scans.

### DIFF
--- a/src/platforms/hosted/platform.c
+++ b/src/platforms/hosted/platform.c
@@ -643,41 +643,31 @@ void adiv5_mem_read(adiv5_access_port_s *ap, void *dest, uint32_t src, size_t le
 {
 	ap->dp->mem_read(ap, dest, src, len);
 	if (cl_debuglevel & BMP_DEBUG_TARGET) {
-		fprintf(stderr, "ap_memread @ %" PRIx32 " len %" PRIx32 ":", src, (uint32_t)len);
-		uint8_t *p = (uint8_t *)dest;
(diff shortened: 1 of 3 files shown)


esden (approved):
LGTM

PR #1305, opened 2022-11-20
Fix: STM32W device IDs

dragonmux:
## Detailed description

This PR seeks to address the STM32W part ID regression caused by the improved ADIv5 implementation correctness. This also redresses the work-around at the top of the probe routine for the STM32L4/L5/G4/W parts as the reference manuals for this part series indicates it is unnecessary (though reading the special part ID register is still required for L4 and G4 parts).

Unfortunately we don't have any STM32L55 parts (in particular) to test against, so this is untested, however it is based on the prior work by Uwe to address this same problem so should be OK.

### DIFF
--- a/src/target/stm32l4.c
+++ b/src/target/stm32l4.c
@@ -25,15 +25,23 @@
  * On L4, flash and options are written in DWORDs (8-Byte) only.
  *
  * References:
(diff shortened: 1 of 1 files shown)


esden (approved):
LGTM, as it is based on tested code I am pulling it in. It is definitely an improvement, and if there are still issues then unlikely an additional regression, and someone will report if there are problems.

PR #1306, opened 2022-11-20
Fix: STM32F1 option bytes regression

dragonmux:
## Detailed description

In this PR we address the option bytes write regression for STM32F1 as GD32F1 which was privately reported. As it turns out, because 0x5aa5 is the erased state for the first option byte, this fails the erased state check at the top of `stm32f1_option_write_erased()` and results in the code writing this first byte, which is not erased and the write aborts with an error (`FLASH_SR`'s `PGERR` bit being set). Because `stm32f1_flash_busy_wait()` checks for this error on the way out, it returns false, this then causes the write control loop in `stm32f1_option_write()` to abort.

In this regression fix we test if `i != 0` in that control loop as a second condition before aborting the write. In the future a map of the erased option bytes should be written per-part supported, and this used as the erased state check at the top of `stm32f1_option_write_erased()`, but this is an item for v1.10.

### DIFF
--- a/src/target/stm32f1.c
+++ b/src/target/stm32f1.c
@@ -544,7 +544,7 @@ static bool stm32f1_option_write(target_s *const t, const uint32_t addr, const u
 
 	uint16_t opt_val[8];
 	/* Retrieve old values */
-	for (size_t i = 0; i < 16; i += 4) {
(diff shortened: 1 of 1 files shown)


esden (approved):
great sleuthing! :) Thanks for the fix!

PR #1312, opened 2022-11-27
Fix: RP2040 blank flash bringup

dragonmux:
## Detailed description

In this PR we address the issues around how the RP2040 boot ROMs handle when the attached Flash is blank, as the ROMs leave the controllers involved in a state where we cannot access the Flash. This is characterised by seeing something like this when using BMDA run as `src/blackmagic -tv 1`:

```
8 byte SFDP read at 0x0:
        00 00 00 00 00 00 00 00
Flash device ID: 00 00 00
```

After this fix, the result is a reliable working SFDP readout:

```
RP2040 Flash controller in POR state, reconfiguring
8 byte SFDP read at 0x0:
        53 46 44 50 05 01 00 ff
8 byte SFDP read at 0x8:
        00 05 01 10 80 00 00 ff
Flash size: 2MiB
```

### DIFF
--- a/src/target/rp.c
+++ b/src/target/rp.c
@@ -622,11 +622,11 @@ static void rp_flash_connect_internal(target_s *const t)
 	// IO_BANK0 as that does not affect XIP signals)
 	const uint32_t reset = target_mem_read32(t, RP_RESETS_RESET);
 	const uint32_t io_pads_bits = RP_RESETS_RESET_IO_QSPI_BITS | RP_RESETS_RESET_PADS_QSPI_BITS;
-	target_mem_write32(t, RP_RESETS_RESET, reset | io_pads_bits);
-	target_mem_write32(t, RP_RESETS_RESET, reset);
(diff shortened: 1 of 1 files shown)


src/target/rp.c
```diff
@@ -615,19 +638,40 @@ static void rp_spi_read(
 	target_mem_write32(t, RP_SSI_ENABLE, ssi_enabled);
 }
 
-#if 0
+/* Checks if the QSPI and XIP controllers are in their POR state */
+static bool rp_flash_in_por_state(target_s *const t)
+{
+	const uint32_t clkEnables0 = target_mem_read32(t, RP_CLOCKS_WAKE_EN0);
+	const uint32_t clkEnables1 = target_mem_read32(t, RP_CLOCKS_WAKE_EN1);
```
perigoso, line 645: Camel case variables?

dragonmux: good catch - we'll amend this particular commit in the patch set

esden (approved):
LGTM

PR #1313, opened 2022-11-28
hosted/dap: Fixed error in link reset pointer math.

zyp:
This fixes a regression introduced in 52cad38.

## Detailed description

52cad38 rewrote the pointer math and forgot brackets, unintentionally changing the sign of the 2-offset, leading to the command requesting to shift 32 bits more than it provided.

### DIFF
--- a/src/platforms/hosted/dap.c
+++ b/src/platforms/hosted/dap.c
@@ -449,7 +449,7 @@ void dap_reset_link(bool jtag)
 		*p++ = 0x3c;
 		*p++ = 0xe7;
 		*p++ = 0x1f;
-		buf[1] = ((p - buf + 2U) * 8U) - 2U;
+		buf[1] = ((p - (buf + 2U)) * 8U) - 2U;
 	} else {
 		*p++ = 0x9e;
 		*p++ = 0xe7;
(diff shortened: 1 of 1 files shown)


dragonmux (approved):
Interesting, that bit of pointer maths should be entirely transitive.. the brackets.. shouldn't.. matter, however for you to be PR'ing this they clearly do (*writes notes for next time*). Clean patch, LGTM. Merging

zyp:
It's not a matter of transitivity, just signedness. Remember `a - (b + c) == a - b - c` and conversely `a - b + c == a - (b - c)`.

dragonmux:
In other words, the other patch for this that drops the brackets again would be `- 2U` instead of `+ 2U`..

When mentioning the transitive property, we were contemplating the result of `pointer - pointer` - which iwrc is a ptrdiff_t and so realising that if `sizeof(T)` for a `T *` is not 1, the adjustment by 2 would be applied to the raw integer, and not multiplied by `sizeof(T)` first and hence it is not transitive, even if all the signs agreed to allow the brackets to be transitive on its own. What you've pointed out is just a second way in which we can't transitively transpose where the brackets are - the relationship you've shown is non-transitive as well, unlike `a + (b + c) = (a + b) + c` which *is* transitive.

PR #1190, opened 2022-08-10
Fix: USB interface naming and organisation refactor

dragonmux:
## Detailed description

In this PR we address the organisation and naming of the USB components of the firmware, refactoring cdcacm.c/h out into separate descriptors, USB serial, DFU stub and overarching USB interface components.

In addition this addresses the issue that the common platform components were all over the shop in the build system by putting that all in a platform common Makefile.inc

Naming has been picked to try and make it clear what is doing what and what component of the firmware certain global static values are intended for.

Tested working on native hardware.

## Your checklist for this pull request

* [x] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md)
* [x] I've read the [guidelines for contributing](https://github.com/blackmagic-debug/blackmagic/blob/main/CONTRIBUTING.md) to this repository
* [x] It builds for hardware native (`make PROBE_HOST=native`)
* [x] It builds as BMDA (`make PROBE_HOST=hosted`)
* [x] I've tested it to the best of my ability
* [x] My commit messages provide a useful short description of what the commits do

PR #1314, opened 2022-11-28
fix: USB serial DTR handling and resulting USB packet sent on the wrong endpoint regression

### LINKED ISSUE #1310: Incorrect response on connection & handshake of the virtual UART port
Use a serial terminal to reproduce the issue. The steps in connecting to GDBserver via a (virtual) serial port are to 1) open the connection, and 2) set the DTR line (Data Terminal Ready).

This is how version 1.8.2 handles it (and this is the correct output):
- connect to GDBserver port: no output, DSR & CTS set
- toggle DTR on: output = "$OK#9A" (repeated 3 times, because we fail to acknowledge reception)

On the secondary UART, we can also connect and toggle DTR, and nothing is received (which is correct).

However, on version 1.9 (mainline):
- connect to GDBserver port: ouput = [a1 20 00 00 00 00 02 00 03 00 a1 20 00 00 00 00 02 00 03 00 a1 20 00 00 00 00 02 00 03 00]; DSR & CTS *not* set
- toggle DTR on: output = [a1 20 00 00 00 00 02 00 03 00] (DSR & CTS still not set)
- toggle DTR off: same output

- on secondary UART, exactly the same thing happens: a stream of binary data as soon as you open a connection (and exactly the same byte sequence as in the case for the GDBserver port), and a repeat of this binary stream when you toggle DTR.

Hypothesis: The byte sequence looks eerily similar to a SERIAL_STATE notification for the state of the (virtual) control lines, which should go over the control endpoint. It is as if this notification is transmitted on the IN endpoint.

Notes: 

1) GDB is unaffected to this error. When it does not receive the expected "OK" return, it first retries toggling DTR. When that fails too, it carries on and requests the capabilities of the debug probe. There is a good response on that command, so GDB concludes the connection is successful.

2) On version 1.9: if you transmit data to the BMP with DTR off (i.e. not set), serial communication hangs on Windows (on that port). Not an issue with GDB (because it always sets DTR), but not truly an appropriate response either. Version 1.8.2 does not show this behaviour.

dragonmux:
## Detailed description

Fixed a regression caused in #1190 that puts the CDC notification replies out over the wrong endpoint.

The effect of this regression is that BMP erroneously replies to state change requests, like asking for DTR to be set, over the wrong endpoint channel - using the serial data IN endpoint rather than the notification endpoint.

Many thanks to @compuphase for the data in their bug report allowing us to so quickly pinpoint this.

## Closing issues

Fixes #1310

### DIFF
--- a/src/platforms/common/usb_serial.c
+++ b/src/platforms/common/usb_serial.c
@@ -95,7 +95,8 @@ static usbd_request_return_codes_e gdb_serial_control_request(usbd_device *dev, 
 
 	switch (req->bRequest) {
 	case USB_CDC_REQ_SET_CONTROL_LINE_STATE:
-		usb_serial_set_state(dev, req->wIndex, CDCACM_GDB_ENDPOINT);
+		/* Send a notification back on the notification endpoint */
+		usb_serial_set_state(dev, req->wIndex, CDCACM_GDB_ENDPOINT + 1U);
 		gdb_serial_dtr = req->wValue & 1U;
 		return USBD_REQ_HANDLED;
 	case USB_CDC_REQ_SET_LINE_CODING:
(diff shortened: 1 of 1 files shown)


esden (approved):
I went back into the history to see if there was any chance for me to catch this during patch review... but it was deep inside magic number land, there is no chance for me to catch this. Good catch and fix! :)

PR #1323, opened 2022-12-06
Feature/auto flash write buffer size

perigoso:
`writebufsize` is no longer set in the target code, it's now sized automatically up to a ceiling

This reduces overhead and reduces allocation/deallocation count in targets with small `writesize`, while keeping the buffers manageably small

## Detailed description

Targets with `writesize` > `FLASH_WRITE_BUFFER_CEILLING`, use `writesize` as buffer size
Other Targets use a buffer sized to as many `writesize` fit in `FLASH_WRITE_BUFFER_CEILLING`

`FLASH_WRITE_BUFFER_CEILLING` was chosen as `1024` as a fairly arbitrary value, targets with large `writesize` don't benefit much from a multi-write buffer like targets with smaller `writesize`, as the overhead is already comparatively small

This does nothing to bind the max size of the buffer, as it's directly tied to `writesize` and that's not something we can just mess with, to keep that to a reasonable size is a job for correctness in the target implementation. (we might want to ditch the automatic `writesize` to `blocksize` when absent and make that value mandatory for the target)

note: **_this code was not tested!!_** the code logic is trivial, and the target flash code should be ready for this, but nonetheless some testing is definitely welcome

### DIFF
--- a/src/target/renesas.c
+++ b/src/target/renesas.c
@@ -622,11 +622,9 @@ static void renesas_add_rv40_flash(target_s *t, target_addr_t addr, size_t lengt
 
 	if (code_flash) {
 		f->blocksize = RV40_CF_REGION1_BLOCK_SIZE;
-		f->writebufsize = RV40_CF_WRITE_SIZE * 8U;
 		f->writesize = RV40_CF_WRITE_SIZE;
 	} else {
 		f->blocksize = RV40_DF_BLOCK_SIZE;
-		f->writebufsize = RV40_DF_BLOCK_SIZE * 8U;
 		f->writesize = RV40_DF_WRITE_SIZE;
 	}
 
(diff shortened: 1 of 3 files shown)


dragonmux:
We've decided to put this into v1.10, and backport it to v1.9 if required as a point release. Milestone has been updated to reflect the PR's scheduling

esden:
Just to clarify the reasoning, as I think it is worth recording it:
This is a potentially a bug fix but definitely a cleanup/feature. It will prevent issues with targets allocating too large of a buffer in the future, and it might fix some currently unknown bug with some target that blows up the ram. That said @dragonmux and I think that it is best to merge it in the next dev cycle instead. (I hope this makes sense :) )

perigoso:
Actually, it is unlikely this will fix any bugs, as it isn't bounding the size of the buffers, that would be a different cleanup, what it might do is improve flash speeds in a couple targets

v1.10 milestone is what I expected, that is fine and makes sense :)

PR #1331, opened 2022-12-15
Fix: Cortex-M SWD attach regression

### LINKED ISSUE #1319: S32K148: Device fails to attach
With `Black Magic Probev1.9.0-rc0-3-gd5e59b94, Hardware Version 6`, the S32K simply fails to attach. It scans just fine, but attaching does not work anymore.

```
(gdb) mon ver
Black Magic Probev1.9.0-rc0-3-gd5e59b94, Hardware Version 6
Copyright (C) 2022 Black Magic Debug Project

(gdb) mon s
Target voltage: 3.3V
Available Targets:
No. Att Driver
 1      S32K14x M4
 2      Kinetis Recovery (MDM-AP)
(gdb) att 1
Attaching to Remote target
Attaching to Remote target failed
(gdb) mon s
Target voltage: 3.3V
Available Targets:
No. Att Driver
 1      S32K14x M4
 2      Kinetis Recovery (MDM-AP)
(gdb) att 1
Attaching to Remote target
Attaching to Remote target failed
(gdb) att 2
Attaching to Remote target
Attaching to Remote target failed
(gdb)
```

Loading an older version fixes things.

```
(gdb) tar ext \\.\COM5
Already connected to a remote target.  Disconnect? (y or n) y
Remote debugging using \\.\COM5
(gdb) mon s
Target voltage: 3.3V
Available Targets:
No. Att Driver
 1      S32K14x M4
 2      Kinetis Recovery (MDM-AP)
(gdb) att 1
Attaching to Remote target
warning: No executable has been specified and target does not support
determining executable automatically.  Try using the "file" command.
0x000182d2 in ?? ()
(gdb) mon ver
Black Magic Probe v1.8.0-682-g12bda1e0, Hardware Version 6
Copyright (C) 2022 Black Magic Debug Project

(gdb)
```

dragonmux:
## Detailed description

This PR addresses the issue reported in https://github.com/blackmagic-debug/blackmagic/issues/1319 and on Discord where a user trying to attach to various Cortex-M cores via SWD would get an attach failure from the firmware, the remote protocol going non-responsive (BMDA connected to BMF), or have the debug server abort (BMDA via ST-Link, JLink, CMSIS-DAP). This would happen only if there was a delay between scan and attach.

The underlying issue causing this (something happening on the bus that causes the target to go into the protocol error state defined by ADIv5.2 §B1.2 subtitle "Protocol errors (SW-DP only)", and §B4.2.5) remains unchanged, however this PR gives BMP proper handling for the no-response answer used to indicate this state and a proper error recovery mechanism which means the target ends up happily back in a usable state and we continue on successfully as before.

A follow-up PR for v1.10 will be needed to address this underlying issue.

Tested on native hardware against STM32F4, Tiva-C and other targets. Tested via BMDA using the BMP remote protocol, ST-Link and CMSIS-DAP.

## Closing issues

Fixes #1319

### DIFF
--- a/src/platforms/hosted/cmsis_dap.c
+++ b/src/platforms/hosted/cmsis_dap.c
@@ -318,6 +318,38 @@ int dbg_get_report_size(void)
 	return report_size;
 }
 
+ssize_t dbg_dap_cmd_hid(const uint8_t *const data, const size_t request_length)
+{
+	if (request_length + 1U > report_size) {
+		DEBUG_WARN(
+			"Attempted to make over-long request of %zu bytes, max length is %zu\n", request_length + 1U, report_size);
+		exit(-1);
+	}
+
+	memset(buffer + request_length + 1U, 0xff, report_size - (request_length + 1U));
+	buffer[0] = 0x00; // Report ID??
+	memcpy(buffer + 1, data, request_length);
+
+	const int result = hid_write(handle, buffer, 65);
+	if (result < 0) {
+		DEBUG_WARN("CMSIS-DAP write error: %ls\n", hid_error(handle));
+		exit(-1);
+	}
+
+	int response = 0;
+	do {
+		response = hid_read_timeout(handle, buffer, 65, 1000);
+		if (response < 0) {
+			DEBUG_WARN("CMSIS-DAP read error: %ls\n", hid_error(handle));
+			exit(-1);
+		} else if (response == 0) {
+			DEBUG_WARN("CMSIS-DAP read timeout\n");
+			exit(-1);
+		}
+	} while (buffer[0] != data[0]);
+	return response;
+}
+
 ssize_t dbg_dap_cmd_bulk(uint8_t *const data, const size_t request_length)
 {
 	int transferred = 0;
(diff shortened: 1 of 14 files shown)


src/platforms/hosted/dap.c
```diff
@@ -583,40 +594,41 @@ void dap_ap_write(adiv5_access_port_s *ap, uint16_t addr, uint32_t value)
...
+	mem_access_setup(ap, requests, src, align);
+	requests[3].request = SWD_AP_DRW | DAP_TRANSFER_RnW;
+	requests[4].request = SWD_DP_R_RDBUFF | DAP_TRANSFER_RnW;
+	uint32_t result[2];
+	if (!perform_dap_transfer(ap->dp, requests, 5U, result, 2U)) {
+		DEBUG_WARN("dap_read_single failed\n");
+		memset(dest, 0, sizeof(uint32_t));
+		return;
+	}
+	/* Pull out the data from the *second* response (value of RDBUFF) */
```
zyp, line 566: This is incorrect. CMSIS-DAP handles the AP read latency on the probe side and does an implicit read of RDBUFF, so the host should never need to read it explicitly. The actual read value ends up in the first response slot, and the second contains undefined garbage (all zeroes in my testing). The proper solution here is to simply remove the fifth transfer request and return the first response.

src/platforms/hosted/cmsis_dap.c
```diff
@@ -425,33 +425,34 @@ static void dap_mem_read(adiv5_access_port_s *ap, void *dest, uint32_t src, size
...
-		unsigned int blocksize = (src | 0x3ffU) - src + 1U;
-		if (blocksize > len)
-			blocksize = len;
-		while (blocksize) {
-			unsigned int transfersize = blocksize;
-			if (transfersize > max_size)
-				transfersize = max_size;
-			if (!dap_read_block(ap, dest, src, transfersize, align)) {
+	/* Otherwise proceed blockwise */
+	dap_ap_mem_access_setup(ap, src, align);
```
zyp, line 459: This needs to be called inside the loop with `src + offset` to update AP_TAR at 1024B boundaries, otherwise it'll wrap.

src/platforms/hosted/cmsis_dap.c
```diff
@@ -454,34 +454,37 @@ static void dap_mem_write_sized(adiv5_access_port_s *ap, uint32_t dest, const vo
...
-			blocksize = len;
-		while (blocksize) {
-			unsigned int transfersize = blocksize;
-			if (transfersize > max_size)
-				transfersize = max_size;
-			if (!dap_write_block(ap, dest, src, transfersize, align)) {
-				DEBUG_WARN("mem_write failed %u\n", ap->dp->fault);
-				ap->dp->fault = 1;
+	/* Otherwise proceed blockwise */
+	dap_ap_mem_access_setup(ap, dest, align);
```
zyp, line 493: This needs to be called inside the loop with `dest + offset` to update AP_TAR at 1024B boundaries, otherwise it'll wrap.

src/platforms/hosted/dap.c
dragonmux: Many thanks for the extra information, this should now be fixed

src/platforms/hosted/cmsis_dap.c
dragonmux: This should now be fixed - many thanks :)

src/platforms/hosted/cmsis_dap.c
dragonmux: This should now be fixed - many thanks :)

esden (approved):
LGTM

PR #1341, opened 2022-12-24
BMDA: Fix build on OSX

djix123:
## Detailed description

Fixes BMDA build issues on OSX:

(i) Adds certain compiler flags if not OSX (-Wmaybe-uninitialized -Wstringop-overflow -Wunsafe-loop-optimizations)
(ii) Add 'const' to serial_open declaration on OSX

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -3,16 +3,21 @@ PLATFORM_DIR = platforms/$(PROBE_HOST)
 VPATH += $(PLATFORM_DIR) target
 ENABLE_DEBUG ?=
 
+SYS = $(shell $(CC) -dumpmachine)
+
 ifneq ($(V), 1)
 MAKEFLAGS += --no-print-dir
 Q := @
 endif
 
-CFLAGS += -Wall -Wextra -Werror -Wmaybe-uninitialized \
-	-Wreturn-type -Wstringop-overflow -Wunsafe-loop-optimizations \
+CFLAGS += -Wall -Wextra -Werror -Wreturn-type \
 	-Wno-char-subscripts \
 	-std=c11 -g3 -MD -I./target \
 	-I. -Iinclude -I$(PLATFORM_DIR)
+
+ifeq (filter, macosx darwin, $(SYS))
+CFLAGS += -Wmaybe-uninitialized -Wstringop-overflow -Wunsafe-loop-optimizations 
+endif
 
 ifeq ($(ENABLE_DEBUG), 1)
 CFLAGS += -DENABLE_DEBUG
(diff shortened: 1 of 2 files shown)


zyp:
This looks good to me. Except the conditional logic, these are the exact changes I've been keeping around in my local working tree.

djix123:
Yes, I've been keeping them around locally too. Thought it about time to submit as a patch!

src/Makefile
```diff
@@ -3,17 +3,25 @@ PLATFORM_DIR = platforms/$(PROBE_HOST)
 VPATH += $(PLATFORM_DIR) target
 ENABLE_DEBUG ?=
 
+CC ?= gcc
```
dragonmux, line 6: CC is already set in the platforms and needs to use the form `$(CROSS_COMPILE)gcc`. We are fine with the assignment being moved into this Makefile, but it must not semantically change in its form for now.

src/Makefile
```diff
@@ -3,17 +3,25 @@ PLATFORM_DIR = platforms/$(PROBE_HOST)
...
+SYS = $(shell $(CC) -dumpmachine)
+
 ifneq ($(V), 1)
 MAKEFLAGS += --no-print-dir
 Q := @
 endif
 
-CFLAGS += -Wall -Wextra -Werror -Wmaybe-uninitialized \
-	-Wreturn-type -Wstringop-overflow -Wunsafe-loop-optimizations \
+CFLAGS += -Wall -Wextra -Werror \
```
dragonmux, line 14: Please reflow `-Wreturn-type` back onto this line

dragonmux (changes requested):
Thank yo for the contribution! This looks reasonable, though there are a couple of changes we request get made before we can accept this for v1.9.

Additionally, please follow the contribution guidelines for the commit messages - we suggest the `hosted/serial_unix:` change should be a separate commit, and the Makefile changes are either `makefile:` or `misc:` (probably the latter as in fixing the first review item you'll need to modify all the platform Makefile.inc files)

src/Makefile
djix123: when checking I realized there was no need to assign CC in the Makefile as it is already set - so I've removed from the proposed patch

src/Makefile
djix123: Done

djix123:
I've reworked the pull request into two separate commits. Let me know if I need to change anything else. Thank you for considering this pull request.

dragonmux (approved):
LGTM, merging

djix123:
Thank you!

PR #1347, opened 2022-12-30
Fix: Flashing from the hosted CLI on macOS/*nix

hollinsky:
Using the hosted CLI on macOS, I found I was not able to flash my target. It would say the flash succeeded having written zero bytes.

I tracked this down to the way the file is loaded in, where the size is captured improperly during the mmap.

I also cleaned up that structure, as the `real_size` variable was not necessary on Windows platforms.

With this change, I'm able to flash successfully.


dragonmux (approved):
LGTM! Good catch, and apologies for the silly bug. Merging

PR #1348, opened 2022-12-31
Feature: clean BMDA shutdown

dragonmux:
## Detailed description

In this PR we implement a feature to allow BMDA to be cleanly shut down rather than having to always signal it.

This has been done for two reasons.
The first more minor one is to help with users scripting GDB/BMDA wishing to shut the server back down after completing a GDB script.
The second reason is to make ASAN, LSAN and friends more useful - leak sanitizer can only act after a clean non-signalled shutdown of the program so this provides a mechanism to allow the program to exit normally and trigger LSAN processing. Likewise this allows testing with other sanitizers with this restriction in the future.


esden (approved):
LGTM

PR #1354, opened 2023-01-05
Support for ST-Link v3

### LINKED ISSUE #1008: Firmware for stlink-v3
[Here](https://github.com/blackmagic-debug/blackmagic/tree/master/src/platforms) there are the implementation for some platforms: I see listed STLINK-V1, V2 and V21 but I can't find V3.
Now some new Nucleo board (as an example Nucleo-H743ZI2) has a STLINK-V3 onboard.
In the [stlink subpage](https://github.com/blackmagic-debug/blackmagic/tree/master/src/platforms/stlink) I see some reference to V3, but the overall is not very clear (maybe the README can be improved a little bit, but the situation is not clear to me).
If someone can give details, I can send s PR with some improvements to the README.

jamesturton:
## Detailed description

Rebased and tested version of #818.
I have verified that all basic functionally works. However I am yet to test SLCAN.
I know the newer stock firmware for the STLink disables the SWD interface on the probe, but for those who haven't updated the stock firmware it does make for a nice probe - especially the [STLINK-V3MINIE](https://www.st.com/en/development-tools/stlink-v3minie.html) which I have been testing with! The major advantage I see with this probe is that the STM32F723 supports USB 2.0/Hi-Speed (480 Mbps) which is magnitudes quicker than the USB 1.1/Full-Speed (12 Mbps) interface of the other STM32F1XX chips.

## Closing issues

Closes #1008


dragonmux:
Hi James, before we review this, we have 3 things we'd like addressed that were issues with the original PR too:

1) Could the commit messages please be updated to conform with the current standards for contribution
2) Please fix the lint issues
3) Rather than adding another locm3 target to build for all platforms in the main Makefile, please modify this platform to work more like the f4discovery where the additional locm3 microcontroller target is only built when building this platform

Many thanks for taking the time and effort to overhaul that mess and bring it up to speed! Was something we were dreading doing having done it for so many other PRs. We look forward to being able to land this in some form for v1.10

jamesturton:
>     1. Could the commit messages please be updated to conform with the current standards for contribution

Hopefully these commit messages are looking a little bit more descriptive!

jamesturton:
>     2. Please fix the lint issues

Lint issues should now be fixed

jamesturton:
>     3. Rather than adding another locm3 target to build for all platforms in the main Makefile, please modify this platform to work more like the f4discovery where the additional locm3 microcontroller target is only built when building this platform

And this should be fixed now too!

> Many thanks for taking the time and effort to overhaul that mess and bring it up to speed! Was something we were dreading doing having done it for so many other PRs. We look forward to being able to land this in some form for v1.10

Hopefully this can be included in the v1.10 release and also looking forward to the soon to come v1.9 release too! :smile:

dragonmux (commented):
LGTM, we don't have hardware to test this on but at least all the major things have been resolved so we'll do a detailed review in a bit. Would have been nice if the fixes had been applied directly to the offending commits in the history via `git commit --fixup` but we understand how much extra work that can be so don't worry about it for this :)

src/remote.c
```diff
@@ -309,11 +309,8 @@ static void remote_packet_process_general(unsigned i, char *packet)
 #endif
 		break;
 
-#if !defined(BOARD_IDENT) && defined(BOARD_IDENT)
```
dragonmux, line 312: This goes in the same camp as the other changes to board identification - it should be reverted

dragonmux (changes requested):
As mentioned on Discord, here's a first run-down review of the PR - we'll do further review rounds once these issues have been addressed as we recognise this is already a lot to resolve

src/platforms/common/usb.h
```diff
@@ -27,16 +27,36 @@
 extern usbd_device *usbdev;
 extern uint16_t usb_config;
 
+#if defined(USB_HS)
```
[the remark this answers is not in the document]
jamesturton: Now fixed!

src/platforms/common/usb_descriptors.h
```diff
@@ -40,11 +37,10 @@ static const usb_device_descriptor_s dev_desc = {
 	.bDeviceClass = 0xef, /* Miscellaneous Device */
 	.bDeviceSubClass = 2, /* Common Class */
 	.bDeviceProtocol = 1, /* Interface Association */
-#ifdef LM4F
-	.bMaxPacketSize0 = 64, /*Fixed for icdi*/
-#else
-	.bMaxPacketSize0 = 32,
-#endif
+	/* The USB specification requires that the control endpoint size for high
```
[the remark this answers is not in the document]
jamesturton: Now fixed!

jamesturton:
Ok, I've pulled out changes regarding SLCAN. I can wait until this has one been merged before submitting a new PR for that.
All of the other points on your review I have now addressed, so let me know what you pick up on on your second round of reviews!

dragonmux:
We'll try and get this reviewed today

src/platforms/common/usb_descriptors.h
```diff
@@ -65,7 +68,7 @@ static const usb_endpoint_descriptor_s gdb_comm_endp = {
 	.bEndpointAddress = (CDCACM_GDB_ENDPOINT + 1U) | USB_REQ_TYPE_IN,
 	.bmAttributes = USB_ENDPOINT_ATTR_INTERRUPT,
 	.wMaxPacketSize = 16,
-	.bInterval = 255,
+	.bInterval = MAX_BINTERVAL,
```
dragonmux, line 71: This is technically fine, however we feel it might make a little more sense to name the new macro `USB_MAX_INTERVAL`, especially as the b prefix is not adding any information here

src/platforms/stlinkv3/Readme.md
```diff
@@ -0,0 +1,54 @@
+# Blackmagic firmware for STLINK-V3 Adapters
```
dragonmux, line 1: Let's name the file `README.md` for consistency with the rest of the repo readme files.

Uwe had a habit of using all-caps names with his own hyphenation for the names of tools, which causes user confusion - ST uses `ST-Link v3` as the official name of the tool, we should to, let's change that throughout this readme for consistency.

src/platforms/stlinkv3/Readme.md
```diff
@@ -0,0 +1,54 @@
...
+As simple as
+```make PROBE_HOST=stlinkv3 clean
+make PROBE_HOST=stlinkv3
+```
+
+## Flashing
+Easiest is using the BMP bootloader. Load the BMP firmware easiest with
+scripts/stm32mem.py  or dfu-utils. BMP bootloader must be flashed with SWD
+access. Use CN2 on the V3Set. For V3-mini and V3MODS use soldered connections
+to CN3, see [connection diagramm](https://github.com/RadioOperator/CMSIS-DAP_for_STLINK-V3MINI/blob/master/STLINK-V3MINI/Adaptor/STLINK-V3MINI_GPIOs_v4.JPG)
```
dragonmux, line 39: Is it possible to find an official pinout diagram from the ST website? We'd prefer to not be relying on 3rd part repositories for things like this.

src/platforms/stlinkv3/Readme.md
```diff
@@ -0,0 +1,54 @@
...
+
+It is a good idea to keep a full image of the original flash content as backup!
+
+If you want to keep the original bootloader or access via SWD is disabled, clone
+https://github.com/UweBonnes/stlink-tool/tree/stlinkv21
+make and use like
+```stlink-tool blackmagic.bin```
```
dragonmux, line 46: Either this needs to be converted to single backticks, or the backticks put onto their own lines. As-is this is invalid Markdown, and GH is syntax highlighting it as such.

Ideally the code blocks should have new-lines before and after to be markdown-clean as in:
```
<paragraph>

`​`​`
stlink-tool blackmagic.bin
`​`​`

<paragraph>
```

but we appreciate that's a little arduous.

src/platforms/stlinkv3/platform.c
```diff
@@ -0,0 +1,259 @@
+/*
+ * This file is part of the Black Magic Debug project.
+ *
+ * Copyright (C) 2011-2021 Black Sphere Technologies Ltd.
```
dragonmux, line 4: The copyright notices for the new files should be updated to `Copyright (C) 2022-2023 1BitSquared  <info@1bitsquared.com>`

src/platforms/stlinkv3/platform.c
```diff
@@ -0,0 +1,259 @@
...
+
+#define SCB_DCISW_SET_Pos 5U                             /*!< SCB DCISW: Set Position */
+#define SCB_DCISW_SET_Msk (0x1ffUL << SCB_DCISW_SET_Pos) /*!< SCB DCISW: Set Mask */
+
+#define SCB_DCISW_WAY_Pos 30U                        /*!< SCB DCISW: Way Position */
+#define SCB_DCISW_WAY_Msk (3UL << SCB_DCISW_WAY_Pos) /*!< SCB DCISW: Way Mask */
+
+#define CCSIDR_SETS(x) (((x)&SCB_CCSIDR_NUMSETS_Msk) >> SCB_CCSIDR_NUMSETS_Pos)
+
+static void __DSB(void)
```
dragonmux, line 65: `__` names and especially ones starting with a capital letter are strictly implementation reserved, and it is undefined behaviour to use them. This and `__ISB` must be changed to use non-reserved names and preferably ons that describe what they're doing.

src/platforms/stlinkv3/platform.c
```diff
@@ -0,0 +1,259 @@
...
+{
+	__asm__ volatile("dsb 0xf" ::: "memory");
+}
+
+static void __ISB(void)
+{
+	__asm__ volatile("isb 0xf" ::: "memory");
+}
+
+static void SCB_EnableICache(void)
```
dragonmux, line 75: This and the others like them need renaming to match the code base contribution style requirements. `scb_enable_icache`, etc.

src/platforms/stlinkv3/platform.c
```diff
@@ -0,0 +1,259 @@
...
+{
+	__DSB();
+	__ISB();
+	SCB_ICIALLU = 0UL; /* invalidate I-Cache */
+	__DSB();
+	__ISB();
+	SCB_CCR |= (uint32_t)SCB_CCR_IC_Msk; /* enable I-Cache */
```
dragonmux, line 82: This cast should not be required, please remove it (this goes for other similar lines below). The macro should be all-caps (eg, `SCB_CCR_IC_MASK`)

src/platforms/stlinkv3/platform.c
```diff
@@ -0,0 +1,259 @@
...
+	__DSB();
+
+	ccsidr = SCB_CCSIDR;
+
+	sets = (uint32_t)(CCSIDR_SETS(ccsidr));
+	do {
+		ways = (uint32_t)(CCSIDR_WAYS(ccsidr));
+		do {
+			SCB_DCISW =
+				(((sets << SCB_DCISW_SET_Pos) & SCB_DCISW_SET_Msk) | ((ways << SCB_DCISW_WAY_Pos) & SCB_DCISW_WAY_Msk));
```
dragonmux, line 103: The outer-most pair of parens are redundant and should be removed.

src/platforms/stlinkv3/platform.c
```diff
@@ -0,0 +1,259 @@
...
+#if defined(__CC_ARM)
+			__schedule_barrier();
```
dragonmux, line 105: We don't support building with ARM's proprietary toolchain, so this should be removed

src/platforms/stlinkv3/platform.c
```diff
@@ -0,0 +1,259 @@
...
+int platform_hwversion(void)
+{
+	return hw_version;
+}
+
+void platform_nrst_set_val(bool assert)
+{
+	gpio_set_val(SRST_PORT, SRST_PIN, !assert);
+	if (assert)
+		for (int i = 0; i < 10000; i++)
```
dragonmux, line 126: `volatile size_t` instead of `int` here, and the loop body should then be `continue;` please

src/platforms/stlinkv3/platform.c
```diff
@@ -0,0 +1,259 @@
...
+	static char ret[] = "0.0V";
+	uint8_t channels[] = {
+		ADC_CHANNEL0,
+	};
+	unsigned value;
+
+	adc_set_regular_sequence(ADC1, 1, channels);
+	adc_start_conversion_regular(ADC1);
+	while (!adc_eoc(ADC1))
+		;
```
dragonmux, line 154: Loop body must contain something, add `continue` please

src/platforms/stlinkv3/platform.c
```diff
@@ -0,0 +1,259 @@
...
+#define GND_DETECT_PORT GPIOG
+#define GND_DETECT_PIN  GPIO5
+
+const char *platform_target_voltage(void)
+{
+	/* On the stlinkv3, the target input voltage is divided by two.
+	 * The ADC is sampling at 12 bit resolution.
+	 * Vref+ input is assumed to be 3.3 volts. */
+	static char ret[] = "0.0V";
+	uint8_t channels[] = {
```
dragonmux, line 146: This can be simplified to `uint8_t channel = ADC_CHANNEL0;` and a single `&` added to the use below.

src/platforms/stlinkv3/platform.c
```diff
@@ -0,0 +1,259 @@
...
+		ADC_CHANNEL0,
+	};
+	unsigned value;
```
dragonmux, line 149: C89 style declaration, and use of a raw type - prefer `uint32_t` at time of first initialisation

src/platforms/stlinkv3/platform.c
```diff
@@ -0,0 +1,259 @@
...
+	/* Drive the tck/swck pin low. */
+	gpio_mode_setup(TCK_PORT, GPIO_MODE_OUTPUT, GPIO_PUPD_NONE, TCK_PIN);
+	gpio_set_output_options(TCK_PORT, GPIO_OTYPE_PP, GPIO_OSPEED_2MHZ, TCK_PIN);
+
+	/* Drive direction switch pin. */
+	gpio_mode_setup(TMS_DRIVE_PORT, GPIO_MODE_OUTPUT, GPIO_PUPD_NONE, TMS_DRIVE_PIN);
+	gpio_set_output_options(TMS_DRIVE_PORT, GPIO_OTYPE_PP, GPIO_OSPEED_2MHZ, TMS_DRIVE_PIN);
+	gpio_set(TMS_DRIVE_PORT, TMS_DRIVE_PIN);
+
+#define PWR_EN_PORT GPIOB
```
dragonmux, line 219: Please move all `#define`'s up to the top of file

src/platforms/stlinkv3/platform.h
```diff
@@ -0,0 +1,198 @@
...
+#define SWDIO_MODER      GPIO_MODER(TMS_PORT)
+#define SWDIO_MODER_MULT (1 << (9 << 1))
+
+#define TMS_SET_MODE()                                                    \
+	gpio_mode_setup(TMS_PORT, GPIO_MODE_OUTPUT, GPIO_PUPD_NONE, TMS_PIN); \
+	gpio_set_output_options(TMS_PORT, GPIO_OTYPE_PP, GPIO_OSPEED_2MHZ, TMS_PIN);
+
+#define SWDIO_MODE_FLOAT()                         \
+	do {                                           \
+		uint32_t moder = SWDIO_MODER;              \
```
dragonmux, line 84: `moder` looks like a typo for `mode`, leading to a bit of  head-scratching reading this. Let's correct the naming?

On second reading, it's likely the `r` is supposed to mean `register`, so for the `#define` above (`SWDIO_MODER`), it probably makes sense to rename to `SWDIO_MODE_REG` to make that obvious

src/platforms/stlinkv3/usb_control.c
```diff
@@ -0,0 +1,297 @@
...
+ * multiple of the endpoint max packet size, we need an explicit ZLP.
+ * @param len how much data we want to transfer
+ * @param wLength how much the host asked for
+ * @param ep_size
+ * @return
+ */
+static bool needs_zlp(uint16_t len, uint16_t wLength, uint8_t ep_size)
+{
+	if (len < wLength) {
+		if (len && (len % ep_size == 0)) {
```
dragonmux, line 66: The closing inner paren comes too late (`(len & ep_size) == 0`) and the braces are redundant

src/platforms/stlinkv3/usb_control.c
```diff
@@ -0,0 +1,297 @@
...
+		}
+	}
+	return false;
+}
+
+/* Register application callback function for handling USB control requests. */
+int usbd_register_control_callback(
+	usbd_device *usbd_dev, uint8_t type, uint8_t type_mask, usbd_control_callback callback)
+{
+	int i;
```
dragonmux, line 77: C89 style declaration and raw type. In this case, `size_t` is more suitable in the for loop initialisation block

src/platforms/stlinkv3/usb_control.c
```diff
@@ -0,0 +1,297 @@
...
+
+	for (i = 0; i < MAX_USER_CONTROL_CALLBACK; i++) {
+		if (usbd_dev->user_control_callback[i].cb) {
```
dragonmux, line 80: Redundant extra braces, please remove them

src/platforms/stlinkv3/usb_control.c
```diff
@@ -0,0 +1,297 @@
...
+			continue;
+		}
+
+		usbd_dev->user_control_callback[i].type = type;
+		usbd_dev->user_control_callback[i].type_mask = type_mask;
+		usbd_dev->user_control_callback[i].cb = callback;
+		return 0;
```
dragonmux, line 87: This is technically abusing ints as boolean return values (the `TRUE` and `FALSE` macros would be more suitable here) but we probably can't avoid the return type being `int` as that would break the API contract from locm3

src/platforms/stlinkv3/usb_control.c
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	return -1;
+}
+
+static void usb_control_send_chunk(usbd_device *usbd_dev)
+{
+	if (usbd_dev->desc->bMaxPacketSize0 < usbd_dev->control_state.ctrl_len) {
```
dragonmux, line 95: This is confusingly written - we would prefer `usbd_dev->control_state.ctrl_len > usbd_dev->desc->bMaxPacketSize0`

src/platforms/stlinkv3/usb_control.c
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	usbd_dev->control_state.ctrl_len += size;
+
+	return packetsize;
+}
+
+static enum usbd_request_return_codes usb_control_request_dispatch(usbd_device *usbd_dev, struct usb_setup_data *req)
+{
+	int i, result = 0;
```
dragonmux, line 131: Same C89 style declaration thing here + defining multiple variables in one statement, which is a clang-tidy lint

src/platforms/stlinkv3/usb_control.c
```diff
@@ -0,0 +1,297 @@
...
+	struct user_control_callback *cb = usbd_dev->user_control_callback;
+
+	/* Call user command hook function. */
+	for (i = 0; i < MAX_USER_CONTROL_CALLBACK; i++) {
+		if (cb[i].cb == NULL) {
+			break;
+		}
+
+		if ((req->bmRequestType & cb[i].type_mask) == cb[i].type) {
+			result = cb[i].cb(usbd_dev, req, &(usbd_dev->control_state.ctrl_buf), &(usbd_dev->control_state.ctrl_len),
```
dragonmux, line 141: `const int ret` here is likely correct and should be preferred. The parens around the control_state expressions are redundant and can be removed as the language guarantees the `&` address-of only applies to the final member access (ctrl_buf, ctrl_len, etc)

dragonmux (changes requested):
We've reviewed down to usb_dwc_common.c in the new platform, but have opted to pause and let you address the items found before we continue this second more detailed review. Of the initial parts that we reviewed, it's looking much better, so this should mostly be new findings in new ground

src/platforms/common/usb_descriptors.h
jamesturton: Agreed, done!

src/platforms/stlinkv3/Readme.md
jamesturton: I actually think that STLINK-V3 is the correct stylising of the name here. See https://www.st.com/en/development-tools/stlink-v3set.html. Will rename the file for consistency.

src/platforms/stlinkv3/Readme.md
jamesturton: Done!

src/platforms/common/aux_serial.c
```diff
@@ -334,7 +340,7 @@ static void aux_serial_receive_isr(const uint32_t usart, const uint8_t dma_irq)
 
 	/* If line is now idle, then transmit a packet */
 	if (is_idle) {
-#ifdef USART_ICR
+#ifdef USART_ICR_IDLECF
```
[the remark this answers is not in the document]
dragonmux: No need to revert this. It seems fine and probably a little safer given that

src/platforms/stlinkv3/Readme.md
dragonmux: They have STLINK-V3SET and they also have ST-Link v3 it seems.. unless we're missing something.. at the very least if you're right that would make it inconsistent with their previous ST-Link tooling which is.. don't like.. we'll noodle on that some more then

src/platforms/stlinkv3/platform.c
dragonmux: Did the cast here `(uint32_t)` wind up being required? we'd prefer to remove the casts if at all possible

src/platforms/stlinkv3/platform.c
```diff
@@ -0,0 +1,233 @@
...
+
+static void scb_enable_d_cache(void)
+{
+	SCB_CCSELR = 0U; /*(0U << 1U) | 0U;*/ /* Level 1 data cache */
+	cm_dsb();
+
+	uint32_t ccsidr = SCB_CCSIDR;
+
+	uint32_t sets = (uint32_t)CCSIDR_SETS(ccsidr);
+	do {
```
dragonmux, line 89: The more we read this, the more this looks like a backwards for loop - the `do {} while ()` pattern here is obscuring meaning and the condition in the while component is bug-prone - for example, what happens if sets is 0 to start with? (same for ways)

This reads like it should be something like:
```c
const uint32_t sets =  CCSIDR_SETS(ccsidr);
const uint32_t ways = CCSIDR_WAYS(ccsidr);
for (uint32_t set = 0; set <= sets; ++set) {
    for (uint32_t way = 0; way <= ways; ++way) {
        CSB_DCISW = ((set << SCB_DCISW_SET_SHIFT) & SCB_DCISW_SET_MASK) |
            ((way << SCB_DCISW_WAY_SHIFT) & SCB_DCISW_WAY_MASK);
    }
}
```

These should be turned into `<` loops instead of `<=` loops if the result is off by one

src/platforms/stlinkv3/platform.h
```diff
@@ -0,0 +1,210 @@
...
+
+#ifdef ENABLE_DEBUG
+#define PLATFORM_HAS_DEBUG
+extern bool debug_bmp;
+#endif
+
+#define PLATFORM_HAS_USBUART
+#define PLATFORM_IDENT "STLINK-V3 "
+
+#define BOOTMAGIC0 0xb007da7a
```
dragonmux, line 43: These should be suffixed with `U`

src/platforms/stlinkv3/platform.h
```diff
@@ -0,0 +1,210 @@
...
+#include "gpio.h"
+#include "timing.h"
+#include "timing_stm32.h"
+
+#include <libopencm3/cm3/common.h>
+#include <libopencm3/stm32/f1/memorymap.h>
+#include <libopencm3/usb/usbd.h>
+
+#ifdef ENABLE_DEBUG
+#define PLATFORM_HAS_DEBUG
```
dragonmux, line 36: This'll need updating for f4eacf8c (the macro was removed)

src/platforms/stlinkv3/usb_control.c
```diff
@@ -0,0 +1,286 @@
...
+	return packetsize;
+}
+
+static enum usbd_request_return_codes usb_control_request_dispatch(usbd_device *usbd_dev, struct usb_setup_data *req)
+{
+	struct user_control_callback *cb = usbd_dev->user_control_callback;
+
+	/* Call user command hook function. */
+	for (size_t i = 0; i < MAX_USER_CONTROL_CALLBACK; i++) {
+		if (cb[i].cb == NULL) {
```
dragonmux, line 128: Braces are redundant, please remove them. Given this is a null check, perhaps `if (!cb[i].cb)`?

src/platforms/stlinkv3/usb_control.c
```diff
@@ -0,0 +1,286 @@
...
+			break;
+		}
+
+		if ((req->bmRequestType & cb[i].type_mask) == cb[i].type) {
+			const int result = cb[i].cb(usbd_dev, req, &usbd_dev->control_state.ctrl_buf,
+				&usbd_dev->control_state.ctrl_len, &usbd_dev->control_state.complete);
+			if (result == USBD_REQ_HANDLED || result == USBD_REQ_NOTSUPP) {
```
dragonmux, line 135: Same comment about braces

src/platforms/stlinkv3/usb_control.c
```diff
@@ -0,0 +1,286 @@
...
+			usbd_dev->control_state.needs_zlp =
+				needs_zlp(usbd_dev->control_state.ctrl_len, req->wLength, usbd_dev->desc->bMaxPacketSize0);
+			/* Go to data out stage if handled. */
+			usb_control_send_chunk(usbd_dev);
+		} else {
+			/* Go to status stage if handled. */
+			usbd_ep_write_packet(usbd_dev, 0, NULL, 0);
+			usbd_dev->control_state.state = STATUS_IN;
+		}
+	} else {
```
dragonmux, line 162: Extra braces on the else block to be removed please

src/platforms/stlinkv3/usb_control.c
```diff
@@ -0,0 +1,286 @@
...
+	if (req->wLength > usbd_dev->ctrl_buf_len) {
+		stall_transaction(usbd_dev);
+		return;
+	}
+
+	/* Buffer into which to write received data. */
+	usbd_dev->control_state.ctrl_buf = usbd_dev->ctrl_buf;
+	usbd_dev->control_state.ctrl_len = 0;
+	/* Wait for DATA OUT stage. */
+	if (req->wLength > usbd_dev->desc->bMaxPacketSize0) {
```
dragonmux, line 179: Neither block should have braces around them

src/platforms/stlinkv3/usb_control.c
```diff
@@ -0,0 +1,286 @@
...
+	/* Note: this is the only difference from the original
+	 * libopencm3 source code, the setup packet is read here,
+	 * and in the original sources it is expected that the
+	 * packet has already been read into the
+	 * usbd_dev->control_state.req buffer. */
+	if (usbd_ep_read_packet(usbd_dev, 0, req, 8) != 8) {
+		stall_transaction(usbd_dev);
+		return;
+	}
+	if (req->wLength == 0) {
```
dragonmux, line 209: None of the 3 blocks should have braces around them

src/platforms/stlinkv3/usb_dwc_common.c
```diff
@@ -0,0 +1,480 @@
...
+	/*
+	 * Configure endpoint address and type. Allocate FIFO memory for
+	 * endpoint. Install callback function.
+	 */
+	uint8_t dir = addr & 0x80;
+	addr &= 0x7f;
+
+	if (addr == 0) { /* For the default control endpoint */
+		/* Configure IN part. */
+		if (max_size >= 64) {
```
dragonmux, line 97: The braces on these blocks should be dropped as redundant

src/platforms/stlinkv3/usb_dwc_common.c
```diff
@@ -0,0 +1,480 @@
...
+			REBASE(OTG_DIEPCTL0) = OTG_DIEPCTL0_MPSIZ_64;
+		} else if (max_size >= 32) {
+			REBASE(OTG_DIEPCTL0) = OTG_DIEPCTL0_MPSIZ_32;
+		} else if (max_size >= 16) {
+			REBASE(OTG_DIEPCTL0) = OTG_DIEPCTL0_MPSIZ_16;
+		} else {
+			REBASE(OTG_DIEPCTL0) = OTG_DIEPCTL0_MPSIZ_8;
+		}
+
+		REBASE(OTG_DIEPTSIZ0) = (max_size & OTG_DIEPSIZ0_XFRSIZ_MASK);
```
dragonmux, line 107: Parens around the & expression are redundant and should be removed

src/platforms/stlinkv3/usb_dwc_common.c
```diff
@@ -0,0 +1,480 @@
...
+		usbd_dev->fifo_mem_top_ep0 = usbd_dev->fifo_mem_top;
+
+		return;
+	}
+
+	if (dir) {
+		REBASE(OTG_DIEPTXF(addr)) = ((max_size / 4) << 16) | usbd_dev->fifo_mem_top;
+		usbd_dev->fifo_mem_top += max_size / 4;
+
+		REBASE(OTG_DIEPTSIZ(addr)) = (max_size & OTG_DIEPSIZ0_XFRSIZ_MASK);
```
dragonmux, line 127: Same with this

src/platforms/stlinkv3/usb_dwc_common.c
```diff
@@ -0,0 +1,480 @@
...
+		REBASE(OTG_DIEPCTL(addr)) |=
+			/*OTG_DIEPCTL0_EPENA |*/ OTG_DIEPCTL0_SNAK | (type << 18) | OTG_DIEPCTL0_USBAEP | OTG_DIEPCTLX_SD0PID |
+			(addr << 22) | max_size;
+
+		if (callback) {
```
dragonmux, line 132: Braces [we'll let you clean up the rest of this file for this lint rather than keeping writing it out]

src/platforms/stlinkv3/usb_dwc_common.c
```diff
@@ -0,0 +1,480 @@
...
+			usbd_dev->user_callback_ctr[addr][USB_TRANSACTION_OUT] = (void *)callback;
+		}
+	}
+}
+
+void dwc_endpoints_reset(usbd_device *usbd_dev)
+{
+	int i;
```
dragonmux, line 151: C89 decl and size_t loop counter

src/platforms/stlinkv3/usb_dwc_common.c
```diff
@@ -0,0 +1,480 @@
...
+{
+	if (addr == 0) {
+		if (stall) {
+			REBASE(OTG_DIEPCTL(addr)) |= OTG_DIEPCTL0_STALL;
+		} else {
+			REBASE(OTG_DIEPCTL(addr)) &= ~OTG_DIEPCTL0_STALL;
+		}
+	}
+
+	if (addr & 0x80) {
```
dragonmux, line 179: Please suffix the hex constants here with U to prevent integral conversions

src/platforms/stlinkv3/usb_dwc_common.c
```diff
@@ -0,0 +1,480 @@
...
+			REBASE(OTG_DOEPCTL(addr)) &= ~OTG_DOEPCTL0_STALL;
+			REBASE(OTG_DOEPCTL(addr)) |= OTG_DOEPCTLX_SD0PID;
+		}
+	}
+}
+
+uint8_t dwc_ep_stall_get(usbd_device *usbd_dev, uint8_t addr)
+{
+	/* Return non-zero if STALL set. */
+	if (addr & 0x80) {
```
dragonmux, line 201: same here

src/platforms/stlinkv3/usb_dwc_common.c
```diff
@@ -0,0 +1,480 @@
...
+
+#if defined(__ARM_ARCH_6M__)
+	/* Take care of word-aligned and non-word-aligned buffers */
+	if (((uint32_t)buf8 & 0x3) == 0) {
+		for (i = len; i > 0; i -= 4) {
+			REBASE(OTG_FIFO(addr)) = *buf32++;
+		}
+	} else {
+		for (i = len; i > 0; i -= 4) {
+			memcpy(&word32, buf8, 4);
```
dragonmux, line 300: word32's declaration should be moved down into this loop

src/platforms/stlinkv3/usb_dwc_common.c
```diff
@@ -0,0 +1,480 @@
...
+}
+
+uint16_t dwc_ep_write_packet(usbd_device *usbd_dev, uint8_t addr, const void *buf, uint16_t len)
+{
+	const uint32_t *buf32 = buf;
+#if defined(__ARM_ARCH_6M__)
+	const uint8_t *buf8 = buf;
+	uint32_t word32;
+#endif /* defined(__ARM_ARCH_6M__) */
+	int i;
```
dragonmux, line 231: Declarations of `i` should occur in the loops they're needed in

src/platforms/stlinkv3/usb_dwc_common.c
```diff
@@ -0,0 +1,480 @@
...
+	 * prior to writing the data to the usb core fifo memory.
+	 */
+	uint32_t saved_interrupt_mask = REBASE(OTG_GINTMSK);
+	REBASE(OTG_GINTMSK) = 0;
+	REBASE(OTG_DIEPCTL(addr)) |= OTG_DIEPCTL0_EPENA | OTG_DIEPCTL0_CNAK;
+
+	/* Copy buffer to endpoint FIFO, note - memcpy does not work.
+	 * ARMv7M supports non-word-aligned accesses, ARMv6M does not. */
+#if defined(__ARM_ARCH_7M__) || defined(__ARM_ARCH_7EM__)
+	for (i = len; i > 0; i -= 4) {
```
dragonmux, line 287: This and several other of the loops here would be better served and less confusing to be written in the form
```c
for (uint16_t i =0; i < len; i += 4)
```
which would also be less bugprone. let's change them over.

src/platforms/stlinkv3/usb_dwc_common.c
```diff
@@ -0,0 +1,480 @@
...
+	len = MIN(len, packet->packet_length);
+	packet->is_packet_present = false;
+	memcpy(buf, packet->packet_data, len);
+	return len;
+}
+
+/* TODO: this does not currently handle the case for __ARM_ARCH_6M__ */
+void dwc_ep_read_packet_internal(usbd_device *usbd_dev, int ep)
+{
+	int i;
```
dragonmux, line 326: C89 decls

`i` should be declared just before the loop it's for given the loop condition followed by checking it after the loop

src/platforms/stlinkv3/usb_dwc_common.c
```diff
@@ -0,0 +1,480 @@
...
+	struct incoming_packet *packet = stashed_packets + ep;
+	uint32_t *buf32 = (uint32_t *)packet->packet_data;
+	uint32_t extra;
+	uint16_t len = sizeof packet->packet_data;
```
dragonmux, line 330: `const uint16_t len = MIN(sizeof(packet->packet_data), usbd_dev->rxbcnt);` please

src/platforms/stlinkv3/usb_f723.c
```diff
@@ -0,0 +1,178 @@
...
+	// ??? What is this??? It is not documented in the manual
+	OTG_HS_GCCFG |= OTG_GCCFG_PHYHSEN;
+
+	/* ??? The st header files have this:
+	 * #define USB_HS_PHYC_LDO_ENABLE                   USB_HS_PHYC_LDO_DISABLE
+	 * ...go figure...
+	 */
+	OTG_HS_PHYC_LDO |= OTG_PHYC_LDO_DISABLE;
+	while (!(OTG_HS_PHYC_LDO & OTG_PHYC_LDO_STATUS))
+		;
```
dragonmux, line 123: Lots of loops needing `continue` statements added

src/platforms/stlinkv3/usb_f723.c
```diff
@@ -0,0 +1,178 @@
...
+	/* This setting is for a HSE clock of 25 MHz. */
+	OTG_HS_PHYC_PLL1 = 5 << 1;
+	OTG_HS_PHYC_TUNE |= 0x00000f13U;
+	OTG_HS_PHYC_PLL1 |= OTG_PHYC_PLL1_ENABLE;
+	/* 2ms Delay required to get internal phy clock stable
+	 * Used by DFU too, so platform_xxx not available.
+	 * Some Stlinkv3-Set did not cold start w/o the delay
+	 */
+	volatile int i = 200 * 1000;
```
dragonmux, line 132: Let's turn this into a for loop with a `volatile size_t` loop counter counting up to the value assigned here

src/platforms/stlinkv3/usbdfu.c
```diff
@@ -0,0 +1,96 @@
...
+	rcc_periph_clock_enable(RCC_GPIOA);
+	/* On the Mini, NRST is on the footprint for the 1.27 mm Jumper
+	 * to the side of the USB connector.
+	 * With debugger connected, ignore reset. Use debugger to enter!
+	 */
+	bool force_bootloader;
+	force_bootloader = (!(SCS_DHCSR & SCS_DHCSR_C_DEBUGEN) && ((RCC_CSR & RCC_CSR_RESET_FLAGS) == RCC_CSR_PINRSTF));
+	RCC_CSR |= RCC_CSR_RMVF;
+	RCC_CSR &= ~RCC_CSR_RMVF;
+	if (force_bootloader || ((magic[0] == BOOTMAGIC0) && (magic[1] == BOOTMAGIC1))) {
```
dragonmux, line 51: the extra inner parens around each `magic[N]` check should be removed as redundant

src/platforms/stm32/dfucore.c
```diff
@@ -33,9 +33,15 @@
...
 #elif defined(STM32F4) || defined(STM32F7)
-#define DFU_IFACE_PAGESIZE      128
+#define DFU_IFACE_PAGESIZE 128
+#if APP_START == 0x08020000
+#define DFU_IFACE_STRING_OFFSET 62
+#define DFU_IFACE_STRING        "@Internal Flash   /0x08000000/1*016Ka,3*016Ka,1*064Ka,1*128Kg,002*128Kg"
+#elif APP_START == 0x08004000
 #define DFU_IFACE_STRING_OFFSET 54
 #define DFU_IFACE_STRING        "@Internal Flash   /0x08000000/1*016Ka,3*016Kg,1*064Kg,000*128Kg"
+#else
```
dragonmux, line 43: Empty `#else` block can be dropped

src/platforms/stm32/traceswoasync_f723.c
```diff
@@ -0,0 +1,140 @@
...
+/* Packets arrived from the SWO interface */
+static uint8_t trace_rx_buf[NUM_TRACE_PACKETS * TRACE_ENDPOINT_SIZE];
+/* Packet pingpong buffer used for receiving packets */
+static uint8_t pingpong_buf[2 * TRACE_ENDPOINT_SIZE];
+/* SWO decoding */
+static bool decoding = false;
+
+void trace_buf_drain(usbd_device *dev, uint8_t ep)
+{
+	static volatile char inBufDrain;
```
dragonmux, line 54: This should be changed to use C11 atomics instead of compiler built-ins please.

https://en.cppreference.com/w/c/atomic/atomic_flag provides the reference required for this particular pattern. You can also find it in use in https://github.com/blackmagic-debug/blackmagic/blob/c6890e120d9bc1761095fc90820c646f22e8af52/src/platforms/stm32/traceswoasync.c#L57

src/platforms/stm32/traceswoasync_f723.c
```diff
@@ -0,0 +1,140 @@
...
+
+	/* If we are already in this routine then we don't need to come in again */
+	if (__atomic_test_and_set(&inBufDrain, __ATOMIC_RELAXED))
+		return;
+	/* Attempt to write everything we buffered */
+	if (w != r) {
+		uint16_t rc;
```
dragonmux, line 61: `rc` -> `result` or similar?

src/platforms/stm32/traceswoasync_f723.c
```diff
@@ -0,0 +1,140 @@
...
+#include "traceswo.h"
+
+#include <libopencmsis/core_cm3.h>
+#include <libopencm3/cm3/nvic.h>
+#include <libopencm3/stm32/timer.h>
+#include <libopencm3/stm32/rcc.h>
+#include <libopencm3/stm32/usart.h>
+#include <libopencm3/stm32/dma.h>
+
+static volatile uint32_t w; /* Packet currently received via UART */
```
dragonmux, line 43: Let's please use proper names for these so the comments are redundant to understanding what's going on here? https://github.com/blackmagic-debug/blackmagic/blob/c6890e120d9bc1761095fc90820c646f22e8af52/src/platforms/stm32/traceswoasync.c#L46-L47 provides a suitable template for how that might look

src/platforms/stm32/traceswoasync_f723.c
```diff
@@ -0,0 +1,140 @@
...
+	rcc_periph_clock_enable(RCC_GPIOD);
+	gpio_mode_setup(SWO_UART_PORT, GPIO_MODE_AF, GPIO_PUPD_NONE, SWO_UART_RX_PIN);
+	gpio_set_af(SWO_UART_PORT, SWO_UART_PIN_AF, SWO_UART_RX_PIN);
+	/* Pull SWO pin high to keep open SWO line ind uart idle state!*/
+	gpio_set(SWO_UART_PORT, SWO_UART_RX_PIN);
+	nvic_set_priority(SWO_DMA_IRQ, IRQ_PRI_SWO_DMA);
+	nvic_enable_irq(SWO_DMA_IRQ);
+	traceswo_setspeed(baudrate);
+	traceswo_setmask(swo_chan_bitmask);
+	decoding = (swo_chan_bitmask != 0);
```
dragonmux, line 139: The parens here are redundant and should be dropped

dragonmux (changes requested):
Sorry there are still quite a large number of review items! Same deal as last time though and we have now done a detailed review of all the files present in the change set, so we are making progress. Very happy with the clean-up done so far, at this rate this should be ready for merge later this week :)

src/platforms/stlinkv3/platform.c
```diff
@@ -0,0 +1,233 @@
...
+#define SCB_CCSIDR_ASSOCIATIVITY_SHIFT 3U                                     /*!< SCB CCSIDR: Associativity Position */
+#define SCB_CCSIDR_ASSOCIATIVITY_MASK  (0x3ffUL << SCB_CCSIDR_ASSOCIATIVITY_SHIFT) /*!< SCB CCSIDR: Associativity Mask */
+#define CCSIDR_WAYS(x)                 (((x)&SCB_CCSIDR_ASSOCIATIVITY_MASK) >> SCB_CCSIDR_ASSOCIATIVITY_SHIFT)
+#define CCSIDR_SETS(x)                 (((x)&SCB_CCSIDR_NUMSETS_MASK) >> SCB_CCSIDR_NUMSETS_SHIFT)
+#define SCB_DCISW_SET_SHIFT            5U                               /*!< SCB DCISW: Set Position */
+#define SCB_DCISW_SET_MASK             (0x1ffUL << SCB_DCISW_SET_SHIFT) /*!< SCB DCISW: Set Mask */
+#define SCB_DCISW_WAY_SHIFT            30U                              /*!< SCB DCISW: Way Position */
+#define SCB_DCISW_WAY_MASK             (3UL << SCB_DCISW_WAY_SHIFT)     /*!< SCB DCISW: Way Mask */
+
+static void cm_dsb(void)
```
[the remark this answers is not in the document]
jamesturton: The `dsb` and `isb` instructions are memory barriers for ensuring execution concurrency. They are actually provided as macros by CMSIS, hence the old syntax of the functions. I've now added some comments with what they do, which I hope help.

... And I hear you ask ... "But we are on a single core chip, why do we need to ensure execution concurrency?" Well due to the inherent memory ordering of all loads and stores on  Cortex-M0,  Cortex-M0+, Cortex-M3 and Cortex-M4 we don't have to have to worry about memory barrier instructions on these platform. But as this Cortex-M7 device with the Art Accelerator, all bets are off, thus the memory barriers are required.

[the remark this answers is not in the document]
dragonmux: Ah, excellent, so to borrow from C++, this is the C-available way to implement https://en.cppreference.com/w/cpp/thread/barrier

Agreed that memory ordering is important especially to prevent the compiler re-ordering or an interrupt screwing things up here

src/platforms/stlinkv3/platform.c
jamesturton: What you suggest looks much better! I assume there is no issue with invalidating the cache set-ways in the reverse order - actually that must be true as the DSB comes after the `do...while` loop in the original code.

src/platforms/stlinkv3/usb_control.c
jamesturton: IMO it would be much better for the files `usb_dwc_common.c`, `usb_control.c` and `usb_f723.c` to be merged upstream in libopencm3. Therefore I am a little hesitant to modify them too far away from their existing form so it is easier to diff the original file. That being said, it seems that development of libopencm3 has practically ground to a halt, and, given the backlog of PRs over there, I do not feel that confident that changes would be merged in a timely manner, so maybe maintaining our own versions of these files is the easiest thing to do in the long term? Does @esden have any input regarding this perhaps?

src/platforms/stlinkv3/usb_dwc_common.c
jamesturton: It's written this way to support non-word-aligned accesses. We check the remainder of `i` after the for loop to sweep up any leftover bytes

jamesturton: Oops, I was looking in the `ep_read_packet_internal()` function. This one seems ok to change

jamesturton:
After looking, there seem to be plenty of existing PRs to add USB HS support for the f7 series waiting to be reviewed in libopencm3, so it looks like we are stuck with that code living here for now.

I've cleaned up some of the mess of those libopencm3 patches so that the platform only relies on a new usb_f723 driver - `usb_dwc_common.c` and `usb_control.c` no longer need to be patched. I've also tried to reuse as many of the dwc_common implementations as possible to try to keep it readable.

dragonmux:
Excellent, we'll re-review later today then.

@esden: sounds like it'll be worth, when you've got some cycles spare, reviewing the locm3 PRs for the F7 series, identifying the most complete/best of them, getting that merged, and then us submitting a patch from James' work to get the USB implementation up-streamed and out of this platform.

dragonmux (changes requested):
This is looking great, think with this round of review items fixed we should be ready for merge

jamesturton:
Hopefully that's everything addressed that you picked out in your review! Thanks for taking the time to review such a colossal PR :smile:

src/platforms/stlinkv3/usb_control.c
dragonmux: We'll leave this unresolved for Esden to see and respond to, but it's non-blocking to merge

dragonmux (approved):
LGTM, merging. Many thanks James for sticking with this and helping sort out this giant PR! No easy thing when you weren't the person to write the code in the first place.

jamesturton:
> > @esden: sounds like it'll be worth, when you've got some cycles spare, reviewing the locm3 PRs for the F7 series, identifying the most complete/best of them, getting that merged, and then us submitting a patch from James' work to get the USB implementation up-streamed and out of this platform.
> 
> I've added two commits to my branch. The first is just [adding more register definitions](https://github.com/jamesturton/libopencm3/commit/b13ff8b1092f94cb780c11f5b47128b676d958d6) which I really hope won't cause too many objections to merging and will hopefully be easy to review.
> 
> The second commit [adds the usb_f732 driver](https://github.com/jamesturton/libopencm3/commit/42a2c4c3ccc8c53fe0c4f7b141e749e0f1c5a114) included in this PR and doesn't modify any of the other code. I hope this way it is more acceptable to be merged as not to change any existing usb implementations, but just add the USB HS support for this chip.

@esden I have opened a PR to upstream the stm32f7 HS USB driver in locm3 https://github.com/libopencm3/libopencm3/pull/1478 if you had the possibility to review this at some point. It would be great if this could be upstreamed :smile:

Issue #1353, opened 2023-01-05
Can't attach to Access Port on nRF52


giobauermeister:
I'm doing some tests with v1.9.0-rc1 and found that I can't attach to Access Port on nRF52.

I tested v1.8.2 https://github.com/blackmagic-debug/blackmagic/commit/5cd430647ecfb6dab4a2ed858fb98567723de699 and v1.9.0-rc0 https://github.com/blackmagic-debug/blackmagic/commit/14e9df69fa44f44a4f3906db581d1289c2329bc6. They seem to be working fine.

```
(gdb) target extended-remote /dev/ttyBmpGdb
Remote debugging using /dev/ttyBmpGdb
(gdb) monitor swdp_scan
Target voltage: 3.18V
Available Targets:
No. Att Driver
 1      Nordic nRF52 M4
 2      Nordic nRF52 Access Port 
(gdb) att 2
Attaching to Remote target
Ignoring packet error, continuing...
```

zyp:
I can reproduce this with BMDA:

```
(gdb) target ext :2000
Remote debugging using :2000
(gdb) mon swdp_scan
Available Targets:
No. Att Driver
 1      Nordic nRF52 M4
 2      Nordic nRF52 Access Port 
(gdb) attach 2
Attaching to Remote target
Remote connection closed
```

BMDA is segfaulting:
```
Black Magic Debug App v1.9.0-rc1
 for Black Magic Probe, ST-Link v2 and v3, CMSIS-DAP, JLink and libftdi/MPSSE
Using 1209:3443 2FCC8092A2189006 Orbcode
 Orbtrace
Listening on TCP port: 2000
scan_multidrop: false
Segmentation fault: 11
```

dragonmux:
Ack, we'll check the other special AP functions (this isn't the only one) and verify they set the target structure up properly. We're surprised this worked before as we replaced the tdesc stuff like-for-like - we'll dive into a suitable way to fix this, even if that ends up being an if guard in target_regs_description:

https://github.com/blackmagic-debug/blackmagic/blob/98915144944d46cb2dc1daedc734b0134dbde7fb/src/target/target.c#L471-L474

zyp:
> We're surprised this worked before as we replaced the tdesc stuff like-for-like…

It had an if-guard in the form of a ternary before:

```c
return t->tdesc ? t->tdesc : "";
```

dragonmux:
[...]

PR #1356, opened 2023-01-06
Fix: target default functions undefined behaviour

dragonmux:
## Detailed description

In this PR we address the secondary problem encountered by zyp in https://github.com/blackmagic-debug/blackmagic/issues/1353

This issue is caused by the undefined and strictly forbidden casting of incompatible function pointer types to assign to the target structure function pointers as defaults. The net result is that on platforms where this visibly matters, it generates segfaults/crashes. On other platforms it generates unpredictable and unreliable behaviour.

This has been fixed by introducing a couple more no-op functions with the correct return types to use, and assigning suitable ones (without cast, where possible) to the various function pointers.

### DIFF
--- a/src/target/target.c
+++ b/src/target/target.c
@@ -497,9 +497,9 @@ unsigned int target_part_id(target_s *t)
 
 uint32_t target_mem_read32(target_s *t, uint32_t addr)
 {
-	uint32_t ret;
-	t->mem_read(t, &ret, addr, sizeof(ret));
-	return ret;
+	uint32_t result = 0;
+	t->mem_read(t, &result, addr, sizeof(result));
+	return result;
 }
 
 void target_mem_write32(target_s *t, uint32_t addr, uint32_t value)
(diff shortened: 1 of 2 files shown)


src/target/target.c
```diff
@@ -64,19 +79,20 @@ target_s *target_new(void)
...
+	t->mem_read = target_nop_void;
+	t->mem_write = target_nop_void;
+	t->reg_read = target_nop_zero;
+	t->reg_write = target_nop_zero;
+	t->regs_read = target_nop_void;
+	t->regs_write = target_nop_void;
+	t->reset = target_nop_void;
+	t->halt_request = target_nop_void;
+	t->halt_poll = target_nop_poll;
+	/* XXX: For some reason, if the cast is omitted, this one line generates a type mismatch error. */
```
zyp, line 93: I suspect it's an ABI thing. Functions with undefined argument lists can be passed to function pointers taking integral arguments, but not `bool` or `float`, so I'm guessing it warns when the argument has a type that's not necessarily compatible with the integral ABI.

Example: https://godbolt.org/z/9bf8rG9sv

While we're at it, it might be worthwhile to just make default functions that's taking the actual argument types.

dragonmux: Good point regarding bool/float.. thank you for giving us a working reason as to why we saw the behaviour we did. Regarding actual types, we thought about that and for the amount of extra code that generates, it would be better to entirely get rid of these default implementations and instead do NULL checks in the target API if we go there.. this already increased firmware size a fair bit. At least with argument-less functions the stack is properly dealt with which is why this current implementation works.

dragonmux: We've replaced the approach with proper NULL checking and fixed some extra bugs in the target memory functions in the process :)

esden (approved):
LGTM, merging

PR #1361, opened 2023-01-10
Feature: qStartNoAckmode

perigoso:
## Detailed description

This is the start of no acknowledge mode (**QStartNoAckMode**) support.

The BMP uses a USB connection with the CDC class (Bulk transfers) for GDB data transfer, this means the connection *should* be reliable, with integrated integrity checks and data delivery guarantees (provided there are no bugs in the class handling side of things in the firmware), this means that acknowledgment and checksum verification is redundant and should be safe to ignore, hopefully providing a performance benefit, GDB provides support for this with **QStartNoAckMode**.

See [GDB: Packet-Acknowledgment](https://sourceware.org/gdb/onlinedocs/gdb/Packet-Acknowledgment.html#Packet-Acknowledgment) for context.

The intent is to test if this brings improvements in performance, particularly in the flashing routines.

As of right now, this consists only of refactoring of the gdb_getpacket to make use of a state machine, in preparation of the actual implementation, reviews are not necessary but feedback is always welcome, shallow testing showed this to work, so testing is also welcome

### DIFF
--- a/src/gdb_packet.c
+++ b/src/gdb_packet.c
@@ -44,6 +44,17 @@ static bool noackmode = false;
 /* https://sourceware.org/gdb/onlinedocs/gdb/Packet-Acknowledgment.html */
 void gdb_set_noackmode(bool enable)
 {
+	/* 
+	 * If we were asked to disable NoAckMode, and it was previously enabled,
+	 * it might mean we got a packet we determined to be the first of a new
+	 * GDB session, and as such it was not acknoledged (before GDB enabled NoAckMode),
+	 * better late than never.
+	 * 
+	 * If we were asked after the connection was terminated, sending the ack will have no effect.
+	 */
+	if (!enable && noackmode)
+		gdb_if_putchar(GDB_PACKET_ACK, 1U);
+
 	DEBUG_GDB("%s NoAckMode\n", enable ? "Enabling" : "Disabling");
 	noackmode = enable;
 }
--- a/src/gdb_packet.c
+++ b/src/gdb_packet.c
@@ -31,124 +31,196 @@
 
 #include <stdarg.h>
 
-void consume_remote_packet(char *const packet, const size_t size)
+typedef enum packet_state {
+	PACKET_IDLE,
+	PACKET_GDB_CAPTURE,
+	PACKET_GDB_ESCAPE,
+	PACKET_GDB_CHECKSUM_UPPER,
+	PACKET_GDB_CHECKSUM_LOWER,
+} packet_state_e;
+
+packet_state_e consume_remote_packet(char *const packet, const size_t size)
 {
 #if PC_HOSTED == 0
 	/* We got what looks like probably a remote control packet */
 	size_t offset = 0;
-	bool getting_remote_packet = true;
-	while (getting_remote_packet) {
+	while (true) {
 		/* Consume bytes until we either have a complete remote control packet or have to leave this mode */
-		const char c = gdb_if_getchar();
-		switch (c) {
-		case REMOTE_SOM: /* Oh dear, packet restarts */
+		const char rx_char = gdb_if_getchar();
+
+		switch (rx_char) {
+		case '\x04':
+			/* EOT (end of transmission) - connection was closed */
+			return PACKET_IDLE;
+
+		case REMOTE_SOM:
+			/* Oh dear, restart remote packet capture */
 			offset = 0;
 			break;
 
-		case REMOTE_EOM: /* Complete packet for processing */
-			packet[offset] = 0;
-			remote_packet_process(offset, packet);
-			getting_remote_packet = false;
-			break;
+		case REMOTE_EOM:
+			/* Complete packet for processing */
 
-		case GDB_PACKET_START: /* A 'real' gdb packet, best stop squatting now */
-			packet[0] = GDB_PACKET_START;
-			getting_remote_packet = false;
-			break;
+			/* Null terminate packet */
+			packet[offset] = '\0';
+			/* Handle packet */
+			remote_packet_process(offset, packet);
+
+			/* Restart packet capture */
+			return PACKET_IDLE;
+
+		case GDB_PACKET_START:
+			/* A 'real' gdb packet, best stop squatting now */
+			return PACKET_GDB_CAPTURE;
 
 		default:
 			if (offset < size)
-				packet[offset++] = c;
+				packet[offset++] = rx_char;
 			else
-				/* Who knows what is going on...return to normality */
-				getting_remote_packet = false;
-			break;
+				/* Buffer overflow, restart packet capture */
+				return PACKET_IDLE;
 		}
 	}
-#endif
+#else
+	(void)packet;
 	(void)size;
-	/*
-	 * Reset the packet buffer start character to zero, because function
-	 * 'remote_packet_process()' above overwrites this buffer, and
-	 * an arbitrary character may have been placed there. If this is a '$'
-	 * character, this will cause this loop to be terminated, which is wrong.
-	 */
-	packet[0] = '\0';
+
+	/* Hosted builds ignore remote control packets */
+	return PACKET_IDLE;
+#endif
 }
 
 size_t gdb_getpacket(char *const packet, const size_t size)
 {
-	unsigned char csum;
-	char recv_csum[3];
+	packet_state_e state = PACKET_IDLE; /* State of the packet capture */
+
 	size_t offset = 0;
+	uint8_t checksum = 0;
+	uint8_t rx_checksum = 0;
 
 	while (true) {
-		char start_char = '\0';
-		/* Wait for packet start */
-		while (start_char != GDB_PACKET_START && start_char != REMOTE_SOM) {
-			start_char = gdb_if_getchar();
-			if (start_char == '\x04')
-				return 1;
-		}
-		packet[0] = start_char;
-
-		/* If the packet appears to start a remote control message, consume it */
-		if (start_char == REMOTE_SOM) {
-			consume_remote_packet(packet, size);
-			continue;
+		const char rx_char = gdb_if_getchar();
+		if (rx_char == '\x04') {
+			/* EOT (end of transmission) - connection was closed */
+			packet[0] = '\x04';
+			packet[1U] = 0;
+			return 1U;
 		}
 
-		/* Handle a normal GDB packet */
-		offset = 0;
-		csum = 0;
-		char c = '\0';
-		/* Capture packet data into buffer */
-		while (c != GDB_PACKET_END) {
-			c = gdb_if_getchar();
-			if (c == GDB_PACKET_END)
-				break;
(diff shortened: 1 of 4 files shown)


perigoso:
QStartNoAckmode now implemented, testing didn't show any noticeable difference in speeds, but further testing is needed and welcome

src/gdb_packet.c
```diff
@@ -31,121 +31,163 @@
 
 #include <stdarg.h>
 
+typedef enum packet_state_e {
```
dragonmux, line 34: the _e on the end here is wrong (but correct below)

src/gdb_packet.c
```diff
@@ -31,121 +31,163 @@
...
-			 * Spin waiting for a start of packet character - either a gdb
-             * start ('$') or a BMP remote packet start ('!').
-			 */
-			do {
-				/* Smells like bad code */
-				packet[0] = gdb_if_getchar();
-				if (packet[0] == '\x04')
-					return 1;
-			} while (packet[0] != '$' && packet[0] != REMOTE_SOM);
+		const char c = gdb_if_getchar();
```
dragonmux, line 57: We're going to be interested to see how this interacts with https://github.com/blackmagic-debug/blackmagic/tree/fix/remote-packet-handling-organisation

dragonmux (changes requested):
A quick check over reveals just one style mistake and something we need to consider between this branch and another that we're contemplating for v1.10. Generally though this looks great and improves the GDB packet handler handily

src/gdb_packet.c
perigoso: ups, that was the find and replace

src/gdb_packet.c
perigoso: ah, didn't realize there was pending work on this, could you clear up what you mean by "interacts with"? I think these are conflicting changes

dragonmux: They are currently conflicting, but we're thinking if there's a way to make either that patch play nice with this, or this play nice with that.. abstracting out the remote protocol handling like that branch does is something we'd been considering a while and we think helps even with no-ack mode as the remote protocol is not and should not be influenced by no-ack mode

perigoso: ah, I follow,  I will take a look eventually, though as is no ack mode doesn't afect the remote protocol at all

perigoso:
@dragonmux rebase over your organization commit, theres still one issue to solve and adding a buildsystem option to enable this feature, but its in a ok state to star rev

src/gdb_packet.c
```diff
@@ -31,119 +31,220 @@
...
+{
+	/* 
+	 * If we were asked to disable no ack mode, and it was previously enabled,
+	 * it might mean we got a paket we determined to be the first of a new
+	 * GDB session, and as such it was not acknoledged (before GDB enabled no ack mode),
+	 * better late than never.
+	 * 
+	 * If we were asked after the connection was terminated, sending the ack have no effect.
+	 */
+	if (enable == false && noackmode == true) {
```
dragonmux, line 55:
```suggestion
	if (!enable && noackmode)
```
and drop the extra braces

src/gdb_packet.c
```diff
@@ -31,119 +31,220 @@
...
+		gdb_if_putchar(GDB_PACKET_ACK, 1U);
+	}
+
+	DEBUG_GDB("%s no ack mode\n", enable ? "Enable" : "Disable");
+
```
dragonmux, line 60: Please remove the extra new line - these two statements feel like they group together fine

src/gdb_packet.c
```diff
@@ -31,119 +31,220 @@
...
+	noackmode = enable;
+}
+
+packet_state_e consume_remote_packet(char *const packet, const size_t size)
+{
+	/* We got what looks like probably a remote control packet */
+
+#if PC_HOSTED == 0
```
dragonmux, line 68: Putting the `#if `check above the comment makes more sense as the entire routine is turned off by PC_HOSTED==1

src/gdb_packet.c
```diff
@@ -31,119 +31,220 @@
...
+			/* Oh dear, restart remote packet capture */
+			offset = 0;
+			break;
+
+		case REMOTE_EOM:
+			/* Complete packet for processing */
+
+			/* Null terminate packet */
+			packet[offset] = '\0';
+
```
dragonmux, line 90: Think it's ok to drop this extra new line too as calling into remote_packet_process() feels grouped to finishing setting up the buffer for the call

src/gdb_packet.c
```diff
@@ -31,119 +31,220 @@
...
+}
+
 size_t gdb_getpacket(char *const packet, const size_t size)
 {
-	unsigned char csum;
-	char recv_csum[3];
+	packet_state_e state = PACKET_IDLE; /* State of the packet capture */
+
 	size_t offset = 0;
+	uint8_t rx_checksum = 0, checksum = 0;
```
dragonmux, line 123: Think it's a clang-tidy lint too, but one definition per line please

src/gdb_packet.c
```diff
@@ -31,119 +31,220 @@
...
-			do {
-				/* Smells like bad code */
-				packet[0] = gdb_if_getchar();
-				if (packet[0] == '\x04')
-					return 1;
-			} while (packet[0] != '$' && packet[0] != REMOTE_SOM);
-#if PC_HOSTED == 0
-			if (packet[0] == REMOTE_SOM) {
-				/* This is probably a remote control packet - get and handle it */
+		// const char rx_char = state != PACKET_GDB_COMPLETE ? gdb_if_getchar() : '\0';
```
dragonmux, line 126: Please drop the commented out line of code

src/gdb_packet.c
```diff
@@ -31,119 +31,220 @@
...
+			packet[1U] = 0;
+			return 1U;
+		}
+
+		switch (state) {
+		case PACKET_IDLE:
+			if (rx_char == GDB_PACKET_START) {
+				/* Start of GDB packet */
+				state = PACKET_GDB_CAPTURE;
 				offset = 0;
```
dragonmux, line 140: It looks like offset and checksum should be set 0 above the if block here inside the case so that things wind up properly reset if the remote protocol handler returns us back in state PACKET_GDB_CAPTURE

src/gdb_packet.c
```diff
@@ -31,119 +31,220 @@
...
-				csum += c + '}';
-				packet[offset++] = (char)((uint8_t)c ^ 0x20U);
-				continue;
+			if (rx_char == GDB_PACKET_END) {
+				/* End of GDB packet */
+				// if (noackmode)
+				// 	/* knowlingly ignore the checksum that follows in no ack mode (as per spec) */
+				// 	state = PACKET_GDB_COMPLETE;
+				// else
+				/* Move to checksum capture */
```
dragonmux, line 169: This should probably be changed to a block comment and the commented out dead code needs dropping

src/gdb_packet.c
```diff
@@ -31,119 +31,220 @@
...
+			/* Resolve escaped char */
+			packet[offset++] = rx_char ^ GDB_PACKET_ESCAPE_XOR;
+
+			/* Return to normal packet capture */
+			state = PACKET_GDB_CAPTURE;
+			break;
+
+		case PACKET_GDB_CHECKSUM_UPPER:
+			/* Checksum upper nibble */
+			rx_checksum = unhex_digit(rx_char) << 4; /* This also clears the lower nibble */
```
dragonmux, line 199: `4` -> `4U`

src/gdb_packet.c
```diff
@@ -31,119 +31,220 @@
...
+			state = PACKET_GDB_CHECKSUM_LOWER;
+			break;
+
+		case PACKET_GDB_CHECKSUM_LOWER:
+			/* Checksum lower nibble */
+			rx_checksum |= unhex_digit(rx_char); /* BITWISE OR lower nibble with upper nibble */
+
+			// 	state = PACKET_GDB_COMPLETE;
+			// 	break;
+			// case PACKET_GDB_COMPLETE:
```
dragonmux, line 209: Dead code?

dragonmux (changes requested):
This is looking great, nice job integrating our refactor of the remote packet handling!

There are a few items we've spotted, mostly commented out dead code.

src/gdb_packet.c
perigoso: This is part of a temporary commit to get things works, this is why I noted "theres still an issue" this wont be part of the final submission

src/gdb_packet.c
perigoso: This is part of a temporary commit to get things works, this is why I noted "theres still an issue" this wont be part of the final submission

src/gdb_packet.c
perigoso: This is part of a temporary commit to get things works, this is why I noted "theres still an issue" this wont be part of the final submission

src/gdb_packet.c
dragonmux: Ah, excellent. Thank you :)

src/gdb_packet.c
perigoso: you're right, good catch, I even added a comment that was plain wrong

src/gdb_packet.c
```diff
@@ -31,124 +31,222 @@
...
+	 */
+	if (!enable && noackmode) {
+		gdb_if_putchar(GDB_PACKET_ACK, 1U);
+	}
+
+	DEBUG_GDB("%s no ack mode\n", enable ? "Enable" : "Disable");
+	noackmode = enable;
+#else
+	if (enable)
+		DEBUG_ERROR("Tried to set no ack mode but it's not enabled in this build\n");
```
dragonmux, line 67: Given space is at a premium, might we suggest that this is replaced by a `#if` check in `exec_q_noackmode()` which handles the packet like the unhandled packet handler would?

At least for `PC_HOSTED == 0` - this is nicer behaviour when space doesn't matter (BMDA)

dragonmux (commented):
We're happy to take this feature when you're happy with the implementation. Update the PR title and mark it as ready for review when you're ready.

We've only seen the one remaining item having re-checked through the cleanups and fixes you've performed and it is only a suggestion not a requested change.

src/gdb_packet.c
perigoso: I opted to go with a build option that controls whether the feature is *advertised*, it will always be compiled in, meaning we can always manually enable in via GDB, but the default behavior depends on the build option (if advertised the feature will be active by default, otherwise only manually)

This means the feature will incur a fixed code size cost, but it's small enough to not worry about (I calculated 164B)

dragonmux: That works 🙂  and it gives us options for how to use this, including in tooling using the remote protocol.

dragonmux (approved):
LGTM, We'll run a couple of tests locally just to be sure things work again GDB 13, and then we'll get this merged ASAP. Excellent work and an awesome contribution, thank you!

PR #1362, opened 2023-01-17
target: MindMotion MM32

koendv:
## Detailed description
This PR provides initial support for a new target: MindMotion MM32, with initial mcu's: MM32SPIN27, MM32L073, and MM32F3273. Other processors are easily added.

- To add support for a new MM32 Cortex-M0 processor, first check the product id at address 0x40013400.
E.g. MM32L073:
```
(gdb) set mem inaccessible-by-default off
(gdb) x 0x40013400
0x40013400:	0xcc568091
```
then add a new case to the switch statement in file src/target/stm32f1.c, function mm32l0xx_probe() :
```
case 0xcc568091U:
        name = "MM32L07x";
        flash_kbyte = 128;
        ram_kbyte = 8;
        break;
```

- To add support for a new MM32 Cortex-M3 processor, first check the product id at address 0x40007080.
E.g. MM32F3273:
```
(gdb) set mem inaccessible-by-default off
(gdb) x 0x40007080
0x40007080:	0xcc9aa0e7
```
then add a new case to the switch statement in file src/target/stm32f1.c, function mm32f3xx_probe() :
```
case 0xcc9aa0e7U:
        name = "MM32F3270";
        flash_kbyte = 512;
        ram1_kbyte = 128;
        break;
```

If different models exist that have the same id but different amounts of flash or ram, choose the largest amount of flash or ram.

Suggestions welcome.


src/target/adiv5.c
```diff
@@ -1072,6 +1072,41 @@ void firmware_mem_write_sized(adiv5_access_port_s *ap, uint32_t dest, const void
 	adiv5_dp_read(ap->dp, ADIV5_DP_RDBUFF);
 }
 
+void mm32l0_firmware_mem_write_sized(adiv5_access_port_s *ap, uint32_t dest, const void *src, size_t len, align_e align)
```
dragonmux, line 1075: This function should be in stm32f1.c as it's not ADIv5-generic, and drop the `_firmware_` - the naming of several of the functions in this file needs rework as they're wrong and that is part of how. (`firmware_mem_read`, for example, should be `adiv5_mem_read`).

src/target/adiv5.c
```diff
@@ -1072,6 +1072,41 @@ void firmware_mem_write_sized(adiv5_access_port_s *ap, uint32_t dest, const void
...
+{
+	uint32_t odest = dest;
+
+	len >>= align;
+	ap_mem_access_setup(ap, dest, align);
+	while (len--) {
+		uint32_t tmp = 0;
+		/* Pack data into correct data lane */
+		switch (align) {
```
dragonmux, line 1084: Please use the switch-case style of adiv5_pack_data() here. It results in better code and avoids an undefined behaviour (UB) cast + unaligned access to src in the ALIGN_DWORD and ALIGN_WORD cases. (we are aware of the contents of firmware_mem_write_sized(), it needs replacing after renaming.. it's duplicating adiv5_pack_data() for no good reason)

It looks like the only change this needs from that code is to replace ALIGN_HALFWORD with:
```c
uint16_t value;
memcpy(&value, src, sizeof(value));
tmp = value;
```
This will already automatically widen correctly because the types all agree on being unsigned, etc.

src/target/adiv5.h
```diff
@@ -404,6 +404,8 @@ const void *adiv5_pack_data(uint32_t dest, const void *src, uint32_t *data, alig
 void firmware_mem_write_sized(adiv5_access_port_s *ap, uint32_t dest, const void *src, size_t len, align_e align);
 void firmware_mem_read(adiv5_access_port_s *ap, void *dest, uint32_t src, size_t len);
 void firmware_ap_write(adiv5_access_port_s *ap, uint16_t addr, uint32_t value);
+void mm32l0_firmware_mem_write_sized(
```
dragonmux, line 407: Don't forget to remove this when you put the mm32-specific write function into stm32f1.c

src/target/stm32f1.c
```diff
@@ -254,6 +258,87 @@ bool at32fxx_probe(target_s *t)
...
+		ram_kbyte = 12;
+		break;
+	case 0x0:
+	case 0xffffffff:
+		return false;
+	default:
+		DEBUG_WARN("mm32_probe: unknown mm32 dev_id 0x%" PRIx32 "\n", mm32_id);
+		return false;
+	}
+	t->part_id = mm32_id & 0xfff;
```
dragonmux, line 294: Missing "U" on the end of this mask constant

src/target/stm32f1.c
```diff
@@ -254,6 +258,87 @@ bool at32fxx_probe(target_s *t)
...
+		name = "MM32L07x";
+		flash_kbyte = 128;
+		ram_kbyte = 8;
+		break;
+	case 0xcc56a097U:
+		name = "MM32SPIN27";
+		flash_kbyte = 128;
+		ram_kbyte = 12;
+		break;
+	case 0x0:
```
dragonmux, line 287: Please write the constant out in full for consistency and apply "U" markings to both this and the next (especially the next due to sign extension problems!)

src/target/stm32f1.c
```diff
@@ -254,6 +258,87 @@ bool at32fxx_probe(target_s *t)
...
+/* Identify MM32 devices (Cortex-M0) */
+bool mm32l0xx_probe(target_s *t)
+{
+	uint32_t mm32_id;
+	const char *name = "?";
+	size_t flash_kbyte = 0;
+	size_t ram_kbyte = 0;
+	size_t block_size = 0x400;
+
+	if (!t)
```
dragonmux, line 270: This check is not required, t cannot be NULL here. Remove it

src/target/stm32f1.c
```diff
@@ -254,6 +258,87 @@ bool at32fxx_probe(target_s *t)
...
+		return false;
+	default:
+		DEBUG_WARN("mm32_probe: unknown mm32 dev_id 0x%" PRIx32 "\n", mm32_id);
+		return false;
+	}
+	t->part_id = mm32_id & 0xfff;
+	t->driver = name;
+	t->mass_erase = stm32f1_mass_erase;
+	target_add_ram(t, 0x20000000, ram_kbyte * 1024U);
+	stm32f1_add_flash(t, 0x8000000, flash_kbyte * 1024U, block_size);
```
dragonmux, line 298: Please write the extra 0 on the front of the address constant - as written even we had to do a double take and count the 0's to check it wasn't `0x80000000`. Also, please add `U` on the end of both address constants

src/target/stm32f1.c
```diff
@@ -254,6 +258,87 @@ bool at32fxx_probe(target_s *t)
...
+	target_add_commands(t, stm32f1_cmd_list, name);
+	if (cortexm_ap(t) && cortexm_ap(t)->dp)
```
dragonmux, line 300: Turn all use of `cortexm_ap(t)` here into use of a local. This is hard to reason about and needlessly pessimises performance on debug builds. It's also not possible for the adiv5_access_port_t structure pointer to be NULL here, nor the DP one. These are guarantees from the adiv5 driver. Remove the extra checks.

src/target/stm32f1.c
```diff
@@ -254,6 +258,87 @@ bool at32fxx_probe(target_s *t)
...
+/* Identify MM32 devices (Cortex-M3) */
+bool mm32f3xx_probe(target_s *t)
+{
+	uint32_t mm32_id;
+	const char *name = "?";
+	size_t flash_kbyte = 0;
+	size_t ram_kbyte = 0;
+	size_t block_size = 0x400;
+
+	if (!t)
```
dragonmux, line 314: Same comment about how t cannot be NULL

src/target/stm32f1.c
```diff
@@ -254,6 +258,87 @@ bool at32fxx_probe(target_s *t)
...
+		DEBUG_WARN("mm32xx_probe: read error at 0x%" PRIx32 "\n", (uint32_t)DBGMCU_IDCODE_MM32F3);
+		return false;
+	}
+	switch (mm32_id) {
+	case 0xcc9aa0e7U:
+		name = "MM32F3270";
+		flash_kbyte = 512;
+		ram_kbyte = 128;
+		break;
+	case 0x0:
```
dragonmux, line 326: Same comment about writing the constant out consistently and applying `U` annotation

src/target/target_probe.h
```diff
@@ -39,6 +39,8 @@ bool rp_rescue_probe(adiv5_access_port_s *ap);
 
 bool ch32f1_probe(target_s *t);  // will catch all the clones
 bool at32fxx_probe(target_s *t); // STM32 clones from Artery
+bool mm32l0xx_probe(target_s *t);
+bool mm32f3xx_probe(target_s *t);
```
dragonmux, line 43: Please add these to target_probe.c too.

dragonmux (changes requested):
This is looking like a good start. There are some items to be addressed (below), but we'd be happy to merge this with those items fixed in v1.10

src/target/stm32f1.c
```diff
@@ -254,6 +257,136 @@ bool at32fxx_probe(target_s *t)
...
+
+bool mm32l0xx_probe(target_s *t)
+{
+	uint32_t mm32_id;
+	const char *name = "?";
+	size_t flash_kbyte = 0;
+	size_t ram_kbyte = 0;
+	size_t block_size = 0x400U;
+
+	if (target_mem_read(t, &mm32_id, DBGMCU_IDCODE_MM32L0, sizeof(mm32_id))) {
```
dragonmux, line 313: Just so we understand it better, what's the motivation of using target_mem_read() here instead of target_mem_read32() + target_check_error()?

dragonmux (commented):
This looks good to us now. We'll merge this in v1.10. In the meantime, there's one question (below) which popped out as we were reading the patch. It's not requiring a change, we just want to better understand the way the ID code readouts are expressed

koendv:
> Just so we understand it better, what's the motivation of using
> target_mem_read() here instead of target_mem_read32() +
> target_check_error()?
> 

You're right; the addresses where MM32 stores the id code only allow 32-bit access, so target_mem_read32() is more suited.
Fixed.

koen

dragonmux (approved):
Ok, fair enough. Approved for merge when the v1.10 window opens

dragonmux (approved):

PR #1369, opened 2023-01-29
hosted/cmsis_dap: Fixes CMSIS-DAP adapters connected with USB HID.

dmsc:
## Detailed description

When using HID as transport, the responses are always the same length, so we only need to check that the received length is more than or equal that the expected response length.

This fixes debugging with [dap42](https://github.com/devanlai/dap42) adapter, tested with an stm32f103 target.


dragonmux (approved):
This looks good to us on first blush. We'll need to test this locally, but it's a small enough change that we'll try and merge it in time for the v1.9 release this weekend

dragonmux:
After further noodling on this, we've decided to just get this merged for v1.9 as it won't break existing functionality.

A note to self though: We should turn getting more data than expected into a DEBUG_WARN() so we can track when this happens, however this is purely for logging. This new logic seems safe enough as long as ASAN builds don't explode when this condition is triggered

PR #1380, opened 2023-02-11
Feature: Initial RISC-V debug support (v0.13 spec)

### LINKED ISSUE #794: Is it possible to support riscv arch?
Gdb and jtag facility can be reused. And I find that the quality of the code in this project is very high. I think this project have the potential to support more arch.

dragonmux:
## Detailed description

In this PR we implement support for debugging RISC-V targets complying with v0.13.2 of the RISC-V debug specification. This includes support for both rv32 and rv64 targets and intentionally fails to work with rv128 targets (though we still must detect those).

This PR also introduces support for the GD32VF1 series through minimal changes to stm32f1.c, providing a test target to play with.

Known limitations include:

* The target must support both abstract register and abstract memory access commands. Reasonable hook points have been left in riscv32.c and riscv64.c to implement system bus access for memory commands in the future. This will probably be addressed in a follow-up PR as using the system bus is not that difficult and determining if it can be used is a trivial DM register write/read.
* RISC-V debug spec v1.0 targets are likely only partially functional (the DMI is the same, but the DM contains differences not addressed here)
* RISC-V debug spec v0.11 devices will not work with this. The correct specification SiFive used in early devices is referenced but support is not addressed at this time.

This PR replaces #924.

## Closing issues

Fixes #794


esden (approved):
Fantastic! Finally! Thank you!

PR #1384, opened 2023-02-12
Fix: JTAG correctness

dragonmux:
## Detailed description

In this PR we address several JTAG bugs (including a regression from PR #1131) which were impacting the ability for the firmware to correctly talk to RISC-V targets, among others. There are still issues with the timings being very tight on many devices, but this does at least appear to be reliable and gives reasonable clocking waveforms too as shown in the capture below.

image

This blocks PR #1380 till merged and must be applied on top of that PR if wishing to test RISC-V with BMP or other BMD firmware platforms.

### DIFF
--- a/src/platforms/common/jtagtap.c
+++ b/src/platforms/common/jtagtap.c
@@ -48,10 +48,9 @@ void jtagtap_init(void)
 	jtag_proc.tap_idle_cycles = 1;
 
 	/* Ensure we're in JTAG mode */
-	for (size_t i = 0; i <= 50U; ++i)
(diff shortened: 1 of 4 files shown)


esden (approved):
LGTM as far as I can tell. Only way to know for sure is to use it on as many targets as possible. And that is only possible by merging it to main at the moment. :crossed_fingers: (HITL CI wen? ;) )

dragonmux:
Ahahaha, indeed! and unfortunately not the last time we're going to have to touch jtagtap.c either as we know there's at least one more problem based on how JTAG-PDI ran today..

PR #1386, opened 2023-02-13
Fix: STM32F1 option bytes handling correctness

dragonmux:
## Detailed description

This PR is a follow-up to https://github.com/blackmagic-debug/blackmagic/pull/1306 which sought to address an issue with how we handle STM32F{0,1,3} option bytes that turned up when `stm32f1_flash_busy_wait` was made to properly check the Flash controller status and report it, and we made the Flash and options bytes handling code use and act on that status return.

The gist of what was wrong is that on the GD32F103, the options byte erase never actually erases option byte slot 0 (read protect), a request to write the option bytes then tries to program that slot back to its original value and fails by triggering the controller to report a programming error due to the the slot not being erased (0xffff) in hardware. This translated into it failing the write and reporting an error.

The old PR patched this by only failing if programming reported an error and we weren't trying to program slot 0, which worked but is not really correct. This replacement patch moves the check into `stm32f1_option_write_erased`, and verifies that the status register specifically reads back a program failure when writing slot 0 (preserving failure handling of write protection errors) before turning that into a success return.

### DIFF
--- a/src/target/stm32f1.c
+++ b/src/target/stm32f1.c
@@ -487,23 +487,24 @@ bool stm32f1_probe(target_s *target)
 	return true;
 }
 
-static bool stm32f1_flash_unlock(target_s *t, uint32_t bank_offset)
+static bool stm32f1_flash_unlock(target_s *target, uint32_t bank_offset)
 {
-	target_mem_write32(t, FLASH_KEYR + bank_offset, KEY1);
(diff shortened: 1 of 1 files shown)


esden (approved):
LGTM
Tested with the BMP provisioning tools and the option byte erase and write work fine.

PR #1387, opened 2023-02-14
Fix: JTAG IR quirk handling

dragonmux:
## Detailed description

As part of investigating further RISC-V support (ESP32-C3), and to address a long-standing issue with the handling of Xilinx parts, and other parts, this PR rewrites how we do JTAG scan so we first read out ID codes, then the IR chain w/ quirks to handle misbehaving parts, and finally the BYPASS DR chain to validate chain length.

We have tested this against all JTAG targets on our desk and have not seen any breakage, and it does fix the scan issues with the ESP32-C3 once a quirk is provided for that part.

This is suggested for v1.10 ahead of the RISC-V support so it gets time to be tested and stablised.


esden (approved):
LGTM

PR #1388, opened 2023-02-14
hosted/probe_info: don't use malloc.h

tlyu:
Use stdlib.h instead of the non-standard malloc.h. Darwin doesn't have it.

## Detailed description

This fixes the build for `PROBE_HOST=hosted` on macOS (I have Ventura 13.2), which otherwise breaks due to a missing `malloc.h`. Previous OS versions or macOS SDKs might have had a compatibility `malloc.h` that is now removed; I'm not sure.

## Closing issues

N/A

### DIFF
--- a/src/platforms/hosted/probe_info.c
+++ b/src/platforms/hosted/probe_info.c
@@ -31,7 +31,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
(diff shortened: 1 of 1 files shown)


dragonmux (approved):
Our bad! This is because there is still some documentation around that to get malloc() and friends, you have to include malloc.h on Linux, which.. if we'd checked the C standard, would have quickly put paid to that. Thank you for the fix! Merging as soon as the checks pass

PR #1389, opened 2023-02-16
Fix: ARM JTAG regressions

### LINKED ISSUE #858: BMP remote protocoll and exception/timeout handling
Exception and timeout handling with BMP remote protocoll  needs to be improved.This is seen e.g. when scanning the ROMTABLE of a STM32WLE5. It is the same chip as the STM32WL55 with an M0 at AP1. AP1 is somehow disabled on theSTM32WLE5, but AP1 registers are available. Reading CIDR on the STM32WLE5 gives a timeout, the remote protocoll probably aborts early for its read timeout and the exception is not set.

dragonmux:
## Detailed description

This PR addresses several regressions in ARM JTAG handling detailed below and necessarily rewrites the ADIv5 high level remote protocol.

The regressions fixed are as follows:

* Fixed a regression introduced with 6dcc5d41 which broke non-multiple-of-8 TDI-TDO sequences
* Fixed an issue whereby CMSIS-DAP adaptors such as ORBTrace should never have worked with JTAG ADIv5 devices due to initialising the IR length information too late
* Fixed a timing issue in the TDI sequence function much like the TDI-TDO sequence one
* Fixed the remote protocol going "everything's ok" even when dp->fault is set non-zero and everything is not ok
* Fixed the remote protocol pretending exceptions don't exist and the ADIv5 JTAG and SWD layers can't throw them (they can) resulting in permanent comms loss when one occurs
* Fixed the BMDA side of the remote protocol never doing protocol recovery and driving the remote into a state due to not handling any errors at all

As part of the remote protocol rewrite we have pre-emptively split the high-level component of the protocol into generic high-level commands and ADIv5-specific ones to make organisational space for the RISC-V specific ones when v2.0 happens. This happens to also make reading BMDA probe logs easier.

## Closing issues

Fixes #858

### DIFF
--- a/src/remote.c
+++ b/src/remote.c
@@ -23,6 +23,7 @@
 
 #include "general.h"
 #include "remote.h"
+#include "gdb_main.h"
 #include "gdb_packet.h"
 #include "jtagtap.h"
 #include "swd.h"
@@ -359,23 +360,6 @@ static void remote_packet_process_high_level(unsigned i, char *packet)
 	remote_ap.dp = &remote_dp;
 	packet += 2;
 	switch (index) {
-	case REMOTE_MEM_READ_CSW: /* HM = Read from Mem and set csw */
-		remote_ap.csw = remote_hex_string_to_num(8, packet);
-		packet += 6;
-		/*fall through*/
-	case REMOTE_MEM_READ: { /* Hh = Read from Mem */
-		const uint32_t address = remote_hex_string_to_num(8, packet);
-		packet += 8;
-		const uint32_t count = remote_hex_string_to_num(8, packet);
-		adiv5_mem_read(&remote_ap, src, address, count);
-		if (remote_ap.dp->fault == 0) {
-			remote_respond_buf(REMOTE_RESP_OK, src, count);
-			break;
-		}
-		remote_respond(REMOTE_RESP_ERR, 0);
-		remote_ap.dp->fault = 0;
-		break;
-	}
 	case REMOTE_MEM_WRITE_CSW: /* Hm = Write to memory and set csw */
 		remote_ap.csw = remote_hex_string_to_num(8, packet);
 		packet += 6;
@@ -462,6 +446,26 @@ static void remote_packet_process_adiv5(const char *const packet)
 		remote_adiv5_respond(NULL, 0U);
 		break;
 	}
+	/* Memory access commands */
+	case REMOTE_MEM_READ: { /* Am = Read from memory */
+		/* Grab the CSW value to use in the access */
+		remote_ap.csw = remote_hex_string_to_num(8, packet + 6);
+		/* Grab the start address for the read */
+		const uint32_t address = remote_hex_string_to_num(8, packet + 14U);
+		/* And how many bytes to read, validating it for buffer overflows */
+		const uint32_t length = remote_hex_string_to_num(8, packet + 22U);
+		if (length > 1024U) {
+			remote_respond(REMOTE_RESP_PARERR, 0);
+			break;
+		}
+		/* Get the aligned packet buffer to reuse for the data read */
+		void *data = gdb_packet_buffer();
+		/* Perform the read and send back the results */
+		adiv5_mem_read(&remote_ap, data, address, length);
+		remote_adiv5_respond(data, length);
+		break;
+	}
+
 	default:
 		remote_respond(REMOTE_RESP_ERR, REMOTE_ERROR_UNRECOGNISED);
 		break;
--- a/src/remote.c
+++ b/src/remote.c
@@ -184,7 +184,6 @@ static void remote_packet_process_jtag(unsigned i, char *packet)
 	uint64_t DO = 0;
 	size_t ticks;
 	uint64_t DI = 0;
-	jtag_dev_s jtag_dev;
 	switch (packet[1]) {
 	case REMOTE_INIT: /* JS = initialise ============================= */
 		remote_dp.dp_read = fw_adiv5_jtagdp_read;
@@ -242,23 +241,6 @@ static void remote_packet_process_jtag(unsigned i, char *packet)
 		else {
 			const bool tdo = jtag_proc.jtagtap_next(packet[2] == '1', packet[3] == '1');
 			remote_respond(REMOTE_RESP_OK, tdo ? 1U : 0U);
-		}
-		break;
-
-	case REMOTE_ADD_JTAG_DEV: /* JJ = fill firmware jtag_devs */
-		if (i < 22U) {
-			remote_respond(REMOTE_RESP_ERR, REMOTE_ERROR_WRONGLEN);
-		} else {
-			memset(&jtag_dev, 0, sizeof(jtag_dev));
-			const uint32_t index = remote_hex_string_to_num(2, &packet[2]);
-			jtag_dev.dr_prescan = remote_hex_string_to_num(2, &packet[4]);
-			jtag_dev.dr_postscan = remote_hex_string_to_num(2, &packet[6]);
-			jtag_dev.ir_len = remote_hex_string_to_num(2, &packet[8]);
-			jtag_dev.ir_prescan = remote_hex_string_to_num(2, &packet[10]);
-			jtag_dev.ir_postscan = remote_hex_string_to_num(2, &packet[12]);
-			jtag_dev.current_ir = remote_hex_string_to_num(8, &packet[14]);
-			jtag_add_device(index, &jtag_dev);
-			remote_respond(REMOTE_RESP_OK, 0);
 		}
 		break;
 
@@ -342,7 +324,6 @@ static void remote_packet_process_general(unsigned i, char *packet)
 
 static void remote_packet_process_high_level(unsigned i, char *packet)
 {
-	(void)i;
 	SET_IDLE_STATE(0);
 	char index = packet[1];
 	if (index == REMOTE_HL_CHECK) {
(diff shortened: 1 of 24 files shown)


tlyu:
Tested on BMP with v1.9.0, with SWD to a GD32 target. Fallback seems to work. With BMP updated to firmware built from this PR, scan succeeds, but attach fails, with errors similar to what I previously reported on Discord.
[bmp-pr-1389-updated-fw.log](https://github.com/blackmagic-debug/blackmagic/files/10762112/bmp-pr-1389-updated-fw.log)

tlyu:
Update: with `ENABLE_DEBUG=1` on the BMP firmware (I had to disable some targets to get it to fit), I get

```
Recovering and re-trying access
```

on the aux CDC channel around the time that BMDA is showing

```
swdptap_seq_in          3 clock_cycles: 00000007
!SI20#
       PFFFFFFFF
swdptap_seq_in_parity  32 clock_cycles: ffffffff ERR
DP Error 0x00000000
ap_mem_write_sized @ e000edf0 len 4, align 4: 03 00 5f a0
!AM00002300004002e000edf00000000403005FA0#
Timeout while waiting for BMP response
remote_adiv5_mem_write_bytes comms error: -4
remote_adiv5_mem_write_bytes error around 0xe000edf0
```

Which reinforces my suspicion that BMP is attempting the recovery procedure, but is somehow becoming uncommunicative with BMDA during or after.

That's about all I have time to debug at the moment.

dragonmux:
Thank you, that's incredibly helpful as we weren't certain if that wasn't still going to be a problem post-patch (it clearly is) so giving us something more to dig into to finish stabilising this branch before Esden reviews it with a view to merge

dragonmux:
So, error recovery was still broken under JTAG, which is now fixed, but we haven't yet dived into error recovery under SWD. Unless we can figure out something simple today, we're inclined to punt fixing that into a new PR so Esden gets a chance to review what's here, test it and merge it. This PR is intentionally not touching SWD as such as it was JTAG that was quite so broken

tlyu:
> So, error recovery was still broken under JTAG, which is now fixed, but we haven't yet dived into error recovery under SWD. Unless we can figure out something simple today, we're inclined to punt fixing that into a new PR so Esden gets a chance to review what's here, test it and merge it. This PR is intentionally not touching SWD as such as it was JTAG that was quite so broken

Thanks for looking into it! I probably won't have time to test your latest changes today, so I'm fine with leaving the SWD fixes for the high-level protocol for separate PR or issue.

tlyu:
I just tried the latest changes, and there seems to be a regression that causes `blackmagic -t` to fail to detach the target after the scan. Connecting directly to BMP in GDB, then attaching and detaching the target, allows the target to resume.

I haven't tracked down what causes it yet, but from preliminary tests, it looks like the recent "target resumption" change is _not_ the cause.

tlyu:
I went back to b865bd8c87939d7f7267af5b5774643c2835eef8 on both BMDA and BMP, which I had previously tested, and the regression on `blackmagic -t` shows up there as well. I guess I didn't test `blackmagic -t` at that time.

dragonmux:
Any chance you could share a `src/blackmagic -tv 17` log of that behaviour so we can dig into it? We'll aim to add a fix to this PR if possible

tlyu:
I'll try to get to that soon. Another regression is that it's no longer possible to use `blackmagic -t -H` to disable the high-level remote protocol for scanning.

tlyu:
I'll upload logs from `blackmagic -tv 17` once I recreate them, because I don't remember which runs caused the target to remain halted or not. I can't see any obvious differences that would cause it.

I do have some suspicions that the clock may be gated off too quickly after sending the detach commands, due to an unintended timing change happening as a side effect of these changes. (ADIv5.2 §B4.1.1 requires at least 8 clocks of idle if the clock is gated off after a command, to ensure that it's fully shifted in to the SW-DP.)

tlyu:
Logs from various combinations of BMDA and BMP firmware. The "pull 1405" happened to be what I had flashed most recently. I can repeat it with main if you want, but I think the changes there aren't going to affect this behavior.

I'm almost tempted to hook up a logic analyzer to see if there's something weird with the clock disable with the firmware built from this patch. It would have to be different between native and remote, despite a lot of shared code, though.

I did try adding `swd_proc.seq_out(0, 8);` in a couple of places after SWD reads, on the theory that a turnaround plus the required idle cycles would make things behave, but it didn't seem to help.

BMDA pull 1405, firmware pull 1405: no halt

[bmp-t-pull-1405-no-halt.log](https://github.com/blackmagic-debug/blackmagic/files/10832175/bmp-t-pull-1405-no-halt.log)

BMDA pull 1389, firmware pull 1405: no halt

[bmp-t-pull-1389-fw-pull-1405-no-halt.log](https://github.com/blackmagic-debug/blackmagic/files/10832178/bmp-t-pull-1389-fw-pull-1405-no-halt.log)

BMDA pull 1389, firmware pull 1405, `-H` flag: no halt

[bmp-t-H-pull-1389-fw-pull-1405-no-halt.log](https://github.com/blackmagic-debug/blackmagic/files/10832180/bmp-t-H-pull-1389-fw-pull-1405-no-halt.log)

BMDA pull 1389, firmware pull 1389: halt

[bmp-t-pull-1389-fw-pull-1389-halt.log](https://github.com/blackmagic-debug/blackmagic/files/10832181/bmp-t-pull-1389-fw-pull-1389-halt.log)

src/remote.c
```diff
@@ -331,102 +320,152 @@ static void remote_packet_process_general(unsigned i, char *packet)
...
+		/* Validate the alignment is suitable */
+		if (length & ((1U << align) - 1U)) {
+			remote_respond(REMOTE_RESP_PARERR, 0);
 			break;
 		}
-		remote_respond(REMOTE_RESP_OK, 0);
+		/* Get the aligned packet buffer to reuse for the data to write */
+		void *data = gdb_packet_buffer();
+		/* And decode the data from the packet into it */
+		unhexify(data, packet, length);
```
tlyu, line 462: This should probably be `unhexify(data, packet + 32U, length)`. I finally got a logic analyzer on the SWD lines (had to slow it down because it was either loading the lines or coupling too much noise) and saw that the data bits actually written on the wire were always the same, regardless of the request.

A local patch with that change seems to resolve at least some of the regressions for me.

tlyu (changes requested):
See file comment for the possible cause of the regression with high-level SWD.

tlyu:
There's still a regression where `blackmagic -t -H` fails, but I haven't finished investigating that yet.

src/remote.c
```diff
@@ -51,62 +54,66 @@ uint64_t remotehston(const uint32_t max, const char *const str)
...
-	gdb_if_putchar(REMOTE_RESP, 0);
-	gdb_if_putchar(respCode, 0);
+	/* This defines space for exactly a 64-bit number expressed in hex */
+	char response[16];
+	size_t idx = 0;
 
-	do {
-		*p++ = NTOH(param & 0x0fU);
+	/* Convert the response to hexadecimal */
+	for (; idx < 64 && param; ++idx) {
```
tlyu, line 92: This bound is incorrect; it should be 16.

src/remote.c
```diff
@@ -51,62 +54,66 @@ uint64_t remotehston(const uint32_t max, const char *const str)
...
+	gdb_if_putchar(REMOTE_RESP, false);
+	gdb_if_putchar(response_code, false);
 
-	gdb_if_putchar(REMOTE_RESP, 0);
-	gdb_if_putchar(respCode, 0);
+	/* This defines space for exactly a 64-bit number expressed in hex */
+	char response[16];
+	size_t idx = 0;
 
-	do {
```
tlyu, line 86: The original do-while loop produced a different protocol: there was always at least one hex digit in a successful result. Unfortunately, the consuming code in BMDA (`remote_swdptap.c`) wasn't adjusted, so there were spurious errors whenever an operation produced an all-zeros response. (`K` instead of `K0` like the previous version did)

tlyu (changes requested):
The changes to `remote_respond` made a substantial change to the protocol with all-zeros responses returning `K` instead of `K0`. This causes some things to break, including `blackmagic -t -H`. Unfortunately, BMDA wasn't adjusted to expect it. It's probably easiest to revert to the previous protocol.

src/remote.c
dragonmux: You are entirely correct. Fixed :)

src/remote.c
dragonmux: Astute observation, thank you!

src/remote.c
dragonmux: We are split between adjusting this and changing the remote_swdptap.c code - there are a couple of ways we can adjust the new loop to regain the old behaviour but at the cost of code size; it's probably better to adjust BMDA instead

dragonmux: Hmm.. this is actually screwing inadvertently with the old JTAG scan method too because it makes jtag_proc.jtagtap_next() produce different results to what's expected.. we'll check if idx is 0 after the loop and adjust it in a fixup step. That'll give the old protocol behaviour

tlyu: As long as BMDA still accepts `K0`, it'll probably remain compatible with older firmware. I prefer to avoid changing existing protocols except where absolutely necessary, lest there be some subtle compatibility issue later on.

(The existing `remote_swdptap.c` also has an incorrect check in the debug message that caused a `swdptap_seq_in_parity failed, error \n` instead of usefully saying `short response`.)

My thoughts about how to adjust the new loop are to either change the condition to `idx < 16 && (param || !idx)` (which is what I tested) or just move the non-zero check for `param` to after the shift, and have it break the loop

```c
    param >>= 4U;
    if (!param)
        break;
```

I haven't inspected the assembly output to compare them, but it's worth noting that the existing proposed change probably already has two test-and-branch sections in the loop already, which is probably more than the previous code did.

dragonmux: Interestingly, the new code produces overall smaller asm output than the old because the compiler is more readily able to optimise. In this case, we do get to entirely change the protocol if we wish as we bumped the protocol version for that reason - a license to break compatibility - but fixing `K` meaning `K0` in every remote protocol function in BMDA would be far too much work. Please give the current version of this PR a spin as it should have fixed all 3 of the issues you found :)

tlyu:
Thanks for the updates! I did some testing, and it does fix the regressions that I found with this PR.

Unfortunately, the timeout during error recovery on SWD still happens. I think I failed to test my remote protocol fixes without the "add idle clocks after reads on SWD" changes, and those were preventing the error conditions that were triggering the timeouts.

That was a pre-existing problem on main, though, so I'm happy to defer it for another PR or issue.

dragonmux:
Excellent, and yeah - think we'll deal with the SWD problem on a separate PR as mentioned before.

Thank you for your excellent review! Much appreciated and helpful :)

dragonmux:
Apparently we hadn't gone quite far enough fixing the order of operations in adiv5.c's adiv5_dp_init(), with the new order for reading the DPIDR and all that both slightly accelerating the process if high-level functions are in use, and dealing with sticky errors that bit earlier.

esden (approved):
Damn, that is a lot of rework. I tested it to the best of my abilities. It does look good otherwise, and it definitely does fix the JTAG regression we had too. More use/testing needed to make sure, but that should happen in main. :)

PR #1391, opened 2023-02-19
Fix: Attaching to nRF Control Access Port Targets

TheyCallMeNobody:
## Detailed description

This is a fix for a problem, which occured when trying to connect to the Control Access Port Targets on NRF52 chips: After attaching, it ran into an endless loop, because the CAP is no real CPU core and therefore no CPU Register Read is implemented. The fix adds the case, when 0 registers are available and respons to GDB by sending an dummy register with the value 'XX', which indicates, that the register value hasn't been collected.
This check is implemented when handling the 'g' packet.

## Closing issues

fixes [#1353](https://github.com/blackmagic-debug/blackmagic/issues/1353)

### DIFF
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -32,6 +32,7 @@
 #include "gdb_main.h"
 #include "gdb_hostio.h"
 #include "target.h"
+#include "target_internal.h"
 #include "command.h"
 #include "crc32.h"
 #include "morse.h"
@@ -125,9 +126,14 @@ int gdb_main_loop(target_controller_s *tc, char *pbuf, size_t pbuf_size, size_t 
 	/* Implementation of these is mandatory! */
 	case 'g': { /* 'g': Read general registers */
 		ERROR_IF_NO_TARGET();
-		uint8_t gp_regs[target_regs_size(cur_target)];
-		target_regs_read(cur_target, gp_regs);
-		gdb_putpacket(hexify(pbuf, gp_regs, sizeof(gp_regs)), sizeof(gp_regs) * 2U);
+		const size_t reg_size = target_regs_size(cur_target);
+		if (reg_size) {
+			uint8_t gp_regs[reg_size];
+			target_regs_read(cur_target, gp_regs);
+			gdb_putpacket(hexify(pbuf, gp_regs, reg_size), reg_size * 2U);
+		} else {
+			gdb_putpacketz("00");
+		}
 		break;
 	}
 	case 'm': { /* 'm addr,len': Read len bytes from addr */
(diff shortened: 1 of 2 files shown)


dragonmux (changes requested):
This is actually quite close to the fix that was talked about on the BMP channel on Discord, however there's a couple of small changes we'd like to see made before we merge the patch - one of them is formatting, and stylistic, the other is functional as the fix in 'g' handling may not work across GDBs and we actually also need to handle the 'p' and 'P' packets better to guarantee a lack of issues across versions etc.

With these items fixed, we're happy to merge this.

Understanding that the review might be a little difficult to fully grok, here's the version that was batted about on Discord which we'd planned to patch in, showing what we mean:

```c
        const size_t regs_size = target_regs_size(cur_target);
        if (regs_size) {
            uint8_t gp_regs[regs_size];
            target_regs_read(cur_target, gp_regs);
            gdb_putpacket(hexify(pbuf, gp_regs, regs_size), regs_size << 1U);
        } else
            gdb_putpacketz("00");
        break;
```

For the 'p' and 'P' packets, a check that t->reg_read and t->reg_write is not NULL before invoking the normal handler w/ a similar single-byte-0 response when they are, would be an appropriate fix.

src/gdb_main.c
```diff
@@ -148,10 +154,15 @@ int gdb_main_loop(target_controller_s *tc, char *pbuf, size_t pbuf_size, size_t
...
-		target_regs_write(cur_target, gp_regs);
-		gdb_putpacketz("OK");
+		const size_t reg_size = target_regs_size(cur_target);
+		if (reg_size) {
+			uint8_t gp_regs[reg_size];
+			unhexify(gp_regs, &pbuf[1], sizeof(gp_regs));
+			target_regs_write(cur_target, gp_regs);
+			gdb_putpacketz("OK");
+		} else {
+			gdb_putpacketz("00");
```
dragonmux, line 164: This isn't a valid response for 'G' to make per [GDB's docs on the remote protocol](https://docs.jade.fyi/gnu/gdb/gdb.html#index-G-packet); We have two options available here, either returning an error code such as "EFF", or returning "OK" and pretending it worked - we're inclined to the latter, which allows this to be simplified by moving the gdb_putpacketz() below the if block and drop the else.

src/gdb_main.c
```diff
@@ -223,28 +234,36 @@ int gdb_main_loop(target_controller_s *tc, char *pbuf, size_t pbuf_size, size_t
...
+			sscanf(pbuf, "P%" SCNx32 "=%n", &reg, &n);
+			// TODO: FIXME, VLAs considered harmful.
+			uint8_t val[strlen(&pbuf[n]) / 2U];
+			unhexify(val, pbuf + n, sizeof(val));
+			if (target_reg_write(cur_target, reg, val, sizeof(val)) > 0)
+				gdb_putpacketz("OK");
+			else
+				gdb_putpacketz("EFF");
+		} else {
+			gdb_putpacketz("00");
```
dragonmux, line 265: This [suffers the same response problem](https://docs.jade.fyi/gnu/gdb/gdb.html#index-P-packet) and we're going to suggest the same thing here with replying with `gdb_putpacketz("OK")` and pretending the write worked

TheyCallMeNobody:
I hope I changed everything in git correctly

PR #1400, opened 2023-02-23
mm32: support for mm32f5277

koendv:
## Detailed description
Adds support for MindMotion MM32F5277, Arm China STAR-MC1 architecture.
Cleanup and document MM32 Cortex-M0 flashing code.
Tested flashing the following:
- MM32SPIN27 M0
- MM32L073 M0
- MM32F3273 M3
- MM32F5277 STAR-MC1

bmda session log attached: [f5277.txt](https://github.com/blackmagic-debug/blackmagic/files/10817039/f5277.txt)
Additional MM32 processors can be added as described in #1362.

### DIFF
--- a/src/target/adiv5.c
+++ b/src/target/adiv5.c
@@ -270,6 +270,14 @@ static const struct {
 	{0xd21, 0x14, 0x1a14, aa_nosupport, cidc_dc, ARM_COMPONENT_STR("Cortex-M33", "(Cross Trigger)")},
 	{0xd21, 0x13, 0x4a13, aa_nosupport, cidc_dc, ARM_COMPONENT_STR("Cortex-M33", "(Embedded Trace)")},
 	{0xd21, 0x11, 0, aa_nosupport, cidc_dc, ARM_COMPONENT_STR("Cortex-M33", "(Trace Port Interface Unit)")},
+	{0x132, 0x31, 0x0a31, aa_nosupport, cidc_dc, ARM_COMPONENT_STR("STAR-MC1 MTB", "(Execution Trace)")},
+	{0x132, 0x43, 0x1a01, aa_nosupport, cidc_dc, ARM_COMPONENT_STR("STAR-MC1 ITM", "(Instrumentation Trace Module)")},
+	{0x132, 0x00, 0x1a02, aa_nosupport, cidc_dc, ARM_COMPONENT_STR("STAR-MC1 DWT", "(Data Watchpoint and Trace)")},
+	{0x132, 0x00, 0x1a03, aa_nosupport, cidc_dc, ARM_COMPONENT_STR("STAR-MC1 BPU", "(Breakpoint Unit)")},
+	{0x132, 0x14, 0x1a14, aa_nosupport, cidc_dc, ARM_COMPONENT_STR("STAR-MC1 CTI", "(Cross Trigger)")},
+	{0x132, 0x00, 0x2a04, aa_cortexm, cidc_dc, ARM_COMPONENT_STR("STAR-MC1 SCS", "(System Control Space)")},
+	{0x132, 0x13, 0x4a13, aa_nosupport, cidc_dc, ARM_COMPONENT_STR("STAR-MC1 ETM", "(Embedded Trace)")},
+	{0x132, 0x11, 0, aa_nosupport, cidc_dc, ARM_COMPONENT_STR("STAR-MC1 TPIU", "(Trace Port Interface Unit)")},
 	{0xfff, 0x00, 0, aa_end, cidc_unknown, ARM_COMPONENT_STR("end", "end")},
 };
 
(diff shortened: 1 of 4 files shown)


src/target/adiv5.c
```diff
@@ -573,7 +581,7 @@ static void adiv5_component_probe(
 		DEBUG_INFO("%sROM: Table END\n", indent);
 
 	} else {
-		if (designer_code != JEP106_MANUFACTURER_ARM) {
+		if ((designer_code != JEP106_MANUFACTURER_ARM) && (designer_code != JEP106_MANUFACTURER_ARM_CHINA)) {
```
dragonmux, line 584: Drop the extra parens, they aren't necessary and suppress a GCC warning about accidental assignment that we want.

src/target/stm32f1.c
```diff
@@ -272,13 +276,16 @@ void mm32l0_mem_write_sized(adiv5_access_port_s *ap, uint32_t dest, const void *
 		case ALIGN_BYTE: {
 			uint8_t value;
 			memcpy(&value, src, sizeof(value));
-			tmp = (uint32_t)value << (8U * (dest & 3U));
+			tmp = (uint32_t)value;
```
dragonmux, line 279: This makes us wonder if multiplying by 0x01010101 might be faster here - either way is about as obscure, so a comment talking through what this is doing (and perhaps why) would be good. Nothing complicated, something like "Copy the value to write to all 4 bytes of the uint32_t"

src/target/stm32f1.c
```diff
@@ -272,13 +276,16 @@ void mm32l0_mem_write_sized(adiv5_access_port_s *ap, uint32_t dest, const void *
...
+			tmp = tmp | tmp << 8U;
+			tmp = tmp | tmp << 16U;
 			break;
 		}
 		case ALIGN_HALFWORD: {
 			uint16_t value;
 			memcpy(&value, src, sizeof(value));
-			tmp = (uint32_t)value; /* MM32 special, no shift << (8U * (dest & 2U)) */
+			tmp = (uint32_t)value;
+			tmp = tmp | tmp << 16U;
```
dragonmux, line 288: Similar to above - comment such as "Copy the value to writte to both halves of the uint32_t" would be useful

src/target/stm32f1.c
```diff
@@ -257,7 +257,11 @@ bool at32fxx_probe(target_s *t)
 	return false;
 }
 
-/* mm32l0 flash write */
+/* mm32l0 flash write
```
dragonmux, line 260: The codebase multi-line comment style is
```c
/*
 * first line
 * second line
 */
```

Please alter this to match and drop the blank line after the comment so IDEs attach the comment to the function properly.

dragonmux (changes requested):
This is looking good! There are a couple of stylistic and documentation issues to be solved, but with them fixed, we're happy to merge this.

src/target/cortexm.c
```diff
@@ -698,7 +698,7 @@ bool cortexm_probe(adiv5_access_port_s *ap)
 	case JEP106_MANUFACTURER_RENESAS:
 		PROBE(renesas_probe);
 		break;
-	case JEP106_MANUFACTURER_ARM_CHINA:
+	case JEP106_MANUFACTURER_ERRATA_ARM_CHINA:
```
perigoso, line 701: Shouldnt this code be replaced when found during the scan, like the other errata codes?

src/target/adiv5.c
koendv: ok, done.

src/target/stm32f1.c
koendv: Multiplying by 0x01010101: nice.
Kept current code, added comment.

adiv5_mem_write_sized() is used to write stm32 flash, but is broken on mm32 cortex-m0.
mm32l0_mem_write_sized() is like adiv5_mem_write_sized() but now works on mm32 and stm32.
mm32l0_mem_write_sized() working on mm32 and stm32 means that if future mm32 cortex-m0 behave like an stm32, things work out of the box.

src/target/stm32f1.c
koendv: ok, done.

src/target/stm32f1.c
koendv: ok, done.

src/target/cortexm.c
koendv: ok, replaced.

src/target/stm32f1.c
dragonmux: It's still missing the *'s on the beginning of each comment line, but it's close enough that we'll merge and next time we touch stm32f1.c we'll run a fixup pass.

dragonmux (approved):
Your commit message follows the contribution guidelines and you've addressed all review changes. LGTM, merging!

PR #1405, opened 2023-02-24
gdb_main: don't reattach on connect

### LINKED ISSUE #1390: don't reattach last target on connect
With current firmware, the BMP reattaches the most recently attached target when GDB connects to it. I think it should wait until explicitly instructed to attach. Among other things, the official documentation says that connecting using `target extended-remote` in GDB won't attach the target or perform any SWD or JTAG actions. (see https://black-magic.org/usage/gdb-commands.html#connecting-gdb-to-the-black-magic-probe)

I found this to be particularly annoying when configuring VS Code's debugger support to use BMP. I ended up having to write a custom GDB script to test whether connecting to the probe had caused a side effect of reattaching to the target, and if so, skip the scan and attach. (The scan detaches the reattached target, which causes VS Code to think that the target program has exited, and VS Code then exits gdb.)

I discovered this was happening due to the `qXfer:features` and `qXfer:memory-map` commands that are sent by GDB at connect time. I tried some local changes to look up the relevant information in the `last_target` structure without reattaching the target, and they appear to work.

I think it's debatable whether it's even correct to do those lookups on the previous target, instead of returning an error, because the user might have changed the device under test to a different type of target without resetting the BMP. On the other hand, it might be reasonable to run GDB commands that invoke those queries after detaching a target; I'm not sure.

tlyu:
Don't reattach the target when handling `qXfer:features` or `qXfer:memory-map`. GDB uses at least `qXfer:features` when connecting to the debug server, halting the target. Official documentation claims that connecting to the server won't attach the target.

## Detailed description

This change removes an unwanted behavior (see #1390) where reconnecting GDB to a BMP after detaching from a target will sometimes reattach the target. This is contrary to the official documentation. There is a workaround: scanning for targets before disconnecting from GDB in the first session, but that is non-intuitive and undocumented.

In case there is a benefit from retrieving the relevant information from `last_target`, it is still consulted, if available, for these `qXfer` requests, without reattaching the target.

## Closing issues

fixes #1390

### DIFF
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -423,35 +423,34 @@ static void exec_q_supported(const char *packet, const size_t length)
 
 static void exec_q_memory_map(const char *packet, const size_t length)
 {
-	(void)packet;
 	(void)length;
+	target_s *target = cur_target;
+
 	/* Read target XML memory map */
-	if ((!cur_target) && last_target) {
-		/* Attach to last target if detached. */
-		cur_target = target_attach(last_target, &gdb_controller);
-	}
-	if (!cur_target) {
+	if (!target)
+		target = last_target;
+	if (!target) {
 		gdb_putpacketz("E01");
 		return;
 	}
 	char buf[1024];
-	target_mem_map(cur_target, buf, sizeof(buf)); /* Fixme: Check size!*/
+	target_mem_map(target, buf, sizeof(buf)); /* Fixme: Check size!*/
 	handle_q_string_reply(buf, packet);
 }
 
 static void exec_q_feature_read(const char *packet, const size_t length)
 {
 	(void)length;
-	/* Read target description */
-	if (!cur_target && last_target)
-		/* Attach to last target if detached. */
-		cur_target = target_attach(last_target, &gdb_controller);
+	target_s *target = cur_target;
 
-	if (!cur_target) {
+	/* Read target description */
+	if (!target)
+		target = last_target;
+	if (!target) {
 		gdb_putpacketz("E01");
 		return;
 	}
-	const char *const description = target_regs_description(cur_target);
+	const char *const description = target_regs_description(target);
 	handle_q_string_reply(description ? description : "", packet);
 	free((void *)description);
 }


src/gdb_main.c
```diff
@@ -423,35 +423,37 @@ static void exec_q_supported(const char *packet, const size_t length)
 
 static void exec_q_memory_map(const char *packet, const size_t length)
 {
+	target_s *target = cur_target;
```
dragonmux, line 426: Seeing as packet is used in this function, please drop that `(void)packet;` line. Additionally, we have been trying to put all parameter use suppression before the body of the function so please put this new line after the line that turns off the warning that `length`is unused

src/gdb_main.c
```diff
@@ -423,35 +423,37 @@ static void exec_q_supported(const char *packet, const size_t length)
...
+
 	(void)packet;
 	(void)length;
 	/* Read target XML memory map */
-	if ((!cur_target) && last_target) {
-		/* Attach to last target if detached. */
-		cur_target = target_attach(last_target, &gdb_controller);
+	if (!target) {
```
dragonmux, line 431: Not entirely your fault here (the old code needed this fix too), but please drop the braces ({}) as they are unnecessary.

src/gdb_main.c
```diff
@@ -423,35 +423,37 @@ static void exec_q_supported(const char *packet, const size_t length)
...
 	}
 	char buf[1024];
-	target_mem_map(cur_target, buf, sizeof(buf)); /* Fixme: Check size!*/
+	target_mem_map(target, buf, sizeof(buf)); /* Fixme: Check size!*/
 	handle_q_string_reply(buf, packet);
 }
 
 static void exec_q_feature_read(const char *packet, const size_t length)
 {
+	target_s *target = cur_target;
```
dragonmux, line 445: same comment as above about the ordering of parameter use suppression and variable declaration

src/gdb_main.c
```diff
@@ -423,35 +423,37 @@ static void exec_q_supported(const char *packet, const size_t length)
...
+
 	(void)length;
 	/* Read target description */
-	if (!cur_target && last_target)
-		/* Attach to last target if detached. */
-		cur_target = target_attach(last_target, &gdb_controller);
-
-	if (!cur_target) {
+	if (!target) {
```
dragonmux, line 449: Same commentabout the braces

dragonmux (changes requested):
This has a couple of stylistic things that need fixing, but otherwise this looks good. We're still interested in possibly adding a monitor command to optionally opt into the old behaviour, but you are correct that it's not what's documented, so we'll accept this contribution gladly!

src/gdb_main.c
tlyu: I'm trying to stick with the style of putting declarations at the top of the function, but I see now that the rest of the code mixes declarations with statements, so I will try to match that. (The `(void)packet;` was pre-existing, but I'll remove it.)

tlyu:
Thanks for the review! I'll push an update after I've tested the changes.

src/gdb_main.c
dragonmux: C89 style decls are bugprone + pessimising and this is a C11 code base, so we've been trying to migrate it away from that style to prevent and fix bugs :) (there have been several caused by read-before-init and such things)

dragonmux (approved):
LGTM, merging!

PR #1413, opened 2023-03-07
Fix: Tiva-C TM4C129 identification

### LINKED ISSUE #1412: Tiva TM4C129
I found some issues with  TI **MSP432P401R**, while attempting to reflash **TM4C1294NCPDT** firmware implementing builtin XDS 110 JTAG programmer, which was broken during the update and doesn't enter DFU mode.
TI suggest to use JTAG (seems this is the only possible way). As I don't have any TI XDS JTAG programmer, I'm trying to do the same with some other JTAG hardware.
I didn't achieve any success with Altera Byte Blaster neither with DirtyJtag+openocd(patched). It can detect the CPU, but then failed during initialization.
I managed to do so after making some changes of the blackmagic sources.
- Tiva C123 and C129 have different DID0 codes, file _src/target/lmi.c_:

defines:
```
  #define DID0_CLASS_STELLARIS_FURY 0x00010000U
  #define DID0_CLASS_STELLARIS_DUST 0x00030000U
> #define DID0_CLASS_TIVA           0x00050000U
< #define DID0_CLASS_TIVA_C123      0x00050000U
< #define DID0_CLASS_TIVA_C129      0x000a0000U

  #define DID1_LM3S3748      0x1049U
  #define DID1_LM3S5732      0x1096U
```
lmi_probe() function:
```

	        return lm3s_probe(t, did1);
>	case DID0_CLASS_TIVA:
<	case DID0_CLASS_TIVA_C123:
<	case DID0_CLASS_TIVA_C129:
		return tm4c_probe(t, did1);b
```
- function _adiv5_dp_init()_, file _src/target/adiv5.c_ can't detect APs on DP correctly.

It call many times, until the memory exhausted, _adiv5_new_ap()_, which every time return the same AP despite of different _apsel_.  I'm not familiar with this JTAG internals but if I remove the APs probe loop, calling  _adiv5_new_ap()_ only once, everything start working and I can connect to the target with GDB.

dragonmux:
## Detailed description

As identified in https://github.com/blackmagic-debug/blackmagic/issues/1412 , we were missing the DID0 value for Tiva-C TM4C129 parts. This PR fixes this so we can properly detect those pparts again.

## Closing issues

Partially fixes #1412

### DIFF
--- a/src/target/lmi.c
+++ b/src/target/lmi.c
@@ -42,7 +42,8 @@
 #define DID0_CLASS_MASK           0x00ff0000U
 #define DID0_CLASS_STELLARIS_FURY 0x00010000U
 #define DID0_CLASS_STELLARIS_DUST 0x00030000U
-#define DID0_CLASS_TIVA           0x00050000U
+#define DID0_CLASS_TIVA_C123      0x00050000U
+#define DID0_CLASS_TIVA_C129      0x000a0000U
 
 #define DID1_LM3S3748      0x1049U
 #define DID1_LM3S5732      0x1096U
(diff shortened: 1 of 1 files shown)


dragonmux:
If we can figure out a sane way to fix the AP issue in the linked issue, we will add that to this PR

esden (approved):
LGTM

PR #1429, opened 2023-04-03
Fix: Cortex-M halt/resume correctness

### LINKED ISSUE #968: Cortex M DHCSR.C_MASKINTS may be handled incorrectly
I suspect that Cortex M's bit `DHCSR.C_MASKINTS` is subtly mishandled, which leads to unstable behavior if semihosting call `SYS_EXIT` (and its extended counterpart) is used. I must admit here that I don't fully comprehend what's happening under the hood yet but I want to report this anyway because this issue had cost me a few days of debugging and tearing my hair off.

I am working on a baremetal application based on STM32F446 that heavily relies on semihosting (it is actually a unit testing suite, not that it matters here). At the end of the execution, the application makes the final semihosting call `SYS_EXIT` with the appropriate exit code indicating whether the test suite is executed successfully. Everything was functioning normally until I had to add a peripheral interrupt handler to the application, whereupon I discovered that under some mysterious circumstances the handler might not be invoked despite the appropriate pending/enable flags being set in the NVIC registers. I was never able to find a robust way of reproducing this problem: sometimes it would work reliably across power cycling and software resets, sometimes it would fail. Fast forward a few days, the culprit has been tracked down to `DHCSR.C_MASKINTS`; here's my best understanding of what is happening.

First, I power on the board, upload and launch the application:

```gdb
shell ln -fs /dev/serial/by-id/usb-*Black_Magic_Probe*-if00 .gdb_port.tmp
target extended-remote .gdb_port.tmp
monitor connect_srst enable
monitor swdp_scan
attach 1
set mem inaccessible-by-default off
set remote system-call-allowed 1
set confirm off
load
compare-sections
run
```

The versions involved are:

- GDB: `GNU gdb (GNU Arm Embedded Toolchain 10.3-2021.10) 10.2.90.20210621-git`
- Black Magic (native): `Black Magic Probe v1.7.1-131-g560a046, Hardware Version 3`

The original state of DHCSR after a power-on reset before the application is started is reliably observed as 0x30003, which makes sense.

Now, wait for the application to finish. The first run after a power-on reset appears to always succeed (I can't be 100% sure due to the sporadic nature of the problem though).

As I wrote above, at the end of the execution the application calls SYS_EXIT. The handler of that system call [invokes `cortexm_halt_resume` with step=true](https://github.com/blacksphere/blackmagic/blob/5c07d6170fce8404ffd5cf841eb5ec9a1e36c222/src/target/cortexm.c#L1612), which causes the debugger to set `DHCSR.C_MASKINTS`.

The debugger is disconnected at this point --- never physically, but logically, by way of exiting GDB.

Next, without changing anything, I re-run the suite. The debugger is attached again (successfully), but this time occasionally you might observe the original state of DHCSR to be not 0x30003 (where C_MASKINTS is reset) but 0x3000B (where C_MASKINTS is set but C_STEP is not). If DHCSR reads as 0x3000B, the application is also able to function except that the interrupt handlers are masked, which is an error because `run` is expected to unmask interrupts.

Here are the most relevant excerpts from the arm-v7m tech reference manual:

Screenshot_20220105_153835

Screenshot_20220105_153234

What I suspect might be happening is that after SYS_EXIT, the target is left with C_MASKINTS=1. When the debugger is reattached later, the new value is written to DHCSR that changes C_MASKINTS bypassing due sequence, triggering the unpredictable behavior that happens to leave C_MASKINTS set only occasionally. If this is correct, then a possible remedy is to read DHCSR before the first write to ensure the bits are not modified until C_HALT & C_DEBUGEN are guaranteed to be set. I would perhaps also harden `cortexm_halt_resume()` against the risk of state divergence between `t->stepping` and the actual bits stored in the register by always fetching the current value before writing a new one.

While googling around I stumbled upon this random tweet that seems to provide a plausible explanation, but I am not sure if it's valid for Black Magic: https://mobile.twitter.com/azonenberg/status/1162402192793358341

As a tentative workaround, I made a script that resets this bit at every opportinity -- it appears to be effective:

```gdb
define clear_dhcsr_c_maskints
    set {unsigned}0xE000EDF0 = 0xA05F0000 | ({unsigned}0xE000EDF0 & 0b1111111111110111)
    p/x {unsigned}0xE000EDF0
end

catch signal all
commands
    clear_dhcsr_c_maskints
end

define hook-run
    clear_dhcsr_c_maskints
end

define hook-continue
    clear_dhcsr_c_maskints
end
```

dragonmux:
## Detailed description

As prompted by #968, we took a dive into the halt/resume code which was subtly broken in a couple of ways. This PR addresses these issues and documents how this tricky part of the Cortex-M support works better.

The exact issues addressed are:

* Recovery APs wrongly specified non-0 register size information which leads to a crash on main
* Our handling of the C_HALT and C_DEBUGEN bits was dubious at best
* C_MASKINTS (interrupt masking) was being mishandled during detach
* Post-DP-scan resumption was doing bad things to targets not found in the current adiv5_dp_init() call
* DHCSR halt was being done wrongly for minimal DP targets
* State was spilling from DHCSR halt into CIDR readout, making that fail under some circumstances when it shouldn't
* Error messages around CIDR readout failing and DHCSR halt failing were misleading/inaccurate

This has been thoroughly tested on: RP2040, Tiva-C, LPC4370, LPC4332, LPC4350, STM32F411. Testing was conducted via BMP, ORBTrace, and Tigard. With this patch, LPC43xx scan in particular progresses correctly when enumerating the Cortex-M0+ cores where previously DHCSR halt would fail and the cores would not be discovered correctly.

This should eliminate the Cortex-M halt/resume issues

## Closing issues

Fixes #968

### DIFF
--- a/src/target/adiv5.c
+++ b/src/target/adiv5.c
@@ -487,7 +487,7 @@ static void adiv5_component_probe(
 	/* CIDR preamble sanity check */
 	if ((cidr & ~CID_CLASS_MASK) != CID_PREAMBLE) {
 		DEBUG_WARN("%s%" PRIu32 " 0x%08" PRIx32 ": 0x%08" PRIx32 " <- does not match preamble (0x%08" PRIx32 ")\n",
-			indent + 1, num_entry, addr, cidr, CID_PREAMBLE);
+			indent, num_entry, addr, cidr, CID_PREAMBLE);
 		return;
 	}
 
--- a/src/target/adiv5.c
+++ b/src/target/adiv5.c
@@ -358,27 +358,14 @@ static uint32_t cortexm_initial_halt(adiv5_access_port_s *ap)
 		adiv5_mem_write(ap, CORTEXM_DHCSR, &dhcsr_debug_en, sizeof(dhcsr_debug_en));
 	}
 
-	/* Workaround for CMSIS-DAP Bulk orbtrace
-	 * High values of TRNCNT lead to NO_ACK answer from debugger.
-	 *
-	 * However CMSIS/HID even with highest value has few chances to catch
-	 * a STM32F767 mostly sleeping in WFI!
-	 */
-	uint32_t start_time = platform_time_ms();
-	uint32_t trncnt = 0x80U;
 	bool reset_seen = false;
 	while (!platform_timeout_is_expired(&halt_timeout)) {
 		uint32_t dhcsr;
 
 		if (use_low_access) {
 			adiv5_dp_low_access(
-				ap->dp, ADIV5_LOW_WRITE, ADIV5_DP_CTRLSTAT, ctrlstat | (trncnt * ADIV5_DP_CTRLSTAT_TRNCNT));
+				ap->dp, ADIV5_LOW_WRITE, ADIV5_DP_CTRLSTAT, ctrlstat | ADIV5_DP_CTRLSTAT_TRNCNT(0xfffU));
 			adiv5_dp_low_access(ap->dp, ADIV5_LOW_WRITE, ADIV5_AP_DRW, dhcsr_ctl);
-
-			if (trncnt < 0xfffU)
-				trncnt += (platform_time_ms() - start_time) * 8U;
-			else
-				trncnt = 0xfffU;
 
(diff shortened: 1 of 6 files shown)


esden (approved):
I have tested this to the best of my abilities right now. LGTM

PR #1430, opened 2023-04-03
Fix: BMDA's FTDI implementation and Tigard usability

dragonmux:
## Detailed description

The FTDI implementation contained a liturgy of errors around how adaptors were being matched, resulting in Tigard being found but unable to be used. Likewise the implementation contained a significant number of errors in how it was driving both JTAG and SWD bitbanging resulting in no-scan and other issues, particularly for multi-TAP JTAG scan chains.

This PR addresses these issues with a rewrite of significant chunks of the FTDI support, which also fixes a regression in `ftdi_swd_seq_in_parity_mpsse()` that was reported to us privately in Dec 2022 by sys64738.


src/platforms/hosted/ftdi_bmp.c
```diff
@@ -382,7 +382,7 @@ int ftdi_bmp_init(bmda_cli_options_s *cl_opts, bmp_info_s *info)
 	}
 
 	if (!cable->name) {
-		DEBUG_WARN("No cable matching found for %s\n", cl_opts->opt_cable);
```
esden, line 384: This is an old school naming of JTAG debuggers. In commercial software they are often called cables. Not sure how common the `adaptor` naming is for the same thing.

esden (approved):
LGTM, it does fix tigard support for me. I tested JTAG and SWD for functionality with 1Bitsy as target. Works nicely!

esden (approved):
LGTM, tested SWD and JTAG on a Tigard. Works nicely.

(had to resolve a conflict caused by a commit that snuck into this PR by accident... :) )

src/platforms/hosted/ftdi_bmp.c
dragonmux: So, the motivation for changing this (and we actually started this process 2 releases ago) is that in context of BMDA, the "cable" nomenclature results in an incredibly confusing user experience for anyone not already deeply rooted in the FTDI ecosystem (we might be able to get away with this if we said "FTDI Cable" maybe but it's a stretch), and makes the BMDA code more spaghetti and incomprehensible.

Consider, for example, someone who's just got a Tigard and they don't know FTDI's nomenclature at all - they're going to call the board an adaptor (it is, to be fair), look for options to set what adaptor they're using, and expect error messages consistent with that.

Likewise, within BMDA, "cable" winds up resulting in a reading experience that, from our own past experience, results in head scratching and getting very confused across against the other backends supported.

Hopefully that helps explain the motivation for this trajectory.

PR #665, opened 2020-05-17
pc-hosted semihosting

koendv:
As requested, semihosting for "blackmagic_hosted". 
PC-hosted semihosting does keyboard, file and screen i/o on the system where blackmagic_hosted runs, using linux system calls, while semihosting in the probe does keyboard, file and screen i/o on the system where gdb runs, using gdb file i/o calls.

The pc-hosted semihosting was tested with the sketches from the Arduino STM32duino-semihosting library. Output attached. 

[semihosting_testbench.txt](https://github.com/blacksphere/blackmagic/files/4642399/semihosting_testbench.txt)

PR #1432, opened 2023-04-03
Fix: Semihosting

### LINKED ISSUE #1385: gdb_main_loop
In gdb_hostio.c:
int gdb_main_loop(target_controller_s *, bool in_syscall);

In gdb_main.c:
/* execute gdb remote command stored in 'pbuf'. returns immediately, no busy waiting. */
int gdb_main_loop(target_controller_s *tc, char *pbuf, size_t pbuf_size, size_t size, bool in_syscall)

This looks strange.

dragonmux:
## Detailed description

In the wake of #1284, Koen correctly identified that a private (UB) declaration of gdb_main_loop() had the semihosting Host IO code broken with nobody having noticed. This PR addresses this by removing that declaration, putting one in the gdb_main.h header so mismatches with gdb_main.c result in compilation failure, and updating how the Host IO code calls the main loop.

## Closing issues

Fixes #1385

### DIFF
--- a/src/gdb_hostio.c
+++ b/src/gdb_hostio.c
@@ -24,8 +24,6 @@
 #include "gdb_hostio.h"
 #include "gdb_packet.h"
 
-int gdb_main_loop(target_controller_s *, bool in_syscall);
-
 int hostio_reply(target_controller_s *tc, char *pbuf, int len)
 {
 	(void)len;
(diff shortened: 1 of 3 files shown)


dragonmux:
@koendv Please give this a test when you find a minute. It should have solved the bug you reported back in February. We are unfortunately ill-equipped with suitable firmware, etc, to actually test this ourself so please let us know how you get on and if this PR needs to go further.

esden (approved):
LGTM

koendv:
ok. 
As test programs I use the examples of this Arduino library: https://github.com/koendv/STM32duino-Semihosting
Running the test program "SemihostingTest.ino.elf" in a stm32f103 with openocd, a commercial probe, blackmagic hosted and blackmagic firmware ought to give mostly similar results. (I say similar because not all debuggers implement all semihosting calls. E.g. blackmagic firmware implements semi-hosting using the gdb fileio calls, while blackmagic hosted implements semi-hosting using posix system calls.)
Sample output of "SemihostingTest.ino.elf" is the attachment "semihosting_testbench.txt" in #665
I add a zip with the compiled binary of "SemihostingTest".  Tests were run with arm-none-eabi-gdb 12.1.90. 

My impression is it doesn't quite work. 
When using bmp firmware, speed is an order of magnitude slower than I expect; as if something in bmp is looping. Return values are odd; e.g. the flen() semihosting call for determing file length returns -1, independent of file length. 
When using bmp hosted, I get errors "remote_adiv5_dp_read error 2". 
This feels as if there might be more than one bug.

I'd like a second opinion on this one.

koendv:
I would suggest adding to the documentation:

If using black magic debug app with black magic debug, make sure both
have the same major release.

(e.g. not app 1.9 with firmware 1.8)

dragonmux:
Using a newer BMDA with older firmware should work (there are fallbacks in BMDA for this) - if it doesn't, please let us know as that's a bug

PR #1438, opened 2023-04-04
crc32: Round up read length to a multiple of four

zyp:
## Detailed description

Round up read length to a multiple of four so it'll be performed as 32b reads.

When I were working on a LPC4078-based project, `compare-sections` were failing on a section which length was not a multiple of four. Rounding up the read length fixed it.

I have not investigated why non multiple of four reads were not returning correct data; whether we've got a bug somewhere else, or whether this is a limitation of this specific MEM-AP implementation, but there's an argument to be made for performing the read as 32b reads in any case, also from a performance perspective.


dragonmux (approved):
LGTM, we'll merge this and if anything breaks then we'll consider options, but doubt we'll have to go there.

PR #1446, opened 2023-04-09
target: Fix out of bound array access in jtag_scan

francois-berder:
## Detailed description

This pull request fixes an out of bound array access in `jtag_scan` function if `device` variable reaches 32.

## Your checklist for this pull request

* [x] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md)
* [x] I've read the [guidelines for contributing](https://github.com/blackmagic-debug/blackmagic/blob/main/CONTRIBUTING.md) to this repository
* [x] It builds for hardware native (`make PROBE_HOST=native`)
* [x] It builds as BMDA (`make PROBE_HOST=hosted`)
* [x] I've tested it to the best of my ability
* [x] My commit messages provide a useful short description of what the commits do

## Closing issues

This PR does not fix any existing issues.

PR #1447, opened 2023-04-09
Fix: JTAG scan irlens handling

dragonmux:
## Detailed description

Thanks to #1446 spurring us on, we have now addressed the way the manually specified IR lengths logic is handled as between the VLA usage, lack of length checking and other handling issues, the soundness of the IR lengths code was questionable.

This PR addresses this with a partial rewrite and finishes the refactoring work began in #1389 in jtag_scan() + cleans up the ST-Link v2 specific version of this routine. The result should be more consistent and robust with better user feedback.

This PR does slightly alter how the IR lengths are specified - dropping the need for a trailing 0 in the array, and making any 0's specified on the command line errors - but as this fixes a bug in how they could be specified and the user was not visibly aware of this behaviour/need, we think it's a safe change (changing only implementation details),

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -217,7 +217,7 @@ static bool cmd_jtag_scan(target_s *target, int argc, const char **argv)
 	volatile exception_s e;
 	TRY_CATCH (e, EXCEPTION_ALL) {
 #if PC_HOSTED == 1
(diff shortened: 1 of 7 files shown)


esden (approved):
LGTM

Issue #14, opened 2013-02-09
Ubuntu 12.10 upgrade issue


ghulands:
After running the stm32_mem.py script, the red led on the board starts to flash. Then when I run it again, it complains of No devices found and doesn't upgrade the firmware.

Can you please advise what I might try to get my bmpm up to date.

Cheers,
Greg

ghulands@px4:~/blackmagic/src$ sudo ../scripts/stm32_mem.py blackmagic.bin

USB Device Firmware Upgrade - Host Utility -- version 1.1
Copyright (C) 2011  Black Sphere Technologies
License GPLv3+: GNU GPL version 3 or later http://gnu.org/licenses/gpl.html

Device 005: ID 0483:5740 Black Sphere Technologies - Black Magic Probe
Failed to read device state! Assuming APP_IDLE
Run again to upgrade firmware.
ghulands@px4:~/blackmagic/src$ sudo ../scripts/stm32_mem.py blackmagic.bin

USB Device Firmware Upgrade - Host Utility -- version 1.1
Copyright (C) 2011  Black Sphere Technologies
License GPLv3+: GNU GPL version 3 or later http://gnu.org/licenses/gpl.html

No devices found!

Issue #13, opened 2013-02-05
Upload script doesn't work with pyusb 1.0.0a3


ghulands:
Hi,
When installing with $ easy_install pyusb, 1.0.0a3 gets installed. When running the firmware updater, it complains because of missing functions in pyusb.

I would be great if you can update the script to support the latest, or can you let me know which version the script was coded against so I can upload the latest firmware.

Thanks,
Greg

gghmbp:src ghulands$ ../scripts/stm32_mem.py blackmagic.bin 

USB Device Firmware Upgrade - Host Utility -- version 1.1
Copyright (C) 2011  Black Sphere Technologies
License GPLv3+: GNU GPL version 3 or later http://gnu.org/licenses/gpl.html

Traceback (most recent call last):
  File "../scripts/stm32_mem.py", line 68, in <module>
    devs = dfu.finddevs()
  File "/Users/ghulands/blackmagic/scripts/dfu.py", line 165, in finddevs
    for bus in usb.busses():
  File "build/bdist.macosx-10.8-intel/egg/usb/legacy.py", line 343, in busses
  File "build/bdist.macosx-10.8-intel/egg/usb/legacy.py", line 339, in **init**
  File "build/bdist.macosx-10.8-intel/egg/usb/core.py", line 856, in find
ValueError: No backend available

PR #1450, opened 2023-04-11
Renesas RA family Option-Setting

krzysztofgawrys:
## Detailed description

Adds support for Option-Setting Flash memory area for Renesas RA Family

`(gdb) load
Loading section .text, size 0xa84 lma 0x0
Loading section .data, size 0x8 lma 0xa84
Loading section .option_setting_ofs, size 0x10 lma 0x100a100
Loading section .option_setting_sas, size 0x4 lma 0x100a134
Loading section .option_setting_s, size 0xd0 lma 0x100a200
Start address 0x000007f8, load size 2928
Transfer rate: 275 bytes/sec, 418 bytes/write.`

[x] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md)
[x] I've read the [guidelines for contributing](https://github.com/blackmagic-debug/blackmagic/blob/main/CONTRIBUTING.md) to this repository
[x] It builds for hardware native (`make PROBE_HOST=native`)
[x] It builds as BMDA (`make PROBE_HOST=hosted`)
[x] I've tested it to the best of my ability
[x] My commit messages provide a useful short description of what the commits do

### DIFF
--- a/src/target/renesas.c
+++ b/src/target/renesas.c
@@ -195,6 +195,10 @@ typedef enum {
 #define FENTRYR_PE_CF      (1U)
 #define FENTRYR_PE_DF      (1U << 7U)
 
+/* Option-Setting Flash memory for RA4 and RA6, RA2 is different */
+#define RENESAS_OF_START UINT32_C(0x0100A100) /* Start of Option-Setting for RA4 and RA6 Family */
+#define RENESAS_OF_SIZE UINT32_C(0x200) /* Size of Option-Setting area */
+
 /* Renesas RA MCUs can have one of two kinds of flash memory, MF3/4 and RV40 */
 
 #define RENESAS_CF_END UINT32_C(0x00200000) /* End of Flash (maximum possible across families) */
@@ -225,6 +229,7 @@ typedef enum {
 #define RV40_DF_BLOCK_SIZE         (0x40U)
 #define RV40_CF_WRITE_SIZE         (0x80U)
 #define RV40_DF_WRITE_SIZE         (0x4U)
+#define RV40_OF_WRITE_SIZE         (0x10U) /* as stated in Table 44.1 of RA4M2 User's Manual */
 
 /* RV40 Flash Commands */
 #define RV40_CMD               UINT32_C(0x407e0000)
@@ -487,7 +492,8 @@ static bool renesas_rv40_prepare(target_flash_s *f)
 	}
 
 	/* code flash or data flash operation */
-	const bool code_flash = f->start < RENESAS_CF_END;
+	/* Option-Setting flash is CF type as per Table 44.1 of RA4M2 User's Manual */
+	const bool code_flash = f->start < (RENESAS_OF_START + RENESAS_OF_SIZE);
 
 	/* Transition to PE mode */
 	const pe_mode_e pe_mode = code_flash ? PE_MODE_CF : PE_MODE_DF;
@@ -510,6 +516,10 @@ static bool renesas_rv40_flash_erase(target_flash_s *f, target_addr_t addr, size
 
 	/* code flash or data flash operation */
 	const bool code_flash = addr < RENESAS_CF_END;
+
+	/* Erasing Option-Setting flash memory is not allowed, only write is supported */
+	if (RENESAS_OF_START <= addr && addr <= (RENESAS_OF_START + RENESAS_OF_SIZE))
+		return true;
 
 	/* Set Erasure Priority Mode */
 	target_mem_write16(t, RV40_FCPSR, RV40_FCPSR_ESUSPMD);
@@ -558,9 +568,17 @@ static bool renesas_rv40_flash_write(target_flash_s *f, target_addr_t dest, cons
 
 	/* code flash or data flash operation */
 	const bool code_flash = dest < RENESAS_CF_END;
+	bool option_flash = false;
+
+	if (RENESAS_OF_START <= dest && dest <= (RENESAS_OF_START + RENESAS_OF_SIZE))
+		option_flash = true;
 
 	/* write size for code flash / data flash */
-	const uint8_t write_size = code_flash ? RV40_CF_WRITE_SIZE : RV40_DF_WRITE_SIZE;
+	uint8_t write_size = code_flash ? RV40_CF_WRITE_SIZE : RV40_DF_WRITE_SIZE;
+
+	/* Option-Setting write size is different to CF and DF memory */
+	if (option_flash)
+		write_size = RV40_OF_WRITE_SIZE;
 
 	while (len) {
 		/* set block start address */
@@ -571,15 +589,21 @@ static bool renesas_rv40_flash_write(target_flash_s *f, target_addr_t dest, cons
 		len -= write_size;
 
 		/* issue two part Write commands */
-		target_mem_write8(t, RV40_CMD, RV40_CMD_PROGRAM);
-		target_mem_write8(t, RV40_CMD, (uint8_t)(write_size / 2U));
+		if (option_flash) {
+			target_mem_write8(t, RV40_CMD, RV40_CMD_CONFIG_SET_1);
+			target_mem_write8(t, RV40_CMD, RV40_CMD_CONFIG_SET_2);
+		} else {
+			target_mem_write8(t, RV40_CMD, RV40_CMD_PROGRAM);
+			target_mem_write8(t, RV40_CMD, (uint8_t)(write_size / 2U));
+		}
 
 		/* according to reference manual the data buffer full time for 2 bytes is 2 usec.
 		 * this is with a FCLK of 4MHz
 		 * a complete should take less than 1 msec.
 		 */
 		platform_timeout_s timeout;
-		platform_timeout_set(&timeout, 10);
+		/* 200ms is arbitrary number i made up */
+		platform_timeout_set(&timeout, option_flash ? 200: 20);
 
 		/* write one chunk */
 		for (size_t i = 0U; i < (write_size / 2U); i++) {
@@ -620,6 +644,10 @@ static void renesas_add_rv40_flash(target_s *t, target_addr_t addr, size_t lengt
 	f->prepare = renesas_rv40_prepare;
 	f->done = renesas_rv40_done;
 
+	if (RENESAS_OF_START <= addr && addr <= (RENESAS_OF_START + RENESAS_OF_SIZE)) {
+		f->blocksize = RV40_DF_BLOCK_SIZE;
+		f->writesize = RV40_OF_WRITE_SIZE;
+	} else
 	if (code_flash) {
 		f->blocksize = RV40_CF_REGION1_BLOCK_SIZE;
 		f->writesize = RV40_CF_WRITE_SIZE;
@@ -805,6 +833,7 @@ bool renesas_probe(target_s *t)
 	case PNR_SERIES_RA4M2:
 	case PNR_SERIES_RA4M3:
 	case PNR_SERIES_RA4E1:
+		renesas_add_flash(t, RENESAS_OF_START, RENESAS_OF_SIZE); /* Option flash memory 512B 0x0100A100 */
 		renesas_add_flash(t, 0x08000000, 8U * 1024U); /* Data flash memory 8 KB 0x08000000 */
 		target_add_ram(t, 0x20000000, 128U * 1024U);  /* SRAM 128 KB 0x20000000 */
 		target_add_ram(t, 0x28000000, 1024U);         /* Standby SRAM 1 KB 0x28000000 */


src/target/renesas.c
```diff
@@ -195,6 +195,10 @@ typedef enum {
 #define FENTRYR_PE_CF      (1U)
 #define FENTRYR_PE_DF      (1U << 7U)
 
+/* Option-Setting Flash memory for RA4 and RA6, RA2 is different */
+#define RENESAS_OF_START UINT32_C(0x0100A100) /* Start of Option-Setting for RA4 and RA6 Family */
+#define RENESAS_OF_SIZE UINT32_C(0x200) /* Size of Option-Setting area */
```
perigoso, line 200: This needs better naming, the other flash related macros are tied to the flash type, so no extra context is required, here it's hard to know what this is about

Does this have to be hardcoded like this? what is this tied to, the family, perhaps flash type?

src/target/renesas.c
```diff
@@ -487,7 +492,8 @@ static bool renesas_rv40_prepare(target_flash_s *f)
 	}
 
 	/* code flash or data flash operation */
-	const bool code_flash = f->start < RENESAS_CF_END;
+	/* Option-Setting flash is CF type as per Table 44.1 of RA4M2 User's Manual */
+	const bool code_flash = f->start < (RENESAS_OF_START + RENESAS_OF_SIZE);
```
perigoso, line 496: This looks like it breaks parts with flashes that go over that particular option flash adress, I don't remember if thats possible, but I have to assume I chose the original value for  areason

src/target/renesas.c
```diff
@@ -511,6 +517,10 @@ static bool renesas_rv40_flash_erase(target_flash_s *f, target_addr_t addr, size
 	/* code flash or data flash operation */
 	const bool code_flash = addr < RENESAS_CF_END;
 
+	/* Erasing Option-Setting flash memory is not allowed, only write is supported */
+	if (RENESAS_OF_START <= addr && addr <= (RENESAS_OF_START + RENESAS_OF_SIZE))
+		return true;
```
perigoso, line 522: if erasing is not allowed why is there a flash_erase method tied to it, we should omit it, I don't know if that's currently allowed, but It should if not, or have a stub, that said retaining the sanity check might still be good

src/target/renesas.c
```diff
@@ -558,9 +568,17 @@ static bool renesas_rv40_flash_write(target_flash_s *f, target_addr_t dest, cons
...
+	if (RENESAS_OF_START <= dest && dest <= (RENESAS_OF_START + RENESAS_OF_SIZE))
+		option_flash = true;
 
 	/* write size for code flash / data flash */
-	const uint8_t write_size = code_flash ? RV40_CF_WRITE_SIZE : RV40_DF_WRITE_SIZE;
+	uint8_t write_size = code_flash ? RV40_CF_WRITE_SIZE : RV40_DF_WRITE_SIZE;
+
+	/* Option-Setting write size is different to CF and DF memory */
+	if (option_flash)
+		write_size = RV40_OF_WRITE_SIZE;
```
perigoso, line 581: It's starting to look like it might be worth to store this information within the flash struct, in particular because after the rework a while ago we are guaranteed fixed sizes here so a lot of code could be pruned

src/target/renesas.c
```diff
@@ -571,15 +589,21 @@ static bool renesas_rv40_flash_write(target_flash_s *f, target_addr_t dest, cons
...
+		}
 
 		/* according to reference manual the data buffer full time for 2 bytes is 2 usec.
 		 * this is with a FCLK of 4MHz
 		 * a complete should take less than 1 msec.
 		 */
 		platform_timeout_s timeout;
-		platform_timeout_set(&timeout, 10);
+		/* 200ms is arbitrary number i made up */
+		platform_timeout_set(&timeout, option_flash ? 200: 20);
```
perigoso, line 606: surely the expected time is documented somehow, it was for most other flash related things, why 200 over 20, was 20 too short?

perigoso (commented):
Good work!

I think the way you implemented this with the hardcoded value for RA4/6 might negatively affect other targets, I was very careful with possible address collisions originally, and read through all existing families, and I want to make sure you have the same confidence here, note that this is just a suspicion I can't back up ATM.

krzysztofgawrys:
Converted this to draft as there is too many problems with this code.

perigoso:
Went through all of the families documentation, here is a table with the extracted address spaces

| family | code end   | option start | option size | option 2 start | option2 size | option3 start | option3 size | data flash | data size  | factory flash | factory size | sramhs     | sramhs size | sram0      | sram0 size | stdby sram | stdby sram size | type | pnr loc |
|--------|------------|--------------|-------------|----------------|--------------|---------------|--------------|------------|------------|---------------|--------------|------------|-------------|------------|------------|------------|-----------------|------|---------|
| ra2l1  | 0x00040000 | 0x01010010   | 0x01010034  | -              | -            | -             | -            | 0x40100000 | 0x40102000 | -             | -            | -          | -           | 0x20000000 | 0x20008000 | -          | -               | MF4  | FL1     |
| ra2e1  | 0x00020000 | 0x01010010   | 0x01010034  | -              | -            | -             | -            | 0x40100000 | 0x40101000 | -             | -            | -          | -           | 0x20004000 | 0x20008000 | -          | -               | MF4  | FL1     |
| ra2e2  | 0x00010000 | 0x01010010   | 0x01010034  | -              | -            | -             | -            | 0x40100000 | 0x40100800 | -             | -            | -          | -           | 0x20004000 | 0x20006000 | -          | -               | MF4  | FL1     |
| ra2a1  | 0x00040000 | 0x01010008   | 0x01010034  | -              | -            | 0x407FB19C    | 0x407FB1A0   | 0x40100000 | 0x40102000 | -             | -            |            |             | 0x20000000 | 0x20008000 | -          | -               | MF3  | FRT     |
| ra4m1  | 0x00040000 | 0x01010008   | 0x01010034  | -              | -            | 0x407FB19C    | 0x407FB1A0   | 0x40100000 | 0x40102000 | -             | -            | -          | -           | 0x20000000 | 0x20008000 | -          | -               | MF3  | -       |
| ra4m2  | 0x00080000 | 0x0100A100   | 0x0100A300  | -              | -            | -             | -            | 0x08000000 | 0x08002000 | 0x010080F0    | 0x010081B4   | -          | -           | 0x20000000 | 0x20020000 | 0x28000000 | 0x28000400      | RV40 | FL2     |
| ra4m3  | 0x00100000 | 0x0100A100   | 0x0100A300  | -              | -            | -             | -            | 0x08000000 | 0x08002000 | 0x010080F0    | 0x010081B4   | -          | -           | 0x20000000 | 0x20020000 | 0x28000000 | 0x28000400      | RV40 | FL2     |
| ra4e1  | 0x00080000 | 0x0100A100   | 0x0100A300  | -              | -            | -             | -            | 0x08000000 | 0x08002000 | 0x010080F0    | 0x010081B4   | -          | -           | 0x20000000 | 0x20020000 | 0x28000000 | 0x28000400      | RV40 | FL2     |
| ra4e2  | 0x00020000 | 0x0100A100   | 0x0100A300  | -              | -            | -             | -            | 0x08000000 | 0x08001000 | 0x010080F0    | 0x010081B4   | -          | -           | 0x20000000 | 0x2000A000 | 0x28000000 | 0x28000400      | RV40 | FL2     |
| ra4w1  | 0x00080000 | 0x01010008   | 0x01010034  | -              | -            | 0x407FB19C    | 0x407FB1A0   | 0x40100000 | 0x40102000 | -             | -            | -          | -           | 0x20000000 | 0x20018000 | -          | -               | MF3  | -       |
| ra6m1  | 0x00080000 | 0x01007000   | 0x01008000  | 0x0100A150     | 0x0100A168   | 0x407FB17C    | 0x407FB1A0   | 0x40100000 | 0x40102000 | -             | -            | 0x1FFE0000 | 0x20000000  | 0x20000000 | 0x20020000 | 0x200FE000 | 0x20100000      | RV40 | FRT     |
| ra6m2  | 0x00100000 | 0x01007000   | 0x01008000  | 0x0100A150     | 0x0100A168   | 0x407FB17C    | 0x407FB1A0   | 0x40100000 | 0x40108000 | -             | -            | 0x1FFE0000 | 0x20000000  | 0x20000000 | 0x20040000 | 0x200FE000 | 0x20100000      | RV40 | FRT     |
| ra6m4  | 0x00280000 | 0x0100A100   | 0x0100A300  | -              | -            | -             | -            | 0x08000000 | 0x08002000 | 0x010080F0    | 0x010081B4   | -          | -           | 0x20000000 | 0x20040000 | 0x28000000 | 0x28000400      | RV40 | FL2     |
| ra6m5  | 0x00300000 | 0x0100A100   | 0x0100A300  | -              | -            | -             | -            | 0x08000000 | 0x08002000 | 0x010080F0    | 0x010081B4   | -          | -           | 0x20000000 | 0x20080000 | 0x28000000 | 0x28000400      | RV40 | FL2     |
| ra6e1  | 0x00280000 | 0x0100A100   | 0x0100A300  | -              | -            | -             | -            | 0x08000000 | 0x08002000 | 0x010080F0    | 0x010081B4   | -          | -           | 0x20000000 | 0x20040000 | 0x28000000 | 0x28000400      | RV40 | FL2     |
| ra6e2  | 0x00040000 | 0x0100A100   | 0x0100A300  | -              | -            | -             | -            | 0x08000000 | 0x08001000 | 0x010080F0    | 0x010081B4   | -          | -           | 0x20000000 | 0x2000A000 | 0x28000000 | 0x28000400      |      | FL2     |
| ra6t1  | 0x00080000 | 0x01007000   | 0x01008000  | 0x0100A150     | 0x0100A168   | 0x407FB17C    | 0x407FB1A0   | 0x40100000 | 0x40102000 | -             | -            | 0x1FFE0000 | 0x1FFF0000  | -          | -          | -          | -               | RV40 | FRT     |
| ra6t2  | 0x00080000 | 0x0100A100   | 0x0100A300  | -              | -            | -             | -            | 0x08000000 | 0x08004000 | 0x010080F0    | 0x010081B4   | -          | -           | 0x20000000 | 0x20010000 | 0x28000000 | 0x28000400      | RV40 | FL2     |

perigoso:
After reading quite a bit, and re-learning these MCUs, I'm questioning if the option-memory is something we should expose like this, as a flash block, and I'm guessing it was the reason I didn't include these memories on the initial implementation...
These memories are very special purpose, they contain reserved registers, read only registers, and registers with bits that cannot be set after clearing, that's a lot of potential for undesired behavior, and not at all what a user expects when dealing with a flash memory, potentially bricking devices in the worst cases.
The way I would maybe tackle this is with something like is being done on other MCUs like stm32 or sam3U, with dedicated commands to work with the options, that said I think the project would need some work, providing some sort of framework to do that, otherwise we end up with a lot of very specific and hard to maintain code like we already do...

perigoso (changes requested):
With my last comment in consideration, let's refactor this PR, reduce the scope. Let's add the necessary functionality to work with those pages, but not expose them as flashes to the user

krzysztofgawrys:
I do not agree we should hide that memory from user, there are option settings that are essential for weird MCU startup configurations, I can agree on making part of this option-memory read-only in driver, with additional option to make them writable.

krzysztofgawrys:
So, just tested something, e2 Studio produces load with option-settings included and that is understandable as IWDT, WDT, HOCO and Voltage Detection are part of this settings, it finds out we need to add option-setting flash like I did in this PR, otherwise, gdb doesn't know what to do with this sections and fails immediately

> (gdb) load
> Loading section .option_setting_ofs, size 0x10 lma 0x100a100
> Load failed

my aproach is that I will expose that flash but ignore writes except for "safe" registers

src/target/renesas.c
krzysztofgawrys: not allowed to not add erase handler in driver, it will crash

> #0  0x0000000000000000 in ?? ()
> #1  0x000055555557a732 in target_flash_erase (t=0x5555555a9630, addr=16818432, len=64) at target/target_flash.c:149
> #2  0x0000555555565320 in handle_v_packet (plen=29, packet=0x5555555a3630 <pbuf> "vFlashErase:0100a100,00000040") at gdb_main.c:624
> #3  gdb_main_loop (tc=<optimized out>, pbuf=0x5555555a3630 <pbuf> "vFlashErase:0100a100,00000040", pbuf_size=<optimized out>, size=29, in_syscall=<optimized out>) at gdb_main.c:339
> #4  0x000055555555c04b in bmp_poll_loop () at main.c:68
> #5  main (argc=<optimized out>, argv=<optimized out>) at main.c:84

krzysztofgawrys: @dragonmux do we have stub for such situations?

dragonmux: We do not - you could modify `target_flash_erase()` to check the pointer for NULL before calling it (treating nullptr as returning `true`).

One note of feedback about the check done in the snippet above, because `RENESAS_OF_START + RENESAS_OF_SIZE` gives an address that is more than the address range of the region, you want the second comparison to just be `addr < `, not `addr <= `

krzysztofgawrys:
here is compiled map is option-setting registers locations (excluding registers in code flash for some MCU series), red marked are registers with OTP, it seems we can allow write to all of option-setting flash except for three ranges
image

src/target/renesas.c
```diff
@@ -78,17 +78,142 @@ typedef enum {
...
+			NONE, NONE, NONE, FLASH_RV40, PNR_FRT}},
+	{PNR_SERIES_RA6T2,
+		{0x0100a100, 512, NONE, NONE, 0x08000000, 0x08004000, 0x010080f0, 196, NONE, NONE, 0x20000000, 0x20010000, NONE,
+			NONE, 0x28000000, 0x28000400, FLASH_RV40, PNR_FL2}},
+};
+
+renesas_ra_family_details_s renesas_ra_family_lookup(pnr_series_t series)
+{
+	renesas_ra_family_details_s details = {};
+	for (uint32_t i = 0U; i < sizeof(renesas_ra_family) / 19; i++) {
```
perigoso, line 209:
```suggestion
	for (size_t i = 0; i < ARRAY_LENGTH(renesas_ra_family); ++i) {
```
use `ARRAY_LENGTH()` (aka `(sizeof(arr) / sizeof(arr[0]))`) here to prevent mistakes, it's defined in `general.h`, and `size_t` for the type

src/target/renesas.c
```diff
@@ -78,17 +78,142 @@ typedef enum {
...
+	uint32_t factory_flash_start;
+	uint32_t factory_size;
+	uint32_t sramhs_start;
+	uint32_t sramhs_end;
+	uint32_t sram0_start;
+	uint32_t sram0_end;
+	uint32_t sram1_start;
+	uint32_t sram1_end;
+	uint32_t stdby_sram_start;
+	uint32_t stdby_sram_end;
```
perigoso, line 124: replace `_end` addresses with size, I missed that In the table I shared, we might be able to store only 16b for sizes at least on some variables

src/target/renesas.c
```diff
@@ -784,88 +914,42 @@ bool renesas_probe(target_s *t)
...
-	case PNR_SERIES_RA4M1:
-		renesas_add_flash(t, 0x40100000, 8U * 1024U); /* Data flash memory 8 KB 0x40100000 */
-		target_add_ram(t, 0x20000000, 32U * 1024U);   /* SRAM 32 KB 0x20000000 */
-		break;
-
-	case PNR_SERIES_RA2E1:
-		renesas_add_flash(t, 0x40100000, 4U * 1024U); /* Data flash memory 4 KB 0x40100000 */
-		target_add_ram(t, 0x20004000, 16U * 1024U);   /* SRAM 16 KB 0x20004000 */
-		break;
+	priv_storage->details = renesas_ra_family_lookup(priv_storage->series);
```
perigoso, line 917: Check that you got a valid family, we can't know that we have every one mapped

src/target/renesas.c
```diff
@@ -637,47 +798,16 @@ static void renesas_add_flash(target_s *t, target_addr_t addr, size_t length)
...
 
-	switch (priv_storage->series) {
-	case PNR_SERIES_RA2L1:
-	case PNR_SERIES_RA2E1:
-	case PNR_SERIES_RA2E2:
-	case PNR_SERIES_RA2A1:
-	case PNR_SERIES_RA4M1:
-	case PNR_SERIES_RA4W1:
+	case FLASH_MF4:
+		DEBUG_WARN("Found renesas chip with Flash type MF4, not implemented\n");
```
perigoso, line 807: You can merge these two, AFAIR, the driver will be the same when time comes to implement it, and the extra debug info is not useful as we have already have the family, better so save memory and use 1 string for both

perigoso (changes requested):

perigoso:
Looking better :) good work

krzysztofgawrys:
@perigoso @dragonmux 

what did I broke?

> *** stack smashing detected ***: terminated
> 
> Thread 1 "blackmagic" received signal SIGABRT, Aborted.
> __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
> 50      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
> (gdb) bt
> #0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
> #1  0x00007ffff7db4859 in __GI_abort () at abort.c:79
> #2  0x00007ffff7e1f26e in __libc_message (action=action@entry=do_abort, fmt=fmt@entry=0x7ffff7f4908f "*** %s ***: terminated\n") at ../sysdeps/posix/libc_fatal.c:155
> #3  0x00007ffff7ec1aba in __GI___fortify_fail (msg=msg@entry=0x7ffff7f49077 "stack smashing detected") at fortify_fail.c:26
> #4  0x00007ffff7ec1a86 in __stack_chk_fail () at stack_chk_fail.c:24
> #5  0x000055555557b97f in target_mem_read8 (t=t@entry=0x5555555ac630, addr=addr@entry=1082122256) at target/target.c:533
> #6  0x000055555557a4c5 in renesas_rv40_error_check (t=t@entry=0x5555555ac630, error_bits=error_bits@entry=16384) at target/renesas.c:554
> #7  0x000055555557a889 in renesas_rv40_prepare (f=0x5555555ac980) at target/renesas.c:610
> #8  0x000055555557bf8e in flash_prepare (f=0x5555555ac980) at target/target_flash.c:93
> #9  flash_prepare (f=0x5555555ac980) at target/target_flash.c:86
> #10 target_flash_erase (t=0x5555555ac630, addr=0, len=32768) at target/target_flash.c:150
> #11 0x0000555555565baf in handle_v_packet (plen=29, packet=0x5555555a6050 <pbuf> "vFlashErase:00000000,00008000") at gdb_main.c:624
> #12 gdb_main_loop (tc=<optimized out>, pbuf=0x5555555a6050 <pbuf> "vFlashErase:00000000,00008000", pbuf_size=<optimized out>, size=29, in_syscall=<optimized out>) at gdb_main.c:339
> #13 0x000055555555c07e in bmp_poll_loop () at main.c:68
> #14 main (argc=<optimized out>, argv=<optimized out>) at main.c:84

dragonmux:
Enable ASAN with `make PROBE_HOST=hosted ASAN=1 ENABLE_DEBUG=1` and re-run to find out - from just "stack smashing" we can't know, but ASAN will trap the moment w/e bad thing happened happens and print proper stack traces and diagnostics on the bad write

krzysztofgawrys:
@dragonmux here is report with ASAN enabled, strangely it shows DEBUG macro, it seems not related to what I'm doing in this PR, because when i commented out this whole debug section of `send_recv` the bug went away.
I think this deserves its own Issue.

```=================================================================
==5469==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffd067a9871 at pc 0x562643bb0335 bp 0x7ffd067a96c0 sp 0x7ffd067a96b0
READ of size 1 at 0x7ffd067a9871 thread T0
    #0 0x562643bb0334 in send_recv platforms/hosted/bmp_libusb.c:437
    #1 0x562643bb2cfc in read_retry platforms/hosted/stlinkv2.c:355
    #2 0x562643bb2cfc in stlink_mem_read platforms/hosted/stlinkv2.c:871
    #3 0x562643b810a8 in adiv5_mem_read platforms/hosted/platform.c:642
    #4 0x562643b98ac8 in target_mem_read8 target/target.c:532
    #5 0x562643b95518 in renesas_rv40_error_check target/renesas.c:534
    #6 0x562643b9599d in renesas_rv40_prepare target/renesas.c:590
    #7 0x562643b99cd6 in flash_prepare target/target_flash.c:93
    #8 0x562643b99cd6 in flash_prepare target/target_flash.c:86
    #9 0x562643b99cd6 in target_flash_erase target/target_flash.c:147
    #10 0x562643b6d400 in handle_v_packet /home/kgawrys/blackmagic/src/gdb_main.c:624
    #11 0x562643b6d400 in gdb_main_loop /home/kgawrys/blackmagic/src/gdb_main.c:339
    #12 0x562643b56f70 in bmp_poll_loop /home/kgawrys/blackmagic/src/main.c:68
    #13 0x562643b56f70 in main /home/kgawrys/blackmagic/src/main.c:84
    #14 0x7f055ecc6082 in __libc_start_main ../csu/libc-start.c:308
    #15 0x562643b5781d in _start (/home/kgawrys/blackmagic/src/blackmagic+0x3181d)

Address 0x7ffd067a9871 is located in stack of thread T0 at offset 33 in frame
    #0 0x562643b989f3 in target_mem_read8 target/target.c:529

  This frame has 1 object(s):
    [32, 33) 'result' (line 530) <== Memory access at offset 33 overflows this variable
HINT: this may be a false positive if your program uses some custom stack unwind mechanism, swapcontext or vfork
      (longjmp and C++ exceptions *are* supported)
SUMMARY: AddressSanitizer: stack-buffer-overflow platforms/hosted/bmp_libusb.c:437 in send_recv
Shadow bytes around the buggy address:
  0x100020ced2b0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x100020ced2c0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x100020ced2d0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x100020ced2e0: 00 00 00 00 00 00 00 00 00 00 00 00 f1 f1 f1 f1
  0x100020ced2f0: 00 00 f3 f3 00 00 00 00 00 00 00 00 00 00 00 00
=>0x100020ced300: 00 00 00 00 00 00 00 00 00 00 f1 f1 f1 f1[01]f3
  0x100020ced310: f3 f3 00 00 00 00 00 00 00 00 00 00 00 00 f1 f1
  0x100020ced320: f1 f1 04 f3 f3 f3 00 00 00 00 00 00 00 00 00 00
  0x100020ced330: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x100020ced340: 00 00 00 00 00 00 f1 f1 f1 f1 f1 f1 01 f2 04 f2
  0x100020ced350: 04 f2 04 f2 04 f2 04 f2 04 f2 04 f2 04 f2 04 f2
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
  Shadow gap:              cc
==5469==ABORTING```

**Comment by dragonmux** [comment]
Agreed, that should be its own separate issue as a bug in BMDA and the ST-Link backend. If you could open a new issue for that and also test and see if it reproduces with feature/improved-debug-logging (in case we've accidentally fixed it.. but we doubt that) that'd be fantastic.

File: src/target/renesas.c (hunk 5 above)
Comment (krzysztofgawrys) [reply src/target/renesas.c]: 20ms was way too short, I ended up on 100ms, tested 50 gave me errors, 75 was stable, but I could only test RA4M2, not sure what other series would behave like, also this timeout should be as high as we feel good with it, this is just timeout, there is `RV40_FSTATR_RDY` check afterwards anyway, so we are not making things go slow.

File: src/target/target_flash.c [hunk 10]
```diff
@@ -134,6 +134,10 @@ bool target_flash_erase(target_s *t, target_addr_t addr, size_t len)
 			return false;
 		}
 
+		/* Target is not implementing flash erase function */
+		if (f->erase == NULL)
+			return true;
+
```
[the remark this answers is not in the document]
krzysztofgawrys: you solution is working, using it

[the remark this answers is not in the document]
perigoso: Theres a mistake here on my part I forgot to edit out, in the else block, we should do some more arithmetic for correctness, It's a bit pedantic but I want to make sure this is correct (I we start at the middle of the flash, we want to subtract only what we theoretically erased, not the full size anyway)

```suggestion
		} else {
			/* This flash does not implement an erase function, skip it */
			len -= MIN((f->start + f->length) -  addr, len);
			addr = f->start + f->length;
		}
```

src/target/renesas.c
```diff
@@ -281,16 +378,19 @@ typedef enum {
...
 	{"uid", renesas_uid, "Prints unique id"},
+	{"option", renesas_cmd_option, "Manipulate option settings"},
 	{NULL, NULL, NULL},
 };
 
 typedef struct renesas_priv {
 	uint8_t pnr[17]; /* 16-byte PNR + 1-byte null termination */
 	renesas_pnr_series_e series;
 	target_addr_t flash_root_table; /* if applicable */
+	renesas_family_s details;
```
perigoso, line 393: now that `renesas_family_s` also holds the `renesas_pnr_series_e` we are storing redundant information, we can remove `series` here and refactor the code to use the `details`

krzysztofgawrys: agree, at this stage we can flatten `renesas_pnr_series_e` completely and remove `details` too

krzysztofgawrys: agreed

src/target/renesas.c
```diff
@@ -78,17 +78,142 @@ typedef enum {
...
+
+typedef struct renesas_ra_family_details {
+	uint32_t option_start;
+	uint32_t option_size;
+	uint32_t option_start_2;
+	uint32_t option_size_2;
+	uint32_t data_flash_start;
+	uint32_t data_flash_end;
+	uint32_t factory_flash_start;
+	uint32_t factory_size;
```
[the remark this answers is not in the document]
krzysztofgawrys: factory flash contains registers for FIXEDn_PNR, UID and MCUVER on some MCUs, other do have FRT, I think this could be also redone to get address of there registers from family table and do not keep them as defines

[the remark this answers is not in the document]
krzysztofgawrys: factory flash aligns only with FIXED2 registers

krzysztofgawrys:
Fun fact, it works with e2 Studio now!

perigoso:
Sorry I left you hanging @krzysztofgawrys, I have been very busy with work, I will try to look at this soon

src/target/renesas.c
```diff
@@ -948,3 +992,100 @@ static bool renesas_uid(target_s *t, int argc, const char **argv)
 
 	return true;
 }
+
+static bool renesas_print_cmd_usage(target_s *t)
+{
+	tc_printf(t, "command help goes here\n");
+	return false;
```
perigoso, line 999: Are you planning on implementing this?

src/target/renesas.c
```diff
@@ -948,3 +992,100 @@ static bool renesas_uid(target_s *t, int argc, const char **argv)
...
+}
+
+static bool renesas_cmd_option(target_s *t, int argc, const char **argv)
+{
+	(void)argc;
+	(void)argv;
```
perigoso, line 1005: variables are being used

perigoso (commented):
This is generally looking good, @dragonmux might be the time you start to take a look at this as I don't have much to add

I will note that the commits are a bit of a mess and need to be cleaned up before we consider merging

dragonmux:
Well, before we dive into reviewing the code, the one thing that need fixing from our side is the PR needs rebasing against main and the contribution guidelines on commit messages followed

src/target/renesas.c
```diff
@@ -91,6 +91,102 @@ typedef enum {
...
+		0x28000000, 0x400, RENESAS_FLASH_RV40, RENESAS_LOCATION_FIXED2},
+};
+
+renesas_family_s renesas_family_lookup(renesas_pnr_series_e series)
+{
+	renesas_family_s details = {};
+	for (size_t i = 0; i < ARRAY_LENGTH(renesas_family); ++i)
+		if (renesas_family[i].series == series)
+			return renesas_family[i];
+	return details;
```
dragonmux, line 187: It would be better for this to construct `details` on when it's actually needed and not on every call by doing:
```suggestion
	return (renesas_family_s){0};
```

src/target/renesas.c
```diff
@@ -445,6 +544,17 @@ static bool renesas_rv40_pe_mode(target_s *const t, const pe_mode_e pe_mode)
 	return true;
 }
 
+static bool renesas_check_restricted_option(target_addr_t dest)
+{
+	for (size_t i = 0; i < ARRAY_LENGTH(renesas_restricted_setting); i++)
+		if (dest < renesas_restricted_setting[i].addr_start || dest > renesas_restricted_setting[i].addr_end)
+			return false;
+		else
+			return true;
```
dragonmux, line 553: This branch of the if statement will always execute on the first loop through where the if condition doesn't match, making this loop always execute just once - drop the else. Also, please put braces on the for body (but not the if)

src/target/renesas.c
```diff
@@ -490,6 +600,10 @@ static bool renesas_rv40_prepare(target_flash_s *const f)
 {
 	target_s *const t = f->t;
 
+	renesas_priv_s *priv_storage = (renesas_priv_s *)t->target_storage;
+	if (!priv_storage)
```
dragonmux, line 604: if priv_storage is always allocated in the probe function and is already NULL checked there, it cannot be NULL here, drop the check.

src/target/renesas.c
```diff
@@ -514,13 +632,33 @@ static bool renesas_rv40_done(target_flash_s *const f)
 	return renesas_rv40_pe_mode(t, PE_MODE_READ);
 }
 
+static bool renesas_check_option_setting(target_s *t, target_addr_t addr)
+{
+	renesas_priv_s *priv_storage = (renesas_priv_s *)t->target_storage;
+	if (!priv_storage)
```
dragonmux, line 638: Same comment about impossible NULL as above

src/target/renesas.c
```diff
@@ -514,13 +632,33 @@ static bool renesas_rv40_done(target_flash_s *const f)
...
 	target_s *const t = f->t;
 
+	uint16_t block_size;
+
 	/* Code flash or data flash operation */
-	const bool code_flash = addr < RENESAS_CF_END;
+	if (addr < RENESAS_CF_END)
+		block_size = addr < RV40_CF_REGION0_SIZE ? RV40_CF_REGION0_BLOCK_SIZE : RV40_CF_REGION1_BLOCK_SIZE;
+	else
+		block_size = f->blocksize;
```
dragonmux, line 661: This looks like it's worth lifting into its own function so  you can say `const uint16_t block_size = renesas_rv40_block_size(f->blocksize, addr);` or so - this will improve codegen and `const`-correctness

src/target/renesas.c
```diff
@@ -565,13 +697,19 @@ static bool renesas_rv40_flash_write(target_flash_s *const f, target_addr_t dest
...
 
-	/* code flash or data flash operation */
-	const bool code_flash = dest < RENESAS_CF_END;
-
 	/* write size for code flash / data flash */
-	const uint8_t write_size = code_flash ? RV40_CF_WRITE_SIZE : RV40_DF_WRITE_SIZE;
+	const uint8_t write_size = f->writesize;
+	bool option_flash = false;
+
+	if (renesas_check_option_setting(t, dest)) {
```
dragonmux, line 704: Drop the braces on this if and the one in the while loop - they're unnecessary.

src/target/renesas.c
```diff
@@ -762,22 +873,22 @@ bool renesas_probe(target_s *t)
 		 */
 
 		if (renesas_pnr_read(t, RENESAS_FIXED2_PNR, pnr)) {
-			DEBUG_WARN("Found renesas chip (%.*s) with pnr location RENESAS_FIXED2_PNR and unsupported Part ID %x "
-					   "please report it\n",
+			DEBUG_WARN("Found Renesas chip (%.*s) with pnr location RENESAS_FIXED2_PNR and unsupported Part ID %" PRIx16
```
dragonmux, line 876: `PRIx16` doesn't do anything special here, please stick to using `%x`.

src/target/renesas.c
```diff
@@ -948,3 +992,100 @@ static bool renesas_uid(target_s *t, int argc, const char **argv)
...
+	tc_printf(t, "command help goes here\n");
+	return false;
+}
+
+static bool renesas_cmd_option(target_s *t, int argc, const char **argv)
+{
+	(void)argc;
+	(void)argv;
+
+	const uint8_t arglen = strlen(argv[1]);
```
dragonmux, line 1007: arglen here should be a size_t as this is the result of strlen()

src/target/renesas.c
```diff
@@ -948,3 +992,100 @@ static bool renesas_uid(target_s *t, int argc, const char **argv)
...
+		if (!renesas_check_option_setting(t, option_addr))
+			return renesas_print_cmd_usage(t);
+
+		/* The address must be aligned */
+		if (option_addr % 0x10 > 0)
+			return renesas_print_cmd_usage(t);
+
+		/* Configuration-Set command always needs 16 bytes of data */
+		const uint8_t len = 16UL;
+		uint8_t data[len];
```
dragonmux, line 1040: Please avoid VLAs. `uint8_t data[16U];` and then when you need the array length, use the `ARRAY_LENGTH()` macro

src/target/renesas.c
```diff
@@ -948,3 +992,100 @@ static bool renesas_uid(target_s *t, int argc, const char **argv)
...
+
+		for (size_t i = 0; i < len;) {
```
dragonmux, line 1042: `i += 4U` should be a part of this loop statement, not in the body of the loop. All constants used here should have the L suffix dropped especially as that might not even result in the right integer type (like in the uint8_t assignment above, where `-Wconversion` would warn about it truncating):

```suggestion
		for (size_t i = 0; i < ARRAY_LENGTH(data); i += 4U) {
```

src/target/renesas.c
```diff
@@ -948,3 +992,100 @@ static bool renesas_uid(target_s *t, int argc, const char **argv)
...
+			target_mem_write32(t, RV40_FSADDR, option_addr);
+
+			/* issue two part Write commands */
+			target_mem_write8(t, RV40_CMD, RV40_CMD_CONFIG_SET_1);
+			target_mem_write8(t, RV40_CMD, RV40_CMD_CONFIG_SET_2);
+
+			platform_timeout_s timeout;
+			platform_timeout_set(&timeout, 200);
+
+			for (size_t i = 0U; i < len;) {
```
dragonmux, line 1065: Same comment as above about the loop increment value. Additionally, drop the braces from the loop and use `read_le2()` from the buffer_utils.h header for constructing the values to pass to `target_mem_write16()`.

src/target/renesas.c
```diff
@@ -948,3 +992,100 @@ static bool renesas_uid(target_s *t, int argc, const char **argv)
...
+			renesas_rv40_pe_mode(t, PE_MODE_READ);
+			return false;
+		}
+	} else
+		return renesas_print_cmd_usage(t);
+
+	renesas_rv40_pe_mode(t, PE_MODE_READ);
+
+	return true;
+}
```
dragonmux, line 1091: This file is missing the final newline, please add it (that is what the "no entry" icon is about in the GitHub interface)

src/target/renesas.c
```diff
@@ -91,6 +91,102 @@ typedef enum {
...
+	RENESAS_FLASH_RV40,
+} renesas_flash_type_e;
+
+typedef enum {
+	RENESAS_LOCATION_FIXED1,
+	RENESAS_LOCATION_FIXED2,
+	RENESAS_LOCATION_FRT,
+} renesas_pnr_location_e;
+
+typedef struct renesas_family {
```
dragonmux, line 106: The order things appear in this structure result in a lot of packing fields being inserted - for example after each uint16_t. Please reorder the fields to put the size information after the start information to reduce the Flash footprint it causes. See the `-Wpacked` compiler flag, or clang-tidy's lints for more information.

dragonmux (changes requested):
First off, thank you for the contribution! This is looking decent but there are a few things that need addressing in addition to Perigoso's outstanding review comments before we can merge this.

PR #1453, opened 2023-04-15
blackpillv2/platform: add function platform_target_voltage_sense

lenvm:
## Detailed description
This pull request fixes the compiler error ```undefined reference to `platform_target_voltage_sense'``` when PLATFORM_HAS_POWER_SWITCH is defined for the blackpillv2.

It fixes the error by adding the function platform_target_voltage_sense as it was previously undefined.

As the blackpillv2 does not sense the voltage on the power pin, the function is implemented by returning a 0.

* [x] It builds for hardware blackpillv2 (`make PROBE_HOST=blackpillv2`)
* [x] It builds for all platforms (`make all_platforms`)

### DIFF
--- a/src/platforms/blackpillv2/platform.c
+++ b/src/platforms/blackpillv2/platform.c
@@ -149,6 +149,22 @@ void platform_target_set_power(const bool power)
 {
 	gpio_set_val(PWR_BR_PORT, PWR_BR_PIN, !power);
 }
+
+/**
+ * @brief Senses the voltage on the power pin in tenths of a volt (so 33 means 3.3V).
+ * 
+ * However it returns 0 for blackpillv2, as it does not sense the voltage on the power pin.
+ * This function is only needed for implementations that allow the target to be powered
+ * from the debug probe.
+ * 
+ * @param void
+ * @return voltage sensed on the power pin. 0 for blackpillv2.
+ */
+uint32_t platform_target_voltage_sense(void)
+{
+	return 0;
+}
+
 #endif
 
 void platform_target_clk_output_enable(bool enable)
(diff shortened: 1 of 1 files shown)


src/platforms/blackpillv2/platform.c
```diff
@@ -149,6 +149,22 @@ void platform_target_set_power(const bool power)
 {
 	gpio_set_val(PWR_BR_PORT, PWR_BR_PIN, !power);
 }
+
+/**
```
dragonmux, line 153: Please do not use Doxygen style comments - a simple block comment more than suffices. For example:
```c
/*
 * A dummy implementation of platform_target_voltage_sense as
 * the Blackpillv2 has no ability to sense the voltage on the power pin.
 * This function is only needed for implementations that allow the target
 * to be powered from the debug probe.
 */
```

src/platforms/blackpillv2/platform.c
```diff
@@ -149,6 +149,22 @@ void platform_target_set_power(const bool power)
...
+ * from the debug probe.
+ * 
+ * @param void
+ * @return voltage sensed on the power pin. 0 for blackpillv2.
+ */
+uint32_t platform_target_voltage_sense(void)
+{
+	return 0;
+}
+
```
dragonmux, line 167: Please remove the excess new line

dragonmux (changes requested):
Good catch. Technically nothing wrong with your change, however using a Doxygen comment goes against the comment style of the entire rest of the code base, so we need you to strip it back to a block comment. We are never going to be generating documentation from the code base in that way as it's not particularly useful in the context, at least for a platform like this.

Once the review items have been addressed however, we are happy to merge this. Thank you for the contribution!

src/platforms/blackpillv2/platform.c
lenvm: Thanks for your reply! The change to a block comment is implemented in d32b652.

src/platforms/blackpillv2/platform.c
lenvm: This is implemented in d32b652.

dragonmux (approved):
LGTM, thank you! Merging

PR #1459, opened 2023-04-18
Feature: better remote protocol fallbacks

dragonmux:
## Detailed description

In the wake of #1389 which introduced remote protocol v3 to address the issue of error propagation from the firmware into BMDA, we patched BMDA to only talk the newest remote protocol, making firmware running protocol v2 and older fully fall back onto the raw JTAG/SWD routines without any ADIv5 acceleration. Needless to say, this results in a more than 5x slowdown on the older firmware.

Knowing that we will be having to bump the remote protocol again in v2.0 and possibly again in the v1.10 cycle, we've opted to build a system where new revisions of the remote protocol can be easily plopped into BMDA side-by-side, allowing complete graceful fallback all the way to the original remote protocol if necessary.

This PR implements this new system, complete with implementations of the original remote protocol (v0), the ADIv5 acceleration enhanced v0 protocol (v0+), the first multi-drop aware version of the remote protocol (v1), the old remote protocol (v2), and the current remote protocol (v3). The v2 and v3 implementations have been tested, but the others are not. Implementations of the older protocols have been blended from the repo history and the current protocol implementation.

It is intended that this new mechanism will make extending and changing the remote protocol going forward significantly easier with a minimum of extra code and a clean way to create fallbacks.


esden (approved):
LGTM, it will definitely improve the experience. We will need to make sure that the warning about "outdated" firmware on the BMP is more prominent. But this fallback mechanism will improve user experience. Thanks!

PR #1460, opened 2023-04-19
Reset pulse length, fix ST-Link v2 clone nRST pin

silbe:
## Detailed description

This patch series fixes two issues I had using the BMP firmware on an ST-Link v2 clone with a target that requires using SRST:
- The nRST pulse length is too short and didn't trigger a target reset at all in my case.

  This is addressed by adding an optional parameter to `cmd_reset` specifying the pulse length in milliseconds, defaulting to the previous behaviour of no delay (0ms).
- Unlike the original ST-Link v2 hardware which has nRST connected to PB0 or PB1, the "clones" use PB6.

  Despite having completely different PCB layouts and ICs, all three different types of "clones" I examined don't have nRST hooked up to PB0 or PB1 but rather to PB6. Some have PB6 and PB7 connected together.
  The clones are now assigned their own hardware revision (0x101 to leave plenty of room for more variations of the original hardware). Based on this hardware revision, BMP will now use PB6 as nRST for the clones.

*N.B.*: Building the BMP firmware for ST-Link clones currently requires disabling some of the targets as otherwise it's too large to fit into the 128KiB flash of the clones (build time error). This requires manual changes to `src/Makefile`. For testing this patch series I disabled the `lpc*` targets.

* [x] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md) that has no contact method filled in.

## Closing issues
None.

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -91,7 +91,7 @@ const command_s cmd_list[] = {
 	{"morse", cmd_morse, "Display morse error message"},
 	{"halt_timeout", cmd_halt_timeout, "Timeout (ms) to wait until Cortex-M is halted: (Default 2000)"},
 	{"connect_rst", cmd_connect_reset, "Configure connect under reset: (enable|disable)"},
-	{"reset", cmd_reset, "Pulse the nRST line - disconnects target"},
+	{"reset", cmd_reset, "Pulse the nRST line - disconnects target: (pulse_len_ms, default 0)"},
 	{"tdi_low_reset", cmd_tdi_low_reset,
 		"Pulse nRST with TDI set low to attempt to wake certain targets up (eg LPC82x)"},
 #ifdef PLATFORM_HAS_POWER_SWITCH
@@ -454,10 +454,12 @@ static bool cmd_halt_timeout(target_s *t, int argc, const char **argv)
 static bool cmd_reset(target_s *t, int argc, const char **argv)
 {
 	(void)t;
-	(void)argc;
-	(void)argv;
+	uint32_t pulse_len_ms = 0;
+	if (argc > 1)
+		pulse_len_ms = strtoul(argv[1], NULL, 0);
 	target_list_free();
 	platform_nrst_set_val(true);
+	platform_delay(pulse_len_ms);
 	platform_nrst_set_val(false);
 	return true;
 }
--- a/src/platforms/stlink/platform.c
+++ b/src/platforms/stlink/platform.c
@@ -51,12 +51,19 @@ void platform_init(void)
 	led_idle_run = GPIO13;
 	nrst_pin = NRST_PIN_V1;
 #else
-	if (rev == 0) {
+	switch (rev) {
+	case 0:
 		led_idle_run = GPIO8;
 		nrst_pin = NRST_PIN_V1;
-	} else {
+		break;
+	case 0x101:
+		led_idle_run = GPIO9;
+		nrst_pin = NRST_PIN_CLONE;
+		break;
+	default:
 		led_idle_run = GPIO9;
 		nrst_pin = NRST_PIN_V2;
+		break;
 	}
 #endif
 	/* Setup GPIO ports */
@@ -73,7 +80,7 @@ void platform_init(void)
 	SCB_VTOR = (uintptr_t)&vector_table;
 
 	platform_timing_init();
-	if (rev > 1U) /* Reconnect USB */
+	if ((rev & 0xff) > 1U) /* Reconnect USB */
 		gpio_set(GPIOA, GPIO15);
 	blackmagic_usb_init();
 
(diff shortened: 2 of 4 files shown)


dragonmux (approved):
Unless someone with a ST-Link v2 or one of the clones can chime in saying why we should not merge this as-stands, this looks good to us and we're happy to merge it.

We'll give it a day or two till we pull the trigger on the merge just to let any of the stlink platform users chime in if this somehow breaks things

silbe:
> Unless someone with a ST-Link v2 or one of the clones can chime in saying why we should not merge this as-stands, this looks good to us and we're happy to merge it.
> 
> We'll give it a day or two till we pull the trigger on the merge just to let any of the stlink platform users chime in if this somehow breaks things

Great, thanks! Being careful about not breaking stuff for other users is quite appreciated. :+1: 

> Edit: please fix the whitespace error introduced in the patch to command.c then we'll re-approve this

Done. Also finally flipped the default for C code to use TABs as it's more common across the code bases I'm working with at the moment, plus I installed editorconfig so projects with a corresponding config should use proper settings.

dragonmux (dismissed):

silbe:
Accidentally pushed my local hack for disabling the lpc targets as well. Sorry for that, fixed now.

Should probably add a pre-push hook that recognises these commits but that's not straightforward as there can be only one hook dir, so per-repository hooks would break or need to be chain-loaded from the global hooks directory...

dragonmux:
Checking for unintended changes to the Makefiles is a component of normal review for us anyway as a hook can't tell the difference between unintended changes by way of removals vs quite intentional changes such as a file being renamed, removed on purpose, etc. Easier to do a review and ask questions when there's any uncertainty.

sidprice (commented):
It is not clear that the second parameter of the "mon/reset" command is optional. Perhaps change the text to: target: (pulse_len_ms [, default 0])

silbe:
> It is not clear that the second parameter of the "mon/reset" command is optional. Perhaps change the text to: target: (pulse_len_ms [, default 0])

I'm not opposed to change it in general, but I don't think that variant is an improvement. It looks like the default value of the parameter is optional rather than the parameter itself. :wink: 

From what I can tell the parentheses (`(param)`) are already supposed to indicate that the parameter is optional. Most CLI help texts use brackets (`[param]`) for optional parameters so it's not immediately obvious (at least it wasn't to me). The difference between the new `cmd_reset` description and most other commands is that the default value is documented as well. The only other command with documented default value is `halt_timeout` and that one is quite confusing IMO: It looks like leaving the parameter away would set a value of 2000ms while in reality it will print the current value. So I'm not sure how to improve on the new `cmd_reset` description.

I thinks the reason most of the command descriptions are slightly confusing to me is that they are a mix of syntax and semantics.

silbe:
> Checking for unintended changes to the Makefiles is a component of normal review for us anyway as a hook can't tell the difference between unintended changes by way of removals vs quite intentional changes such as a file being renamed, removed on purpose, etc. Easier to do a review and ask questions when there's any uncertainty.

:+1: 

The hook would be a local thing for me; not for BMP. I already labelled the commit as a local hack that's not supposed to go upstream, I just forgot to check before pushing. :blush:

dragonmux:
> From what I can tell the parentheses ((param)) are already supposed to indicate that the parameter is optional. Most CLI help texts use brackets ([param]) for optional parameters so it's not immediately obvious (at least it wasn't to me). The difference between the new cmd_reset description and most other commands is that the default value is documented as well. The only other command with documented default value is halt_timeout and that one is quite confusing IMO: It looks like leaving the parameter away would set a value of 2000ms while in reality it will print the current value. So I'm not sure how to improve on the new cmd_reset description.

In Black Magic Debug, we use `[]` exclusively for optional parameters. This is consistent with BNF, Regex, etc too. It would be consistent to use this markup for the help text, especially as we need to keep the text as short as we reasonably can to conserve Flash space.

silbe:
> In Black Magic Debug, we use `[]` exclusively for optional parameters. This is consistent with BNF, Regex, etc too. It would be consistent to use this markup for the help text, especially as we need to keep the text as short as we reasonably can to conserve Flash space.

If there's consensus on it I'm more than happy to provide a follow-up MR to switch all command help texts to brackets for optional parameters.

dragonmux:
That works for us :)

dragonmux (approved):

dragonmux:
We'll merge this once the builds complete now we know you've seen the response about the follow-up PR

silbe:
Rebased on `main`; no code changes. Follow-up change to unify help syntax is in #1460.

silbe:
Thanks for the thorough review! Much appreciated.

PR #1474, opened 2023-04-26
Unify help syntax

silbe:
## Detailed description

As discussed in #1460 modify command help output to use the same syntax as used by BMD `--help` output and other applications (e.g. Python `argparse`). This makes it easier to understand which parameters are optional.

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -84,32 +84,34 @@ const command_s cmd_list[] = {
 	{"version", cmd_version, "Display firmware version info"},
 	{"help", cmd_help, "Display help for monitor commands"},
 	{"jtag_scan", cmd_jtag_scan, "Scan JTAG chain for devices"},
-	{"swdp_scan", cmd_swdp_scan, "Scan SW-DP for devices"},
+	{"swdp_scan", cmd_swdp_scan, "Scan SW-DP for devices: [TARGET_ID]"},
 	{"auto_scan", cmd_auto_scan, "Automatically scan all chain types for devices"},
-	{"frequency", cmd_frequency, "set minimum high and low times"},
+	{"frequency", cmd_frequency, "set minimum high and low times: [FREQ]"},
 	{"targets", cmd_targets, "Display list of available targets"},
 	{"morse", cmd_morse, "Display morse error message"},
-	{"halt_timeout", cmd_halt_timeout, "Timeout (ms) to wait until Cortex-M is halted: (Default 2000)"},
-	{"connect_rst", cmd_connect_reset, "Configure connect under reset: (enable|disable)"},
-	{"reset", cmd_reset, "Pulse the nRST line - disconnects target: (pulse_len_ms, default 0)"},
+	{"halt_timeout", cmd_halt_timeout, "Timeout to wait until Cortex-M is halted: [TIMEOUT, default 2000ms]"},
+	{"connect_rst", cmd_connect_reset, "Configure connect under reset: [enable|disable]"},
+	{"reset", cmd_reset, "Pulse the nRST line - disconnects target: [PULSE_LEN, default 0ms]"},
 	{"tdi_low_reset", cmd_tdi_low_reset,
 		"Pulse nRST with TDI set low to attempt to wake certain targets up (eg LPC82x)"},
 #ifdef PLATFORM_HAS_POWER_SWITCH
-	{"tpwr", cmd_target_power, "Supplies power to the target: (enable|disable)"},
+	{"tpwr", cmd_target_power, "Supplies power to the target: [enable|disable]"},
 #endif
 #ifdef ENABLE_RTT
-	{"rtt", cmd_rtt, "enable|disable|status|channel 0..15|ident (str)|cblock|ram|poll maxms minms maxerr"},
+	{"rtt", cmd_rtt,
+		"[enable|disable|status|channel [0..15 ...]|ident [STR]|cblock|ram [RAM_START RAM_END]|poll [MAXMS MINMS "
+		"MAXERR]]"},
 #endif
 #ifdef PLATFORM_HAS_TRACESWO
 #if defined TRACESWO_PROTOCOL && TRACESWO_PROTOCOL == 2
-	{"traceswo", cmd_traceswo, "Start trace capture, NRZ mode: (baudrate) (decode channel ...)"},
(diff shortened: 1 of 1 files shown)


dragonmux:
If you rebase this on main now the first PR is merged, we'll do a review with a view to merging this once it's ready

silbe:
Applied both patches output by `clang-format` (PR checks). The first one used spaces instead of TABs for the second indented line but after the second run it uses TABs as I would have expected.

silbe:
Looks like with the slightly longer help output (120 bytes according to `size-diff`) we're now exceeding the flash size even on the native hardware when RTT is enabled. :frowning_face: 

I guess it's time to make the targets configurable at build-time but I doubt I'll get around to that any time soon.

dragonmux:
Rebuilding the build system and making targets configurable in that way is a part of v2.0, so for now reducing texts and always-embedded strings and keeping leanness in mind is the goal. We'll review these proposed changes and see if we can suggest ways to accomplish this without compromising on the PR goal.

We'd like to keep resorting to disabling targets as a last ditch effort if possible for now.

src/command.c
```diff
@@ -83,33 +83,35 @@ static bool cmd_shutdown_bmda(target_s *t, int argc, const char **argv);
 const command_s cmd_list[] = {
 	{"version", cmd_version, "Display firmware version info"},
 	{"help", cmd_help, "Display help for monitor commands"},
-	{"jtag_scan", cmd_jtag_scan, "Scan JTAG chain for devices"},
-	{"swdp_scan", cmd_swdp_scan, "Scan SW-DP for devices"},
+	{"jtag_scan", cmd_jtag_scan, "Scan JTAG chain for devices: [IR_LENGTH ...]"},
```
dragonmux, line 86: While this is correct, we're also dropping the parameter in #1477 so you can drop this particular change to help save some space

src/command.c
```diff
@@ -83,33 +83,35 @@ static bool cmd_shutdown_bmda(target_s *t, int argc, const char **argv);
...
+	{"swdp_scan", cmd_swdp_scan, "Scan SW-DP for devices: [TARGET_ID]"},
 	{"auto_scan", cmd_auto_scan, "Automatically scan all chain types for devices"},
-	{"frequency", cmd_frequency, "set minimum high and low times"},
+	{"frequency", cmd_frequency, "set minimum high and low times: [FREQUENCY]"},
```
dragonmux, line 89: Perhaps lets use the shortening "FREQ" here for the []'d part

src/command.c
```diff
@@ -83,33 +83,35 @@ static bool cmd_shutdown_bmda(target_s *t, int argc, const char **argv);
...
 	{"targets", cmd_targets, "Display list of available targets"},
 	{"morse", cmd_morse, "Display morse error message"},
-	{"halt_timeout", cmd_halt_timeout, "Timeout (ms) to wait until Cortex-M is halted: (Default 2000)"},
-	{"connect_rst", cmd_connect_reset, "Configure connect under reset: (enable|disable)"},
-	{"reset", cmd_reset, "Pulse the nRST line - disconnects target: (pulse_len_ms, default 0)"},
+	{"halt_timeout", cmd_halt_timeout, "Timeout to wait until Cortex-M is halted: [TIMEOUT_MS, default 2000]"},
```
dragonmux, line 92: it's only one character, but perhaps `[TIMEOUT, default 2000ms]`? that still expresses the units used (similar suggestion for the pulse length one two lines down)

dragonmux (changes requested):
We have 3 suggestions on wording, but otherwise this looks good. We're hoping that with the space savings made in #1477, that allows this all to fit properly with RTT enabled. If not, we'll make RTT builds failing non-fatal in CI as it's mostly a nice to have without disabling targets

silbe:
> Rebuilding the build system and making targets configurable in that way is a part of v2.0, so for now reducing texts and always-embedded strings and keeping leanness in mind is the goal. We'll review these proposed changes and see if we can suggest ways to accomplish this without compromising on the PR goal.
> 
> We'd like to keep resorting to disabling targets as a last ditch effort if possible for now.

BTW, I saw that new BMP boards were produced recently (or at least the PnP machine was getting ramped up for production). Are these still using STM32F103CB with 128KiB flash or do they have a different chip with more flash (e.g. STM32F303Cx)?

dragonmux:
They are using the GD32F103 which has the same amount of Flash as the STM32F103 used before.

src/command.c
dragonmux: it should reduce the code size by a byte per (for 2 bytes total), given we're suggesting `[TIMEOUT_MS, default 2000]` becomes `[TIMEOUT, default 2000ms]`. Interesting point regarding it sounding like you can specify units - that's kind of a problem either way in our opinion and something that the limited number of characters to describe the command causes.

One way we could deal with this is apply our suggestion and then also add some documentation to the website that explains that the commands have fixed units so if people have a query, we can point them to the documentation that can then greatly expand on these strings - how would you feel about that?

silbe: Fine with me. Follow-up MR: blackmagic-debug/black-magic-org#21. Wasn't sure where to place it so I just added a sentence after the example `monitor help` output.

silbe:
> They are using the GD32F103 which has the same amount of Flash as the STM32F103 used before.

OK, too bad. We could move the strings to the new SPI(?) flash but that seems to only save ~6KiB so not really worth the additional complexity and effort.

(Feel free to ignore; just brainstorming ideas on how to cope with limited flash size).

dragonmux:
It is a SPI Flash, but its intended for other purposes such as a stand-alone mode where you can then plug BMP into a target, click a button, and have it program firmware from the Flash to the target.

The space problem is actually a part of why the desire to revamp the build system in v2.0 so we can configure multiple default builds for the firmware, and is why bmputil was born as we didn't just want to simplify the user experience for updating the firmware, but there are plans for allowing the user to switch between pre-built firmware images using the tool kinda like a Minecraft launcher or an app store but for the firmware.

dragonmux (approved):
LGTM, merging despite the build failures as we hope #1477 gets merged soon which should help get the firmware back below the threshold where RTT builds succeed without intervention, and then we'll look at changing CI so we only do RTT builds for platforms with more Flash for now

PR #1464, opened 2023-04-21
Fix: BMDA ST-Link v2 crash

dragonmux:
## Detailed description

This PR fixes the ST-Lijnk backend for BMDA writing 2 bytes to a single byte of storage when doing a target_mem_read8() and entirely overhauls this backend, bringing it mostly up to speed with the other backends.

A side effect of this PR is that both the ST-Link and J-Link backends should now run considerably faster due to their improved use of libusb Bulk transfers.

Tested against the ST-Link v2.1 that is part of the STM32F411 Nucleo board.

This PR also pre-emptively fixes several other issues including the frequency selection/calculation logic and splits the backend up into JTAG-specific, SWD-specific and general components to match the other backends more closely

## Your checklist for this pull request

* [x] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md)
* [x] I've read the [guidelines for contributing](https://github.com/blackmagic-debug/blackmagic/blob/main/CONTRIBUTING.md) to this repository
* [x] It builds for hardware native (`make PROBE_HOST=native`)
* [x] It builds as BMDA (`make PROBE_HOST=hosted`)
* [x] I've tested it to the best of my ability
* [x] My commit messages provide a useful short description of what the commits do

## Closing issues

Fixes: #1455

PR #1475, opened 2023-04-26
Fix DWT mask calculation

silbe:
## Detailed description

DWT can match power-of-two sized memory regions up to an implementation-defined limit. Properly calculate the mask length for region sizes of more then 4 bytes rather than writing reserved bits.

Factored out `stlink_ulog2()` from [`src/platforms/hosted/stlinkv2.c`](https://github.com/blackmagic-debug/blackmagic/blob/5a925ee79c36a2a3f13d32570fd48b890b4a4f4e/src/platforms/hosted/stlinkv2.c) as `ulog2()` into a new source file `src/compat.c` so we can use it in `src/target/cortexm.c`. I didn't find a good location for this function (the closest match I could find was `hex_utils.c`) so I created a new file.

### DIFF
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -1261,22 +1261,22 @@ bool cortexm_run_stub(target_s *t, uint32_t loadaddr, uint32_t r0, uint32_t r1, 
 	return bkpt_instr & 0xffU;
 }
 
-/* The following routines implement hardware breakpoints and watchpoints.
+/*
+ * The following routines implement hardware breakpoints and watchpoints.
  * The Flash Patch and Breakpoint (FPB) and Data Watch and Trace (DWT)
- * systems are used. */
+ * systems are used.
+ */
 
+/*
+ * DWT only supports powers of two as size. Convert length in bytes to
+ * number of least-significant bits of the address to ignore during
+ * match (maximum 31).
+ */
 static uint32_t dwt_mask(size_t len)
 {
-	switch (len) {
-	case 1:
-		return CORTEXM_DWT_MASK_BYTE;
-	case 2:
-		return CORTEXM_DWT_MASK_HALFWORD;
-	case 4:
-		return CORTEXM_DWT_MASK_WORD;
-	default:
-		return -1;
-	}
+	if (len < 2)
+		return 0;
+	return MIN(ulog2(len - 1), 31);
 }
 
 static uint32_t dwt_func(target_s *t, target_breakwatch_e type)
--- a/src/Makefile
+++ b/src/Makefile
@@ -53,6 +53,7 @@ SRC =              \
 	lpc55xx.c      \
 	kinetis.c      \
 	main.c         \
+	maths_utils.c  \
 	morse.c        \
 	msp432.c       \
 	nrf51.c        \
--- a/src/include/general.h
+++ b/src/include/general.h
@@ -39,6 +39,7 @@
 #include <inttypes.h>
 #include <sys/types.h>
 
+#include "maths_utils.h"
 #include "platform.h"
 #include "platform_support.h"
 
--- a/src/include/maths_utils.h
+++ b/src/include/maths_utils.h
@@ -1,0 +1,41 @@
+/*
+ * This file is part of the Black Magic Debug project.
+ *
+ * Copyright (C) 2023 1BitSquared <info@1bitsquared.com>
+ * Written by Rachel Mant <git@dragonmux.network>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its
+ *    contributors may be used to endorse or promote products derived from
+ *    this software without specific prior written permission.
+ *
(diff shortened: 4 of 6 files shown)


dragonmux:
We're glad to see that the ulog2 implementation is coming in handy! `compat.c` though sounds like we're providing compatibility with something, so may we suggest instead something like `maths_utils.c` or `int_maths.c`, with an intention that maths functions will be concentrated there over time?

We'll give this an in-depth review once the help output PR has been reviewed

silbe:
> We're glad to see that the ulog2 implementation is coming in handy! `compat.c` though sounds like we're providing compatibility with something, so may we suggest instead something like `maths_utils.c` or `int_maths.c`, with an intention that maths functions will be concentrated there over time?

The idea was that it's for compatibility between different compilers (as `ulog2()` is little more than a wrapper around `__builtin_clz()` / `__lzcnt()` plus a fallback implementation) but a math functions collection would be an equally good fit. At least some other built-ins we might want to use in the future also fit the math umbrella (e.g. `ffs()`) so I'm fine with either choice. Just let me know which one you prefer.

> We'll give this an in-depth review once the help output PR has been reviewed

:+1:

dragonmux:
We consider the count leading zeros intrinsic to only be a means to an end, with additional maths having to be performed around the answers the instruction gives to construct the actual logarithm, so it being a maths function in a maths function file gives us the most flexibility for where to put additional maths functions in the future without it being confusing to the reader why that function's in the file

src/compat.c
```diff
@@ -0,0 +1,62 @@
+/*
+ * This file is part of the Black Magic Debug project.
+ *
+ * Copyright (C) 2019-2020 Uwe Bonnes <bon@elektron.ikp.physik.tu-darmstadt.de>
```
dragonmux, line 4: Please use the copyright notice from sfdp.c here, with the year updated to 2023 - the ulog2 function is code we contributed in #1464 (this should also serve as a reminder to ourself to fix the stlinkv2.c header).

We'd ask you to update the license terms to the same as sfdp.c but that's entirely optional.

src/compat.c
```diff
@@ -0,0 +1,62 @@
...
+ * built-ins. */
+
+#include "compat.h"
+
+#ifdef __GNUC__
+#define unlikely(x) __builtin_expect(x, 0)
+#else
+#define unlikely(x) x
+#endif
+
```
dragonmux, line 30: The extra blank line here introduces a whitespace error, please remove it

src/compat.c
```diff
@@ -0,0 +1,62 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.	 See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.	 If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* Compatibility functions wrapping compiler-specific code, especially
```
dragonmux, line 20: The code base comment style for multi-line comments is
```c
/*
 * comment
 * lines
 */
```

src/target/cortexm.c
```diff
@@ -1265,18 +1265,14 @@ bool cortexm_run_stub(target_s *t, uint32_t loadaddr, uint32_t r0, uint32_t r1,
  * The Flash Patch and Breakpoint (FPB) and Data Watch and Trace (DWT)
  * systems are used. */
 
+/* DWT only supports powers of two as size. Convert length in bytes to
```
dragonmux, line 1268: Same comment as in compat.c about multi-line comments. Please fix this comment in the same way

src/include/compat.h
```diff
@@ -0,0 +1,25 @@
+/*
+ * This file is part of the Black Magic Debug project.
+ *
```
dragonmux, line 3: Even though the header's very short, please include a copyright notice.

dragonmux (changes requested):
Beyond our conversation about compat.c's naming, we've done a quick look over of the contribution. It's looking good - the main set of changes that need to occur are stylistic with just the two other adjustments needing making as a copyright notice is wrong, and another missing.

With those items fixed, this'll be ready for merge after the help text unification PR

src/compat.c
silbe: Sorry, somehow messed up using `git blame` and traced it back to when `stlink.c` was first introduced. Couldn't find it in the other referenced projects so I incorrectly attributed it to Uwe Bonnes. Fixed now.

src/compat.c
silbe: Fixed. While at it I also fixed the directly preceding comment in `cortexm.c`.

src/compat.c
silbe: Done. Personally I prefer some visual separation between defines and functions but that's a matter of taste.

src/include/compat.h
silbe: Uses the same license header as `maths_utils.c` now.

src/compat.c
dragonmux: No worries :) it happens.

src/compat.c
dragonmux: The code base style does not include this extra space and uses a single blank line per the clang-format configuration - so it's more what we enforce across the whole code base for consistency rather than taste. The project has had a history of people cargo-culting their personal style in which caused a huge number of headaches in the lead-up to v1.8.

dragonmux:
Almost all the items have been addressed fully, we'll mark this ready for merge behind the help syntax PR once the final items are sorted.

src/compat.c
silbe: Oops, didn't notice the leading blank line, only the trailing one. Fixed now. Thanks for your patience!

dragonmux: No worries, thanks for yours!

dragonmux (approved):
LGTM, thanks for your patience and the contribution! We'll merge this once the other PR is in

PR #1478, opened 2023-04-28
Fix selecting between multiple identical CMSIS-DAP devices

zyp:
## Detailed description

Unlike ST-Link and JLink's init functions, CMSIS-DAP's `dap_init_bulk()` uses `libusb_open_device_with_vid_pid()` which just grabs the first matching device without filtering by serial, which makes it impossible to select other devices than the first.

Rather than transplanting the opening/filtering code from one of the other interfaces, this PR leverages the fact that we've already found and decided on which USB device we want to use in `find_debuggers()`, so all we have to do is to keep a reference to the selected `libusb_device` so that the interface drivers can pass that to `libusb_open()` without having to do a new scan.

The natural next step would be to simplify the ST-Link and JLink interface drivers to also make use of this, but I'm not able to test those right now, so I've left those out of the scope for this PR.

### DIFF
--- a/src/platforms/hosted/cmsis_dap.c
+++ b/src/platforms/hosted/cmsis_dap.c
@@ -199,9 +199,9 @@ static bool dap_init_hid(const bmp_info_s *const info)
 static bool dap_init_bulk(const bmp_info_s *const info)
 {
 	DEBUG_INFO("Using bulk transfer\n");
-	usb_handle = libusb_open_device_with_vid_pid(info->libusb_ctx, info->vid, info->pid);
-	if (!usb_handle) {
-		DEBUG_ERROR("libusb_open_device_with_vid_pid() failed\n");
+	int res = libusb_open(info->libusb_dev, &usb_handle);
+	if (res != LIBUSB_SUCCESS) {
+		DEBUG_ERROR("libusb_open() failed (%d): %s\n", res, libusb_error_name(res));
 		return false;
 	}
 	if (libusb_claim_interface(usb_handle, info->interface_num) < 0) {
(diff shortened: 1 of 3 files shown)


src/platforms/hosted/cmsis_dap.c
```diff
@@ -199,9 +199,9 @@ static bool dap_init_hid(const bmp_info_s *const info)
 static bool dap_init_bulk(const bmp_info_s *const info)
 {
 	DEBUG_INFO("Using bulk transfer\n");
-	usb_handle = libusb_open_device_with_vid_pid(info->libusb_ctx, info->vid, info->pid);
-	if (!usb_handle) {
-		DEBUG_ERROR("libusb_open_device_with_vid_pid() failed\n");
+	int res = libusb_open(info->libusb_dev, &usb_handle);
+	if (res != LIBUSB_SUCCESS) {
+		DEBUG_ERROR("libusb_open() failed\n");
```
dragonmux, line 204: To aid in debugging when this happens, could the format string be expanded to include the value of 'res' and the `libusb_error_name()` string for the error to provide the user with more information please

dragonmux (changes requested):
Looks good - the only change requested is an expansion of an error message to help the user in debugging the failure path.

In regards simplifying the ST-Link backend, that would be delightful and we're most happy to test the results as we do have an ST-Link v2.1

Many thanks for the contribution!

dragonmux (commented):
LGTM, thank you! We'll merge this just as soon as the build completes.

dragonmux (approved):

PR #1479, opened 2023-04-29
make: Don't delete version.h in archives.

esden:
## Detailed description

Together with releases we also provide sourcecode archives. The archives contain the `version.h` file but no `.git` as it is not an active git repository. When the code is cleaned in a source archive the version.h should not be deleted. Also we should not require `git` to be installed so we check if it is in the path before running `git desscribe`.

This patch checks if git exists in the PATH and is executable, and makes cleaning of `version.h` optional.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -144,10 +144,18 @@ endif
 
 .PHONY:	clean host_clean all_platforms clang-format FORCE
 
+
+GIT_VERSION := $(shell if [ -d "../.git" ]; then git describe --always --dirty --tags; fi)
+
 clean:	host_clean
 	$(Q)echo "  CLEAN"
 	-$(Q)$(RM) *.o *.d *.elf *~ $(TARGET) $(HOSTFILES)
-	-$(Q)$(RM) platforms/*/*.o platforms/*/*.d mapfile include/version.h
+	-$(Q)$(RM) platforms/*/*.o platforms/*/*.d mapfile
+ifeq ($(GIT_VERSION),)
+	@echo Git not found, not deleting include/version.h
+else
+	-$(Q)$(RM) include/version.h
+endif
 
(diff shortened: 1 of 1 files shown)


src/Makefile
```diff
@@ -144,10 +144,18 @@ endif
 
 .PHONY:	clean host_clean all_platforms clang-format FORCE
 
+
+GIT_VERSION := $(shell if ! [ -x "$(command -v git)" ]; then git describe --always --dirty --tags; fi)
```
dragonmux, line 148: If the error output from `git describe` is redirected via `2>/dev/null` then this should suppress the error line caused in the Make output. Alternatively you could add `-a -d .git` to the if [] clause to check that the .git directory exists as well as `git` being available

dragonmux (changes requested):
Looks good, we've got one small tweak to suggest to improve the Make output, but otherwise this looks ready to merge

src/Makefile
esden: If we check for .git then we don't really have to check for git binary being there do we?

dragonmux: If, like us, one has a clone of BMD on a network accessible share, and tries doing a build from a machine other than the one that cloned the repo, that would result in there being a .git dir but not necessarily `git` being avaiable

esden: Ok, changed the logic to check for the .git directory only. This way it resolves the issue. If we have the .git directory we need to have git installed anyways so a double check is moot in this case.

dragonmux (approved):
Looks good, merging

src/Makefile
dragonmux: Ok, given that will result in a command-not-found when that's not the case, and that's on stderr iwrc, that works. We'll go with that

PR #1481, opened 2023-04-29
Feature: restructure platforms folder

lenvm:
## Detailed description

Previously the folders `stm32` and `tm4c` looked like platforms. One might therefore expect that firmware can build for the with `make PROBE_HOST=tm4c`. Actually these folders contain code common to multiple platforms.

Previously also the linker scripts of several files were stored in the `stm32` folder, which was unintuitive.

This pull request changes the following:
- It changes the folder names `stm32` to `common/stm32` and `tm4c` to `common/tm4c`.
- It moves the linker script of each platforms to the folder of that platform.
- It adds new linker scripts for the platforms that did not have their own linker script.

* [x] It builds for all_platforms (`make all_platforms`)


esden:
Thank you @lenvm for this. I think it is a good improvement for the codebase clarity.

I think it would be better to add another directory level here, and put the architecture common files in `common/stm32` and `common/tm4c` respectively.

lenvm:
@esden Thanks for your feedback! I have updated the pull request accordingly.

@dragonmux can you review this pull request and merge it if it is ok?

src/platforms/launchpad-icdi/Makefile.inc
```diff
@@ -9,11 +9,11 @@ CPU_FLAGS = -mcpu=cortex-m4 -mthumb -mfpu=fpv4-sp-d16 -mfloat-abi=hard
 CFLAGS += $(INCLUDES) $(CPU_FLAGS) -DSERIAL_NO=$(SERIAL_NO) -DTARGET_IS_BLIZZARD_RB1 -DLM4F -DPART_TM4C123GH6PM
 CFLAGS += -DDFU_SERIAL_LENGTH=9
 
-LINKER_SCRIPT="platforms/tm4c/tm4c.ld"
+LINKER_SCRIPT="platforms/common/tm4c/tm4c.ld"
```
dragonmux, line 12: Given the other platforms have had their linker scripts moved into the platform directory, does it not make sense to do that for this one too for consistency?

dragonmux (changes requested):
There's only one consistency issue we spotted, the rest of this looks great. We'll merge one that one item is addressed.

src/platforms/launchpad-icdi/Makefile.inc
lenvm: Thanks for the feedback! I was thinking that for the `tm4c` platform there only seems to be one chip, and therefore only one ram and rom size, but out of consistency considerations I also agree it makes sense to place it in the directory of the respective platform. I have implemented this.

Additionally, for consistency, I have also renamed the linker script of the native platform.

dragonmux (approved):
Good call with the native platform too - happy with this, so merging.

dragonmux (approved):

dragonmux:
As a note for the future, please use `git rebase -i main` rather than a merge commit. Thankfully, GitHub will do that for us in merging this.

PR #1482, opened 2023-04-29
imxrt: use PRIu32 for debug print

xobs:
This fixes the build on ESP32.

### DIFF
--- a/src/target/imxrt.c
+++ b/src/target/imxrt.c
@@ -463,8 +463,8 @@ static void imxrt_spi_wait_complete(target_s *const target)
 #ifdef ENABLE_DEBUG
 		/* Read out the status code and display it */
 		const uint32_t status = target_mem_read32(target, IMXRT_FLEXSPI1_STAT1);
-		DEBUG_TARGET(
-			"Error executing sequence, offset %u, error code %u\n", (status >> 16U) & 0xfU, (status >> 24U) & 0xfU);
+		DEBUG_TARGET("Error executing sequence, offset %" PRIu8 ", error code %" PRIu8 "\n",
+			(uint8_t)(status >> 16U) & 0xfU, (uint8_t)(status >> 24U) & 0xfU);
 #endif
 		/* Now clear the error (this clears the status field bits too) */
 		target_mem_write32(target, IMXRT_FLEXSPI1_INT, IMXRT_FLEXSPI1_INT_CMD_ERR);


src/target/imxrt.c
```diff
@@ -463,8 +463,8 @@ static void imxrt_spi_wait_complete(target_s *const target)
 #ifdef ENABLE_DEBUG
 		/* Read out the status code and display it */
 		const uint32_t status = target_mem_read32(target, IMXRT_FLEXSPI1_STAT1);
-		DEBUG_TARGET(
-			"Error executing sequence, offset %u, error code %u\n", (status >> 16U) & 0xfU, (status >> 24U) & 0xfU);
+		DEBUG_TARGET("Error executing sequence, offset %" PRIu32 ", error code %" PRIu32 "\n", (status >> 16U) & 0xfU,
```
dragonmux, line 466: In the interests of keeping Flash usage down, perhaps it would be better to cast the shifted status values to uint8_t's?

dragonmux (commented):
Patch looks good, we have one query as shown in the review below but once that's taken care of we're happy to merge this.

Thanks for the contribution!

src/target/imxrt.c
xobs: I've made the change. Please let me know if that affects flash at all.

dragonmux: From the CI run: ```
Run head/scripts/diff_size.py base/build.log head/build.log
rom: +0 / 131072 B (+0.00%)
ram: +0 / 20480 B (+0.00%)
```
So that seems perfect

File: src/target/imxrt.c [hunk 2]
```diff
@@ -463,8 +463,8 @@ static void imxrt_spi_wait_complete(target_s *const target)
 #ifdef ENABLE_DEBUG
 		/* Read out the status code and display it */
 		const uint32_t status = target_mem_read32(target, IMXRT_FLEXSPI1_STAT1);
-		DEBUG_TARGET(
-			"Error executing sequence, offset %u, error code %u\n", (status >> 16U) & 0xfU, (status >> 24U) & 0xfU);
+		DEBUG_TARGET("Error executing sequence, offset %" PRIu8 ", error code %" PRIu8 "\n",
+			(unsigned char)((status >> 16U) & 0xfU), (unsigned char)((status >> 24U) & 0xfU));
```
dragonmux, line 467: Please us `uint8_t` here, not the raw C type. This is great otherwise. Technically, you don't need to enclose the expression in brackets because of that final masking operation, but we understand why you've done it this way. `(uint8_t)(status >> 16U) & 0xfU` will execute identically though.

```suggestion
		DEBUG_TARGET("Error executing sequence, offset %u, error code %u\n",
			(uint8_t)(status >> 16U) & 0xfU, (uint8_t)(status >> 24U) & 0xfU);
```

dragonmux (changes requested):
We'd really like to avoid using raw C types (especially in casts) but hopefully the suggestion provided helps show what we were intending

src/target/imxrt.c
xobs: Alright, I've made that change, and removed the extra parenthesis.

dragonmux (approved):
Thank you, LGTM now. We'll merge this as soon as the build finishes

zyp:
This broke the build for me on MacOS:
```
  CC      target/imxrt.c
target/imxrt.c:467:4: error: format specifies type 'unsigned char' but the argument has type 'unsigned int' [-Werror,-Wformat]
                        (uint8_t)(status >> 16U) & 0xfU, (uint8_t)(status >> 24U) & 0xfU);
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
include/general.h:110:40: note: expanded from macro 'DEBUG_TARGET'
#define DEBUG_TARGET(...) debug_target(__VA_ARGS__)
                                       ^~~~~~~~~~~
target/imxrt.c:467:37: error: format specifies type 'unsigned char' but the argument has type 'unsigned int' [-Werror,-Wformat]
                        (uint8_t)(status >> 16U) & 0xfU, (uint8_t)(status >> 24U) & 0xfU);
                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
include/general.h:110:40: note: expanded from macro 'DEBUG_TARGET'
#define DEBUG_TARGET(...) debug_target(__VA_ARGS__)
                                       ^~~~~~~~~~~
2 errors generated.
make: *** [imxrt.o] Error 1
```

Looks like the bitwise AND after the cast might promote the expression back to a larger size. The obvious fix is to do the cast after the AND.

dragonmux:
The fix we'd apply is to use just `"%u"` as varags integer promotion will always re-promote to `unsigned` - we were worried that might happen with using `"%" PRIu8`

xobs:
I applied this fix in #1486

PR #1499, opened 2023-05-27
Feature: SPI remote protocol

dragonmux:
## Detailed description

This PR implements support for having BMD talk SPI using its remote protocol, allowing access to both the on-board Flash of BMP v2.3+ and externally connected Flash on supported probes.

Additionally, this rationalises the existing SPI Flash communications code to help reduce duplication and get back some of the Flash usage of the duplicated code, making adding new targets that use SPI Flash for their firmware storage easier.

The changes for RP2040, LPC43xx and i.MXRT have been briefly tested, and the remote protocol implementation has been thoroughly tested against blackmagic-debug/bmpflash which is the complementary tool to this PR that enables use of this new SPI remote protocol.


esden:
This does not seem to fit onto native any more :(

dragonmux:
You will need to use the ARM-provided GCC 12.2.Rel1 compiler, as the older compiler is unable to optimise quite as well and indeed, fails to make the binary fit the device. We did our best to limit the effects of this problem, but there's only so much we can do.

esden (approved):
LGTM

PR #1506, opened 2023-06-06
Fix: RTT CI for Flash limitations

dragonmux:
## Detailed description

As the BMD project grows, and has been growing, we have started to hit the limitations on what we can fit in the Flash of the various processors supported by the platforms the project can be built for.

This PR limits which platforms we build RTT-enabled firmware for to nagivate this problem while keeping coverage as high as possible.

In the current build system we can only toggle whether features like RTT are enabled or not, but cannot currently pick subsets of targets to include in the build to tailor the firmware and keep things within the bounds of the available Flash.


esden (approved):
LGTM

PR #1509, opened 2023-06-07
Feature: Microsoft OS USB descriptors

dragonmux:
## Detailed description

For the longest time on Windows, users would plug their shiny new BMP in and get up and running with the serial ports just fine, but be left with both the DFU and Trace interfaces showing a yellow warning triangle in Device Manager because Windows didn't know how to properly bind a driver to either interface. This has resulted in many bug reports and confused users thinking their device was misbehaving or non-functional, and caused just as many for users trying to upgrade their firmware with having to get things just right in Zadig.

The project has shipped .inf files in the drivers directory which rectifies this, but users tend to either not know to use and apply them, or be unable to on modern Windows due to changes in the driver model. This PR seeks to address at least part of this problem by shipping Microsoft OS Descriptors in the firmware which Windows will then read out and use to create driver bindings. We accomplish this by packaging up a Big Object Store (BOS) descriptor with the appropriate subcomponents and advertising it suitably during enumeration.

This has been tested working on Windows 10. It **only** solves the problem for the driver binding and does not currently solve for interface naming as the latter is surprisingly more complicated due to how these descriptors work. A further patch is in development for this latter problem, and we will aim to have it ready for the v2.0 merge window.


esden (approved):
LGTM

PR #1512, opened 2023-06-12
Fix: Replace strtol with strtoul

ALTracer:
## Detailed description

* The PR tries to optimise flash size by slightly reducing the set of functions required from libc.
* Precisely, I've identified the three places where `strtol` is called in firmware with help of Puncover. `strtoul` seems to be sufficient to store into uint32_t results in all three cases, and the `base` argument is 16 or 0 (for hexadecimal) anyway.

man 3 strtoul: 
>Negative values are considered valid input and are silently converted to the equivalent unsigned long value.

Even though newlib provides `_strtol_r` & `_strtoul_r` as different implementations (not calling each other, unlike atoi()/atol() wrappers), the resulting program ends up referencing both functions via (si)scanf implementation anyway (newlib & newlib-nano). There is only an effectively 20-40-byte .text reduction from dropping the `strtol` wrapper.
Still offering this PR in the event siscanf (1.5k with deps) from newlib-nano somehow gets replaced with an optimized version.

## Closing issues
None known.
Testing would involve fiddling with option bytes of stm32f1 and issuing a swdp_scan of a given TARGETID.

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -245,7 +245,7 @@ bool cmd_swdp_scan(target_s *t, int argc, const char **argv)
 	(void)t;
 	volatile uint32_t targetid = 0;
 	if (argc > 1)
-		targetid = strtol(argv[1], NULL, 0);
+		targetid = strtoul(argv[1], NULL, 0);
 	if (platform_target_voltage())
 		gdb_outf("Target voltage: %s\n", platform_target_voltage());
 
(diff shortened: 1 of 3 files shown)


dragonmux (approved):
LGTM, thank you for noticing and fixing this consistency as that will be very helpful in the future when we replace printf()/sscanf() usage with a custom optimised solution

PR #1515, opened 2023-06-17
platforms/blackpill-f4: fix inverted tpwr command

lenvm:
This pull request fixes a bug where the command `monitor tpwr disable` would enable tpwr, and `monitor tpwr enable` would disable tpwr.

With this fix, running `monitor tpwr disable` sets the voltage of the tpwr pin to 0.00 V, and running `monitor tpwr enable` sets the voltage of the tpwr pin to 3.30 V.

> **Note**
> A remaining issue is that tpwr is enabled upon startup, whereas `monitor tpwr` reports `Target Power: disabled`. This was already the case before this pull request, and remains an issue. @dragonmux Do you perhaps know what could cause this, and what could be a fix?

* [x] It builds for all platforms (`make all_platforms`)

### DIFF
--- a/src/platforms/common/blackpill-f4/blackpill-f4.c
+++ b/src/platforms/common/blackpill-f4/blackpill-f4.c
@@ -96,7 +96,7 @@ void platform_init(void)
 	gpio_mode_setup(LED_PORT_UART, GPIO_MODE_OUTPUT, GPIO_PUPD_NONE, LED_UART);
 
 #ifdef PLATFORM_HAS_POWER_SWITCH
-	gpio_set(PWR_BR_PORT, PWR_BR_PIN);
+	gpio_clear(PWR_BR_PORT, PWR_BR_PIN); // Set the pin of the given GPIO port to 0.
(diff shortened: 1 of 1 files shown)


dragonmux (approved):
Hi lenvm, your patch looks good and is approved for merge, but we'll hold off if you'd like to address the boot-up behaviour.

Regarding where that's happening, we'd be looking at [the platform startup code](https://github.com/blackmagic-debug/blackmagic/blob/main/src/platforms/common/blackpill-f4/blackpill-f4.c) at how PWR_BR_{PORT,PIN} is being initialised - https://github.com/blackmagic-debug/blackmagic/blob/df014ff488bf869ff09461cb85c2e649b501a16b/src/platforms/common/blackpill-f4/blackpill-f4.c#L99 looks like it should be modified

lenvm:
Hi @dragonmux, thanks for the tip! I've found what was the issue, and fixed it. I also found that the command to get the power was inverted, such that tpwr would state "enabled" if it was actually disabled, and vice versa.

dragonmux (approved):
LGTM, merging!

PR #1524, opened 2023-06-26
Invalid sector_num fixed for dfu erase stm32f4

Electronshik:
## Detailed description

I think I found bug in Black Magic DFU bootloader for stm32f4 platform, where I'm testing BMP firmware.
I've encountered with DFU not erasing flash memory, where after manually erased chip, dfu bootloader successfully loaded BMP firmware. The commit where erase flow was changed  SHA:
3647a627bf1f2d42560a007085e7c06d1a82cb28
02_bmp

After the changes I've made I was able to download BMP firmware with DFU bootloader successfully.

* [* ] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md)
* [*] I've read the [guidelines for contributing](https://github.com/blackmagic-debug/blackmagic/blob/main/CONTRIBUTING.md) to this repository
* [* ] It builds for hardware native (`make PROBE_HOST=native`)
* [* ] It builds as BMDA (`make PROBE_HOST=hosted`)
* [* ] I've tested it to the best of my ability
* [* ] My commit messages provide a useful short description of what the commits do

### DIFF
--- a/src/platforms/common/stm32/dfu_f4.c
+++ b/src/platforms/common/stm32/dfu_f4.c
@@ -60,6 +60,7 @@ static void get_sector_num(uint32_t addr)
 
 	if (!sector_addr[i])
 		return;
+	--i;
 	sector_num = i & 0x1fU;
 }
 


ALTracer:
Can you clarify which hardware do you work with? There are known PROBE_HOST platforms in BMD repo, such as 32F4Discovery with F407 core, Hydrabus with F405/415, and WeAct MiniF4 with F401/411 cores. "stm32f4" does not match a single known platform. Is this a custom board your team developed separately?

F429 may be compatible with F407 code as it used 8/168MHz clocks.

I have problems with BMP DFU on blackpill-f411ce, I did try to preemptively mass-erase, it did not help. Your findings may be useful since you claim to have tested the DFU code on some board.

src/platforms/common/stm32/dfu_f4.c
```diff
@@ -52,10 +52,11 @@ static void get_sector_num(uint32_t addr)
 {
 	if (addr < sector_addr[0])
 		return;
-	size_t i;
-	for (i = 1; sector_addr[i]; ++i) {
-		if (addr < sector_addr[i])
+	size_t i = 0;
+	while (sector_addr[i + 1]) {
```
dragonmux, line 56: The for-loop transformation which we applied here is the correct thing to do - even if apparently we got either the loop condition, or more likely the post-state invariant, wrong. This is a for loop in while loop clothing.

Given what is changing here, it suggests that what was missing is really `--i;` after the for loop (line 60) so the if statement on line 62 then has the `i` value it's expecting.

dragonmux (changes requested):
Thank you for the contribution and letting us know about this bug, Electronshik!

There are a couple of things we need you to change to make this merge-able: Please follow the contribution guidelines and alter your commit message to match - it is presently missing its suffix (`common/stm32/dfu_f4:`) and could do with being a bit more descriptive; We would appreciate some refinement to the patch - see the review note that follows.

With those items fixed, we would be happy to merge this PR.

ALTracer: The reason they haven't specified which platform specifically this applies to is that it's a bug in *all* STM32F4 based platforms, and specifically the bootloader. dfu_f4.c is the common processor-specific DFU code for them all.

Electronshik:
> Can you clarify which hardware do you work with? There are known PROBE_HOST platforms in BMD repo, such as 32F4Discovery with F407 core, Hydrabus with F405/415, and WeAct MiniF4 with F401/411 cores. "stm32f4" does not match a single known platform. Is this a custom board your team developed separately?
> 
> F429 may be compatible with F407 code as it used 8/168MHz clocks.
> 
> I have problems with BMP DFU on blackpill-f411ce, I did try to preemptively mass-erase, it did not help. Your findings may be useful since you claim to have tested the DFU code on some board.

Hi! I'm using custom draft board based on stm32f411CEU6 (similar to Blue Pill board) with custom BMP target. But I think this chunk of code applies for all stm32f4 platforms. I did exactly the same with DFU util, tried mass-erase before downloading with no success:
`dfu-util.exe -d 1D50:DF11,:6017 -s 0x08004000:force:mass-erase:leave -D firmware.bin`

I hope this fix may help you with your board. My command for DFU is:
`dfu-util.exe -d 1D50:DF11,:6017 -s 0x08004000:leave -D firmware.bin`

src/platforms/common/stm32/dfu_f4.c
Electronshik: Thank you for your reply! I'edit the PR in accordance with the guidelines and test it more thoroughly (seems that `--i;` will be enough, I'll check it). Thanks for the great BMP project!

ALTracer:
I very well understand this code applies for all F4 chips running BMPBootloader DFU.
Maybe you're right, a simple `i--` makes this testcase resolve right sector addresses.
https://gist.github.com/ALTracer/31ebfa8e8225af0dab9046ef62f51616

src/platforms/common/stm32/dfu_f4.c
Electronshik: I've tested decrement `--i;` and, as you said, it seems sufficient to find correct sector number. I've placed it right after `if (!sector_addr[i])` check so that sector number 12 would never returned from the function. As I readed from docs, sector 12 exists in stm32f4 with more than 1Mb flash memory and for that memory sectors 12-23 needed so I thought it is not necessary for now. I've attached the chunk of code where I checked the output of the function depending from input addresses so anyone can check it too. And I debugged that chunk with my hardware (blackmagic debugging blackmagic :)), DFU downloads were stable.

If I've done something wrong please correct me. it's a pleasure to be involved in the black magic project!

https://onlinegdb.com/g83JE4cmG8

BMP-01
BMP-02

dragonmux (approved):
LGTM! The logic of that function makes a lot more sense now.

It might be worth adding a comment to the front of the loop such as `/* Search for the end of the sector containing the address */`, a comment to the if check such as `/* Fail if the address is past the end */` and a comment on what the `i` logic at the end does.

However, these all firmly fall under the category of enhancements. We'll get this PR merged as it fixes a regression.

PR #1525, opened 2023-06-27
Feature: Implement SPI support for blackpill-f4

ALTracer:
## Detailed description
This PR implements support for the internal (on-board) SPI Flash found on the Blackpill series boards, mirroring the work done for "native" (BMP).
* A new feature was implemented in #1499 for accessing probe-attached SPI flash chips via BMP remote protocol (and desktop cross-platform bmpflash client).
* However, it only implements SPI platform code for `native` hardware, providing no-op stubs for all the other platforms.
* I propose implementing platform_spi_init/deinit/chip_select/xfer for `blackpill-f4` hardware, which I own and test on.

This change results in ~~+320~~ +448 bytes flash footprint for `blackpill-f4x1cx`. Testing included `sfdp` and `read` from onboard w25q64 to desktop dualboot host (Linux Mint 21.1, Windows 11) using latest manually compiled bmpflash/substrate/fmt/libusb-1.0.
Read speed of 8MiB flash chip: Linux 200KiB/s, Win11 175KiB/s. File seems to match remains of NuttX SmartFS I kept there from previous development tests.

Misc. notes:
* At APB2=96MHz and SPI1 DIV8, line clock frequency should be 12MHz which is reasonable (`native`/F1 uses APB1=36;/8=4.5MHz for EXT SPI1 and APB2=72;/8=9Mhz for AUX SPI2).
* w25q64 datasheet states the chip supports accesses at clock frequency, when powered from 3.3v, in Standard SPI mode, `03h Read Data` up to 50MHz and `0Bh Fast Read` up to 104MHz.
* ~~I might add SPI2 on PB12/13/14/15 (F4 APB1=48Mhz) later as external/aux/etc. bus, these pins are not claimed yet.~~ Added SPI2 with same divisor, 48/8=6MHz.
* Other WeAct Studios designed hardware (BluePill Plus) also has SOIC-8 footprints on the bottom of PCB, so this PR might be useful in case somebody implements SPI support for that hardware later. It even maps that to the same GPIOs and HW SPI.

* [x] It builds for hardware native (`make PROBE_HOST=native`) -- changes another platform
* [x] It builds as BMDA (`make PROBE_HOST=hosted`) -- not applicable

## Closing issues
No issues raised on SPI functionality by users of blackpill-f4 boards, because it was absent from platform code.

### DIFF
--- a/src/platforms/common/blackpill-f4/blackpill-f4.c
+++ b/src/platforms/common/blackpill-f4/blackpill-f4.c
@@ -37,6 +37,7 @@
 #include <libopencm3/cm3/systick.h>
 #include <libopencm3/cm3/cortex.h>
 #include <libopencm3/usb/dwc/otg_fs.h>
+#include <libopencm3/stm32/spi.h>
 
 jmp_buf fatal_error_jmpbuf;
 volatile uint32_t magic[2] __attribute__((section(".noinit")));
@@ -174,26 +175,69 @@ void platform_target_clk_output_enable(bool enable)
 
 bool platform_spi_init(const spi_bus_e bus)
 {
-	(void)bus;
-	return false;
+	if (bus != SPI_BUS_INTERNAL)
+		return false;
+
+	/* Set up onboard flash SPI GPIOs: PA5/6/7 as SPI1 in AF5, PA4 as nCS output push-pull */
+	const uint32_t gpioport = OB_SPI_PORT;
+	const uint16_t gpios = (OB_SPI_SCLK | OB_SPI_MISO | OB_SPI_MOSI);
+	gpio_mode_setup(gpioport, GPIO_MODE_AF, GPIO_PUPD_NONE, gpios);
+	gpio_mode_setup(gpioport, GPIO_MODE_OUTPUT, GPIO_PUPD_NONE, OB_SPI_CS);
+	gpio_set_output_options(gpioport, GPIO_OTYPE_PP, GPIO_OSPEED_50MHZ, gpios | OB_SPI_CS);
+	gpio_set_af(gpioport, GPIO_AF5, gpios);
+	/* Deselect the onboard flash chip */
+	gpio_set(gpioport, OB_SPI_CS);
+
+	rcc_periph_clock_enable(RCC_SPI1);
+	rcc_periph_reset_pulse(RST_SPI1);
+
+	const uint32_t controller = OB_SPI;
+	/* Set up hardware SPI: master, PCLK/8, Mode 0, 8-bit MSB first */
+	spi_init_master(controller, SPI_CR1_BAUDRATE_FPCLK_DIV_8, SPI_CR1_CPOL_CLK_TO_0_WHEN_IDLE,
+		SPI_CR1_CPHA_CLK_TRANSITION_1, SPI_CR1_DFF_8BIT, SPI_CR1_MSBFIRST);
+	spi_enable(controller);
+	return true;
 }
 
 bool platform_spi_deinit(const spi_bus_e bus)
 {
-	(void)bus;
-	return false;
+	if (bus != SPI_BUS_INTERNAL)
+		return false;
+	const uint32_t controller = OB_SPI;
+	spi_disable(controller);
+	/* Gate SPI1 APB clock */
+	rcc_periph_clock_disable(RCC_SPI1);
+	/* Unmap GPIOs */
+	const uint32_t gpioport = OB_SPI_PORT;
+	const uint16_t gpios = (OB_SPI_SCLK | OB_SPI_MISO | OB_SPI_MOSI);
+	gpio_mode_setup(gpioport, GPIO_MODE_INPUT, GPIO_PUPD_NONE, gpios | OB_SPI_CS);
+	return true;
 }
 
 bool platform_spi_chip_select(const uint8_t device_select)
 {
-	(void)device_select;
-	return false;
+	const uint8_t device = device_select & 0x7fU;
+	const bool select = !(device_select & 0x80U);
+	uint32_t port;
+	uint16_t pin;
+	switch (device) {
+	case SPI_DEVICE_INT_FLASH:
+		port = OB_SPI_CS_PORT;
+		pin = OB_SPI_CS;
+		break;
+	default:
+		return false;
+	}
+	gpio_set_val(port, pin, select);
+	return true;
 }
 
 uint8_t platform_spi_xfer(const spi_bus_e bus, const uint8_t value)
 {
-	(void)bus;
-	return value;
+	if (bus != SPI_BUS_INTERNAL)
+		return false;
+	const uint32_t controller = OB_SPI;
+	return spi_xfer(controller, value);
 }
 
 int platform_hwversion(void)
--- a/src/platforms/common/blackpill-f4/blackpill-f4.c
+++ b/src/platforms/common/blackpill-f4/blackpill-f4.c
@@ -175,39 +175,35 @@ void platform_target_clk_output_enable(bool enable)
 
 bool platform_spi_init(const spi_bus_e bus)
 {
-	uint32_t gpioport = 0;
-	uint16_t gpio_hwspi = 0;
-	uint16_t gpio_cs = 0;
 	uint32_t controller = 0;
 	if (bus == SPI_BUS_INTERNAL) {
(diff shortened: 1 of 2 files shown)


dragonmux (changes requested):
This is looking good, but there are a few stylistic things to fix and a few queries about how things are being done we'd like answered before we can merge. Nice work though!

As a side note in the future: Your PR description goes to great lengths to talk about speed, testing, etc, but at no point does it actually describe as a single easily read paragraph what this PR implements and does. In the future, please start the description with something like:
```md
This PR implements support for the internal (on-board) SPI Flash found on the Blackpill series boards, mirroring the work done for "native" (BMP).
```

If you open with something like that, it'll help provide context before you drop into technical details that are more a curiosity than providing us with information needed to review the PR itself.

src/platforms/common/blackpill-f4/blackpill-f4.c
```diff
@@ -173,24 +174,67 @@ void platform_target_clk_output_enable(bool enable)
...
 bool platform_spi_init(const spi_bus_e bus)
 {
-	(void)bus;
-	return false;
+	if (bus != SPI_BUS_INTERNAL)
+		return false;
+
+	/* Set up onboard flash SPI GPIOs: PA5/6/7 as SPI1 in AF5, PA4 as nCS output push-pull */
+	const uint32_t gpioport = OB_SPI_PORT;
+	const uint16_t gpios = (OB_SPI_SCLK | OB_SPI_MISO | OB_SPI_MOSI);
```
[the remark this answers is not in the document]
ALTracer: Agreed, in this case parens play no syntactic role (no higher logic expression).

src/platforms/common/blackpill-f4/blackpill-f4.c
```diff
@@ -173,24 +174,67 @@ void platform_target_clk_output_enable(bool enable)
...
+	return true;
 }
 
 bool platform_spi_deinit(const spi_bus_e bus)
 {
-	(void)bus;
-	return false;
+	if (bus != SPI_BUS_INTERNAL)
+		return false;
+	const uint32_t controller = OB_SPI;
```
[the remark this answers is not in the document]
ALTracer: Substituted the controller values.

src/platforms/common/blackpill-f4/blackpill-f4.c
```diff
@@ -173,24 +174,67 @@ void platform_target_clk_output_enable(bool enable)
...
+	spi_disable(controller);
+	/* Gate SPI1 APB clock */
+	rcc_periph_clock_disable(RCC_SPI1);
+	/* Unmap GPIOs */
+	const uint32_t gpioport = OB_SPI_PORT;
+	const uint16_t gpios = (OB_SPI_SCLK | OB_SPI_MISO | OB_SPI_MOSI);
```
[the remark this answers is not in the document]
ALTracer: Inlined the gpioport, removed parens of gpios. On inlining the four pins clang-format breaks the line, do you consider this necessary in two-bus case?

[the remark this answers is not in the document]
dragonmux: As the two bus case probably involves two entirely separate controllers, we'd expect the gpio_mode_set() call to be duplicated, branched on an if statement querying which bus we're reconfiguring. Given that means entirely separate sets of pins, it makes no sense to us to set variables to get one call - we're not sure the compiler won't a) optimise this for us or b) generate more code this way

Plus still, having the chip select in the call and not the rest.. it's a very strange split - we don't see it serving any purpose.

ALTracer:
Applied suggested changes from first review pass, awaiting additional review comments or approval; then I will {fast-forward fork's `main`, rebase this PR, squash three commits into one}. Please don't merge this as-is until then, because a) fixups present; b) honestly SPI1 & SPI2 is a single feature.

dragonmux:
We'll aim to review the new code today then. Please **do not** squash the commits - as they are right now is better for `git bisect`, and `git blame` and allowing us to better follow the diff and peel one of them off if it ever needs reverting.

We very explicitly want 3 commits here because it's not about "by feature", it's about small atomic changes that are easily reviewed, etc.. the commits as they are, are a bit big anyway. Ideally we'd want them to be even more granular.

src/platforms/common/blackpill-f4/blackpill-f4.c
```diff
@@ -173,24 +174,98 @@ void platform_target_clk_output_enable(bool enable)
...
+
+	if (bus == SPI_BUS_INTERNAL) {
+		rcc_periph_clock_enable(RCC_SPI1);
+		rcc_periph_reset_pulse(RST_SPI1);
+		controller = OB_SPI;
+	} else if (bus == SPI_BUS_EXTERNAL) {
+		rcc_periph_clock_enable(RCC_SPI2);
+		rcc_periph_reset_pulse(RST_SPI2);
+		controller = EXT_SPI;
+	}
```
dragonmux, line 207: Rather than this being a separate if block, especially with where the `controller` variable is defined, it would make more sense for this to be integrated with the above block with a line of whitespace between the GPIO setup and the RCC pokes. That would make this flow better.

src/platforms/common/blackpill-f4/blackpill-f4.c
```diff
@@ -173,24 +174,98 @@ void platform_target_clk_output_enable(bool enable)
...
+	spi_enable(controller);
+	return true;
 }
 
 bool platform_spi_deinit(const spi_bus_e bus)
 {
-	(void)bus;
-	return false;
+	if (bus == SPI_BUS_INTERNAL) {
+		/* Notice: spi_clean_disable() hangs */
```
dragonmux, line 219: Note that `spi_clean_disable()` is designed for when there could be stuff waiting in the data register, which is why it hangs, as we already ensure that not to be the case. Please remove this comment as we don't want to lead future readers to that function as it's explicitly the wrong tool here.

dragonmux (commented):
There are a couple of small things that need tweaking, and once tweaked and keeping this as 3 separate commits (please), we're happy to merge this.

src/platforms/common/blackpill-f4/blackpill-f4.c
ALTracer: After modifying gpio setup to conform to code style like in native, I agree, moving these inside works better.

src/platforms/common/blackpill-f4/blackpill-f4.c
ALTracer: This is remains of me trying to use libopencm3 APIs. Maybe I was mislead by their documentation, but it actually hangs the probe with no timeouts, I didn't expect it to. Will remove the commit

dragonmux: They don't actually mention, to be fair, that it will hang if there's no data in the DR to read, but.. it'll be hanging on the RXNE check:

https://github.com/blackmagic-debug/libopencm3/blob/0c6c7c2895c1c2b81c157242a754f99964de828d/lib/stm32/common/spi_common_all.c#L104-L120

ALTracer:
For edits in this review would you like a new commit or amending?
I usually squash-rebase the (fixup) commits after reviews, but you always have the final say as the reviewer and maintainer.

dragonmux:
Amends please, both changes are quite small and clean up the diffs

src/platforms/common/blackpill-f4/blackpill-f4.c
ALTracer: Propagated RCC upwards, keeping the `controller` variable. I could duplicate the SPI init code, too, setting different divisors (because SPI1 & SPI2 are on APB2 & APB1 and one is clocked twice as fast than the other), but for that I think a better scheme is needed which allows e.g. instructing BMP from `bmpflash` on divisors in a `monitor frequency` fashion.

ALTracer:
Run-tested (`sfdp` & `read`) both before and after fast-forward rebase. Works for `int` bus, raises a proper error for `ext` bus without me connecting them with 4 duponts). Ready for merge to upstream.

Afterthoughts: there's no LED indication of SPI activity as blackpill has a single LED. BMF doesn't touch IDLE_RUN during remote-spi, and I don't touch PC13 in platform spi code. For now rely on `bmpflash` reporting progress or general PC performance monitors.

dragonmux (approved):
LGTM, merging.

PR #1526, opened 2023-06-29
Fix: Refactor LPC43xx target sram detection code

ALTracer:
## Detailed description

* No new features implemented.
* The pull request is aimed at flash size/footprint reduction.
* I propose to change lpc43x0_detect and lpc43xx_detect logic to use local stack variables and fewer calls, which leads to measurable firmware .text shrink.

Disclaimer: I don't have NXP LPC43xx targets and hence cannot test the change on them. But this approach is already present in other targets code and improves readability by removing duplication/syntactic noise. +12b stack usage (if spilled) should not be a problem.

### DIFF
--- a/src/target/lpc43xx.c
+++ b/src/target/lpc43xx.c
@@ -414,39 +414,48 @@ static void lpc43x0_add_spi_flash(target_s *const target, const size_t length)
 
 static void lpc43x0_detect(target_s *const t, const lpc43xx_partid_s part_id)
 {
+	uint32_t sram1_size;
+	uint32_t sram2_size;
+	uint32_t sram_ahb_size;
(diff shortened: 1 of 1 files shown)


dragonmux (approved):
A quick review says this looks good. We'll test it on our small trove of LPC43xx parts locally in the next day or so and get this merged once we have.

ALTracer:
hmm, sizediff shows only -36 bytes?
First commit is -52 bytes (which is a useful measurement of what to expect from dropping 9 calls with 2-3 arguments).
Second commit is +16 bytes because it's complicated. GCC becomes clever at -Os. I decided not to complicate the code with /1024 and *1024 operations further just to fit the _SRAM_SIZE macros into uint16_t as GCC seems capable of doing this itself.

dragonmux (approved):
LGTM, merging

PR #1527, opened 2023-06-29
Fix: target: Deduplicate message strings

ALTracer:
## Detailed description

* No new feature implemented.
* This pull request is aimed at flash size/footprint reduction.
* Repeated parts of message strings in stm32l4 & samx5x could be output via `%s` of tc_printf/gdb_voutf/vasprintf, and this allows GCC to fold string literal constants, saving 200 bytes in `.rodata`.

Disclaimer: I don't have targets handled by stm32l4 code available (and certainly not samx5x), so I can't runtime-test whether the probe now crashes at this stack depth etc.
Additionally, I didn't find a `tc_puts()` or something similar to print simple `\n\0`-terminated string literals. GCC is known to optimise such calls of `printf("...\n")` to `puts("...")` (see `gimple_fold_builtin_printf()`). Some of the changes could be alternatively split into two print statements, the second printing the common/folded message part.

### DIFF
--- a/src/target/samx5x.c
+++ b/src/target/samx5x.c
@@ -648,7 +648,7 @@ static bool samx5x_cmd_lock_flash(target_s *t, int argc, const char **argv)
 		tc_printf(t, "Error writing NVM page\n");
 		return false;
 	}
-	tc_printf(t, "Flash locked. The target must be reset for this to take effect.\n");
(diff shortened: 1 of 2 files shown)


dragonmux (approved):
This looks good to us, and we would not expect any stack issues here regardless. Merging

PR #1529, opened 2023-07-02
Fix: TraceSWO Async improvements

ALTracer:
## Detailed description

* This is kinda a new feature -- this provides traceswoasync for STM32F4 and similar (newer than F1 chip-based) platforms. 
* This PR has two changes and tries to solve two problems. 
* Problem F4. STM32F4 platforms had only timer-based RZ/Manchester traceswo available.
* Solution F4. The PR extends logic in traceswoasync with F4-specific macros to properly support newer DMA IP (and maybe GPIO) like aux_serial.
* Problem F1. The only two known users of traceswoasync are stlink and swlink platforms (stlinkv3 has a forked traceswoasync_f723), and they become very RAM-constrained after permanently assigning 8k of 20k to `trace_rx_buf[]`. (4k static .data/.bss, 8k trace_rx_buf, ~-4k stack, and only like 4k left for newlib-nano heap)
* Solution F1. The first commit in PR adds a `memalign()` call and makes the trace_rx_buf a dynamic buffer allocated on heap only when requested.

* [x] It builds for hardware native (`make PROBE_HOST=native`) -- no change as async not supported yet
* [x] It builds as BMDA (`make PROBE_HOST=hosted`) -- no change to hosted

## Closing issues

should fix issues triggered by heap exhaustion on stlink & swlink platforms. Can't find exact issue numbers.

## Miscellaneous notes

The new `traceswo_deinit()` is there to a) allow switching between RZ & NRZ in the future (after refactoring both); b) remind developers to possibly free the buffer. This does NOT shrink the heap on newlib-nano, because malloc_trim is unimplemented, but MAY on newlib after adjusting mallopt or manually calling malloc_trim. At the moment `nano.specs` is used by native, stlink, swlink, f072, f3, stlinkv3.

The alloc and blackpil-f4 changesets may be split into a second PR, but I needed a way to test the unified traceswoasync unit on the two platforms I own (bluepill & blackpill-f4) against regressions. I did not test anything of this yet besides building `all_platforms`. I tried `posix_memalign`, it does not exist in newlib.

### DIFF
--- a/src/platforms/common/blackpill-f4/blackpill-f4.h
+++ b/src/platforms/common/blackpill-f4/blackpill-f4.h
@@ -27,7 +27,9 @@
 #include "timing.h"
 #include "timing_stm32.h"
 
-#define PLATFORM_HAS_TRACESWO
+#define PLATFORM_HAS_TRACESWO 1
+#define NUM_TRACE_PACKETS     512U /* 32K buffer */
+//#define TRACESWO_PROTOCOL     2U   /* 1 = RZ/Manchester, 2 = NRZ/async/uart */
 
 /* Error handling for ALTERNATIVE_PINOUT
  * If ALTERNATIVE_PINOUT has a value >= 4 (undefined), or <= 0, an error is thrown.
@@ -202,11 +204,27 @@
 #define IRQ_PRI_USBUSART     (2U << 4U)
 #define IRQ_PRI_USBUSART_DMA (2U << 4U)
 #define IRQ_PRI_TRACE        (0U << 4U)
+#define IRQ_PRI_SWO_DMA      (0U << 4U)
 
 #define TRACE_TIM          TIM3
 #define TRACE_TIM_CLK_EN() rcc_periph_clock_enable(RCC_TIM3)
 #define TRACE_IRQ          NVIC_TIM3_IRQ
 #define TRACE_ISR(x)       tim3_isr(x)
+
+/* On F411 use USART1 remapped to PB7 for async capture */
+#define SWO_UART        USBUSART1
+#define SWO_UART_CLK    USBUSART1_CLK
+#define SWO_UART_DR     USBUSART1_DR
+#define SWO_UART_PORT   GPIOB
+#define SWO_UART_RX_PIN GPIO7
+
+/* Bind to the same DMA Rx channel */
+#define SWO_DMA_BUS    USBUSART1_DMA_BUS
+#define SWO_DMA_CLK    USBUSART1_DMA_CLK
+#define SWO_DMA_CHAN   USBUSART1_DMA_RX_CHAN
+#define SWO_DMA_IRQ    USBUSART1_DMA_RX_IRQ
+#define SWO_DMA_ISR(x) USBUSART1_DMA_RX_ISRx(x)
+#define SWO_DMA_TRG    DMA_SxCR_CHSEL_4
 
 #define SET_RUN_STATE(state)      \
 	{                             \
--- a/src/platforms/common/blackpill-f4/blackpill-f4.h
+++ b/src/platforms/common/blackpill-f4/blackpill-f4.h
@@ -29,7 +29,7 @@
 
 #define PLATFORM_HAS_TRACESWO 1
 #define NUM_TRACE_PACKETS     512U /* 32K buffer */
-//#define TRACESWO_PROTOCOL     2U   /* 1 = RZ/Manchester, 2 = NRZ/async/uart */
+#define TRACESWO_PROTOCOL     2U   /* 1 = RZ/Manchester, 2 = NRZ/async/uart */
 
 /* Error handling for ALTERNATIVE_PINOUT
  * If ALTERNATIVE_PINOUT has a value >= 4 (undefined), or <= 0, an error is thrown.
--- a/src/platforms/common/stm32/traceswoasync.c
+++ b/src/platforms/common/stm32/traceswoasync.c
@@ -42,6 +42,20 @@
 #include <libopencm3/stm32/usart.h>
 #include <libopencm3/stm32/dma.h>
 
+#if defined(DMA_STREAM0)
+#define dma_channel_reset(dma, channel)   dma_stream_reset(dma, channel)
+#define dma_enable_channel(dma, channel)  dma_enable_stream(dma, channel)
+#define dma_disable_channel(dma, channel) dma_disable_stream(dma, channel)
+
+#define DMA_PSIZE_8BIT DMA_SxCR_PSIZE_8BIT
+#define DMA_MSIZE_8BIT DMA_SxCR_MSIZE_8BIT
+#define DMA_PL_HIGH    DMA_SxCR_PL_HIGH
+#else
+#define DMA_PSIZE_8BIT DMA_CCR_PSIZE_8BIT
+#define DMA_MSIZE_8BIT DMA_CCR_MSIZE_8BIT
+#define DMA_PL_HIGH    DMA_CCR_PL_HIGH
+#endif
+
 #define TRACESWOASYNC_ALLOCS
 //#define TRACESWOASYNC_DEALLOCS
 
@@ -96,11 +110,18 @@ void traceswo_setspeed(uint32_t baudrate)
 	/* Set up DMA channel */
 	dma_channel_reset(SWO_DMA_BUS, SWO_DMA_CHAN);
 	dma_set_peripheral_address(SWO_DMA_BUS, SWO_DMA_CHAN, (uint32_t)&SWO_UART_DR);
+#if defined(DMA_STREAM0)
+	dma_set_transfer_mode(SWO_DMA_BUS, SWO_DMA_CHAN, DMA_SxCR_DIR_PERIPHERAL_TO_MEM);
+	dma_channel_select(SWO_DMA_BUS, SWO_DMA_CHAN, SWO_DMA_TRG);
+	dma_set_dma_flow_control(SWO_DMA_BUS, SWO_DMA_CHAN);
+	dma_enable_direct_mode(SWO_DMA_BUS, SWO_DMA_CHAN);
+#else
 	dma_set_read_from_peripheral(SWO_DMA_BUS, SWO_DMA_CHAN);
+#endif
 	dma_enable_memory_increment_mode(SWO_DMA_BUS, SWO_DMA_CHAN);
-	dma_set_peripheral_size(SWO_DMA_BUS, SWO_DMA_CHAN, DMA_CCR_PSIZE_8BIT);
-	dma_set_memory_size(SWO_DMA_BUS, SWO_DMA_CHAN, DMA_CCR_MSIZE_8BIT);
-	dma_set_priority(SWO_DMA_BUS, SWO_DMA_CHAN, DMA_CCR_PL_HIGH);
+	dma_set_peripheral_size(SWO_DMA_BUS, SWO_DMA_CHAN, DMA_PSIZE_8BIT);
+	dma_set_memory_size(SWO_DMA_BUS, SWO_DMA_CHAN, DMA_MSIZE_8BIT);
+	dma_set_priority(SWO_DMA_BUS, SWO_DMA_CHAN, DMA_PL_HIGH);
 	dma_enable_transfer_complete_interrupt(SWO_DMA_BUS, SWO_DMA_CHAN);
 	dma_enable_half_transfer_interrupt(SWO_DMA_BUS, SWO_DMA_CHAN);
 	dma_enable_circular_mode(SWO_DMA_BUS, SWO_DMA_CHAN);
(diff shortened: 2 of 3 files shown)


dragonmux:
Ah, apologies for not having reviewed this yet. We'll do so today and see if this is something that can still land in v1.10.0 or if it needs bumping to v2.0 and be one of the early merge items.

ALTracer:
Rebased from v1.9.0 era to current `main`. What is the plan w.r.t. `memalign` changes -- drop/move? And F4 UART/DMA changes -- keep (to consider merging after PR1827 makes way)? Or the opposite, keep memalign for F103, and extract F4 Async support into a fresh PR? I have only tested the combination of `swo_alloc`+`endpoints`+`over8` on `blackpill-f411ce`.

src/platforms/common/blackpill-f4/blackpill-f4.h
```diff
@@ -27,7 +27,9 @@
 #include "timing.h"
 #include "timing_stm32.h"
 
-#define PLATFORM_HAS_TRACESWO
+#define PLATFORM_HAS_TRACESWO 1
```
dragonmux, line 30: This macro is only ever used with a `#ifdef`/`#if defined()` check, never for its absolute definition value. Just defining it without value is perfectly reasonable enough.

src/platforms/common/blackpill-f4/blackpill-f4.h
```diff
@@ -272,12 +274,28 @@ extern bool debug_bmp;
 #define IRQ_PRI_USBUSART     (2U << 4U)
 #define IRQ_PRI_USBUSART_DMA (2U << 4U)
 #define IRQ_PRI_TRACE        (0U << 4U)
```
dragonmux, line 276: While this becomes unused in this switch-over, it's a fair note that it remains useful when we get both modes compiling into the firmware together.

src/platforms/common/stm32/traceswoasync.c
```diff
@@ -29,6 +29,8 @@
 /* TDO/TRACESWO signal comes into the SWOUSART RX pin. */
 
 #include <stdatomic.h>
+#include <malloc.h>
+#include <errno.h>
```
dragonmux, line 33: These in particular should be included after `general.h` so bad things don't happen with the feature macros. Ideally the `stdatomic.h` include should really also be moved after.

src/platforms/common/stm32/traceswoasync.c
```diff
@@ -41,10 +43,31 @@
...
+#define DMA_PSIZE_8BIT DMA_SxCR_PSIZE_8BIT
+#define DMA_MSIZE_8BIT DMA_SxCR_MSIZE_8BIT
+#define DMA_PL_HIGH    DMA_SxCR_PL_HIGH
+#else
+#define DMA_PSIZE_8BIT DMA_CCR_PSIZE_8BIT
+#define DMA_MSIZE_8BIT DMA_CCR_MSIZE_8BIT
+#define DMA_PL_HIGH    DMA_CCR_PL_HIGH
+#endif
+
+#define TRACESWOASYNC_ALLOCS
```
dragonmux, line 60: This should be determined by the platform header, not in this file. Having this alternative mode is certainly quite interesting, but it should not be the default for all platforms we think. We're also of the opinion that if init allocates, deinit must deallocate in this paradigm, otherwise that's a permanent memory usage ever after turning Trace/SWO capture back off, eliminating the benefit, particularly if the probe is then used to scan a large JTAG or SWD chain after.

(Perhaps this should be enabled on F1 based platforms, but not the F4's? for example)

src/platforms/common/stm32/traceswoasync.c
```diff
@@ -106,35 +136,90 @@ void traceswo_setspeed(uint32_t baudrate)
...
+	}
+	if (dma_get_interrupt_flag(SWO_DMA_BUS, SWO_DMA_CHAN, DMA_TCIF)) {
+		dma_clear_interrupt_flags(SWO_DMA_BUS, SWO_DMA_CHAN, DMA_TCIF);
+		memcpy(
+			&trace_rx_buf[write_index * TRACE_ENDPOINT_SIZE], &pingpong_buf[TRACE_ENDPOINT_SIZE], TRACE_ENDPOINT_SIZE);
+	}
+	write_index = (write_index + 1U) % NUM_TRACE_PACKETS;
+	trace_buf_drain(usbdev, TRACE_ENDPOINT | USB_REQ_TYPE_IN);
+}
+#endif
```
dragonmux, line 169: We can't see any difference between this and the IRQ handler above - if there is no difference or it's very minimal, we suggest merging these together with much more targeted macro usage.

src/platforms/common/stm32/traceswoasync.c
```diff
@@ -106,35 +136,90 @@ void traceswo_setspeed(uint32_t baudrate)
...
 
 void traceswo_init(uint32_t baudrate, uint32_t swo_chan_bitmask)
 {
+#ifdef TRACESWOASYNC_ALLOCS
+	if (trace_rx_buf == NULL) {
+		uint8_t *newbuf = memalign(TRACE_ENDPOINT_SIZE, NUM_TRACE_PACKETS * TRACE_ENDPOINT_SIZE);
```
dragonmux, line 175: `memalign()` is not a C standard library function (it's not defined by the standard, that is), and we're not sure what the impact will be on newlib under a very constrained memory environment. How important is it that the allocation be aligned to the trace endpoint size? We don't recall seeing any source buffer alignment requirements, so unless there are any, we strongly suggest using `malloc()` instead here.

src/platforms/common/stm32/traceswoasync.c
```diff
@@ -106,35 +136,90 @@ void traceswo_setspeed(uint32_t baudrate)
...
+#endif
+
 	nvic_set_priority(SWO_DMA_IRQ, IRQ_PRI_SWO_DMA);
 	nvic_enable_irq(SWO_DMA_IRQ);
 	traceswo_setspeed(baudrate);
 	traceswo_setmask(swo_chan_bitmask);
 	decoding = (swo_chan_bitmask != 0);
 }
+
+void traceswo_deinit(void)
```
dragonmux, line 207: What, if anything, currently calls this? If nothing, please include usage in this PR, such as a command in command.c that turns Trace/SWO capture off with this.

src/platforms/common/stm32/traceswoasync.c
```diff
@@ -106,35 +136,90 @@ void traceswo_setspeed(uint32_t baudrate)
...
+{
+	/* Stop peripherals servicing */
+	nvic_disable_irq(SWO_DMA_IRQ);
+	dma_disable_channel(SWO_DMA_BUS, SWO_DMA_CHAN);
+	usart_disable(SWO_UART);
+	/* Dump the buffered remains */
+	trace_buf_drain(usbdev, TRACE_ENDPOINT | USB_REQ_TYPE_IN);
+#ifdef TRACESWOASYNC_DEALLOCS
+	/* Release this chunk of precious SRAM. FIXME: does heap shrink after free/malloc_trim? */
```
dragonmux, line 216: Heap never shrinks under newlib-nano, but it's also not helpful or useful to either worry about or force - when the allocation is `free()`'d back to the allocation pool, it will automatically be re-used by future allocations, either whole or in parts. -nano's allocator cannot un-`sbrk()` memory and will aggressively re-use chunks in the free list in preference for `sbrk()`'ing new memory, as long as it can find enough contiguous spare in the list.

dragonmux (changes requested):
Sorry that reviewing this has taken so long to get to. Hopefully the items we've spotted are able to be addressed fairly quickly as it would be good to get this merged (there's a lot of great stuff in here) and clear the way for some of your other Trace/SWO PRs.

src/platforms/common/blackpill-f4/blackpill-f4.h
ALTracer: `IRQ_PRI_TRACE` will be used once PR1833 moves meaningfully, and I'd rename it to `IRQ_PRI_TRACETIM` or something to make it carry meaning related to Timer-based Manchester capture.

src/platforms/common/stm32/traceswoasync.c
ALTracer: Could move.

src/platforms/common/stm32/traceswoasync.c
ALTracer: What happened there is probably I copied the block and started replacing the pieces to tailor F4, but forgot to keep the old F1 one intact. The two impls look similar enough to me, so if the functions are static inline helpers, like macros, then I could drop one. Otherwise there would be a lot of jumping around in IRQ context, and STM32F1 Prefetch doesn't like that (GD32F1 will happily hop in 0WS)

src/platforms/common/stm32/traceswoasync.c
ALTracer: This buffer used to have static storage duration (and linkage), but despite it being uint8_t[] BFD linker guarantees some alignment, likely 32 bit (check this). Now that it's allocated dynamically, either malloc or memalign should ensure 32-bit alignment for 1) DMA to work properly. Although USART Rx streams are configured with `DMA_MSIZE_8BIT`. 2) This will be memcpy'ed in IRQ context to USB buffers, either 16-bit PMA or OTG FIFO, so I want 32-bit alignment for an efficient memcpy4 (`__aeabi_memcpy4`?) to pump data faster than a byte-wise memcpy1. Alignment for **64 bytes** is not supported by anything, but I wanted to see if it's honored at all by newlib.

src/platforms/common/stm32/traceswoasync.c
ALTracer: Oops, nothing calls it. `command.c` should, but back then I ran out of determination to hook it into argv parser, which is already untrivial with all the parameters. The idea was to deinit on qRcmd upon stop or mode switch. I'll need some help on UI/naming. `monitor traceswo stop`? halt/deinit/reset/free?

src/platforms/common/stm32/traceswoasync.c
ALTracer: Yes, by now I know newlib-nano `free` does no trimming and there's no `malloc_trim` on everything F1 in-tree. I wanted to make space for possible (descending) stack growth, as some Big things in BMF are allocated on stack. This is less relevant in light of PR1830. I see how sbrk is forward-only, and it will eventually reach end-of-SRAM (not MSP stack), and after a traceswo_deinit, any subsequent `malloc` may pick this space.
The code might be useful where newlib is available (F4), and in out-of-tree projects (with RTOS and complex stuff). If this were NuttX, I might've asked to allocate in DMA-able memory (e.g. not DTCM).
I can remove the comments, but I don't think the code itself needs to go. Are there nano-specific feature macros I could test this against? `mallopt` has to go before this `free`, might go in initialisation fwiw. `malloc_trim` has to be hidden.

src/platforms/common/stm32/traceswoasync.c
dragonmux: Given it's `mon traceswo` to enable it right now.. and that's not clear that it will actually enable decoding.. perhaps `enable` for the enable side, and `disable` for the deinit/stop? This would be consistent with other parts of BMD too then, including target power and such.

ALTracer: Added extra command logic. I attempted to sidestep untangling the existing logic. Enabling in-probe ITM decode was optional and required the extra verbs (`mon traceswo 2250000 decode 2 3 30` or just `mon traceswo decode 2 3 30` for default baudrate). I think a real switch-case (or sequential ifs) parser is needed here for subcommands, especially for picking timer/uart modes.

dragonmux: Agreed, which is something we have to tackle to get both back-ends compiling and working in the same build anyway - a long-term goal, but not something this PR needs to concern itself with.

dragonmux:
This is looking much better. With the outstanding couple of review items taken care of, and a rebase on main, this is looking fairly close to ready to merge.

src/platforms/common/stm32/traceswoasync.c
ALTracer: Dropped memalign.
I have two suggestions, 1) `return false` to command.c (requires changing prototypes) on allocation failure, else user sees "Trace enabled" which is a lie. 2) Attempt to allocate for a smaller buffer size (down to 2x64 bytes) and optionally report this via `gdb_voutf()`. Which of those are necessary -- will only be visible in testing.

src/platforms/common/stm32/traceswoasync.c
ALTracer: Dropped code incompatible with nano, reworded comment.

src/platforms/common/stm32/traceswoasync.c
dragonmux: Definitely thing that, if you're happy to do a prototype change, having the commands able to report failures has the potential to be extremely useful and good for reducing what they have to do - "failed? just return false". Quite happy for that to be done.

ALTracer:
One year later, give or take, to attempt to break the stalled state, I decided to split out the F4-related changes. Rationale: statistically I've noticed that smaller PRs take less time to review and test; and e.g. ctxLink may need a small subset of changes from here, best disentangled from here.

dragonmux (approved):
All review items addressed - with thanks. LGTM, merging. Thank you for the contribution and improving the state of the UART (Async) Trace/SWO implementation so it can be more broadly applicable!

PR #1530, opened 2023-07-02
Fix: Decouple headers general.h/platform.h

ALTracer:
## Detailed description

* This is not a feature.
* Incremental builds after touching `platform.h` on my laptop take >10s (`user` time over ~3 of 4 Tremont N6000 CPU cores) because of `target/*.c` rebuilds. Not using ccache.
* The PR solves this problem by breaking the include-dependency of `target/` drivers on `platform.h` header and by propagating that into its actual users, with inclusion order `general.h` then `platform.h` as before. There also are a couple related/required/revealed codebase cleanup changes.

* [x] It builds for hardware native (`make PROBE_HOST=native`) -- fixed encountered breakage
* [x] It builds as BMDA (`make PROBE_HOST=hosted`) -- fixed encountered breakage

## Closing issues

No issues were reported on incremental build times by foreign platform developers yet, I'm the only one complaining.

## Miscellaneous notes

This reduces incremental build times from 10s for 64 translation units to 3s for 16 units. CI full builds are not affected.
This PR is aimed specifically at the old/current Makefile build system, but future Kconfig or meson systems may benefit from this as well. I can't say whether other major PRs will need rebasing.

### DIFF
--- a/src/platforms/96b_carbon/platform.c
+++ b/src/platforms/96b_carbon/platform.c
@@ -36,6 +36,11 @@
 
 jmp_buf fatal_error_jmpbuf;
 
+int platform_hwversion(void)
+{
+	return 0;
+}
+
 void platform_init(void)
 {
 	rcc_clock_setup_pll(&rcc_hse_16mhz_3v3[RCC_CLOCK_3V3_84MHZ]);
--- a/src/platforms/96b_carbon/platform.c
+++ b/src/platforms/96b_carbon/platform.c
@@ -21,6 +21,7 @@
 /* This file implements the platform specific functions for the 96Boards Carbon implementation. */
 
 #include "general.h"
+#include "platform.h"
 #include "usb.h"
 #include "aux_serial.h"
 #include "morse.h"
--- a/src/platforms/common/blackpill-f4/blackpill-f4.c
+++ b/src/platforms/common/blackpill-f4/blackpill-f4.c
@@ -194,3 +194,8 @@ uint8_t platform_spi_xfer(const spi_bus_e bus, const uint8_t value)
 	(void)bus;
 	return value;
 }
+
+int platform_hwversion(void)
+{
+	return 0;
+}
--- a/src/platforms/common/blackpill-f4/blackpill-f4.c
+++ b/src/platforms/common/blackpill-f4/blackpill-f4.c
@@ -21,6 +21,7 @@
 /* This file implements the platform specific functions for the blackpill-f4 implementation. */
 
 #include "general.h"
+#include "platform.h"
 #include "usb.h"
 #include "aux_serial.h"
 #include "morse.h"
--- a/src/platforms/f072/platform.c
+++ b/src/platforms/f072/platform.c
@@ -121,6 +121,11 @@ void platform_request_boot(void)
 
 #pragma GCC diagnostic pop
 
+int platform_hwversion(void)
+{
+	return 0;
+}
+
 void platform_target_clk_output_enable(bool enable)
 {
 	(void)enable;
--- a/src/platforms/f072/platform.c
+++ b/src/platforms/f072/platform.c
@@ -20,6 +20,7 @@
 /* This file implements the platform specific functions for the STM32F072-IF implementation. */
 
 #include "general.h"
+#include "platform.h"
 #include "usb.h"
 #include "aux_serial.h"
 #include "morse.h"
--- a/src/platforms/f3/platform.c
+++ b/src/platforms/f3/platform.c
@@ -34,6 +34,11 @@
 
 extern uint32_t _ebss; // NOLINT(bugprone-reserved-identifier,cert-dcl37-c,cert-dcl51-cpp)
 
+int platform_hwversion(void)
+{
+	return 0;
+}
+
 void platform_init(void)
 {
 	volatile uint32_t *magic = &_ebss;
--- a/src/platforms/f3/platform.c
+++ b/src/platforms/f3/platform.c
@@ -20,6 +20,7 @@
 /* This file implements the platform specific functions for the STM32F3-IF implementation. */
 
 #include "general.h"
+#include "platform.h"
 #include "usb.h"
 #include "aux_serial.h"
 #include "morse.h"
--- a/src/platforms/f4discovery/platform.c
+++ b/src/platforms/f4discovery/platform.c
@@ -21,6 +21,7 @@
 /* This file implements the platform specific functions for the F4 Discovery implementation. */
 
 #include "general.h"
+#include "platform.h"
 #include "usb.h"
 #include "aux_serial.h"
 #include "morse.h"
--- a/src/platforms/f4discovery/platform.c
+++ b/src/platforms/f4discovery/platform.c
@@ -39,6 +39,11 @@
 jmp_buf fatal_error_jmpbuf;
 extern uint32_t _ebss; // NOLINT(bugprone-reserved-identifier,cert-dcl37-c,cert-dcl51-cpp)
 
+int platform_hwversion(void)
+{
+	return 0;
+}
+
 void platform_init(void)
 {
 	volatile uint32_t *magic = &_ebss;
--- a/src/platforms/hydrabus/platform.c
+++ b/src/platforms/hydrabus/platform.c
@@ -21,6 +21,7 @@
 /* This file implements the platform specific functions for the Hydrabus implementation. */
 
 #include "general.h"
+#include "platform.h"
 #include "usb.h"
 #include "aux_serial.h"
 #include "morse.h"
--- a/src/platforms/hydrabus/platform.c
+++ b/src/platforms/hydrabus/platform.c
@@ -36,6 +36,11 @@
 
 jmp_buf fatal_error_jmpbuf;
 
+int platform_hwversion(void)
+{
+	return 0;
+}
+
 void platform_init(void)
 {
 	/* Check the USER button */
--- a/src/platforms/launchpad-icdi/platform.c
+++ b/src/platforms/launchpad-icdi/platform.c
@@ -48,6 +48,11 @@ uint32_t platform_time_ms(void)
 	return time_ms;
 }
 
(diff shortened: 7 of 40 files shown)


ALTracer:
Additionally, when building with `/usr/bin/time -v colormake CROSS_COMPILE="ccache arm-none-eabi-" all_platforms -j4`, 
I see 1:30 elapsed time and 800/800 hits/misses from `ccache --show-stats`. 
Was 2:00 elapsed time and 8/1600 hits/misses. Interesting. This is probably due to 3 blackpill variants having similar enough headers and CFLAGS, and maybe native/stlink/swlink which are pretty similar etc. I did not wipe the ccache in between runs. After wiping it's 1:40 and 200/1535. Maybe my testing methodology is flawed for this.

Consequential rebuilds of both branches hit everything `arm-none-eabi-gcc` did from ccache and take ~10s only because of `hosted`.

src/platforms/96b_carbon/platform.c
```diff
@@ -36,6 +37,11 @@
 
 jmp_buf fatal_error_jmpbuf;
 
+int platform_hwversion(void)
```
dragonmux, line 40: Is there a particular reason for removing this from the platform header? This will impair the optimiser's ability to optimise this function out to just a value in all call sites

src/platforms/common/blackpill-f4/blackpill-f4.c
```diff
@@ -194,3 +195,8 @@ uint8_t platform_spi_xfer(const spi_bus_e bus, const uint8_t value)
 	(void)bus;
 	return value;
 }
+
+int platform_hwversion(void)
```
dragonmux, line 199: Is there a particular reason for removing this from the platform header? This will impair the optimiser's ability to optimise this function out to just a value in all call sites

src/platforms/f072/platform.c
```diff
@@ -121,6 +122,11 @@ void platform_request_boot(void)
 
 #pragma GCC diagnostic pop
 
+int platform_hwversion(void)
```
dragonmux, line 125: Is there a particular reason for removing this from the platform header? This will impair the optimiser's ability to optimise this function out to just a value in all call sites

src/platforms/f3/platform.c
```diff
@@ -34,6 +35,11 @@
 
 extern uint32_t _ebss; // NOLINT(bugprone-reserved-identifier,cert-dcl37-c,cert-dcl51-cpp)
 
+int platform_hwversion(void)
```
dragonmux, line 38: Is there a particular reason for removing this from the platform header? This will impair the optimiser's ability to optimise this function out to just a value in all call sites

src/platforms/f4discovery/platform.c
```diff
@@ -39,6 +40,11 @@
 jmp_buf fatal_error_jmpbuf;
 extern uint32_t _ebss; // NOLINT(bugprone-reserved-identifier,cert-dcl37-c,cert-dcl51-cpp)
 
+int platform_hwversion(void)
```
dragonmux, line 43: Is there a particular reason for removing this from the platform header? This will impair the optimiser's ability to optimise this function out to just a value in all call sites

src/platforms/hydrabus/platform.c
```diff
@@ -36,6 +37,11 @@
 
 jmp_buf fatal_error_jmpbuf;
 
+int platform_hwversion(void)
```
dragonmux, line 40: Is there a particular reason for removing this from the platform header? This will impair the optimiser's ability to optimise this function out to just a value in all call sites

src/platforms/launchpad-icdi/platform.c
```diff
@@ -48,6 +49,11 @@ uint32_t platform_time_ms(void)
 	return time_ms;
 }
 
+int platform_hwversion(void)
```
dragonmux, line 52: Is there a particular reason for removing this from the platform header? This will impair the optimiser's ability to optimise this function out to just a value in all call sites

dragonmux (commented):
This looks pretty good. There are a few questions that come from moving `platform_hwversion()` calls around that need answering before we can merge this, but we can see the logic applied for things like the timing header changes, so with those questions answered we will then merge this.

src/platforms/common/blackpill-f4/blackpill-f4.c
ALTracer: For `blackpill-f4` "all call sites" boil down to `cmd_version()` in src/command.c and `jtagtap_reset()`. First one prints board revision to gdb, second manipulates jTRST pin if available (and incurs a 10000 loops delay anyway). Flash size is not affected by this diff hunk.

src/platforms/f072/platform.c
ALTracer: For `f072` the only call site is `cmd_version()` which is cold. This diff hunk is +12 `.text` bytes for that platform, and nothing is spilled on the stack.
`(F072-IF)` does not have a `TRST_PIN` defined, but if added, then same reasoning as for `blackpill-f4` applies.

src/platforms/common/blackpill-f4/blackpill-f4.c
dragonmux: We aren't asking for speed reasons, we're asking for reasons of code size and generated code complexity and the ability for the optimiser to potentially just entirely optimise things away as a result.

src/platforms/96b_carbon/platform.c
ALTracer: Yes, the mismatching declarations for `platform_hwversion()` lead to a compiler warning.
```
platforms/96b_carbon/platform.h:148:19: error: static declaration of 'platform_hwversion' follows non-static declaration
  148 | static inline int platform_hwversion(void)
      |                   ^~~~~~~~~~~~~~~~~~
In file included from include/general.h:44,
                 from command.c:27:
include/platform_support.h:51:5: note: previous declaration of 'platform_hwversion' with type 'int(void)'
   51 | int platform_hwversion(void);
      |     ^~~~~~~~~~~~~~~~~~
```
Otherwise see the blackpill-f4 review response.

src/platforms/common/blackpill-f4/blackpill-f4.c
ALTracer: Code size is not affected significantly here. Complexity is reduced in my opinion as now I can clearly see where the function is actually used for a given platform by checking the callers in Puncover. LTO is the ultimate optimizer, but we're not going there yet.

dragonmux: You can find all call sites without doing this change by using an IDE like VSCode and right-clicking the function definition and "Find All References", meaning leaking symbols into the output binary is never a reason to do this. The increase in complexity comes from now having a split function definition vs declaration - ie, we'd like to avoid doing this unless there's a compelling reason to do so.

Heck, for this, if we were in C++, we'd be asking you to mark the header definition `constexpr` even

ALTracer:
On the uninlining: I do not suggest inlining the 180-byte `platform_hwversion()` into the 23 callsites on `native` platform with 16-byte (Thumb) hot path and unconditional stack spillage, as that is impossible would be a disaster. There could be a small getter extracted from it which simply returns the cached (on platform_init) value, and that is inlinable.
Only `native` uses this so extensively. `swlink` has two callers for it, `stlink` one (no `TRST_PIN`).

dragonmux:
We aren't suggesting inlining `native`'s platform_hwversion(), we are just very reticent to remove the inlining on a single `return 0;` version of the function because it shuts the door to so many potential optimisations.

You've provided a good reason in your justification of the 96b_carbon platform one by way of an existing declaration of the function existing in the platform support header, leading to UB and an inconsistent definition. That's your technical justification.

dragonmux (approved):
LGTM now that all the queries are resolved. Merging

src/platforms/common/blackpill-f4/blackpill-f4.c
ALTracer: Agreed on `constexpr` qualifier for returning 0 for the 7 changed platforms, maybe even `consteval`.
It won't fly for `stlink` and `swlink` because they have a variable (set on `detect_rev()`) and have to to an indirect load to return that.
I am specifically not touching `native` as that's a whole another web of version checks and runtime-conditional behaviour.

dragonmux: Right, but for `constexpr` to work (no need for consteval, that serves a different purpose) the functions would need to be put back in the headers. Hence why our wanting to be sure there was a good reason for moving to the implementation TUs. There is: a conflicting existing declaration for the function.

ALTracer:
Thank you for merging this and improving developer experience for foreign platform contributors.
Two remarks on this so far:
1. I explicitly mentioned in the commit message two weeks ago that "There is a declaration conflict from platform_support.h", maybe I should've been more explicit by saying "-Werror triggered".
2. There were 23 commits merged to `main` since this was posted, and I was not asked to rebase locally, so now I don't get to see the size diffs for all platforms or other build-pr CI output (without re-running it on my desktop runner). Because I hide "platform.h" visibility from most everything that #included general.h, there may be side-effects such as linker pulling giant stdio functions with `float/double` support on platforms without `--nano.specs`. This will be resolved separately, because re-introducing integer stdio requires a different approach to keep the build system benefits.

dragonmux:
Regarding the first point, per the PR template: "Please provide enough information so that others can review your pull request.
Information embedded in the description part of the commits doesn't count."

We don't get to see that there is information embedded in the details of commit messages, The whole point of the description block in the template is to include that kind of information as part of telling us what the PR does/fixes/etc. This goes un-noted in the PR description above.

Nothing that went to main since this PR was opened has impacted the size differences this will have caused. If you really want to figure those stats out, then you can take main from just before this merged, look at the CI for it, note the numbers; then repeat on main now, note the numbers; compare. The size information is present in the CI output for main both with and without `ENABLE_RTT=1`.

PR #1537, opened 2023-07-12
Fix: Factor out integer-only stdio into a common header

ALTracer:
## Detailed description

* The previous PR #1530 reduced visibility of "newlib integer-only stdio" redefinitions to only those translation units which `#include platform.h` This lead to flash size increase for 7 platforms (all not using newlib-nano) by 37 KiB because of linker pulling float/double capable stdio implementations (and softfp routines for double etc).
* This PR ties these redefinitions back to `platform_support.h` visible via `general.h` to every platform except `hosted`.

## Testing
To reproduce/verify, `make all_platforms` and compare binary sizes of files under `src/artifacts/v1.9.0-...` for relevant revisions.
Also Puncover to check for functions pulled from libc. 
*I don't know yet how to easily automatically perform checks for ENABLE_DEBUG=1 builds without encountering build failures or invoking make per-platform manually, but these binaries are Big anyways.*

## Miscellaneous
If there is a better header to place this in than `platform_support.h`, then I'll edit this branch.
I could also edit the copyright string, author and filename for "stdio_newlib.h" as I'm only reorganizing the header contents and not actually writing implementation code in this PR. The touched `platform.h` headers contain at least three authors in "Written by" strings.

### DIFF
--- a/src/include/platform_support.h
+++ b/src/include/platform_support.h
@@ -24,6 +24,9 @@
 #error "Include 'general.h' instead"
 #endif
 
+#if PC_HOSTED == 0
+#include "stdio_newlib.h"
+#endif
 #include "target.h"
 #include "spi_types.h"
 
--- a/src/include/stdio_newlib.h
+++ b/src/include/stdio_newlib.h
@@ -1,0 +1,46 @@
+/*
+ * This file is part of the Black Magic Debug project.
+ *
+ * Copyright (C) 2023 1BitSquared <info@1bitsquared.com>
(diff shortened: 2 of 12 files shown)


dragonmux (approved):
LGTM. We were contemplating if "stdio_newlib.h" is necessarily the best name for this new header - perhaps "newlib_stdio_defs.h" or something might be better.. but as this fixes both a huge amount of duplication and a bug, we're going to get this merged so `main` gets unbroken.

PR #1544, opened 2023-07-19
Fix: ADIv5 DP init

perigoso:
## Detailed description

This PR does a much needed cleanup in `adiv5_dp_init`, related to the `TRY_CATCH` blocks

`TRY_CATCH` brings clobbering errors, forcing us to volatile certain variables and thus blocking certain optimizations

This contains the problematic code and unblocks optimizations, bringing a small code size saving

Tested with BMDA (J-Link  V8) and native on a RP2040 target

### DIFF
--- a/src/target/adiv5.c
+++ b/src/target/adiv5.c
@@ -773,6 +773,19 @@ static void adiv5_dp_clear_sticky_errors(adiv5_debug_port_s *dp)
 		adiv5_dp_error(dp);
 }
 
+/* Keep the TRY_CATCH funkiness contained to avoid clobbering and reduce the need for volatiles */
+static uint32_t adiv5_dp_read_dpidr(adiv5_debug_port_s *const dp)
+{
+	volatile uint32_t dpidr = 0;
+	volatile exception_s e;
+	TRY_CATCH (e, EXCEPTION_ALL) {
+		dpidr = adiv5_dp_read(dp, ADIV5_DP_DPIDR);
+	}
+	if (e.type)
+		return 0;
+	return dpidr;
+}
+
 void adiv5_dp_init(adiv5_debug_port_s *const dp, const uint32_t idcode)
 {
 	/*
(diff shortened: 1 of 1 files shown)


src/target/adiv5.c
```diff
@@ -822,16 +831,22 @@ void adiv5_dp_init(adiv5_debug_port_s *const dp, const uint32_t idcode)
...
 
-		/* Check for a valid DPIDR / designer */
-		if (dp->designer_code != 0) {
+		/* 
+		 * Check DPIDR validity
+		 * Designer code 0 is not a valid JEP-106 code
+		 * Version 0 is reserved for DPv0 which does not implement DPIDR
+		 * Bit 0 of DPIDR is read as 1
+		 */
+		if (dp->designer_code != 0 && dp->version > 0 && dpidr & 1U) {
```
dragonmux, line 843: The masking of `dipid` here needs parens to get the right order of operations:
```suggestion
		if (dp->designer_code != 0 && dp->version > 0 && (dpidr & 1U)) {
```

Without this, it's equivalent to having written:
```c
		if ((dp->designer_code != 0 && dp->version > 0 && dpidr) & 1U) {
```

dragonmux (changes requested):
LGTM, there's only one item we've spotted in review and with that fixed this looks good to merge. Really great job!

src/target/adiv5.c
perigoso: Actually, the Bitwise AND should take precedence over Logical AND, at least according to the [cheat sheet](https://en.cppreference.com/w/c/language/operator_precedence)

Do let me know if I'm reading this wrong or using the wrong resource tho.

dragonmux: Ah, you're right it seems. This is the problem with & - if you're doing, eg, `(dpidr & 1U) != 0` then the parens are absolutely required.. the confusing precedence order is as a result a lint on many compilers.

Looks like technically what you wrote is correct then, so we'll put the request this way: For clarity so this doesn't become a constant query, how about we include the parens in this case so the reader doesn't make the same mistake we did, even though they're not strictly required? We suspect we'll forget this order of operations clarification 6mo down the line, so making it more explicit can only be a good thing in context.

perigoso: Sounds perfectly reasonable.

dragonmux (approved):
LGTM, merging.

ALTracer (commented):
Testing this as part of rebasing PR1546 yields a possible regression for me with STM32MP15.

src/target/adiv5.c
```diff
@@ -795,22 +808,18 @@ void adiv5_dp_init(adiv5_debug_port_s *const dp, const uint32_t idcode)
...
-			dpidr = adiv5_dp_read(dp, ADIV5_DP_DPIDR);
-	}
-	if (e.type) {
-		DEBUG_ERROR("DP not responding!...\n");
-		free(dp);
-		return;
-	}
+	if (idcode != JTAG_IDCODE_ARM_DPv0) {
+		const uint32_t dpidr = adiv5_dp_read_dpidr(dp);
+		if (!dpidr) {
```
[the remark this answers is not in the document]
dragonmux: As it turned out, it's not quite so much a regression as much as this PR asserts the correct behaviour here and this was always just a touch broken

PR #1546, opened 2023-07-19
Feature: Identify STM32MP15x_CM4 coprocessor

ALTracer:
## Detailed description

* This is a new feature.
* BMD can scan and attach to the Cortex-M4 coprocessor of STM32MP157 platform, but there's no memory map.
* This PR solves that by implementing MP15x-specific detection and registering a SRAM map.

STM32MP157, MP153, MP151 is a MPU, microprocessor SoC with dual Cortex-A7 cores and one Cortex-M4 microcontroller/coprocessor core all connected to a lot of STM32-standard peripherals over quite a few multilayer buses. It has a proper SWJ-DP exposing JTAG and SWD capable pins. These are used for custom platform bring-up (CA7 secure bootloader chain) and for runtime-debugging coprocessor firmware.

The `stm32mp15_probe()` I implement is sufficient to work with coprocessor firmware. I also pulled a couple functions from H7/H5 target driver code to read out silicon revision and electronic signature unique ID.

MP15x has a complicated SWJ Access Port setup which interferes with debugging the CA7 cores by BMD. The recommended solution by ST is to use upstream OpenOCD or their fork, and that accesses a few registers over all three APs.
MP13x doesn't have a coprocessor and is outside this PR's scope.
MP2 is not released to General Availability yet.

Engineering Boot exposes the CM4 immediately (spinning in a loop in RETRAM), but Production Boot starts up the SoC with CM4 hidden. From my experience it appears later when OpenSTLinux boots, initializes remoteproc framework and manipulates DBGMCU_CR.

We did not significantly alter TrustZone-A settings in our software platform, but take note it can manifest in problems, too. Same on sleep states -- not using any, so no problems triggered by C-Sleep/C-Stop.


src/target/target_probe.h
```diff
@@ -48,6 +48,7 @@ bool stm32f1_probe(target_s *target);
 bool stm32f4_probe(target_s *target);
 bool stm32h5_probe(target_s *target);
 bool stm32h7_probe(target_s *target);
+bool stm32mp15_probe(target_s *target);
```
dragonmux, line 51: This change needs mirroring in target_probe.c

src/target/stm32mp15.c
```diff
@@ -0,0 +1,154 @@
...
+ */
+
+#include "general.h"
+#include "target.h"
+#include "target_internal.h"
+#include "cortexm.h"
+
+/* Access from processor address space.
+ * Access via the debug APB is at 0xe0081000 over AP1. */
+#define DBGMCU_IDCODE      0x50081000U
```
dragonmux, line 37: Same comment for all the DBG* macros here as in #1547 - pleas prefix with the MCU series name like you did on the next macro down.

Additionally, as you're using this as a base address, it should be `STM32MP15_DBGMCU_BASE` and the macro you've named _IDC, should be `_IDCODE`.

src/target/stm32mp15.c
```diff
@@ -0,0 +1,154 @@
...
+#define STM32MP15_UID_BASE 0x5c005234U
+
+#define DBGMCU_IDC (DBGMCU_IDCODE + 0U)
+#define DBGMCU_CR  (DBGMCU_IDCODE + 4U)
```
dragonmux, line 41: `_CR` -> `_CTRL`? We assume that's what this register is, which is why the "CR" nomenclature is bad as we could be wrong.

src/target/stm32mp15.c
```diff
@@ -0,0 +1,154 @@
...
+#define DBGSLEEP   (1U << 0U)
+#define DBGSTOP    (1U << 1U)
+#define DBGSTBY    (1U << 2U)
+
+#define ID_STM32MP15x        0x5000U /* RM0436 */
```
dragonmux, line 46: Please prefix this with a comment stating what § and pg this comes from in the RM like we did in the STM32H5 support:
https://github.com/blackmagic-debug/blackmagic/blob/640e853d2478c72a09514b4513da39e1beec2b46/src/target/stm32h5.c#L108-L109

src/target/stm32mp15.c
```diff
@@ -0,0 +1,154 @@
...
+
+const command_s stm32mp15_cmd_list[] = {
+	{"uid", stm32mp15_uid, "Print unique device ID"},
+	{"revision", stm32mp15_cmd_rev, "Returns the Device ID and Revision"},
+	{NULL, NULL, NULL},
+};
+
+static void stm32mp15_detach(target_s *target)
+{
+	stm32mp15_priv_s *ps = (stm32mp15_priv_s *)target->target_storage;
```
dragonmux, line 64: `ps` -> `priv` please - trying to avoid more single/two letter variable names.

src/target/stm32mp15.c
```diff
@@ -0,0 +1,154 @@
...
+	priv_storage->dbg_cr = target_mem_read32(target, DBGMCU_CR);
+	target->target_storage = priv_storage;
+
+	/* Build the RAM map */
+	/* Figure 4. Memory map RM0436, pg158 */
+	target_add_ram(target, 0x00000000, 0x10000); /* RETRAM,     64kiB */
+	target_add_ram(target, 0x10000000, 0x20000); /* AHB SRAM1, 128kiB */
+	target_add_ram(target, 0x10020000, 0x20000); /* AHB SRAM2, 128kiB */
+	target_add_ram(target, 0x10040000, 0x10000); /* AHB SRAM3,  64kiB */
+	target_add_ram(target, 0x10050000, 0x10000); /* AHB SRAM4,  64kiB */
```
dragonmux, line 90: These constants should all be suffixed with `U` to make them unsigned.

src/target/stm32mp15.c
```diff
@@ -0,0 +1,154 @@
...
+};
+
+static void stm32mp15_detach(target_s *target)
+{
+	stm32mp15_priv_s *ps = (stm32mp15_priv_s *)target->target_storage;
+	target_mem_write32(target, DBGMCU_CR, ps->dbg_cr);
+	cortexm_detach(target);
+}
+
+bool stm32mp15_probe(target_s *target)
```
dragonmux, line 69: Given this is for the Cortex-M core, and there are Cortex-A cores in this part that will need their own, separate, probe routine, we suggest including that information in the probe routine's name - `stm32mp15_cm4_probe`. That way there's room left for a `stm32mp15_ca7_probe` or similar without having to come back and rename this.

dragonmux (changes requested):
This is looking good. There are a few items of refinement we'd like to see and a couple of mistakes need fixing, but this is excellent work.

dragonmux:
Additional review note: Per the contribution guidelines, the `target/` prefix on the target reference in the commit message is not required, `stm32mp15:` instead of `target/stm32mp15:` is sufficient as a prefix

src/target/stm32mp15.c
ALTracer: Prefixed as suggested, for the `STM32MP15_DBGMCU_BASE` and `STM32MP15_UID_BASE`
Not touching the _IDC and _CR (yet) because that's what these registers were always called across most of STM32 lines of MCUs, and the layout seems to match across stm32** BMP target drivers.

src/target/stm32mp15.c
ALTracer: `_CTRL` was a suffix which Arterytek decided to use for their FLASH controller (#1547), so unrelated. STM32 always used `_CR` for "Control Registers" in most of peripherals' documentation, and would not like to be the one to break the convention. `DBGMCU_CR` allows inhibiting CM4 sleep states and is not part of CoreSight ROM table.

src/target/stm32mp15.c
ALTracer: Done for both of the observed ID values, referring to `TARGETID==IDCODE` match and `CM4ROM_PIDR4` errata.

src/target/stm32mp15.c
ALTracer: Applied the item. 
Is it okay to declare so many (contigious) regions? SRAM1-2-3 could be merged, but SRAM4 is by default reserved for CA7 OpenSTLinux DMA chaining and should not be clobbered by CM4. It's up to Developer SDK user to reclaim this back from CA7 to CM4, which is what I did for our system. SRAM3 is often used for OpenAMP remote syscall buffers.

src/target/stm32mp15.c
ALTracer: Renamed as suggested. Though the CA7 is only accessible in Engineering Boot for boot chain debugging (BootROM traces, TF-A BL2 or U-Boot SPL, TF-A SP-MIN or OP-TEE) during platform bring-up, and Linux kernel is often debugged by its KDB/KGDB stub over UART.
This item was the first thing I did revisiting this PR, but apparently I lost the first commit in this chain when moving patches around. Re-done.

src/target/target_probe.h
ALTracer: Mirrored the renamed probe in `_WEAK_NOP`'s.

ALTracer:
> Additional review note: Per the contribution guidelines, the `target/` prefix on the target reference in the commit message is not required, `stm32mp15:` instead of `target/stm32mp15:` is sufficient as a prefix

The first commit introduces a new target driver, so I thought to write the full prefix for context. Following commits omit the `target/`, but I might have said `stm32mp1` in some of them. This was how it used to be referred to prior to introduction of STM32MP13x in 2022.

Applied the review changes trying to keep logic intact, as separate small commits specific to review items.

Tested with a stm32f103cb bluepill running `stlink` platform (and as `hosted` over that via bmp-remote, HL/non-HL) against the STM32MP157C target idling in Engineering Boot. This seems to still work both on old `main` and current `main`, after sorting out some minor breakage from DPv0 and DP power-down code (with your help, much appreciated!). Builds for `native` and `hosted`, +460 bytes of flash for `native` and `stlink`, both still fit.

Please let me know whether I can squash these cosmetic commits into one, and when I can fast-forward the `main` in my fork and rebase this on top of current `main` regardless of squashing.

src/target/stm32mp15.c
dragonmux: We were suggesting `priv` specifically (we're actually working on dropping the `_storage` elsewhere) as while the target member has storage in its name (accurate for that structure) the result of the cast is the private structure - not storage for that structure. Hence as it's called `stm32mp15_priv_s` -> `priv` for the variable name.

src/target/stm32mp15.c
dragonmux: It's fine to declare them like this, but given it results in one giant memory region and they're all RAM, it's also absolutely fine for you to define them as one large region to save Flash space.

Further to the original review comment, please define these as proper constants at the top of file to name these otherwise magic numbers. We'd suggest:
```c
STM32MP15_RETRAM_BASE 0x00000000U
STM32MP15_RETRAM_SIZE 0x00010000U
STM32MP15_SRAM_BASE 0x10000000U
STM32MP15_RETRAM_SIZE 0x00060000U
```

src/target/stm32mp15.c
dragonmux: We honestly don't really care what Arterytek decided to do as we're trying to accomplish harmonised convention across BMD without having to know insane detail of the manufacturer's foibles and nomenclature to read and understand the different target support, and getting blind-sided each time one switches file with "yet another new convention".

In the rest of the code base the `_CTRL` suffix is used most consistently to refer to a control register. `_CR` doesn't mean anything as it's been used to mean things other than control registers and is too short. The convention we've been establishing in BMD is:

`CHIP_PERIPH_BASE` -> memory base address for a peripheral or memory region
`CHIP_PERIPH_CTRL` -> control register associated with that peripheral
`CHIP_PERIPH_STATUS` -> status register associated with that peripheral

where `CHIP` is the chip suffix (`STM32MP15` here) and `PERIPH` is the name of the peripheral from the datasheet (`DBGMCU` here).

src/target/stm32mp15.c
ALTracer: Accepted pending rename in detach and in probe to `priv`.
Should I also factor out this dbg_cr manipulation at end of probe into a separate _attach to decouple sleep-inhibition from scans? (this will also use the priv)

src/target/stm32mp15.c
dragonmux: This naming convention is something we've been working on fixing. Changing this to match the convention being established reduces the number of fixups we have to do later.

src/target/stm32mp15.c
dragonmux: It'd definitely be more hygienic, and especially as this currently makes detaching one-shot with scan required to bring the part back into a good state to attach again. 👍🏼

src/target/stm32mp15.c
dragonmux: This item still needs applying to lines 40 through 44, along with the register naming convention point in the next review comment down

src/target/stm32mp15.c
dragonmux: It seems like your pending fix for this never got pushed?

dragonmux:
It seems like after addressing some of the review items above in your working copy, you never pushed, so we've added notes where that is obvious to us. We'd like to get this merged ASAP.

Regarding the structuring of the fixup commits, we'd appreciate the fixes being applied as fixups to the base commits of the PR and we'll then rebase merge the resulting commit series. For example, an initial commit that provides the probe changes, the next adding memory maps, and another couple after that implementing Flash routines (erase, then write), is what we typically aim for in this kind of scenario.

src/target/stm32mp15.c
ALTracer: priv_storage rename was pushed in 88cccfd4, and I kinda forgot about _attach. Will revisit soon. Should be trivial.

src/target/stm32mp15.c
ALTracer: These regions are visible to CA7 cores at different addresses. For now I'm just reminding that there will be different aliases in cortexa counterpart code. Maybe these macros should be suffixed.

dragonmux: That's fine, these were only suggested names. Given the Cortex-A point, we'd insert `_CM4` after the chip name to sort that problem out and be consistent with the naming in the rest of this code.

ALTracer: Proposed change: insert `_CM4_` for bases (because of different memory maps) but not for sizes. Clang-formatted.
```diff
diff --git a/src/target/stm32mp15.c b/src/target/stm32mp15.c
index 99955cc6..ae118844 100644
--- a/src/target/stm32mp15.c
+++ b/src/target/stm32mp15.c
@@ -33,10 +33,10 @@
 #include "cortexm.h"
 
 /* Memory map constants for STM32MP15x */
-#define STM32MP15_RETRAM_BASE  0x00000000U
-#define STM32MP15_RETRAM_SIZE  0x00001000U /* RETRAM, 64 KiB */
-#define STM32MP15_AHBSRAM_BASE 0x10000000U
-#define STM32MP15_AHBSRAM_SIZE 0x00060000U /* AHB SRAM 1+2+3+4, 128+128+64+64 KiB */
+#define STM32MP15_CM4_RETRAM_BASE  0x00000000U
+#define STM32MP15_RETRAM_SIZE      0x00001000U /* RETRAM, 64 KiB */
+#define STM32MP15_CM4_AHBSRAM_BASE 0x10000000U
+#define STM32MP15_AHBSRAM_SIZE     0x00060000U /* AHB SRAM 1+2+3+4, 128+128+64+64 KiB */
 
 /* Access from processor address space.
  * Access via the debug APB is at 0xe0081000 over AP1. */
@@ -105,8 +105,8 @@ bool stm32mp15_cm4_probe(target_s *target)
 	target->target_storage = priv;
 
 	/* Figure 4. Memory map from §2.5.2 in RM0436 rev 6, pg158 */
-	target_add_ram(target, STM32MP15_RETRAM_BASE, STM32MP15_RETRAM_SIZE);
-	target_add_ram(target, STM32MP15_AHBSRAM_BASE, STM32MP15_AHBSRAM_SIZE);
+	target_add_ram(target, STM32MP15_CM4_RETRAM_BASE, STM32MP15_RETRAM_SIZE);
+	target_add_ram(target, STM32MP15_CM4_AHBSRAM_BASE, STM32MP15_AHBSRAM_SIZE);
 
 	return true;
 }
```
This doesn't feel right.

src/target/stm32mp15.c
ALTracer: Please reopen this on updated patchset contents. I don't quite understand how much level of detail is required here.

src/target/stm32mp15.c
dragonmux: Iff the sizes between both cores are the same, then this is pretty perfect - if they are not, then prefix the sizes too. Looks good otherwise.

ALTracer:
This should look better. If something else is wrong, please submit additional review on top of updated patchset contents, preferably with (updated) line numbers. Otherwise, since the feature/style pass is complete, I am aiming to commence the rebase pass this week to get into v1.10 release.

> Regarding the structuring of the fixup commits, we'd appreciate the fixes being applied as fixups to the base commits of the PR and we'll then rebase merge the resulting commit series. For example, an initial commit that provides the probe changes, the next adding memory maps, and another couple after that implementing Flash routines (erase, then write), is what we typically aim for in this kind of scenario.

There is only one base commit I see, the other 9 are minor edits (fixups). There is no flash in MP1. To reduce the size of one giant `+170` commit, I propose splitting it into three functional compileable parts: 
1. Identification, probe hierarchy hooks, new TU (with license header) of minimal functionality (RAM map);
2. UID & rev/IDCODE commands implementation (mostly H7 code + MP1 values);
3. DBGMCU operations on attach/detach, priv

As of now this patchset takes up +432 bytes on `native` and still fits when cherry-picked verbatim to latest `main` (1236 bytes left). Using 12.2.Rel1 Toolchain.

dragonmux:
Your proposed patch series structure seems fine - we'll let you get the rebase done and review the result as we think that'll probably then be accepting the change set for merge.

src/target/stm32mp15.c
ALTracer: I mean... It's the same region. See ST devicetrees, https://github.com/STMicroelectronics/linux/blob/8e756f0b4a005e9a0374ab2ffb20df8c3ed4ed63/arch/arm/boot/dts/stm32mp151.dtsi#L2025-L2029 and more importantly https://github.com/STMicroelectronics/linux/blob/8e756f0b4a005e9a0374ab2ffb20df8c3ed4ed63/arch/arm/boot/dts/stm32mp15xx-dkx.dtsi#L25-L70 for aliases and default OpenAMP remoteproc splits. I doubt the debugger needs to know about both aliases, but just for good measure also referencing this https://wiki.st.com/stm32mpu/wiki/STM32MP15_RAM_mapping

dragonmux: This item is still outstanding in the diff - the base addresses still need prefixing as you noted due to them being core-dependant.

dragonmux (commented):
The only item preventing this PR from being ready for merge, is the one on the RAM base macro naming. Once that's solved this looks ready to go and we can approve for merge.

dragonmux (approved):
LGTM, we'll get this merged after we merge the Cortex-A fix PR.

PR #1540, opened 2023-07-17
Feature: Identify AT32F435/437

ALTracer:
## Detailed description

* A new feature is implemented: detection of AT32F435 cores by DBGMCU_IDCODE with assignment of respective total internal Flash sizes and SRAM in default NZW split.
* The pull request solves an existing problem of none/generic detection of these cores as "Cortex-M4".
* The pull request solves that by introducing a list of ID codes from reference manual for mentioned parts.

Arterytek / Artery / ARTERY AT32F435 is a more powerful (than F403A/F407/F415) core variant with possible Flash sizes of 256, 1024, 448 or 4032 KiB and bank split at 512 or 2048 KiB watermark. 
AT32F437 is a F435 with an Ethernet "EMAC" core.
Note that at the moment flash (re)programming is unsupported/unreliable, although the page (or "sector") sizes should be correct -- I couldn't get BMF or BMDA to erase the target flash of AT32F435ZMT6.

## Your checklist for this pull request

* [x] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md)
* [x] I've read the [guidelines for contributing](https://github.com/blackmagic-debug/blackmagic/blob/main/CONTRIBUTING.md) to this repository
* [x] It builds for hardware native (`make PROBE_HOST=native`)
* [x] It builds as BMDA (`make PROBE_HOST=hosted`)
* [x] I've tested it to the best of my ability
* [x] My commit messages provide a useful short description of what the commits do

## Miscellaneous
This PR may be improved by redacting comments, now that I see the errors after a couple weeks since writing the patch.
Functionality may depend on having distinct (from stm32f1) erase & write routines, although the FLASH controller seems very similar to stm32f1-style one.

PR #1547, opened 2023-07-19
Feature: AT32F435/F437 flash programming

### LINKED ISSUE #1542: AT32F4 flash handling
## Summary

How should all these different parts be handled? 

## Description

~~#1150~~ actually #1151 introduced detection of ARTERY AT32F403A, F407 (and F407A) and F415 parts by listing their DBGMCU_IDCODE values, or `part_id` as referred in sources in `src/target/stm32f1.c`. I agree with the `at32f41_detect()` routine which assigns different flash sizes to F415 parts, but disagree with `at32f40_detect()` which makes them all appear as 256 KiB parts.

In the PR discussion and in the code comments there are concerns about "external memory" and "default layout".
1. From my understanding of RM_AT32F403A_407_EN reference manual the MCUs have a "SPIM" external Quad I/O-like NOR Flash memory controller which can map into 0x0840_0000 base and perform transparent encryption. However, I'm now interested in internal flash only. 256 & 512 KiB flash parts have it all in a single bank in 2048-byte pages. 1 MiB parts employ two banks, the split address is at 0x0808_0000 (+512 KiB) and the second bank has individual FLASH_CTRL2 etc. registers. 
2. ZW & NZW split is a characteristic feature of most AT32F4 MCUs: first 128 KiB of Flash are always shadowed by 128 KiB of on-chip SRAM which cannot be reclaimed. Upwards from that, [by default] another (next) 128 KiB of Flash are shadowed/cached by 128 KiB of SRAM and this margin can be moved lower (or higher if there's more SRAM, like on F435). On F403A/F407, 96 leftover KiB of SRAM are by default available as work SRAM (but 96+128=224 max). This is controlled by option bytes. NZW flash runs at 72 MHz, for 240-288 MHz cores this incurs 4 wait states without caching (on jumps).

* Is it possible to use all of 512..1024 KiB flash on AT32F403A, particularly on WeActStudio.BlackPill AT32F403ACGU7, by registering bigger sizes / page counts in `stm32f1_add_flash()`?
* AT32F435/7 parts with 256 KiB are single-bank with 2048-byte pages, too; but parts with 1024 KiB flash are dual-bank, split equally, at 0x0808_0000 address (+512 KiB). These targets can be treated flash-wise like the F403A.
* But AT32F435/7 parts with 4096-byte pages (also 448 KiB, single-bank -- simple, not considered here) of 4032 KiB dual-bank flash are split at 0x0820_0000 address (+2048 KiB) which is unusual. 

`stm32f1.c` has branches on 0x430U XL-density in mass_erase, flash_erase & flash_write routines:
https://github.com/blackmagic-debug/blackmagic/blob/c4b988b6319a9a038301609bfe01ca577047c6ea/src/target/stm32f1.c#L619-L621
but it relies on macros for +0x40 control registers offset (which is suitable) and +512 KiB second bank offset
https://github.com/blackmagic-debug/blackmagic/blob/c4b988b6319a9a038301609bfe01ca577047c6ea/src/target/stm32f1.c#L66-L68

There are other targets already supporting dual-bank MCUs. Like stm32h7.c has a simple dual-bank support logic, or stm32f4.c has a rather convoluted one.
I am writing #1540 and I need recommendations on how to handle this situation. Maybe all of AT32F4 should be split out as a separate target, or a subset of functions (but stm32f1 flash API is not public), or just a set of distinct write/erase/mass_erase routines living in stm32f1.c will be enough (using its own macros for register manipulations)?

ALTracer:
## Detailed description

* This can be considered a new feature, but in essence is old code cloned/ported/refactored to suit Artery targets.
* The previous PR #1540 attached `stm32f1_flash` routines to AT32F435 targets and this doesn't work.
* This PR implements mandatory erase/write routines specifically for AT32F43x so that it's actually possible to reprogram target flash.

ARTERY AT32F435 (and F437) series chips have a flash controller which is similar in operation to `stm32f1` but have its register base at 0x40023C00 like `stm32f4`. It has even sectors (all the same size, 2048 or 4096 byte pages) unlike most `stm32f4` (or `gd32f4`) targets, and requires selecting sectors to erase by their address, not sector number. I considered using code in one of these target drivers but the difference are too significant.

I ended up porting relevant routines from `stm32f1`, incorporating runtime dual-bank split argument passing (for controller register base and for physical address split in internal flash) and updating naming/code style where I noticed it. Some of `stm32f4` code was helpful but not directly applicable.

Tested on AT32F437ZMT7 (4032 KiB in 4 KiB pages) of AT-START-F437 by wiping and loading a glorified blinky project occupying 12 KiB using AT32IDE with its xpack gdb, (*and in ST 10.3 GNU MCU Tools GDB, also Ubuntu/jammy gdb-multiarch*).
Dual bank functionality tested by invoking `hosted` app to write a 3.5 MiB datasheet DS_AT32F435_F437_EN.pdf into the target, and then verify, both of which passed (at like 20 KiB/s speed).

Mass erase takes 7 seconds per 2 MiB bank, as the datasheet says, and BMF (blackpill-f411ce) indicates "progress" with dots twice per second. Reprogramming with BMF (or verifying `compare-sections`) displays performance of 15-19 KiB/s (15 for JTAG, 19 for SWDP), which is twice as fast compared to onboard AT-Link-EZ in CMSIS-DAP mode implementing SWD-only at "5000 kHz" when used with Arterytek-patched OpenOCD (sources?) from AT32IDE.

I did not aim to implement logic for option bytes ("USD, user data"), EPPS erase/program protection, SLib, NZW/ZW split manipulation and other Artery-specific features. These can be implemented later by interested contributors.

## Closing issues

fixes #1542

### DIFF
--- a/src/target/at32f43x.c
+++ b/src/target/at32f43x.c
@@ -32,6 +32,39 @@
 #include "target_internal.h"
 #include "cortexm.h"
 
+static bool at32f43_flash_prepare(target_flash_s *flash);
+static bool at32f43_flash_erase(target_flash_s *flash, target_addr_t addr, size_t len);
+static bool at32f43_flash_write(target_flash_s *flash, target_addr_t dest, const void *src, size_t len);
+static bool at32f43_flash_done(target_flash_s *flash);
+static bool at32f43_mass_erase(target_s *target);
+
+/* Flash memory controller register map */
+#define AT32F43x_FLASH_REG_BASE 0x40023c00U
+#define AT32F43x_FLASH_UNLOCK   (AT32F43x_FLASH_REG_BASE + 0x04U)
+#define AT32F43x_FLASH_STS      (AT32F43x_FLASH_REG_BASE + 0x0cU)
+#define AT32F43x_FLASH_CTRL     (AT32F43x_FLASH_REG_BASE + 0x10U)
+#define AT32F43x_FLASH_ADDR     (AT32F43x_FLASH_REG_BASE + 0x14U)
+/* There is a second set of identical registers at +0x40 offset for Bank 2 */
+
+#define AT32F43x_FLASH_BANK1_REG_OFFSET 0x00U
+#define AT32F43x_FLASH_BANK2_REG_OFFSET 0x40U
+
+/* Flash registers bit fields */
+#define AT32F43x_FLASH_CTRL_FPRGM   (1U << 0U)
+#define AT32F43x_FLASH_CTRL_SECERS  (1U << 1U)
+#define AT32F43x_FLASH_CTRL_BANKERS (1U << 2U)
+#define AT32F43x_FLASH_CTRL_ERSTR   (1U << 6U)
+#define AT32F43x_FLASH_CTRL_OPLK    (1U << 7U)
+/* CTRL bits [8:11] are reserved, parallellism x8/x16/x32 (don't care) */
+
+/* OBF is BSY, ODF is EOP */
+#define AT32F43x_FLASH_STS_OBF     (1U << 0U)
+#define AT32F43x_FLASH_STS_PRGMERR (1U << 2U)
+#define AT32F43x_FLASH_STS_ODF     (1U << 5U)
+
+#define AT32F43x_FLASH_KEY1 0x45670123U
+#define AT32F43x_FLASH_KEY2 0xcdef89abU
+
 #define DBGMCU_IDCODE 0xe0042000U
 
 #define AT32F4x_IDCODE_SERIES_MASK 0xfffff000U
@@ -39,33 +72,42 @@
 #define AT32F43_SERIES_4K          0x70084000U
 #define AT32F43_SERIES_2K          0x70083000U
 
-static void at32f43_add_flash(target_s *const t, const uint32_t addr, const size_t length, const size_t blocksize,
-	const uint8_t base_sector, const uint8_t split)
+typedef struct at32f43_flash {
+	target_flash_s target_flash;
+	target_addr_t bank_split; /* Address of first page of bank 2 */
+	uint32_t bank_reg_offset; /* Flash register offset for this bank */
+} at32f43_flash_s;
+
+static void at32f43_add_flash(target_s *const target, const target_addr_t addr, const size_t length,
+	const size_t pagesize, const target_addr_t bank_split, const uint32_t bank_reg_offset)
 {
 	if (length == 0)
 		return;
 
-	target_flash_s *f = calloc(1, sizeof(*f));
-	if (!f) { /* calloc failed: heap exhaustion */
+	at32f43_flash_s *flash = calloc(1, sizeof(*flash));
+	if (!flash) { /* calloc failed: heap exhaustion */
 		DEBUG_ERROR("calloc: failed in %s\n", __func__);
 		return;
 	}
 
-	f->start = addr;
-	f->length = length;
-	f->blocksize = blocksize;
-	f->erase = NULL;
-	f->write = NULL;
-	(void)base_sector;
-	(void)split;
-	f->writesize = 1024;
-	f->erased = 0xffU;
-	target_add_flash(t, f);
+	target_flash_s *target_flash = &flash->target_flash;
+	target_flash->start = addr;
+	target_flash->length = length;
+	target_flash->blocksize = pagesize;
+	target_flash->prepare = at32f43_flash_prepare;
+	target_flash->erase = at32f43_flash_erase;
+	target_flash->write = at32f43_flash_write;
+	target_flash->done = at32f43_flash_done;
+	target_flash->writesize = 1024U;
+	target_flash->erased = 0xffU;
+	flash->bank_split = bank_split;
+	flash->bank_reg_offset = bank_reg_offset;
+	target_add_flash(target, target_flash);
 }
 
 static bool at32f43_detect(target_s *target, const uint16_t part_id)
 {
-	/* 
+	/*
 	 * AT32F435 EOPB0 ZW/NZW split reconfiguration unsupported,
 	 * assuming default split ZW=256 SRAM=384.
 	 * AT32F437 also have a working "EMAC" (Ethernet MAC)
@@ -83,7 +125,7 @@ static bool at32f43_detect(target_s *target, const uint16_t part_id)
 	case 0x054fU: // LQFP144 w/Eth
 	case 0x0552U: // LQFP100 w/Eth
 	case 0x0555U: // LQFP64 w/Eth
-		// Flash (G): 4032 KB in 2 banks (2048+1984), 4KB per sector.
+		// Flash (M): 4032 KB in 2 banks (2048+1984), 4KB per sector.
 		flash_size_bank1 = 2048U * 1024U;
 		flash_size_bank2 = 1984U * 1024U;
 		sector_size = 4096;
@@ -109,7 +151,7 @@ static bool at32f43_detect(target_s *target, const uint16_t part_id)
 	case 0x0350U: // LQFP144 w/Eth
 	case 0x0353U: // LQFP100 w/Eth
 	case 0x0356U: // LQFP64 w/Eth
-		// Flash (M): 1024 KB in 2 banks (equal), 2KB per sector.
+		// Flash (G): 1024 KB in 2 banks (equal), 2KB per sector.
 		flash_size_bank1 = 512U * 1024U;
 		flash_size_bank2 = 512U * 1024U;
 		sector_size = 2048;
@@ -134,11 +176,14 @@ static bool at32f43_detect(target_s *target, const uint16_t part_id)
 	 * Block erase operates on 64 KB at once for all parts.
 	 * Using here only sector erase (page erase) for compatibility.
 	 */
-	at32f43_add_flash(target, 0x08000000, flash_size_bank1, sector_size, 0, 0);
 	if (flash_size_bank2 > 0) {
-		const uint16_t base_sector = flash_size_bank1 / sector_size;
-		at32f43_add_flash(target, 0x08000000 + flash_size_bank1, flash_size_bank2, sector_size, base_sector, 0);
-	}
+		const uint32_t bank_split = 0x08000000 + flash_size_bank1;
+		at32f43_add_flash(
+			target, 0x08000000, flash_size_bank1, sector_size, bank_split, AT32F43x_FLASH_BANK1_REG_OFFSET);
+		at32f43_add_flash(
+			target, bank_split, flash_size_bank2, sector_size, bank_split, AT32F43x_FLASH_BANK2_REG_OFFSET);
+	} else
+		at32f43_add_flash(target, 0x08000000, flash_size_bank1, sector_size, 0, AT32F43x_FLASH_BANK1_REG_OFFSET);
 
 	// SRAM1 (64KB) can be remapped to 0x10000000.
 	target_add_ram(target, 0x20000000, 64U * 1024U);
@@ -149,11 +194,11 @@ static bool at32f43_detect(target_s *target, const uint16_t part_id)
 	 * Out of 640 KB SRAM present on silicon, at least 128 KB are always
 	 * dedicated to "zero-wait-state Flash". ZW region is limited by
 	 * specific part flash capacity (for 256, 448 KB) or at 512 KB.
-	 * AT32F435ZMT default EOPB0=0xffff05fa, 
+	 * AT32F435ZMT default EOPB0=0xffff05fa,
 	 * EOPB[0:2]=0b010 for 384 KB SRAM + 256 KB zero-wait-state flash.
 	 */
 	target->driver = "AT32F435";
-	target->mass_erase = NULL;
+	target->mass_erase = at32f43_mass_erase;
 	return true;
 }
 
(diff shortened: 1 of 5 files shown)


src/target/at32f43x.c
```diff
@@ -0,0 +1,376 @@
...
+/* Flash memory controller register map */
+#define AT32F435_FLASH_REG_BASE 0x40023c00U
+#define AT32F435_FLASH_UNLOCK   (AT32F435_FLASH_REG_BASE + 0x04U)
+#define AT32F435_FLASH_STS      (AT32F435_FLASH_REG_BASE + 0x0cU)
+#define AT32F435_FLASH_CTRL     (AT32F435_FLASH_REG_BASE + 0x10U)
+#define AT32F435_FLASH_ADDR     (AT32F435_FLASH_REG_BASE + 0x14U)
+//#define AT32F435_FLASH_UNLOCK2  (AT32F435_FLASH_REG_BASE + 0x44U)
+//#define AT32F435_FLASH_STS2     (AT32F435_FLASH_REG_BASE + 0x4cU)
+//#define AT32F435_FLASH_CTRL2    (AT32F435_FLASH_REG_BASE + 0x50U)
+//#define AT32F435_FLASH_ADDR2    (AT32F435_FLASH_REG_BASE + 0x54U)
```
dragonmux, line 48: As these are unused, please drop them rather than commenting them out. You'd be welcome to write a comment noting of their existence and being +0x40 from the previous registers, but this is a clang-tidy lint for a reason.

src/target/at32f43x.c
```diff
@@ -0,0 +1,376 @@
...
+
+#define FLASH_BANK1_REG_OFFSET 0x00U
```
dragonmux, line 50: Ideally all these defines want prefixing with `AT32F43_`, and the 5 above can have the 5 dropped or changed to an x in light of the file name

src/target/at32f43x.c
```diff
@@ -0,0 +1,376 @@
...
+	if (length == 0)
+		return;
+
+	at32f43_flash_s *sf = calloc(1, sizeof(*sf));
+	if (!sf) { /* calloc failed: heap exhaustion */
+		DEBUG_ERROR("calloc: failed in %s\n", __func__);
+		return;
+	}
+
+	target_flash_s *flash = &sf->flash;
```
dragonmux, line 98: We suggest renaming `flash` here to `target_flash` so you can rename `sf` to `flash` and avoid introducing yet more short variable names we need to fix.

src/target/at32f43x.c
```diff
@@ -0,0 +1,376 @@
...
+
+	/* Erase range begins in bank 1? Unlock bank 1 */
+	if (addr < sf->bank_split && !at32f43_flash_unlock(target, 0))
+		return false;
+	/* Erase range ends in bank 2? Unlock bank 2 */
+	const target_addr_t end = addr + len - 1U;
+	if (end >= sf->bank_split && !at32f43_flash_unlock(target, FLASH_BANK2_REG_OFFSET))
+		return false;
+
+	for (size_t offset = 0; offset < len; offset += flash->blocksize) {
```
dragonmux, line 283: The target Flash API layer guarantees `len` will be flash->blocksize and `addr` will start on a block boundary, so this loop can be ditched and the return simplified to `return at32f43_flash_busy_wait(target, bank_reg_offset, NULL);`

dragonmux (changes requested):
This is looking great. There are a couple of items we've caught in review, including a simplification of the Flash routines for a little extra space savings over the work you were sharing in Discord regarding the part_id switch-case and other bits.

With these items addressed, this looks good good to merge.

src/target/at32f43x.c
ALTracer: These two are used in 6 & 7 places, am I to rename them?
Does this apply to all the bitfield macros, too? It may introduce more Very Long lines, as the bits are usually used in calls with long (macro) arguments and at some level of nesting/indentation.
`AT32F435` + `AT32F437` = `AT32F43X` or lowercase `AT32F43x`?

src/target/at32f43x.c
dragonmux: you'll need to install clang-tidy and then you can get lints from the tool running `make clang-tidy` or, as that'll spew tons of output and be a bit unhelpful, if you use, say, VSCode then you can run this configuration to have the cpptools extension run clang-tidy for you per-file:
```json
	"C_Cpp.codeAnalysis.runAutomatically": true,
	"C_Cpp.codeAnalysis.clangTidy.enabled": true,
	"C_Cpp.codeAnalysis.clangTidy.useBuildPath": true,
```

You can technically also run clang-tidy manually per-file but getting the command line options right is a bit of a bore which is why there's automation in the repo for it (scripts/run-clang-tidy.py)

src/target/at32f43x.c
ALTracer: Can do. There are 9 other uses of `sf`, should I do something about them, too?
May I inquire why were they originally called `t`, `f` and `sf` in legacy driver code?

src/target/at32f43x.c
dragonmux: Yes, the review item is to rename all in-file macros that define things about the devices' registers to be prefixed with `AT32F43x` and, yes, with the lower-case x there (it's consistent with targets like LPC43xx).

We're not worried about long lines as clang-format will reflow them to keep them <= 120 characters long. Being properly descriptive matters more to us.

src/target/at32f43x.c
dragonmux: We couldn't hope to know what was going on in Gareth's mind nor other contributor's, but as the short variable names have lead to readability issues and confusion and as it's a clang-tidy lint, we've been gradually fixing this over the last couple of releases

src/target/at32f43x.c
ALTracer: This loop was copied verbatim from `stm32f1_flash_erase()` by me not familiar with internal BMD API layers. Do you mean that in current implementation this for-loop will always run exactly once and its contents can be de-indented along with dropping `offset`? I may need to test with `DEBUG_INFO`s before significantly changing driver logic.

dragonmux: Exactly, and the STM32F1 implementation hasn't yet been updated to match that reality.

src/target/at32f43x.c
ALTracer: I installed the pre-commit hook to run clang-format, can `clang-tidy` be run likewise for modified files or even diff context? `clang-format -i src/tu.c` updates sources per-file to conform to codestyle in its settings, and this is useful for me. I don't do codebase-wide refactors (often), and `make clang-tidy` outputs a ton of noise for me.
I run `cppcheck` as Geany lint button, so having `scripts/run-clang-tidy.py` print an example command would allow me to attach it into Geany User actions (it understands %f and other templates).

dragonmux: Unsure what "%f" means in this context and especially as we don't know Geany, but we'd be surprised if it doesn't have some kind of clang-tidy integrations. Regarding pre-commit hooks, we'd have to ask @perigoso about those as we don't know but we also wouldn't include one in the repo for the same reason as why `make clang-tidy` is currently not so helpful.

We can look at having the run-clang-tidy.py script "dry run" things and look at specifying specific files to poke rather than the whole project.

dragonmux:
Additional review note: Per the contribution guidelines, the `target/` prefix on the target reference in the commit message is not required, `at32f43x:` instead of `target/at32f43x:` is sufficient as a prefix

src/target/at32f43x.c
perigoso: Our pre-commit configuration doesn't include a hook for `clang-tidy` at the moment, only `clang-format`, but that can be done and I'm pretty sure there's already one ready to go for us.
I don't remember why I didn't include it from the start, but if I had to guess it was because the code still contained a lot of `clang-tidy` warnings and it would probably cause more trouble than help

src/target/at32f43x.c
```diff
@@ -0,0 +1,376 @@
...
+		return;
+	}
+
+	target_flash_s *flash = &sf->flash;
+	flash->start = addr;
+	flash->length = length;
+	flash->blocksize = pagesize;
+	flash->erase = at32f43_flash_erase;
+	flash->write = at32f43_flash_write;
+	flash->writesize = 1024U; // limited by FLASH_WRITE_BUFFER_CEILING
```
perigoso, line 104: This is not quite correct, `FLASH_WRITE_BUFFER_CEILING` only limits how big of a buffer we *automatically* create for targets that have small write sizes, if your write size is bigger a buffer of that size will be created, What that may cause is allocation issues on targets with small RAM, but that's another story

Following on that not, writesize is meant to be the *minimum* write size, the flash API layer handles the rest (i.e a larger buffer with multiple write sizes for efficiency), I would recommend using that approach unless there are technical reasons not to or it incurs a serious speed decrease

src/target/at32f43x.c
```diff
@@ -0,0 +1,376 @@
...
+	target_s *target = flash->t;
+	const at32f43_flash_s *const sf = (at32f43_flash_s *)flash;
+
+	/* Erase range begins in bank 1? Unlock bank 1 */
+	if (addr < sf->bank_split && !at32f43_flash_unlock(target, 0))
+		return false;
+	/* Erase range ends in bank 2? Unlock bank 2 */
+	const target_addr_t end = addr + len - 1U;
+	if (end >= sf->bank_split && !at32f43_flash_unlock(target, FLASH_BANK2_REG_OFFSET))
+		return false;
```
perigoso, line 281: This code belongs in a `flash_prepare` function, this note goes hand in hand with the note on `writesize`, you can then re-lock the flash in a `flash_done` if needed

src/target/at32f43x.c
perigoso: co-signing this note. the `stm32f1_flash_erase` is "legacy code" that needs to be updated to the current "workflow"

Testing is very much welcome as we can't have too many guarantees that code is working properly, but it is expected to function as noted. (I believe there is come testing "logs" on a discord thread when I originally wrote that code)

src/target/at32f43x.c
```diff
@@ -0,0 +1,376 @@
...
+	if (addr + len > bank_split)
+		return bank_split - addr;
+	return len;
+}
+
+static bool at32f43_flash_write(target_flash_s *flash, target_addr_t dest, const void *src, size_t len)
+{
+	target_s *target = flash->t;
+	const at32f43_flash_s *const sf = (at32f43_flash_s *)flash;
+	const size_t range = at32f43_range_in_bank1(dest, len, sf->bank_split);
```
perigoso, line 314: this variable name leaves a bit to be desired

src/target/at32f43x.c
```diff
@@ -0,0 +1,376 @@
...
+	if (sf->bank_split && remainder) {
+		const uint8_t *data = src;
+		at32f43_flash_clear_eop(target, FLASH_BANK2_REG_OFFSET);
+
+		target_mem_write32(target, AT32F435_FLASH_CTRL + FLASH_BANK2_REG_OFFSET, FLASH_CTRL_FPRGM);
+		cortexm_mem_write_sized(target, dest + range, data + range, remainder, psize);
+
+		if (!at32f43_flash_busy_wait(target, FLASH_BANK2_REG_OFFSET, NULL))
+			return false;
+	}
```
perigoso, line 340: I don't understand the need for this split / remainder handling, maybe you could clear it up. Aren't the 2 banks handled by 2 different flash structs? if so a call to `flash_write` for bank1 flash should never have any data for bank2 flash.

src/target/at32f43x.c
ALTracer: I don't completely understand. There's no lower bound on write sizes, you can literally program a single byte into the page pre-erased some time ago, that's how they offer EEPROM emulation library with alternating between two Flash pages.
A page, or sector as in RM, is either 2048 or 4096 bytes, depending on part, and is the smallest erase unit. They support "block" erases of 65536 bytes for both part varieties as well, and then there's bank erase.
If I declare a 4096-byte write buffer in here, will small adapters crash at runtime (esp. swlink/stlink/native)? What is the recommended writesize for them?

src/target/at32f43x.c
ALTracer: Sounds like this API is newer than the target driver. Which targets use it with internal Flash? I could use `prepare` to unlock both banks, but then there's stm32h5 which does it in enter_flash_mode, so what fits better here?

src/target/at32f43x.c
ALTracer: It was named "offset", and I decided to rename it to something but didn't want to name it e.g. "length_to_write_in_bank1" which is more descriptive but convoluted.

src/target/at32f43x.c
ALTracer: On second thought you're right, a single write call won't cross bank boundary, and that is guaranteed by new target flash API which flushes buffered writes `writesize`-wise. But I still need to dispatch writes to the correct bank (registers) as there's only one routine used with both regions.

src/target/at32f43x.c
perigoso: The recomended `writesize` is the smallest possible write we can do, if that is a 32bit word so be it.

src/target/at32f43x.c
perigoso: maybe `bank1_range` or `bank1_length`

src/target/at32f43x.c
ALTracer: Rename complete.

src/target/at32f43x.c
ALTracer: Performance suffers with 4-byte writes. Reverted to writesize of 1024 because that's what stm32f1 driver uses, and popular low-end platforms can't allocate more (e.g. I tried asking for 8192 in gd32e50x/stm32f1 and it OOMed the swlink probe)

src/target/at32f43x.c
ALTracer: Simplified, please review.

src/target/at32f43x.c
ALTracer: Dropped cross-bank write logic.

src/target/at32f43x.c
dragonmux: Looks good to us, marking as resolved.

src/target/at32f43x.c
```diff
@@ -0,0 +1,373 @@
...
+	at32f43_flash_clear_eop(target, bank_reg_offset);
+
+	target_mem_write32(target, AT32F435_FLASH_CTRL + bank_reg_offset, FLASH_CTRL_FPRGM);
+	cortexm_mem_write_sized(target, dest, src, len, psize);
+
+	/* Datasheet: flash programming takes 50us (typ), 200us (max) */
+	if (!at32f43_flash_busy_wait(target, bank_reg_offset, NULL))
+		return false;
+
+	return true;
```
dragonmux, line 342: This can still be simplified to:
```suggestion
	return at32f43_flash_busy_wait(target, bank_reg_offset, NULL);
```

There should be a clang-tidy lint for this (though getting it to fire in all similar situations can be a challenge).

ALTracer: Simplified as suggested. There was more complicated logic dropped around it.

dragonmux (commented):
From a code perspective, this looks good. Please fix up your commit messages as noted in a previous review and apply the review fix items from the last 3 commits as fixups over the others in the series. We'll approve this for merge once that's done.

dragonmux (approved):
LGTM, we'll get this merged at the very start of the v2.0 merge window opening. Many thanks for your hard work on this contribution!

PR #1548, opened 2023-07-20
J-Link cleanup and Target  Power implementation

perigoso:
## Detailed description

Branch is a fix, but this is actually a feature, all things considered

This is a general cleanup of J-Link code, with the final goal of adding target power functionality

It also includes the change from #1545:
>  J-Link version string was being shown on BMDA startup with verbosity, but not when calling the version command, this addresses that

Tested with a J-Link V8 and OB J-Link (Silabs Dev board)

* [ ] ~~It builds for hardware native (`make PROBE_HOST=native`)~~

## Closing issues

replaces #1545


src/include/platform_support.h
```diff
@@ -55,7 +55,7 @@ int platform_hwversion(void);
 void platform_nrst_set_val(bool assert);
 bool platform_nrst_get_val(void);
 bool platform_target_get_power(void);
-void platform_target_set_power(bool power);
+bool platform_target_set_power(bool power);
```
dragonmux, line 58: This needs to be backed by similar changes in all the platforms, not just BMDA.

src/platforms/hosted/jlink.c
```diff
@@ -67,7 +67,7 @@ int jlink_simple_request(const uint8_t command, const uint8_t operation, void *c
 }
 
 /*
- * This runs JLINK_CMD_IO_TRANSACT transactions, these have the following format:
+ * This runs JLINK_CMD_HW_JTAG3 transactions, these have the following format:
```
dragonmux, line 88: While this is technically correct, what "HW_JTAG3" means is less clear - which is why we'd specifically renamed it for BMDA. A note about how our command names map to the manual might be a good call, but we would prefer to keep the more descriptive naming (it's performing an IO transaction!).

Same for the naming of things like the reset commands. We'd done that for code clarity because we got fed up of having headaches trying to understand the J-Link code every time we dug into it and needing the manual open to understand the code.

dragonmux (changes requested):
This PR does a lot to the J-Link code so we've started small with a couple of items we've spotted that have large knock-on effects. We'll review further once these items have been addressed

perigoso:
Pushed a couple fixups, typos and extra const keywords

src/platforms/hosted/jlink_protocol.h
```diff
@@ -25,52 +26,334 @@
...
+ * └────────────────────────┴────────────────────────────────┘
+ */
+#define JLINK_CMD_CONFIG_READ  0xf2U /* Read the probe configuration */
+#define JLINK_CMD_CONFIG_WRITE 0xf3U /* Write the probe configuration */
+
+/* J-Link hardware version - JLINK_CMD_INFO_GET_HARDWARE_VERSION */
+#define JLINK_HARDWARE_VERSION_TYPE_JLINK    0U
+#define JLINK_HARDWARE_VERSION_TYPE_JTRACE   1U
+#define JLINK_HARDWARE_VERSION_TYPE_FLASHER  2U
+#define JLINK_HARDWARE_VERSION_TYPE_JLINKPRO 3U
```
dragonmux, line 185: These seem to be duplicates of the below definitions and the below ones make a bit more sense in terms of order of definitions. Please remove these.

src/platforms/hosted/jlink.c
```diff
@@ -42,48 +43,71 @@
...
 
-int jlink_simple_query(const uint8_t command, void *const rx_buffer, const size_t rx_len)
+/* J-Link USB protocol functions */
+
+bool jlink_simple_query(const uint8_t command, void *const rx_buffer, const size_t rx_len)
+{
+	return bmda_usb_transfer(info.usb_link, &command, sizeof(command), rx_buffer, rx_len, JLINK_USB_TIMEOUT) >= 0;
+}
+
+bool jlink_simple_request(const uint8_t command, const uint8_t operation, void *const rx_buffer, const size_t rx_len)
```
dragonmux, line 68: Given the below two functions, it would make sense for consistency for this to be `jlink_simple_request_8` we would suggest.

src/platforms/hosted/jlink.c
```diff
@@ -151,91 +175,24 @@ bool jlink_transfer_swd(
...
 /*
  * Try to claim the debugging interface of a J-Link adaptor.
  * On success this copies the endpoint addresses identified into the
  * usb_link_s sub-structure of bmp_info_s (info.usb_link) for later use.
  * Returns true for success, false for failure.
+ *
+ * Note: Newer J-Links use 2 bulk endpoints, one for "IN" (EP1) and one
+ * for "OUT" (EP2) communication whereas old J-Links (V3, V4) only use
+ * one endpoint (EP1) for "IN" and "OUT" communication.
+ * Presently we only support the newer J-Links with 2 bulk endpoints.
```
dragonmux, line 195: How difficult would it be to detect and configure for the older style? If that might even be vaguely useful..

src/platforms/hosted/jlink.c
```diff
@@ -277,6 +234,313 @@ static bool jlink_claim_interface(void)
...
+}
+
+static bool jlink_get_capabilities(void)
+{
+	uint8_t buffer[4U];
+	if (!jlink_simple_query(JLINK_CMD_INFO_GET_PROBE_CAPABILITIES, buffer, sizeof(buffer)))
+		return false;
+
+	jlink.capabilities = read_le4(buffer, 0);
+	DEBUG_INFO("Capabilities: 0x%" PRIx32 "\n", jlink.capabilities);
```
dragonmux, line 321: Does it make sense to make this `0x%08" PRIx32 "\n"`? So that the value is always the same width and so a bit easier to read and parse out visually

src/platforms/hosted/jlink.c
```diff
@@ -277,6 +234,313 @@ static bool jlink_claim_interface(void)
...
+
+	return true;
+}
+
+static bool jlink_interface_available(const uint8_t interface)
```
dragonmux, line 326: We'd suggest `static inline bool` given this is a one-liner and trivially inlined

src/platforms/hosted/jlink.c
```diff
@@ -277,6 +234,313 @@ static bool jlink_claim_interface(void)
...
+		return true;
+
+	/* Select the requested interface */
+	uint8_t buffer[4U];
+	if (!jlink_simple_request(JLINK_CMD_INTERFACE_SET_SELECTED, interface, buffer, sizeof(buffer)))
+		return false;
+
+	/* Give the J-Link some time to switch interfaces */
+	platform_delay(10U);
+
```
dragonmux, line 359: We'd remove this extra new line

src/platforms/hosted/jlink.c
```diff
@@ -277,6 +234,313 @@ static bool jlink_claim_interface(void)
...
+	// Restoring the selected interface seemed to cause issues sometimes (unsure if culprit)
+	// so we don't do it for now, it didn't seem to be necessary, saving in case it proves to be necessary later
+	// if (selected_interface != interface) {
+	// 	/* If the selected interface didn't match the requested interface, restore the selected interface */
+	// 	DEBUG_WARN("Restoring selected interface to %s\n", jlink_interface_to_string(selected_interface));
+
+	// 	/* select the requested interface, we don't consider this a critical failure, as we got what we wanted */
+	// 	if (!jlink_select_interface(selected_interface))
+	// 		DEBUG_ERROR("Failed to restore selected interface to %s\n", jlink_interface_to_string(selected_interface));
+	// }
```
dragonmux, line 441: Rather than committing this as commented out code, it would be better to instead commit this in an `#if 0` block if it's going to be included in the PR

src/platforms/hosted/jlink.c
```diff
@@ -277,6 +234,313 @@ static bool jlink_claim_interface(void)
...
+		DEBUG_WARN("J-Link does not support interface frequency command\n");
+		return false;
+	}
+
+	/* Get the selected interface */
+	const uint8_t selected_interface = jlink_selected_interface();
+
+	if (selected_interface != interface) {
+		/* If the selected interface doesn't match the requested interface, select it, let's hope this doesn't mess something up elsewhere */
+		DEBUG_WARN("Trying to set frequency for interface %s but it is not selected, selecting it\n",
```
dragonmux, line 458: We think this makes a touch more sense as a DEBUG_INFO() statement as it's not so much a warning but informational

src/platforms/hosted/jlink.c
```diff
@@ -277,6 +234,313 @@ static bool jlink_claim_interface(void)
...
+	// Restoring the selected interface seemed to cause issues sometimes (unsure if culprit)
+	// so we don't do it for now, it didn't seem to be necessary, saving in case it proves to be necessary later
+	// if (selected_interface != interface) {
+	// 	/* If the selected interface didn't match the requested interface, restore the selected interface */
+	// 	DEBUG_WARN("Restoring selected interface to %s\n", jlink_interface_to_string(selected_interface));
+
+	// 	/* select the requested interface, we don't consider this a critical failure, as we got what we wanted */
+	// 	if (!jlink_select_interface(selected_interface))
+	// 		DEBUG_ERROR("Failed to restore selected interface to %s\n", jlink_interface_to_string(selected_interface));
+	// }
```
dragonmux, line 497: Same comment as before about `#if 0` blocks

src/platforms/hosted/jlink.c
```diff
@@ -303,61 +567,77 @@ bool jlink_init(void)
...
 		return false;
 	}
-	jlink_info();
+	if (!jlink_get_capabilities() || !jlink_get_version() || !jlink_get_interfaces()) {
+		DEBUG_ERROR("Failed to read J-Link information\n");
+		libusb_release_interface(info.usb_link->device_handle, info.usb_link->interface);
+		libusb_close(info.usb_link->device_handle);
+		return false;
+	}
+	strncpy(info.version, jlink.fw_version, sizeof(info.version));
```
dragonmux, line 576: Rather than using strncpy() which will write past the end of the used part of the version string with a filler character, please use memcpy() here as you already guarantee NUL character termination when reading the firmware version string.

```suggestion
	memcpy(info.version, jlink.fw_version, strlen(jlink.fw_version) + 1U);
```

dragonmux (changes requested):
Thank you for your excellent work on this. It's looking much better as a result - there are some times we've caught in review, but they shouldn't be complicated to sort. We look forward to merging this.

src/platforms/hosted/jlink_protocol.h
perigoso: Yep, unintended dupe

src/platforms/hosted/jlink.c
perigoso: Probably fairly straight forward, at least the raw protocol, but I believe it will be extremely unlikely that one is found in the wild, I also don't have one to test, and to pile on, I think older J-Links might not have some of the newer CMD_IO commands we are using

Perhaps we could add a warning when a single endpoint probe is detected, but I'm starting to burnout on J-Link code so I won't

src/platforms/hosted/jlink.c
perigoso: sure, sounds good, capabilities are always 32 bits, there are also the extended capabilities that we don't read right now, but those are also fixed size 256 bits

src/platforms/hosted/jlink.c
perigoso: I went for warning because I considered this to be a potentially problematic action, hence the commented out "restoring" code included, so far it didn't prove to be an issue so perhaps you're right

src/platforms/hosted/jlink.c
perigoso: good catch

src/platforms/hosted/jlink.c
dragonmux: No worries, this can be a later enhancement and a low priority one at that then.

src/platforms/hosted/jlink.c
dragonmux: Ahhh, that makes sense. Sure, this can be left as a warning then.

dragonmux (approved):

PR #1549, opened 2023-07-20
Feature: SWD error handling deduplication

perigoso:
## Detailed description

Noticed while working on the multi-drop handling, that a couple error handling routines could be replaced by the default error handling routine (J-Link and DAP), but this needs to be tested and validated to be true. 

### DIFF
--- a/src/platforms/hosted/dap_swd.c
+++ b/src/platforms/hosted/dap_swd.c
@@ -55,7 +55,6 @@ static void dap_swd_seq_out_parity(uint32_t tms_states, size_t clock_cycles);
 static bool dap_swd_configure(dap_swd_turnaround_cycles_e turnaround, dap_swd_fault_cfg_e fault_cfg);
 static bool dap_write_reg_no_check(uint16_t addr, uint32_t data);
 static uint32_t dap_read_reg_no_check(uint16_t addr);
-static uint32_t dap_swd_clear_error(adiv5_debug_port_s *target_dp, bool protocol_recovery);
 
 bool dap_swd_init(adiv5_debug_port_s *target_dp)
 {
@@ -85,7 +84,6 @@ bool dap_swd_init(adiv5_debug_port_s *target_dp)
 	/* Set up the accelerated SWD functions for basic target operations */
 	target_dp->read_no_check = dap_read_reg_no_check;
 	target_dp->dp_read = dap_dp_read_reg;
-	target_dp->error = dap_swd_clear_error;
 	target_dp->low_access = dap_dp_low_access;
 	target_dp->abort = dap_dp_abort;
 	return true;
@@ -176,22 +174,6 @@ static bool dap_swd_seq_in_parity(uint32_t *const result, const size_t clock_cyc
 	return !parity;
 }
 
-static void dap_line_reset(void)
-{
-	const uint8_t data[8] = {
-		0xffU,
-		0xffU,
-		0xffU,
-		0xffU,
-		0xffU,
-		0xffU,
-		0xffU,
-		0x0fU,
-	};
-	if (!perform_dap_swj_sequence(64, data))
-		DEBUG_ERROR("line reset failed\n");
-}
-
 static bool dap_write_reg_no_check(const uint16_t addr, const uint32_t data)
 {
 	DEBUG_PROBE("dap_write_reg_no_check %04x <- %08" PRIx32 "\n", addr, data);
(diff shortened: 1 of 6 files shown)


dragonmux:
Please rebase this on main so we can start looking into it

perigoso:
@dragonmux whats the best way of testing swd error handling? I guess the scan has to deal with a couple no-responses so that's a step

I do need help testing the DAP change in particular, as I don't have a probe that implements it

dragonmux:
We can test CMSIS-DAP changes, so that's fine.

Regarding testing SWD error handling, that's an excellent question.. we're going to have to dig through our available targets and try a couple of things like taking a Tiva-C and putting the probe call after the other Cortex-M4 probe routines to force bad accesses to happen and see what goes wrong as a result.

We (you, us, and other contributors) seem to have done such a relatively good job at implementing protocol recovery and error handling that intentionally making BMD trigger an error that uses these routines is quite difficult now.

src/target/adiv5.h
```diff
@@ -235,6 +235,7 @@ struct adiv5_debug_port {
 
 	/* dp_low_write returns true if no OK response, but ignores errors */
 	bool (*dp_low_write)(uint16_t addr, const uint32_t data);
+	uint32_t (*dp_low_read)(uint16_t addr);
```
dragonmux, line 238: Given this addition and the proper inclusion of call-through functions in the header, might we suggest fixing the naming while here - `dp_low_write` -> `write_no_check`, `dp_low_read` -> `read_no_check`

We've been itching for a good opportunity to do this and fix `low_access` which should be `raw_access`

src/target/adiv5_swd.c
```diff
@@ -153,6 +153,7 @@ bool adiv5_swd_scan(const uint32_t targetid)
 	}
 
 	dp->dp_low_write = firmware_dp_low_write;
+	dp->dp_low_read = firmware_dp_low_read;
 	dp->error = firmware_swdp_error;
```
dragonmux, line 157: Given this change set, we propose fixing the naming of the error function here inline with the names of the ones being removed - `adiv5_swd_clear_error`.

src/platforms/hosted/dap_swd.c
```diff
@@ -82,8 +82,9 @@ bool dap_swd_init(adiv5_debug_port_s *target_dp)
 	else
 		target_dp->dp_low_write = NULL;
 	/* Set up the accelerated SWD functions for basic target operations */
+	target_dp->dp_low_read = dap_read_reg_no_check;
 	target_dp->dp_read = dap_dp_read_reg;
-	target_dp->error = dap_swd_clear_error;
+	target_dp->error = firmware_swdp_error;
```
dragonmux, line 87: This line can be dropped as this function is called via `bmda_swd_dp_init()` which runs just after the DP is already set up in the base ADIv5 implementation.

src/platforms/hosted/jlink_swd.c
```diff
@@ -101,8 +95,9 @@ bool jlink_swd_init(adiv5_debug_port_s *dp)
 
 	/* Set up the accelerated SWD functions for basic target operations */
 	dp->dp_low_write = jlink_adiv5_raw_write_no_check;
+	dp->dp_low_read = jlink_adiv5_raw_read_no_check;
 	dp->dp_read = firmware_swdp_read;
-	dp->error = jlink_adiv5_clear_error;
+	dp->error = firmware_swdp_error;
```
dragonmux, line 100: This line can be dropped as this function is called via `bmda_swd_dp_init()` which runs just after the DP is already set up in the base ADIv5 implementation.

dragonmux (changes requested):
Sorry that it's taken quite so long to get to reviewing this properly. The idea of this change looks good, there are just a couple of nomenclature items we'd like to address so we don't loose the nomenclature work we did in the BMDA backends.

src/target/adiv5.h
perigoso: good suggestion, the naming was iching me too

src/target/adiv5_swd.c
perigoso: likewise good suggestion

src/platforms/hosted/dap_swd.c
perigoso: ah, good catch! looks like another method was being redundantly assigned in the jlink also

dragonmux (approved):
LGTM. Given the risks in merging this right before the release of v1.10, we'll schedule it for the start of the v2.0 merge window which should be open in the next ~month. Excellent job!

src/target/adiv5_swd.c
```diff
@@ -234,7 +235,7 @@ bool adiv5_swd_scan(const uint32_t targetid)
 		bool scan_multidrop = targetid || dp->version >= 2U;
 
 #if PC_HOSTED == 1
-	if (scan_multidrop && !dp->dp_low_write) {
+	if (scan_multidrop && !dp->write_no_check) {
 		DEBUG_WARN("Discovered multi-drop enabled target but CMSIS_DAP < v1.2 cannot handle multi-drop\n");
 		scan_multidrop = false;
```
dragonmux, line 240: Just realising that this might need to mark the DP as being "v1" only to inhibit the TARGETSEL logic in error recovery as that's what this is really all about iwrc.. need to very carefully think on how this code interacts with protocol and error recovery.

dragonmux (commented):
Good job updating this against PR #1629 and incorporating that fix. There are a couple of review notes that fell out of that, hopefully nothing too difficult to handle though.

src/target/adiv5_swd.c
perigoso: This particular detail has been iching me for a while, but I never dove into it, so I dont know what goes into it

src/remote.c
```diff
@@ -129,7 +129,7 @@ static void remote_respond_string(const char response_code, const char *const st
  * and remote_packet_process_jtag() for use by remote_packet_process_adiv5() so its faked AP can do the right
  * thing.
  *
- * REMOTE_INIT for SWD and JTAG rewrite the dp_low_write, dp_read, error, low_access and abort function
+ * REMOTE_INIT for SWD and JTAG rewrite the write_no_check, dp_read, error, low_access and abort function
```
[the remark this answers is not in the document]
perigoso: I don't follow, this PR does that rename already, was this a leftover draft review comment?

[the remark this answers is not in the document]
dragonmux: the replaced text only mentions `write_no_check`, while we're suggesting it says, verbatim, `{read,write}_no_check`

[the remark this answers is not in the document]
perigoso: Oh I see, it went completely over my head, sorry. Sure, I should've added read_no_check there to begin with but it was just a mechanical rename so I missed it.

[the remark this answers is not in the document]
dragonmux: No worries 🙂 better to ask and clarify!

src/target/adiv5_swd.c
perigoso: I did some reading, looks like the command used for the sequence in out was only [added in v1.2.0](https://www.keil.com/pack/doc/CMSIS/DAP/html/dap_revisionHistory.html), the command is [DAP_SWD_Sequence](https://www.keil.com/pack/doc/CMSIS/DAP/html/group__DAP__SWD__Sequence.html) and critically, it allows for a input sequence (SWDIO as input), which the previous existing command didn't ([DAP_SWJ_Sequence](https://www.keil.com/pack/doc/CMSIS/DAP/html/group__DAP__SWJ__Sequence.html)), this is required for the target select, which does indeed affect the protocol recovery along with the multi-drop. 

What it looks like to me as that the current implementation is wholesale broken for v < 1.2, as it generally makes use of the sequence command which isn't available, I may be missing some parts though.

As an extra note, I do wonder if its possible to replicate a input sequence with a sequence of [DAP_SWJ_Pins](https://www.keil.com/pack/doc/CMSIS/DAP/html/group__DAP__SWJ__Pins.html), the documentation does not make the behaviour clear (when are the pins configured as inputs, if ever)

dragonmux: We're going to suggest this should be a separate PR then that focuses on fixing the CMSIS-DAP implementation's handling of the protocol version and we'll leave this item unresolved until a PR is made but merge this PR anyway.

dragonmux (approved):

PR #1550, opened 2023-07-20
platforms/f4discovery: Update README.md

franzflasch:
Nothing fancy here. This just updates the README.md of the f4discovery and adds the missing UART pins, which are actually there just not documented. 

Tested on my f4 discovery board. Works just fine.


dragonmux (approved):

PR #1551, opened 2023-07-20
Fix: ST-Link v2 serial number readout

dragonmux:
## Detailed description

The original ST-Link v2 has a firmware bug that remains unfixed which means that instead of the serial number being properly hex encoded UTF-16 returned to the host, the serial number is returned raw as a series of uint16_t's that must then be converted to hex for display and which must bypass and side-step libusb's libusb_get_string_descriptor_ascii() function due to it converting any character with numerical value >= 0x80 to '?'.

This PR addresses this long standing problem, allowing the serial number matching mechanics to work properly with ST-Link v2 serial numbers. This builds on the new probe scanning engine which already stopped us from displaying garbage on the user's console.


esden (approved):
**Seesh** ... what a nasty STLink bug... 🤦 ... thanks. LGTM :D

Issue #959, opened 2021-12-22
Hosted/CMSIS-DAP: CMSID-DAP Interface can come after a CDCACM interface


UweBonnes:
BMP/Hosted with a device like described in https://sourceforge.net/p/openocd/tickets/327/ would probably try to open the wrong port.

PR #1502, opened 2023-05-31
Improved BMDA probe scanning

sidprice:
This PR is a complete rewrite of the BMDA probe scanner and also includes an implementation for FTDI-based probes on Windows. The latter does not require and special driver installation.

* [x] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md)
* [x] I've read the [guidelines for contributing](https://github.com/blackmagic-debug/blackmagic/blob/main/CONTRIBUTING.md) to this repository
* [x] It builds for hardware native (`make PROBE_HOST=native`)
* [x] It builds as BMDA (`make PROBE_HOST=hosted`)
* [x] I've tested it to the best of my ability
* [x] My commit messages provide a useful short description of what the commits do

PR #1552, opened 2023-07-20
Fix: BMDA CMSIS-DAP v2 selection

dragonmux:
## Detailed description

This PR addresses a small regression introduced in https://github.com/blackmagic-debug/blackmagic/pull/1502 which causes CMSIS-DAP probes to be assumed to be v1 even when they might be v2 or support both.

We fix this by re-checking the descriptors of the probe if it's a CMSIS-DAP probe, once it's been selected by the probe discovery code. We then check for the v2 descriptors and extract the correct endpoints to allow the backend to do the right thing.

The way this is done also ensures that we should always pick up the proper interfaces no matter what order they appear in.

## Closing issues

Fixes https://github.com/blackmagic-debug/blackmagic/issues/959


esden (approved):
LGTM

PR #1531, opened 2023-07-07
Fix: JTAG bitbanging-with-delay logic

dragonmux:
## Detailed description

This PR does a couple of things. The first, and most major, thing it addresses is a bug in the delayed versions of the JTAG bitbanging routines that ALTracer kindly reported on Discord.

Additionally, it addresses some cleanup that needed doing in the remote protocol implementation for the JTAG bitbanging routines, and the way that delay timing is calculated and handled as this was all quite wrong and out of whack.

The results have been tested on the LPC4370 and the GD32VF103 and show both working properly.

## Your checklist for this pull request

* [x] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md)
* [x] I've read the [guidelines for contributing](https://github.com/blackmagic-debug/blackmagic/blob/main/CONTRIBUTING.md) to this repository
* [x] It builds for hardware native (`make PROBE_HOST=native`)
* [x] It builds as BMDA (`make PROBE_HOST=hosted`)
* [x] I've tested it to the best of my ability
* [x] My commit messages provide a useful short description of what the commits do

PR #1553, opened 2023-07-20
Fix: remote protocol SWD hangs

dragonmux:
## Detailed description

This PR addresses a regression introduced in PR #1531 which forgot that the SWD bitbanging code needs to deal with target_clk_divider = UINT32_MAX fully. The effect of this mistake was that SWD would "hang" and soft-crash the probe.

We've also taken the opportunity to clean up the SWD remote protocol implementation the same as we did the JTAG implementation, saving a couple of bytes of Flash from that routine.


esden (approved):
LGTM

PR #1560, opened 2023-07-22
Fix: threads-induced IDE hangups

dragonmux:
## Detailed description

Back in v1.8.2 we introduced fake threads support to allow modern GDBs to work with the project. As part of that we had to report to GDB that there's always and forever one thread to match the 1 core no OS reality of debugging bare-metal. Unfortunately a consequence of that was that IDEs that don't properly gate checking for threads against being presently attached to a target would then ask, after detaching, if there were any threads, get told yes, and then go sulk due to their impaired logic.

This PR addresses this by gating our telling GDB there are threads, behind being attached to a target. This means we report an empty thread list when detached, preventing the logic issue in those IDEs. It technically also means we're being more correct in our handling of the request which is why we've marked it as a bug in BMD, but it is papering over a bug in particularly Eclipse-based IDEs like STM32CubeIDE, Eclipse Embedded CDT and AT32IDE.

### DIFF
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -183,7 +183,8 @@ int gdb_main_loop(target_controller_s *tc, char *pbuf, size_t pbuf_size, size_t 
 			gdb_putpacketz("OK");
 		break;
 	}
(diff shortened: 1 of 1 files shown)


esden (approved):
LGTM

PR #1565, opened 2023-07-25
Fix: GCC -fanalyzer lints

dragonmux:
## Detailed description

This PR came about by trying `-fanalyzer` on BMDA and seeing what fell out. In this PR we fix several missing pieces of error handling around allocations, and some logic errors the analyser found resulting in buggy behaviour on a couple of targets and some UB.

While we have not been able to test all of the results, this should be a low-risk set of changes, and definitely solves an issue seen with LPC43x0 targets.

We did not address /all/ lints this compiler option produces yet as a couple of the lints being produced are rather complicated to address. This means there will be a follow-up PR at some future point to look into the remaining lints.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@ blackmagic_upgrade
 src/artifacts/
 
 src/ftd2xx.dll
+
+scripts/swolisten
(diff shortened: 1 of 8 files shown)


esden (approved):
LGTM

PR #1568, opened 2023-07-28
Fix: scan return type + swdp naming phase out

perigoso:
## Detailed description

Followup to #1184 

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -212,13 +212,13 @@ static bool cmd_jtag_scan(target_s *target, int argc, const char **argv)
 	if (connect_assert_nrst)
 		platform_nrst_set_val(true); /* will be deasserted after attach */
 
-	uint32_t devs = 0;
+	bool scan_result = false;
 	volatile exception_s e;
 	TRY_CATCH (e, EXCEPTION_ALL) {
 #if PC_HOSTED == 1
-		devs = bmda_jtag_scan();
+		scan_result = bmda_jtag_scan();
 #else
-		devs = jtag_scan();
+		scan_result = jtag_scan();
 #endif
 	}
 	switch (e.type) {
@@ -230,7 +230,7 @@ static bool cmd_jtag_scan(target_s *target, int argc, const char **argv)
 		break;
 	}
 
-	if (devs == 0) {
+	if (!scan_result) {
 		platform_target_clk_output_enable(false);
 		platform_nrst_set_val(false);
 		gdb_out("JTAG device scan failed!\n");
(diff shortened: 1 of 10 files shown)


src/command.c
```diff
@@ -85,7 +85,7 @@ const command_s cmd_list[] = {
 	{"version", cmd_version, "Display firmware version info"},
 	{"help", cmd_help, "Display help for monitor commands"},
 	{"jtag_scan", cmd_jtag_scan, "Scan JTAG chain for devices"},
-	{"swdp_scan", cmd_swdp_scan, "Scan SW-DP for devices: [TARGET_ID]"},
+	{"swd_scan", cmd_swd_scan, "Scan SWD interface for devices: [TARGET_ID]"},
```
dragonmux, line 88: For compatibility, we need to either keep the command named the same thing or implement a shadow version of the command named "swdp_scan" as otherwise we'll break scripts and automated workflows.

We personally agree with this change but need to think about the release cycle and how to do a deprecation here.

src/include/target.h
```diff
@@ -35,11 +35,11 @@ typedef uint32_t target_addr_t;
 typedef struct target_controller target_controller_s;
 
 #if PC_HOSTED == 1
-uint32_t bmp_swd_scan(uint32_t targetid);
-uint32_t bmda_jtag_scan(void);
+bool bmp_swd_scan(uint32_t targetid);
```
dragonmux, line 38: This  should probably be renamed to `bmda_swd_scan()` for consistency

dragonmux (changes requested):
This is looking good. There are only two items we spotted in review that need addressing and then this should be good to merge.

src/command.c
perigoso: Ah, right! I hadn't thought about that...

src/include/target.h
perigoso: I did wonder why it was named like this

src/command.c
perigoso: I added a simple alias with "Deprecated" noted on the description, ideally we might want a "deeper" alias that also prints a warning

We can leave this out of this PR, let me know

dragonmux: Think that works nicely till we figure out a better way to hang onto and implement deprecated commands so they don't get advertised in the help output but do still work (and print a warning on use).

dragonmux (approved):
LGTM, merging. Nothing super scary here, just a lot of mechanical renaming and return type cleanup.

PR #1577, opened 2023-07-29
Fix: ADIv5 debug bringup reliability

dragonmux:
## Detailed description

During the process of building target support for the Ambiq Apollo 3, Sid wound up hitting and issue where the debug reset code in the ADIv5 implementation would cause the target to quit talking SWD or JTAG till power cycled, preventing scan from working.

After working with zyp, mubes and Sid and reading the ADIv5 spec closely, we have come to the conclusion that we can get the effect of the reset sequence logic by instead powering down the debug engine on the target, then powering it back up. This is guaranteed by the spec for targets that actually implement power control on the APs and debug bus to cause the APs and debug bus to come back up via a reset, getting the same net effect as the prior logic.

This has been tested on a couple of different targets including the STM32F411, LPC4370 and Apollo 3, and appears to be reliable, with the AP bringup code mopping up the only potential stale values this can cause for targets that don't implement power control.

### DIFF
--- a/src/target/adiv5.c
+++ b/src/target/adiv5.c
@@ -714,7 +714,7 @@ adiv5_access_port_s *adiv5_new_ap(adiv5_debug_port_s *dp, uint8_t apsel)
 		return NULL;
 	tmpap.csw = adiv5_ap_read(&tmpap, ADIV5_AP_CSW);
 	// XXX: We might be able to use the type field in ap->idr to determine if the AP supports TrustZone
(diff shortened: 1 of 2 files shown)


esden (approved):
Another instance where I would love to have our HITL CI handy... it is impossible to know if it will work 100% on all targets the same way. But adhering to the "optimistic merging policy" let's merge it so that more people have a chance to test it before the next release.

dragonmux:
Agreed, so far it's been fine across the targets we've tried, but our desk only has so many..

PR #1584, opened 2023-08-02
Fix: ADIv5 DPv0 ID Codes

dragonmux:
## Detailed description

In this PR we address the STM32MP15x having a JTAG ID of 0x6ba00477 indicating a DPv0 TAP, and an ADIv5 non-compliance issue the part has which results in it hanging in the new power down code in adiv5_dp_init().

The latter is caused by the control and status register hanging at the value 0xa0000000 forever contrary to how the acknowledgement bits are supposed to work according to the ADIv5 spec.

@Perigoso given your experience with ADIv5, please check our logic regarding the ID code part of this fix.

### DIFF
--- a/src/target/adiv5.c
+++ b/src/target/adiv5.c
@@ -813,7 +813,7 @@ void adiv5_dp_init(adiv5_debug_port_s *const dp, const uint32_t idcode)
 	 *
 	 * for SWD-DP, we are guaranteed to be DP v1 or later.
 	 */
-	if (idcode != JTAG_IDCODE_ARM_DPv0) {
+	if ((idcode & (JTAG_IDCODE_DESIGNER_MASK | JTAG_IDCODE_PARTNO_MASK)) != JTAG_IDCODE_ARM_DPv0) {
 		const uint32_t dpidr = adiv5_dp_read_dpidr(dp);
 		if (!dpidr) {
 			DEBUG_ERROR("Failed to read DPIDR\n");
--- a/src/target/adiv5.c
+++ b/src/target/adiv5.c
@@ -805,15 +805,37 @@ void adiv5_dp_init(adiv5_debug_port_s *const dp, const uint32_t idcode)
 	bmda_adiv5_dp_init(dp);
 #endif
 
+	/* Check if we have a JTAG ID code and that it's a valid one */
+	if (idcode & 1U) {
+		/*
+		 * We do, so this is a JTAG DP. Let's pull out the designer and part codes then.
+		 *
+		 * Start by pulling out the designer code which will be used to attempt to
+		 * detect a DPv0 DP. This will get overriden later by DPIDR if the DP turns out
+		 * to be DPv1+.
+		 */
+		const uint16_t designer = (idcode & JTAG_IDCODE_DESIGNER_MASK) >> JTAG_IDCODE_DESIGNER_OFFSET;
+		/*
+		 * Now extract the part number and sort out the designer code.
+		 * The JTAG ID code designer is in the form:
+		 * Bits 10:7 - JEP-106 Continuation Code
+		 * Bits 6:0 - JEP-106 Identity Code
+		 * So here we convert that into our internal representation.
+		 * See the JEP-106 code list (jep106.h) for more on that.
+		 */
+		dp->designer_code =
+			((designer & ADIV5_DP_DESIGNER_JEP106_CONT_MASK) << 1U) | (designer & ADIV5_DP_DESIGNER_JEP106_CODE_MASK);
+		dp->partno = (idcode & JTAG_IDCODE_PARTNO_MASK) >> JTAG_IDCODE_PARTNO_OFFSET;
+	}
+
 	/*
 	 * Start by assuming DP v1 or later.
 	 * this may not be true for JTAG-DP (we attempt to detect this with the part ID code)
-	 * in such cases (DPv0) DPIDR is not implemented
-	 * and reads are UNPREDICTABLE.
+	 * in such cases (DPv0) DPIDR is not implemented and reads are UNPREDICTABLE.
 	 *
 	 * for SWD-DP, we are guaranteed to be DP v1 or later.
 	 */
-	if ((idcode & (JTAG_IDCODE_DESIGNER_MASK | JTAG_IDCODE_PARTNO_MASK)) != JTAG_IDCODE_ARM_DPv0) {
+	if (dp->designer_code != JEP106_MANUFACTURER_ARM || dp->partno != JTAG_IDCODE_PARTNO_DPv0) {
 		const uint32_t dpidr = adiv5_dp_read_dpidr(dp);
 		if (!dpidr) {
 			DEBUG_ERROR("Failed to read DPIDR\n");
--- a/src/target/adiv5.c
+++ b/src/target/adiv5.c
@@ -791,7 +791,7 @@ uint32_t adiv5_dp_read_dpidr(adiv5_debug_port_s *const dp)
 	return dpidr;
 }
 
-void adiv5_dp_init(adiv5_debug_port_s *const dp, const uint32_t idcode)
+void adiv5_dp_init(adiv5_debug_port_s *const dp)
 {
 	/*
 	 * We have to initialse the DP routines up front before any adiv5_* functions are called or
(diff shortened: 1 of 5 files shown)


src/target/adiv5.c
perigoso, line 1: So while reviewing this and looking at the `IDCODE` macros, i questioned, why don't we parse the `IDCODE` designer formally, like we do for `DPIDR` and `TARGETID`, in the context of ADIv5 the `IDCODE` should have the designer and it is in the same form as `DPIDR`/`TARGETID`, I noted this when I rewrote the ADIv5 definitions initially, and added `IDCODE` macros to parse it too, just did not implement it at the time because there was no use for it. But with the check for DPv0 there is.

I checked and the conversion of `JTAG_IDCODE_ARM_DPv0` does map to `JEP106_MANUFACTURER_ARM` and `partno`  `0x4BA00` which is exactly the values we expect according to the spec. Note that we ignore the version field, it is implementation defined and does not map to the DP version or anything useful, that was the crux of the initial problem with `DPv0`.

```c
#define JTAG_IDCODE_DPv0_PARTNO 0x4BA00U
```

```c
void adiv5_dp_init(adiv5_debug_port_s *const dp, const uint32_t idcode)
{
...
	/* This doubles as a zero check, and bit 1 set as per ADIv5 spec */
	if (idcode & 1U) {
		/*
		 * This is a JTAG-DP
		 *
		 * Use the IDCODE to determine the designer and part number of the DP
		 * we will use it to try detect DPv0, this will be overriden later by
		 * the DPIDR if we are on a DPv1+ 
		 */
		const uint16_t designer = (idcode & JTAG_IDCODE_DESIGNER_MASK) >> JTAG_IDCODE_DESIGNER_OFFSET;
		/* 
		 * IDCODE DESIGNER is in the form
		 * Bits 10:7 - JEP-106 Continuation code
		 * Bits 6:0 - JEP-106 Identity code
		 * here we convert it to our internal representation, See JEP-106 code list
		 */
		dp->designer_code =
			(designer & ADIV5_DP_DESIGNER_JEP106_CONT_MASK) << 1U | (designer & ADIV5_DP_DESIGNER_JEP106_CODE_MASK);
		dp->partno = (idcode & JTAG_IDCODE_PARTNO_MASK) >> JTAG_IDCODE_PARTNO_OFFSET;
	}

	/*
	 * Start by assuming DP v1 or later.
	 * this may not be true for JTAG-DP (we attempt to detect this with the part ID code)
	 * in such cases (DPv0) DPIDR is not implemented
	 * and reads are UNPREDICTABLE.
	 *
	 * for SWD-DP, we are guaranteed to be DP v1 or later.
	 */
	if (dp->designer_code == JEP106_MANUFACTURER_ARM || dp->partno == JTAG_IDCODE_DPv0_PARTNO) {
...
}
```

Two additional notes:
- With this approach we might want to not override `designer_code` and `partno` immediately, only after we checked that the `DPIDR` is valid, right now if we find the `DPIDR` to not be valid we do a last ditch effort, assume it's a `DPv0` and clear those variables out
- This one is not related to this PR but to the last cleanup I submited to `adiv5_dp_init`, the dp->version == 0 check in else case is redundant, if we detect the `DPv0` based on `IDCODE` we know it is `DPv0` and nothing before this if sets that value.

perigoso (changes requested):

src/target/adiv5.c
```diff
@@ -813,7 +813,7 @@ void adiv5_dp_init(adiv5_debug_port_s *const dp, const uint32_t idcode)
 	 *
 	 * for SWD-DP, we are guaranteed to be DP v1 or later.
 	 */
-	if (idcode != JTAG_IDCODE_ARM_DPv0) {
+	if ((idcode & JTAG_IDCODE_ARM_MASK) != JTAG_IDCODE_ARM_DPv0) {
```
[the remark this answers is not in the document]
dragonmux: It's not even that it's more pedantically correct, it's more clear what we're doing here - `~JTAG_IDCODE_VERSION_MASK` is less code, but hides that what we're doing is masking to oget the designer and part number components only. Going with the suggested change.

src/target/adiv5.c
dragonmux: Well, this certainly isn't going to be a difficult change to look at incorporating in some form, so we'll look at doing this tomorrow as part of this PR.

dragonmux: Please give 10d5848d a look and let us know what you think. Tested against a Tiva-C as we know this has a DPv0 and tested against the LPC4370 as a target with DPv1's. It also seems to clean up scan for STM32H5's.

ALTracer:
STM32MP15x SoC has a JTAG IDCODE `0x6ba00477` and SWD IDCODE `0x6ba02477`, please reflect this information in issue description. This is supported both by RM0436/RM0442/RM0441 and by scanning two real hardware boards.

dragonmux:
Corrected the description JTAG ID code, but the SWD one does not matter for this PR so we aren't going to try working that in.

src/target/adiv5_jtag.c
```diff
@@ -57,7 +57,8 @@ void adiv5_jtag_dp_handler(const uint8_t dev_index)
 	bmda_jtag_dp_init(dp);
 #endif
 
-	if (jtag_devs[dev_index].jd_idcode == JTAG_IDCODE_ARM_DPv0)
+	if ((jtag_devs[dev_index].jd_idcode & (JTAG_IDCODE_PARTNO_MASK | JTAG_IDCODE_DESIGNER_MASK)) ==
```
perigoso, line 60: Given that we have a routine here specific to jtag adiv5, and it has access to the dp struct, since it's created here, we could do the idcode parsing here, that way we could avoid the duplicate macro we have here for DPv0, and would not need to pass idcode to adiv5_dp_init at all anymore.

I'm not sure that would make the code cleaner or more obfuscated, so I'm leaving this not as a suggestion but as a note.

perigoso (approved):
LGTM

src/target/adiv5_jtag.c
dragonmux: That's a fair point. Perhaps that's an enhancement we perform after this PR is merged and battle tested a bit.

dragonmux: As this PR has been tested by various people hitting one of the couple of different bugs it fixes, we're more confident that the code here doesn't create new regressions so we'll work on applying this refactor as suggested.

src/target/adiv5_jtag.c
```diff
@@ -57,11 +57,36 @@ void adiv5_jtag_dp_handler(const uint8_t dev_index)
 	bmda_jtag_dp_init(dp);
 #endif
 
-	if (jtag_devs[dev_index].jd_idcode == JTAG_IDCODE_ARM_DPv0)
+	const uint32_t idcode = jtag_devs[dev_index].jd_idcode;
+
+	/* Check if we have a JTAG ID code and that it's a valid one */
+	if (idcode & 1U) {
+		/*
+		 * We do, so this is a JTAG DP. Let's pull out the designer and part codes then.
```
perigoso, line 65: In this context this part of the comment looses it's purpose, we already knew it's a JTAG DP

src/target/adiv5_jtag.c
```diff
@@ -57,11 +57,36 @@ void adiv5_jtag_dp_handler(const uint8_t dev_index)
...
+		 * Bits 6:0 - JEP-106 Identity Code
+		 * So here we convert that into our internal representation.
+		 * See the JEP-106 code list (jep106.h) for more on that.
+		 */
+		dp->designer_code =
+			((designer & ADIV5_DP_DESIGNER_JEP106_CONT_MASK) << 1U) | (designer & ADIV5_DP_DESIGNER_JEP106_CODE_MASK);
+		dp->partno = (idcode & JTAG_IDCODE_PARTNO_MASK) >> JTAG_IDCODE_PARTNO_OFFSET;
+	}
+
+	if (dp->partno == JTAG_IDCODE_PARTNO_DPv0)
```
perigoso, line 85: Is not checking for for ARM designer code here intentional? Is the JTAG DP guaranteed to be always ARM in this context?

src/target/adiv5.h
```diff
@@ -208,7 +208,8 @@
 #define JTAG_IDCODE_DESIGNER_OFFSET 1U
 #define JTAG_IDCODE_DESIGNER_MASK   (0x7ffU << JTAG_IDCODE_DESIGNER_OFFSET)
 
-#define JTAG_IDCODE_ARM_DPv0 UINT32_C(0x4ba00477)
+#define JTAG_IDCODE_ARM_DPv0    UINT32_C(0x0ba00476)
```
perigoso, line 211: This macro is not used after the changes, I think, so we can remove it as it's redundant

src/target/adiv5_jtag.c
dragonmux: Yes, thanks to the JTAG devices table requiring the ARM manufacturer code to call this function - https://github.com/blackmagic-debug/blackmagic/blob/14f28821bc699bfe9195dd68e13d2abb89aa6fcc/src/target/jtag_devs.c#L28-L35

That means, for now, the ARM part is checked by the scan code prior to entry here - handily!

perigoso (approved):
LGTM, good cleanup!

esden (approved):
LGTM

PR #1588, opened 2023-08-03
Fix morse new null msg

jcamdr:
## Detailed description
Discord 2023-08-02
jcamdr: I can't understand how a morse(NULL, false) can work without dereferencing a NULL pointer inside morse_update() after a morse() with a string.
dragonmux: good catch, that's because it does dereference NULL under that circumstance
jcamdr: Is there any possible atomic issue with morse_msg and morse_index between morse() and morse_update() on some plateform?
dragonmux: technically, yes. morse() should update those 3 variables in exactly the opposite order to how they are currently
that is, morse_repeat, then msg_index then morse_msg. that'll guarantee the update, while not atomic, will apply correctly and do sensible things
jcamdr: Fully agree. All 3 variables probably need to be volatile AFAIK.

### DIFF
--- a/src/include/morse.h
+++ b/src/include/morse.h
@@ -23,7 +23,7 @@
 
 #include <stdbool.h>
 
-extern const char *morse_msg;
+extern volatile const char *morse_msg;
 
 void morse(const char *msg, bool repeat);
 bool morse_update(void);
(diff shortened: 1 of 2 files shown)


dragonmux:
From a code stand-point this looks good. Please update your commit messages to conform to the contribution requirements - `src/morse.c:` -> `morse:` - both `src/` and the `.c` extension should be dropped from the prefix

jcamdr:
Commit messages updated, sorry for overhead.

dragonmux (approved):

No worries re the commit messages, it happens 🙂

Issue #1590, opened 2023-08-10
Enabling extra compile warnings


perigoso:
In the process of working on #1528 the new build-system capabilities were leveraged to enable extra warnings.
This issue lists all desired future extra warnings and tracks whether they pass with the current state of the code-base, along with respective cleanups where needed.

Warnings:
- [ ] [arith-conversion](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Warith-conversion)
- [ ] [bad-function-cast](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wbad-function-cast)
- [x] [cast-align=strict](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wcast-align)
- [x] [cast-function-type](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wcast-function-type)
- [x] [cast-qual](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wcast-qual)
- [ ] [conversion](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wconversion)
- [x] [dangling-else](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wdangling-else)
- [ ] [double-promotion](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wdouble-promotion)
- [x] [duplicated-branches](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wduplicated-branches)
- [x] [float-conversion](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wfloat-conversion)
-Wformat=2 -Wformat-overflow=2 -Wformat-signedness -Wformat-truncation
- [ ] [format](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wformat)
	- [format-overflow=2](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wformat-overflow)
	- [format-signedness](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wformat-signedness)
	- [format-truncation](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wformat-truncation)
[...]

PR #1592, opened 2023-08-10
Fix: warning format

perigoso:
## Detailed description

Fix some warnings for `-Wformat=2 -Wformat-overflow=2 -Wformat-signedness -Wformat-truncation`

See #1590 for context

### DIFF
--- a/src/gdb_hostio.c
+++ b/src/gdb_hostio.c
@@ -28,6 +28,22 @@ int hostio_reply(target_controller_s *const tc, char *const pbuf, const int len)
 {
 	(void)len;
 
+	/* 
+	 * File-I/O Remote Protocol Extension
+	 * See https://sourceware.org/gdb/onlinedocs/gdb/Protocol-Basics.html#Protocol-Basics
+	 * 
+	 * This handles the F Reply Packet, sent by GDB after handling the File-I/O Request Packet.
+	 * 
+	 * The F reply packet consists of the following:
+	 * 
+	 * - retcode, the return code of the system call as hexadecimal value.
+	 * - errno, the errno set by the call, in protocol-specific representation.
+	 * 		Can be omitted if the call was successful.
+	 * - Ctrl-C flag, sent only if user requested a break.
+	 * 		In this case, errno must be sent as well, even if the call was successful.
+	 * 		The Ctrl-C flag itself consists of the character ‘C’: 
+	 */
+
 	const bool retcode_is_negative = pbuf[1U] == '-';
 
 	unsigned int retcode;
--- a/src/gdb_hostio.c
+++ b/src/gdb_hostio.c
@@ -46,16 +46,26 @@ int hostio_reply(target_controller_s *const tc, char *const pbuf, const int len)
 
 	const bool retcode_is_negative = pbuf[1U] == '-';
 
-	unsigned int retcode;
-	unsigned int errno_;
-	char ctrl_c_flag;
-	const unsigned int items =
-		sscanf(pbuf + (retcode_is_negative ? 2U : 1U), "%x,%x,%c", &retcode, &errno_, &ctrl_c_flag);
+	unsigned int retcode = 0;
+	unsigned int errno_ = 0;
+	char ctrl_c_flag = '\0';
+	const int items = sscanf(pbuf + (retcode_is_negative ? 2U : 1U), "%x,%x,%c", &retcode, &errno_, &ctrl_c_flag);
 
-	/* if break is requested */
-	tc->interrupted = items == 3U && ctrl_c_flag == 'C';
+	if (items < 1) {
+		/* 
+		 * Something went wrong with the sscanf or the packet format, avoid UB
+		 * FIXME: how do we properly handle this?
+		 */
+		tc->interrupted = false;
+		tc->errno_ = TARGET_EUNKNOWN;
+		return -1;
+	}
 
-	tc->errno_ = errno_;
+	/* If the call was successful the errno may be omitted */
+	tc->errno_ = items >= 2 ? errno_ : 0;
+
+	/* If break is requested */
+	tc->interrupted = items == 3 && ctrl_c_flag == 'C';
 
 	return retcode_is_negative ? -retcode : retcode;
 }
--- a/src/gdb_hostio.c
+++ b/src/gdb_hostio.c
@@ -24,24 +24,24 @@
 #include "gdb_hostio.h"
 #include "gdb_packet.h"
 
-int hostio_reply(target_controller_s *tc, char *pbuf, int len)
+int hostio_reply(target_controller_s *const tc, char *const pbuf, const int len)
 {
 	(void)len;
-	int retcode, items, errno_;
-	char c, *p;
-	if (pbuf[1] == '-')
-		p = &pbuf[2];
-	else
-		p = &pbuf[1];
-	items = sscanf(p, "%x,%x,%c", &retcode, &errno_, &c);
-	if (pbuf[1] == '-')
-		retcode = -retcode;
+
+	const bool retcode_is_negative = pbuf[1U] == '-';
+
+	unsigned int retcode;
+	unsigned int errno_;
+	char ctrl_c_flag;
+	const unsigned int items =
+		sscanf(pbuf + (retcode_is_negative ? 2U : 1U), "%x,%x,%c", &retcode, &errno_, &ctrl_c_flag);
 
 	/* if break is requested */
-	tc->interrupted = items == 3 && c == 'C';
+	tc->interrupted = items == 3U && ctrl_c_flag == 'C';
+
 	tc->errno_ = errno_;
 
-	return retcode;
+	return retcode_is_negative ? -retcode : retcode;
 }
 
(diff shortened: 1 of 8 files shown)


src/gdb_hostio.c
```diff
@@ -24,24 +24,22 @@
...
-		p = &pbuf[1];
-	items = sscanf(p, "%x,%x,%c", &retcode, &errno_, &c);
-	if (pbuf[1] == '-')
-		retcode = -retcode;
+
+	const bool retcode_is_negative = pbuf[1] == '-';
+
+	unsigned int retcode;
+	unsigned int errno_;
+	char c;
```
dragonmux, line 35: What is `c` do we know and can we get it a better name?

src/gdb_hostio.c
```diff
@@ -24,24 +24,22 @@
...
+	const unsigned int items = sscanf(&pbuf[retcode_is_negative ? 2U : 1U], "%x,%x,%c", &retcode, &errno_, &c);
```
dragonmux, line 36: sscanf() returns `int`, unfortunately. It would be best if the `items` variable was made `int` and checked for being `< 1` to handle sscanf() failures. The post-scanf() code that uses the variables scanned should then also check that items is a high enough value otherwise these are UB uninit variable reads and probably also clobber values with new values that don't make sense. Initialising the values on line 33-35 would help with this, but properly checking before blind assignment seems like a good improvement.

We like the p -> pbuf change but would use `pbuf + (retcode_is_negative ? 2U : 1U)` to make it more clear that you're offsetting the pbuf pointer only here and not doing some additional trickery. This is technically correct though.

src/platforms/hosted/cmsis_dap.c
```diff
@@ -477,7 +477,7 @@ static void dap_mem_read(adiv5_access_port_s *ap, void *dest, uint32_t src, size
 	if (len == 0)
 		return;
 	align_e align = MIN(ALIGNOF(src), ALIGNOF(len));
-	DEBUG_WIRE("dap_mem_read @ %" PRIx32 " len %zu, align %d\n", src, len, align);
+	DEBUG_WIRE("dap_mem_read @ %" PRIx32 " len %zu, align %u\n", src, len, align);
```
dragonmux, line 480: We are surprised that enums are being treated as unsigned here, [because we thought they were supposed to be `int` for underlying type](https://en.cppreference.com/w/c/language/enum#Explanation) until C23.

We are hesitant to take the %d -> %u changes on enums blindly and instead would suggest finding a compiler flag that enforces what the standard says should happen here so we get consistency across compilers (this is otherwise likely something where compiler A says `unsigned` and B says `int` and it becomes a mess).

dragonmux (changes requested):
This is looking good, and particularly the fixes and cleanup in the HostIO code. It'll be nice to eliminate some more UB and get things better sorted out.

There are some changes we'd like to see made in the hostio_reply() code though and a query over format string changes involving enums. Hopefully we can find a suitable solution with the latter though.

src/gdb_hostio.c
perigoso: I did not know what it was, had to go through GDB docs
https://sourceware.org/gdb/onlinedocs/gdb/Protocol-Basics.html#Protocol-Basics
hopefully I got it right

dragonmux: Looks great, and the addition of the block comment is extremely helpful, thank you!

src/platforms/hosted/cmsis_dap.c
perigoso: Valid concern

The thing is enum types can be int type but they are not int in C. On gcc 1), enum types are unsigned int by default.
enum constants are int but enum types are implementation defined.
In your case the enum constant is int but you are giving it a value that does not fit in a int. You cannot have unsigned int enum constants in C as C says they are int.

I had to do some reading, from what I gathered, the type for `enum` is not that well defined, in C99, `enum` *constants* are int (C99 §6.4.4.3), but `enum` *types* are implementation defined (C99 §6.7.2.2/4), GCC defaults to `unsigned int` unless there are signed entries in the `enum` (https://gcc.gnu.org/onlinedocs/gcc/Structures-unions-enumerations-and-bit-fields-implementation.html), I didn't find any useful flags in this regard, but I'm looking at possible workarounds

perigoso: You raise a valid concern...

I had to do some reading, from what I gathered, the type for `enum` is not that well defined, in C99, `enum` *constants* are int (C99 §6.4.4.3), but `enum` *types* are implementation defined (C99 §6.7.2.2/4), GCC defaults to `unsigned int` unless there are signed entries in the `enum` (https://gcc.gnu.org/onlinedocs/gcc/Structures-unions-enumerations-and-bit-fields-implementation.html), I didn't find any useful flags in this regard, but I'm looking at possible workarounds

dragonmux: No worries. Ok, given that we should probably check and see what both MSVC and Clang say on the subject (the former in particular because of Clang-CL which does what MSVC does rather than what Clang normally does) to try and pick out a path that will work into the future and not have us stumble on this problem again.

Then document what we choose and why - making an entry in HACKING.md for example - so we have a reference to go back to and a reasoning for other contributors to read for why we enforce whatever we end up enforcing.

perigoso: Enum changes were reverted for now, added a note to HACKING.md about format specifiers, this can be addressed in a future PR

perigoso:
Reverted changes to enum type format specifiers, given the discussion in the review comments, for now let's reduce the scope of this PR and address the enum type issue separately

src/target/HACKING.md
```diff
@@ -85,3 +86,34 @@ typedef struct lpc43xx_flash {
 	/* contents */
 } lpc43xx_flash_s;
 ```
+
+## Print format specifiers
+
+When writing format specifiers for `printf` and friends care should be taken to ensure they match the variable
+underlying type, taking into consideration that it may change depending on the compiler, to this end some
+standard macros are available to help
```
amyspark, line 94:
```suggestion
When writing format specifiers for `printf` and friends, care should be taken to ensure they match the variables'
underlying types, taking into consideration that they may change depending on the compiler. To this end, some
standard macros are available to help.
```

src/target/HACKING.md
```diff
@@ -85,3 +86,34 @@ typedef struct lpc43xx_flash {
...
+
+Non-exhaustive cheat sheet:
```
amyspark, line 96:
```suggestion
### Non-exhaustive cheat sheet
```

src/target/HACKING.md
```diff
@@ -85,3 +86,34 @@ typedef struct lpc43xx_flash {
...
+  - `float`: `%f`
+  - `size_t`: `%zu`
+* hex:
+  - `uint8_t`: `PRIx8`
+  - `uint16_t`: `PRIx16`
+  - `uint32_t`: `PRIx32`
+
+A note on enums:
+
+The type of `enum` is not well defined in C11, `enum` **constants** have type `int` (C11 §6.4.4.3),
```
amyspark, line 113:
```suggestion
A note on `enum`: its underlying type is not well defined in C11, `enum` **constants** have type `int` (C11 §6.4.4.3),
```

src/target/HACKING.md
```diff
@@ -85,3 +86,34 @@ typedef struct lpc43xx_flash {
...
+but `enum` **types** are implementation defined and can have any type compatible with `char`, a _signed integer type_,
+or an _unsigned integer type_ (C11 §6.7.2.2)
```
amyspark, line 115:
```suggestion
or an _unsigned integer type_ (C11 §6.7.2.2).
```

src/target/HACKING.md
```diff
@@ -85,3 +86,34 @@ typedef struct lpc43xx_flash {
...
+
+GCC efaults to unsigned int unless there are signed entries in the `enum`
+
+This makes it difficult to write portable printf/scanf format strings, for now `%d` should be used
```
amyspark, line 119:
```suggestion
GCC defaults to `unsigned int` unless there are signed entries in the `enum`. This makes it difficult to write portable format specifiers, for now `%d` should be used.
```

amyspark (changes requested):
Hey @perigoso ! @dragonmux showed me this PR, good job! Just found some grammar and typo nitpicks to share with you.

perigoso:
Hey @amyspark, thank you for the review and suggestions! :smiley:

I made a small change to the way it's formatted after applying the the changes, since the "note" "directive" moved inline, the break in the middle felt out of place, so I removed it

I think I'm happy with this

amyspark (approved):
Going to poke @dragonmux to give it the final thumbs up!

dragonmux (approved):
LGTM. thank you Amy for the SPaG pass on the hacking guide change, and thank you Perigoso for putting together such a good PR that fixes so many subtle things.

Merging. The only change here that needs testing, we don't have the capability right now to test.. and that's the HostIO code. It looks correct though, so that's going to have to do with the community giving us feedback.

PR #1595, opened 2023-08-10
Fix: warning pedantic binary const

perigoso:
## Detailed description

Fix (some) warnings for `-Wpedantic`

See #1590 for context


dragonmux (approved):
We'd been contemplating this PR a while now because we weren't sure if switching away from the binary constants was a good idea or not, but the more we've thought about it, the more we've realised that your solution in putting the binary after the morse code conversion for each letter makes perfect sense here.

It perhaps bares writing an explainer at some point on how the binary representation maps to the dots and dashes representation, and why this mapping was chosen, but for now - LGTM, merging.

PR #1599, opened 2023-08-10
Detect iMX RT variant to support more than 1060

tannewt:
## Detailed description
This uses a word of the ROM to distinguish different chips. Once distinguished, we can use the correct FLEXSPI base address.  Specifically, the 1011 has the FLEXSPI at a different address. Tested on the 1011, 1015, 1020, 1040, 1050 and 1060 EVK boards.

This change also checks the length of the SFDP basic parameters table to ensure it is long enough to include page size. The 1011 and 1015 EVKs use the AT25SF128A which has an old SFDP table that is only 9 DWORDS long. Page size is in DWORD 11.

The 1176 is detected but the flash SFDP doesn't work still. So, it doesn't fully work.

### DIFF
--- a/src/target/imxrt.c
+++ b/src/target/imxrt.c
@@ -73,25 +73,29 @@
 #define IMXRT10xx_CCM_CCG6_FLEXSPI_CLK_MASK           0xfffff3ffU
 #define IMXRT10xx_CCM_CCG6_FLEXSPI_CLK_ENABLE         0x00000c00U
 
-#define IMXRT_FLEXSPI1_BASE UINT32_C(0x402a8000)
-/* We only carry definitions for FlexSPI1 Flash controller A1. */
-#define IMXRT_FLEXSPI1_MOD_CTRL0             (IMXRT_FLEXSPI1_BASE + 0x000U)
-#define IMXRT_FLEXSPI1_INT                   (IMXRT_FLEXSPI1_BASE + 0x014U)
-#define IMXRT_FLEXSPI1_LUT_KEY               (IMXRT_FLEXSPI1_BASE + 0x018U)
-#define IMXRT_FLEXSPI1_LUT_CTRL              (IMXRT_FLEXSPI1_BASE + 0x01cU)
-#define IMXRT_FLEXSPI1_CTRL0                 (IMXRT_FLEXSPI1_BASE + 0x060U)
-#define IMXRT_FLEXSPI1_CTRL1                 (IMXRT_FLEXSPI1_BASE + 0x070U)
-#define IMXRT_FLEXSPI1_CTRL2                 (IMXRT_FLEXSPI1_BASE + 0x080U)
-#define IMXRT_FLEXSPI1_PRG_CTRL0             (IMXRT_FLEXSPI1_BASE + 0x0a0U)
-#define IMXRT_FLEXSPI1_PRG_CTRL1             (IMXRT_FLEXSPI1_BASE + 0x0a4U)
-#define IMXRT_FLEXSPI1_PRG_CMD               (IMXRT_FLEXSPI1_BASE + 0x0b0U)
-#define IMXRT_FLEXSPI1_PRG_READ_FIFO_CTRL    (IMXRT_FLEXSPI1_BASE + 0x0b8U)
-#define IMXRT_FLEXSPI1_PRG_WRITE_FIFO_CTRL   (IMXRT_FLEXSPI1_BASE + 0x0bcU)
-#define IMXRT_FLEXSPI1_STAT1                 (IMXRT_FLEXSPI1_BASE + 0x0e4U)
-#define IMXRT_FLEXSPI1_PRG_WRITE_FIFO_STATUS (IMXRT_FLEXSPI1_BASE + 0x0f4U)
-#define IMXRT_FLEXSPI1_PRG_READ_FIFO         (IMXRT_FLEXSPI1_BASE + 0x100U)
-#define IMXRT_FLEXSPI1_PRG_WRITE_FIFO        (IMXRT_FLEXSPI1_BASE + 0x180U)
-#define IMXRT_FLEXSPI1_LUT_BASE              (IMXRT_FLEXSPI1_BASE + 0x200U)
+#define IMXRT106x_FLEXSPI1_BASE UINT32_C(0x402a8000)
+
+/*
+ * We only carry definitions for FlexSPI1 Flash controller A1.
+ * The base address varies across the 10xx line. We store it in the private structure.
+ */
+#define IMXRT_FLEXSPI1_MOD_CTRL0(priv)             ((priv)->flexspi_base + 0x000U)
+#define IMXRT_FLEXSPI1_INT(priv)                   ((priv)->flexspi_base + 0x014U)
+#define IMXRT_FLEXSPI1_LUT_KEY(priv)               ((priv)->flexspi_base + 0x018U)
+#define IMXRT_FLEXSPI1_LUT_CTRL(priv)              ((priv)->flexspi_base + 0x01cU)
+#define IMXRT_FLEXSPI1_CTRL0(priv)                 ((priv)->flexspi_base + 0x060U)
+#define IMXRT_FLEXSPI1_CTRL1(priv)                 ((priv)->flexspi_base + 0x070U)
+#define IMXRT_FLEXSPI1_CTRL2(priv)                 ((priv)->flexspi_base + 0x080U)
+#define IMXRT_FLEXSPI1_PRG_CTRL0(priv)             ((priv)->flexspi_base + 0x0a0U)
+#define IMXRT_FLEXSPI1_PRG_CTRL1(priv)             ((priv)->flexspi_base + 0x0a4U)
+#define IMXRT_FLEXSPI1_PRG_CMD(priv)               ((priv)->flexspi_base + 0x0b0U)
+#define IMXRT_FLEXSPI1_PRG_READ_FIFO_CTRL(priv)    ((priv)->flexspi_base + 0x0b8U)
+#define IMXRT_FLEXSPI1_PRG_WRITE_FIFO_CTRL(priv)   ((priv)->flexspi_base + 0x0bcU)
+#define IMXRT_FLEXSPI1_STAT1(priv)                 ((priv)->flexspi_base + 0x0e4U)
+#define IMXRT_FLEXSPI1_PRG_WRITE_FIFO_STATUS(priv) ((priv)->flexspi_base + 0x0f4U)
+#define IMXRT_FLEXSPI1_PRG_READ_FIFO(priv)         ((priv)->flexspi_base + 0x100U)
+#define IMXRT_FLEXSPI1_PRG_WRITE_FIFO(priv)        ((priv)->flexspi_base + 0x180U)
+#define IMXRT_FLEXSPI1_LUT_BASE(priv)              ((priv)->flexspi_base + 0x200U)
 
 #define IMXRT_FLEXSPI1_MOD_CTRL0_SUSPEND          0x00000002U
 #define IMXRT_FLEXSPI1_INT_PRG_CMD_DONE           0x00000001U
@@ -140,6 +144,7 @@ typedef struct imxrt_flexspi_lut_insn {
 
 typedef struct imxrt_priv {
 	imxrt_boot_src_e boot_source;
+	uint32_t flexspi_base;
 	uint32_t mpu_state;
 	uint32_t flexspi_lut_state;
 	uint16_t flexspi_cached_commands[4];
@@ -171,6 +176,8 @@ bool imxrt_probe(target_s *const target)
 	target->target_storage = priv;
 	target->target_options |= CORTEXM_TOPT_INHIBIT_NRST;
 	target->driver = "i.MXRT10xx";
+
+	priv->flexspi_base = IMXRT106x_FLEXSPI1_BASE;
 
 #if defined(ENABLE_DEBUG) && (PC_HOSTED == 1 || defined(ESP_LOGD))
 	const uint8_t boot_mode = (target_mem_read32(target, IMXRT_SRC_BOOT_MODE2) >> 24U) & 3U;
@@ -256,8 +263,8 @@ static bool imxrt_enter_flash_mode(target_s *const target)
 	priv->mpu_state = target_mem_read32(target, IMXRT_MPU_CTRL);
 	target_mem_write32(target, IMXRT_MPU_CTRL, 0);
 	/* Start by stepping the clocks to ~50MHz and putting the controller in a known state */
-	target_mem_write32(target, IMXRT_FLEXSPI1_MOD_CTRL0,
-		target_mem_read32(target, IMXRT_FLEXSPI1_MOD_CTRL0) | IMXRT_FLEXSPI1_MOD_CTRL0_SUSPEND);
+	target_mem_write32(target, IMXRT_FLEXSPI1_MOD_CTRL0(priv),
+		target_mem_read32(target, IMXRT_FLEXSPI1_MOD_CTRL0(priv)) | IMXRT_FLEXSPI1_MOD_CTRL0_SUSPEND);
 	target_mem_write32(target, IMXRT10xx_CCM_CCG6,
 		target_mem_read32(target, IMXRT10xx_CCM_CCG6) & IMXRT10xx_CCM_CCG6_FLEXSPI_CLK_MASK);
 	target_mem_write32(target, IMXRT10xx_CCM_CSCM1,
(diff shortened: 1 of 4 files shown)


dragonmux:
Please update your commit message to conform to the contribution guidelines and split the SPI and SFDP changes out into their own commits. They are not directly related to the i.MXRT support improvements and so don't belong in the same commit.

src/target/imxrt.c
```diff
@@ -59,21 +59,64 @@
...
+#define IMXRT_11XX_CCM_CLOCK_ROOT20_CONTROL_OSC400M (0x2 << 8)
+#define IMXRT_11XX_CCM_CLOCK_ROOT20_CONTROL_DIV(divisor) ((divisor - 1) << 0)
+#define IMXRT_11XX_CCM_LPCG28 (IMXRT_11XX_CCM_BASE + 0x6000 + (28 * 0x20))
+
+#define SCB_CPUID UINT32_C(0xE000ED00)
+#define SCB_CPUID_PARTNO_Pos                4U                                            /*!< SCB CPUID: PARTNO Position */
+#define SCB_CPUID_PARTNO_Msk               (0xFFFUL << SCB_CPUID_PARTNO_Pos)              /*!< SCB CPUID: PARTNO Mask */
+
+#define CORE_ID_CM33 (0xD21)
+#define CORE_ID_CM7  (0xC27)
```
dragonmux, line 88: These core IDs are already defined in the cortexm.h header - https://github.com/blackmagic-debug/blackmagic/blob/718c3aedd49224f03ee4f34cb9b2668b0311fa03/src/target/cortexm.h#L178-L185

Please make use of them rather than re-defining them here.

src/target/imxrt.c
```diff
@@ -227,6 +278,70 @@ bool imxrt_probe(target_s *const target)
...
+		rom_location = IMXRT_X00_ROM_FINGERPRINT_ADDR;
+	} else if (cpu_partno == CORE_ID_CM7) {
+		if (target->part_id == 0x88c6U) {
+			rom_location = IMXRT_11XX_ROM_FINGERPRINT_ADDR;
+		} else {
+			rom_location = IMXRT_10XX_ROM_FINGERPRINT_ADDR;
+		}
+	} else {
+		DEBUG_ERROR("Unknown core\n");
+	}
```
dragonmux, line 299: These are excess braces, please remove them on all bar the else-if for the Cortex-M7 ID block

src/target/imxrt.c
```diff
@@ -59,21 +59,64 @@
...
+#define IMXRT_10XX_CCM_CCG6_FLEXSPI_CLK_ENABLE         0x00000c00U
+
+#define IMXRT_11XX_CCM_BASE UINT32_C(0x40CC0000)
+#define IMXRT_11XX_CCM_CLOCK_ROOT20_CONTROL (IMXRT_11XX_CCM_BASE + 0x0 + (20 * 0x80))
+#define IMXRT_11XX_CCM_CLOCK_ROOT20_CONTROL_PLL_480M (0x7 << 8)
+#define IMXRT_11XX_CCM_CLOCK_ROOT20_CONTROL_OSC400M (0x2 << 8)
+#define IMXRT_11XX_CCM_CLOCK_ROOT20_CONTROL_DIV(divisor) ((divisor - 1) << 0)
+#define IMXRT_11XX_CCM_LPCG28 (IMXRT_11XX_CCM_BASE + 0x6000 + (28 * 0x20))
+
+#define SCB_CPUID UINT32_C(0xE000ED00)
```
dragonmux, line 83: This should not be required as this is already in the target structure as `target->cpuid` and already in an appropriate format for the Cortex-M header constants and mask value

dragonmux (changes requested):
Thank you for the contribution! This is an initial review and complements the items given in the lint pass from CI. This is looking good, but could do with a little more work before we can merge it.

src/target/imxrt.c
```diff
@@ -154,14 +193,14 @@ static void imxrt_spi_read(target_s *target, uint16_t command, target_addr_t add
 static void imxrt_spi_write(
 	target_s *target, uint16_t command, target_addr_t address, const void *buffer, size_t length);
 static void imxrt_spi_run_command(target_s *target, uint16_t command, target_addr_t address);
+static void imxrt_id_chip(target_s *target, imxrt_priv_s *priv);
 
 bool imxrt_probe(target_s *const target)
 {
 	/* If the part number fails to match, instantly return. */
-	if (target->part_id != 0x88cU)
+	if (target->part_id != 0x88cU && target->part_id != 0x88c6U) {
```
dragonmux, line 201: Please remove the addition of the braces. They're unnecessary.

src/target/imxrt.c
```diff
@@ -227,6 +270,68 @@ bool imxrt_probe(target_s *const target)
...
+	uint32_t rom_location = 0;
+	const uint16_t cpuid_partno = target->cpuid & CPUID_PARTNO_MASK;
+	if (cpuid_partno == CORTEX_M33)
+		rom_location = IMXRT_X00_ROM_FINGERPRINT_ADDR;
+	else if (cpuid_partno == CORTEX_M7)
+		if (target->part_id == 0x88c6U)
+			rom_location = IMXRT_11XX_ROM_FINGERPRINT_ADDR;
+		else
+			rom_location = IMXRT_10XX_ROM_FINGERPRINT_ADDR;
+	else
```
dragonmux, line 288:
```suggestion
	else if (cpuid_partno == CORTEX_M7) {
		if (target->part_id == 0x88c6U)
			rom_location = IMXRT_11XX_ROM_FINGERPRINT_ADDR;
		else
			rom_location = IMXRT_10XX_ROM_FINGERPRINT_ADDR;
	} else
```
You went a touch too far removing braces - this is considered a multi-line block so should still have them. It's single-line blocks that they get dropped from.

src/target/imxrt.c
```diff
@@ -227,6 +270,68 @@ bool imxrt_probe(target_s *const target)
...
+		DEBUG_ERROR("Unknown core %04x\n", cpuid_partno);
+
+	uint32_t fingerprint = target_mem_read32(target, rom_location);
```
dragonmux, line 291: This can be `const`, so please lets make it `const`.

src/target/imxrt.c
```diff
@@ -227,6 +270,68 @@ bool imxrt_probe(target_s *const target)
...
+	case IMXRT_117X_ROM_FINGERPRINT:
+		priv->chip_id = 1176;
+		priv->flexspi_root = IMXRT_117X_FLEXSPI1_BASE;
+		break;
+	default:
+		DEBUG_TARGET("Unknown ROM fingerprint at %08x = %08x\n", rom_location, fingerprint);
+		break;
+	}
+
+	DEBUG_TARGET("Chip id: %d\n", priv->chip_id);
```
dragonmux, line 330: You are storing chip_id as a uint16_t which is unsigned. `%d` is for *signed* integers. Please use `%u` here which is for unsigned integers. We'd replace or at least prefix the "Chip ID:" wording with the name of the function so we can be sure the message has come from the i.MXRT code.

```suggestion
	DEBUG_TARGET("%s: %u\n", __func__, priv->chip_id);
```

src/target/imxrt.c
```diff
@@ -379,6 +501,7 @@ static void imxrt_spi_exec_sequence(
 
 static void imxrt_spi_wait_complete(target_s *const target)
 {
+	const imxrt_priv_s *const priv = (imxrt_priv_s *)target->target_storage;
```
dragonmux, line 504: This is unused in this function, please remove it or use it.

src/target/imxrt.c
```diff
@@ -404,6 +527,7 @@ static void imxrt_spi_wait_complete(target_s *const target)
 static void imxrt_spi_read(target_s *const target, const uint16_t command, const target_addr_t address,
 	void *const buffer, const size_t length)
 {
+	const imxrt_priv_s *const priv = (imxrt_priv_s *)target->target_storage;
```
dragonmux, line 530: This is unused in this function, please remove it or use it.

src/target/imxrt.c
```diff
@@ -418,6 +542,7 @@ static void imxrt_spi_read(target_s *const target, const uint16_t command, const
 static void imxrt_spi_write(target_s *const target, const uint16_t command, const target_addr_t address,
 	const void *const buffer, const size_t length)
 {
+	const imxrt_priv_s *const priv = (imxrt_priv_s *)target->target_storage;
```
dragonmux, line 545: This is unused in this function, please remove it or use it.

src/target/spi.c
```diff
@@ -121,8 +121,10 @@ spi_flash_s *bmp_spi_add_flash(target_s *const target, const target_addr_t begin
 		spi_parameters.sector_size = 4096U;
 		spi_parameters.capacity = length;
 		spi_parameters.sector_erase_opcode = SPI_FLASH_OPCODE_SECTOR_ERASE;
+		DEBUG_WARN("SFDP read failed. Using best guess.\n");
 	}
 	DEBUG_INFO("Flash size: %" PRIu32 "MiB\n", (uint32_t)spi_parameters.capacity / (1024U * 1024U));
+	DEBUG_INFO("Flash page size: %" PRIu32 "B\n", (uint32_t)spi_parameters.page_size);
```
dragonmux, line 127: We can see why you'd want to print this parameter out, but we don't think it adds to the information displayed by this routine for the Flash cost associated with it. The warning above is good though to help indicate control flow to the user.

src/target/sfdp.c
```diff
@@ -82,7 +83,14 @@ static spi_parameters_s sfdp_read_basic_parameter_table(
 			break;
 		}
 	}
-	result.page_size = SFDP_PAGE_SIZE(parameter_table);
+	// The timing and page size DWORD was added in JESD216A. Before then, the
+	// table is 9 DWORDs and we default to page size 256.
+	if (length >= 4 * 11) {
```
dragonmux, line 88: This if-else block has excess braces, please remove them. As mentioned in Discord we'd prefer to use a version check than a length check here because Vendors have a habit of doing silly things that screw these kinds of check up. We'll add a suggestion for how to do that from the SFDP spec when we get a chance to properly review it.

src/target/imxrt.c
```diff
@@ -170,7 +209,11 @@ bool imxrt_probe(target_s *const target)
 	}
 	target->target_storage = priv;
 	target->target_options |= CORTEXM_TOPT_INHIBIT_NRST;
-	target->driver = "i.MXRT10xx";
+
+	imxrt_id_chip(target, priv);
+
+	snprintf(priv->variant_string, sizeof(priv->variant_string), "i.MXRT%d", priv->chip_id);
```
dragonmux, line 215: `%d` -> %u` please.

src/target/imxrt.c
```diff
@@ -140,10 +175,14 @@ typedef struct imxrt_flexspi_lut_insn {
...
 typedef struct imxrt_priv {
 	imxrt_boot_src_e boot_source;
+	uint16_t chip_id;
+	uint32_t flexspi_root;
 	uint32_t mpu_state;
 	uint32_t flexspi_lut_state;
 	uint16_t flexspi_cached_commands[4];
 	imxrt_flexspi_lut_insn_s flexspi_prg_seq_state[4][8];
+	bool flash_in_package;
+	char variant_string[16];
```
dragonmux, line 185: Given the length of this is something that can change, it makes sense to us to make the size component a `#define` above - `#define IMXRT_NAME_MAX_LENGTH 16U` for example - so it's more trivial to adjust. This then gives you a define to use in the snprintf() below rather than having to rely on sizeof() doing something sensible on all possible compilation platforms.

We would also suggest this is the part's name, so `variant_string` -> `name`? It'd make it make more sense below in our opinion.

src/target/imxrt.c
```diff
@@ -59,21 +58,57 @@
 #define IMXRT_MPU_BASE UINT32_C(0xe000ed90)
 #define IMXRT_MPU_CTRL (IMXRT_MPU_BASE + 0x04U)
 
-#define IMXRT_CCM_ANALOG_BASE     UINT32_C(0x400d8000)
-#define IMXRT_CCM_ANALOG_PLL3_PFD (IMXRT_CCM_ANALOG_BASE + 0x0f0U)
+#define IMXRT_10XX_CCM_ANALOG_BASE     UINT32_C(0x400d8000)
```
dragonmux, line 61: Per the naming schemes used in other files in BMD, this should probably be `IMXRT10xx_` not `IMXRT_10XX_` (ditto for each of the defines below)

src/target/imxrt.c
```diff
@@ -31,7 +31,6 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include <string.h>
```
dragonmux, line 34: We know why you removed this from Discord, but if memory serves this then causes include errors in VSCode on some platforms. Might we suggest moving the inclusion below general.h but keeping it present?

dragonmux (changes requested):
Thank you for splitting the contribution into commits. We have now given the code a more proper review, see below. It looks like a lot of items, for which apologies, but most of them should be fairly trivial to address and are quite mechanical in nature.

We look forward to testing this on our i.MXRT1060 board and getting this merged soon!

src/target/imxrt.c
tannewt: This is a consequence of IMXRT_FLEXSPI macro using it. It is used, but not obviously. I'm happy to entertain better ways to do this.

src/target/imxrt.c
tannewt: Would this clang-format setting work to automatically get braces correct? https://clang.llvm.org/docs/ClangFormatStyleOptions.html#removebracesllvm This would be nice to automate.

src/target/sfdp.c
tannewt: I'll update based on version number when you have a suggestion. It wasn't clear to me when I looked because I couldn't find the old spec. The changes described in the current doc doesn't make it super clear what the old spec said for versioning.

src/target/imxrt.c
dragonmux: Unfortunately, and because of that big fat warning listed, it wouldn't quite do the right thing and, as they note, would lead to incorrect code in some instances. That is why we elected to leave it and the next one about excess parens off

dragonmux:
There's a small additional change we'd like to request based on some research @mubes has been doing - it turns out that allowing the Cortex-M probe routine to poke for Kinetis parts before iMXRT parts puts certain iMXRT parts in a bad state when we arrive, so please apply this patch in a new commit as part of this PR:
```diff
diff --git a/src/target/cortexm.c b/src/target/cortexm.c
index 6d80f4f8..28bdc5d9 100644
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -686,8 +686,8 @@ bool cortexm_probe(adiv5_access_port_s *ap)
 
 	switch (t->designer_code) {
 	case JEP106_MANUFACTURER_FREESCALE:
-		PROBE(kinetis_probe);
 		PROBE(imxrt_probe);
+		PROBE(kinetis_probe);
 		break;
 	case JEP106_MANUFACTURER_GIGADEVICE:
 		PROBE(gd32f1_probe);
```

src/target/sfdp.c
dragonmux: Reading the standard's page 27 ("Page 13") - §6.2.2, it shows that JESD26F uses version major 1, version minor 10; JESD26E uses version major 1 and minor 9, JESD25D major 1 and minor 8 and then before that, in JESD25C it was major 0 and an unknown minor number. Your part has major 0 and minor 1:
```
8 byte SFDP read at 0x0:
        53 46 44 50 00 01 01 ff
```
so it's reasonable to assume this is an original standard part or one following the second release.

There are another pair of version numbers in the table header following a similar numbering, and with your part having:
```
8 byte SFDP read at 0x8:
        00 00 01 09 30 00 00 ff
```
which defines the table major to 0 and minor to 1 while the current standard is 1 and 8.

If this is passed into this function as a `const sfdp_parameter_table_header_s *const header` parameter, we can test `if (header->version_major > 1 || (header->version_major == 1 && header->version_minor >= 5))`.

These values are derived from Annex B's (sketchy..) data on what changed in each JESD216 letter suffixed standard, and in A -> B they specify this header goes to minor 6 by incrementing the number, with the no letter -> A change saying the main header gets major 1 and minor 5. It's in this no-to-A change that the parameter table is changed from 9 to 16 uint32_t's and the value we're after here is in the 11th.

src/target/imxrt.c
```diff
@@ -256,35 +361,55 @@ static bool imxrt_enter_flash_mode(target_s *const target)
...
+		// PLL3 is 480 mhz and PFD0 is set to 0x16 which is 480 * (18 / 0x16) = 392 which is then divided by 2.
+		target_mem_write32(target, IMXRT10xx_CCM_ANALOG_PLL3_PFD,
+			(target_mem_read32(target, IMXRT10xx_CCM_ANALOG_PLL3_PFD) & IMXRT10xx_CCM_ANALOG_PLL_PFD0_FRAC_MASK) |
+				0x16U);
+
+		// Ungate the clock.
+		target_mem_write32(target, IMXRT10xx_CCM_CCG6,
+			target_mem_read32(target, IMXRT10xx_CCM_CCG6) | IMXRT10xx_CCM_CCG6_FLEXSPI_CLK_ENABLE);
+	} else if (1100 <= priv->chip_id && priv->chip_id < 1200) {
+		DEBUG_TARGET("0xe0040000 %08x\n", target_mem_read32(target, 0xe0040000));
```
dragonmux, line 383: Did you mean to leave these two debug reads in?

src/target/sfdp.c
```diff
@@ -82,7 +84,14 @@ static spi_parameters_s sfdp_read_basic_parameter_table(
 			break;
 		}
 	}
-	result.page_size = SFDP_PAGE_SIZE(parameter_table);
+	// The timing and page size DWORD was added in JESD216A. It is marked as
+	// version 1.5.
+	if (header->version_major > 1 || (header->version_major == 1 && header->version_minor >= 5)) {
```
dragonmux, line 89: Please remove the braces from this and the else - they're not necessary.

dragonmux (commented):
There are only two items we saw doing a further review. With these taken care of we think this is ready to be merged. Please rebase this on main (`git rebase -i main`), which will eliminate that merge commit, and then squish fixes such as the spi.c print removal into the original commit for that - we are happy to provide help with this if you would like.

Thank you for your hard work with this and persevering!

tannewt:
> we are happy to provide help with this if you would like.

Please feel free to finish this. I've lost the context for it (not working with imx currently.)

dragonmux:
Thank you - we'll get on that in the next day or two then.

esden (approved):
LGTM

PR #1601, opened 2023-08-12
riscv: add single register access, add csr access with an offset, expose some CSRs

mean00:
## Detailed description

Hi this tempative MR adds the following featurettes  and could receive some feedback :
- Allow single register access (rw) . This is done only for RV32 , i dont have RV64 to do & test, should be straightforward.
- Exposes common CSRs trough the xml description. 
- Allow CSRs rw as regular registers. There is an offset so that the it does not clash with GPR registers

Please note that i extracted this MR manually from a larger codebase (including Perigoso wchlink work, the CH32V3xx chip support and flashstubs) to have a small one. It looks alright but it is  hard for me to test on its own.

Finally, it may be a good idea to think about having a generic way to add CSR per MCU (the designer custom ones) ?

There is placeholder for ch32xx_probe under #if 0, if this is bothersome please tell me

Explain the **details** for making this change.
* Is a new feature implemented? Sort of, it was somehow missing and planned according to the comments.
* What existing problem(s) does the pull request solve? Single register access,  CSRs access (the common ones at least)
* How does the pull request solve these problems? Add the missing functions + expose CSRs are general registers with an offset

## Closing issues

N/A

### DIFF
--- a/src/target/riscv32.c
+++ b/src/target/riscv32.c
@@ -67,6 +67,8 @@ typedef struct riscv32_regs {
 #define RV32_MATCH_BEFORE 0x00000000U
 #define RV32_MATCH_AFTER  0x00040000U
 
+static ssize_t riscv32_reg_read(target_s *target, const int c, void *data, const size_t max);
+static ssize_t riscv32_reg_write(target_s *const target, const int c, const void *data, const size_t max);
 static void riscv32_regs_read(target_s *target, void *data);
 static void riscv32_regs_write(target_s *target, const void *data);
 static void riscv32_mem_read(target_s *target, void *dest, target_addr_t src, size_t len);
@@ -91,6 +93,8 @@ bool riscv32_probe(target_s *const target)
 	target->regs_size = sizeof(riscv32_regs_s);
 	target->regs_read = riscv32_regs_read;
 	target->regs_write = riscv32_regs_write;
+	target->reg_write = riscv32_reg_write;
+	target->reg_read = riscv32_reg_read;
 	target->mem_read = riscv32_mem_read;
 	target->mem_write = riscv32_mem_write;
 
@@ -139,6 +143,41 @@ static void riscv32_regs_write(target_s *const target, const void *const data)
 	}
 	/* Special access to poke in the program counter that will be executed on resuming the hart */
 	riscv_csr_write(hart, RV_DPC, &regs->pc);
+}
+
+static inline ssize_t riscv32_bool_to_4(const bool ret)
+{
+	return ret ? 4 : -1;
+}
+
+static ssize_t riscv32_reg_read(target_s *target, const int reg, void *data, const size_t max)
+{
+	if (max != 4)
+		return -1;
+	/* Grab the hart structure  */
+	riscv_hart_s *const hart = riscv_hart_struct(target);
+	if (reg < 32)
+		return riscv32_bool_to_4(riscv_csr_read(hart, RV_GPR_BASE + reg, data));
+	if (reg == 32)
+		return riscv32_bool_to_4(riscv_csr_read(hart, RV_DPC, data));
+	if (reg >= RV_CSR_GDB_OFFSET)
+		return riscv32_bool_to_4(riscv_csr_read(hart, reg - RV_CSR_GDB_OFFSET, data));
+	return -1;
+}
+
+static ssize_t riscv32_reg_write(target_s *const target, const int reg, const void *data, const size_t max)
+{
+	if (max != 4)
+		return -1;
+	/* Grab the hart structure  */
+	riscv_hart_s *const hart = riscv_hart_struct(target);
+	if (reg < 32)
+		return riscv32_bool_to_4(riscv_csr_write(hart, RV_GPR_BASE + reg, data));
+	if (reg == 32)
+		return riscv32_bool_to_4(riscv_csr_write(hart, RV_DPC, data));
+	if (reg >= RV_CSR_GDB_OFFSET)
+		return riscv32_bool_to_4(riscv_csr_write(hart, reg - RV_CSR_GDB_OFFSET, data));
+	return -1;
 }
 
 /* Takes in data from abstract command arg0 and, based on the access width, unpacks it to dest */
--- a/src/target/riscv_debug.c
+++ b/src/target/riscv_debug.c
@@ -137,6 +137,24 @@ static const char *const riscv_gpr_names[RV_GPRS_COUNT] = {
 	"s8", "s9", "s10", "s11",
 	"t3", "t4", "t5", "t6",
 };
+
+typedef struct riscv_csr_descriptor  {
+	const char *name;
+	const uint32_t csr_number; // fits in 16 bits actually (?)
+} riscv_csr_descriptor_s;
+
+static const riscv_csr_descriptor_s riscv_csrs[]={
+	{"mstatus",RV_CSR_STATUS},
+	{"misa",RV_CSR_MISA},
+	{"mie",	RV_CSR_MIE},
+	{"mtvec",	RV_CSR_MTVEC},
+	{"tscratch",RV_CSR_MSCRATCH},
+	{"mepc",	RV_CSR_MEPC},
+	{"mcause",RV_CSR_MCAUSE},
+	{"mtval",	RV_CSR_MTVAL},
+	{"mip",	RV_CSR_MIP},
+};
+
 // clang-format on
 
 /* General-purpose register types */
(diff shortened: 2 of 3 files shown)


src/target/riscv32.c
```diff
@@ -101,6 +105,11 @@ bool riscv32_probe(target_s *const target)
 	case JEP106_MANUFACTURER_RV_GIGADEVICE:
 		PROBE(gd32vf1_probe);
 		break;
+#if 0
+	case JEP106_MANUFACTURER_WCH:
+		PROBE(ch32v3xx_probe);
+		break;
+#endif
```
dragonmux, line 112: This looks to be code left over from another potential PR, please remove it from this one.

src/target/riscv32.c
```diff
@@ -141,6 +150,45 @@ static void riscv32_regs_write(target_s *const target, const void *const data)
 	riscv_csr_write(hart, RV_DPC, &regs->pc);
 }
 
+static ssize_t riscv32_bool_to_4(bool ret)
```
dragonmux, line 153: Please mark this `inline` as in `static inline ssize_t`, the body of the function is so short that we'll likely get better codegen to have the compiler inline this.

src/target/riscv32.c
```diff
@@ -141,6 +150,45 @@ static void riscv32_regs_write(target_s *const target, const void *const data)
...
+{
+	if (ret)
```
dragonmux, line 155: Stylistically this is fine, but a suggestion as this can only ever return two values: Use a ternary as in `return ret ? 4 : -1;`. It does make the body a one-line body but there's just no point here to making it an expandable if form piece of conditional logic as `bool` only ever has two states.

Also, please mark the parameter `const`. Same with the other functions introduced - please mark their read-only parameters const and pointers (but not the things those pointers point to if written) const.

dragonmux (changes requested):
Outside of wondering if it's sensible to be exposing any of the CSRs to GDB both because you can then screw up debug state with them and because they're basically all optional and so problematic to expose..

This PR is looking like a good start that needs some cleanup and stylistic fixes - we've tried to be through in the review but expect we've likely missed some things, so please consider it only a start that might need more work.

Do please note too that in `riscv_build_target_description()` there are a pile of missed checks and handling of `print_size` so that code is likely to explode in some fun ways. We didn't write review notes on it because the code above in the function should serve as an example on how to handle this variable.

src/target/riscv_debug.c
```diff
@@ -942,11 +962,22 @@ static size_t riscv_build_target_description(
 
 	/* XXX: TODO - implement generation of the FPU feature and registers */
 
+	/* Add main CSR registers*/
+	offset += snprintf(buffer + offset, print_size, "</feature>");
+
+	offset += snprintf(buffer + offset, print_size, "<feature name=\"org.gnu.gdb.riscv.csr\">");
+	int nb_csr = sizeof(riscv_csrs) / sizeof(riscv_csr_descriptor_s);
+	for (int i = 0; i < nb_csr; i++) {
```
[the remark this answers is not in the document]
mean00: Before i get back to the comments (thanks!), the problem is that if you dont have access to some CSRs you are blind when  looking into low level interrupt stuff or rescheduling issue for example.
Gdb is even kind enough to decode some of them in human readable form (mstatus, misa,....)
On arm, you most of the time just need plain memory read/write, no special handling needed (except co processors but that's another story).

mean00:
I think i got them all
Please note it is very difficult to test for me,  as it is part of much bigger changeset i have to manually extract from.
Thank you

src/target/riscv32.c
```diff
@@ -141,6 +145,41 @@ static void riscv32_regs_write(target_s *const target, const void *const data)
...
 }
 
+static inline ssize_t riscv32_bool_to_4(const bool ret)
+{
+	return ret ? 4 : -1;
+}
+
+static ssize_t riscv32_reg_read(target_s *target, const int reg, void *data, const size_t max)
+{
+	if (max < 4)
```
dragonmux, line 155: This still checks `max < 4` while the write checks `max != 4` - which is still inconsistent. Is there a good way to be more consistent about this?

src/target/riscv32.c
```diff
@@ -67,8 +67,10 @@ typedef struct riscv32_regs {
 #define RV32_MATCH_BEFORE 0x00000000U
 #define RV32_MATCH_AFTER  0x00040000U
 
+static ssize_t riscv32_reg_read(target_s *target, const int c, void *data, const size_t max);
 static void riscv32_regs_read(target_s *target, void *data);
 static void riscv32_regs_write(target_s *target, const void *data);
+static ssize_t riscv32_reg_write(target_s *const target, const int c, const void *data, const size_t max);
```
dragonmux, line 73: Please move this declaration up to just under the `riscv32_reg_read()` one for consistency.

src/target/riscv_debug.c
```diff
@@ -137,6 +137,25 @@ static const char *const riscv_gpr_names[RV_GPRS_COUNT] = {
 	"s8", "s9", "s10", "s11",
 	"t3", "t4", "t5", "t6",
 };
+
+typedef struct riscv_csr_descriptor  {
+	const char *name;
+	const uint32_t   csr_number; // fits in 16 bytes actually (?)	
+} riscv_csr_descriptor_s;
+
```
dragonmux, line 145: Please remove this excess new line.

src/target/riscv_debug.c
```diff
@@ -942,6 +961,13 @@ static size_t riscv_build_target_description(
 
 	/* XXX: TODO - implement generation of the FPU feature and registers */
 
+	/* Add main CSR registers*/
+	offset += snprintf(buffer + offset, print_size, "</feature><feature name=\"org.gnu.gdb.riscv.csr\">");
```
dragonmux, line 965: This needs prefixing with:
```c
	if (max_length != 0)
		print_size = max_length - (size_t)offset;
```

As does the snprintf() in the loop. Otherwise `print_size` will be set wrong and allow snprintf() to buffer overflow `buffer` in the normal formatting pass. The result is a security issue and a correctness issue.

src/target/riscv_debug.c
```diff
@@ -942,6 +961,13 @@ static size_t riscv_build_target_description(
...
+	for (size_t i = 0; i < ARRAY_LENGTH(riscv_csrs); i++) {
+		offset += snprintf(buffer + offset, print_size,
+			" <reg name=\"%s\" bitsize=\"%u\" type=\"int\" regnum=\"%" PRIu32 "\" group=\"csr\" save-restore=\"no\"/>",
```
dragonmux, line 968: This string adds quite a lot to the size here - and `type="int"` can be inferred by GDB so can be dropped, as can the `group="csr"` part. `save-restore="no"` should be able to be formatted in by %s by grabbing `gdb_reg_save_restore_strings[GDB_SAVE_RESTORE_NO]` which will serve to further compress this string in the binary.

These are all only suggestions to help save some space as this is technically a correct string as-is, just large.

dragonmux (changes requested):
This is looking much better - there are still a couple of items outstanding, captured in this re-review, but you are making excellent progress and with the rationale provided in [your earlier comment](https://github.com/blackmagic-debug/blackmagic/pull/1601#issuecomment-1677262820) we acknowledge the use of providing these CSRs to GDB even if the results run the risk of making some targets very unhappy (we can address this when it happens though..). 👍🏼

mean00:
Hopefully this time it's correct
Thank you

src/target/riscv_debug.c
```diff
@@ -946,6 +964,16 @@ static size_t riscv_build_target_description(
...
 
+	/* Add main CSR registers*/
+	if (max_length != 0)
+		print_size = max_length - (size_t)offset;
+	offset += snprintf(buffer + offset, print_size, "</feature><feature name=\"org.gnu.gdb.riscv.csr\">");
+	for (size_t i = 0; i < ARRAY_LENGTH(riscv_csrs); i++) {
+		if (max_length != 0)
+			print_size = max_length - (size_t)offset;
+		offset += snprintf(buffer + offset, print_size, " <reg name=\"%s\" bitsize=\"%u\" regnum=\"%" PRIu32 "\" />",
+			riscv_csrs[i].name, address_width, riscv_csrs[i].csr_number + RV_CSR_GDB_OFFSET);
```
dragonmux, line 975: When talking about optimising this string, we didn't mean you to remove the save-restore component completely as that's rather important for GDB (it defaults to yes otherwise). This is great otherwise - see this suggested change for what we were intending to suggest.

```suggestion
		offset += snprintf(buffer + offset, print_size, " <reg name=\"%s\" bitsize=\"%u\" regnum=\"%" PRIu32 "\" %s/>",
			riscv_csrs[i].name, address_width, riscv_csrs[i].csr_number + RV_CSR_GDB_OFFSET,
			gdb_reg_save_restore_strings[GDB_SAVE_RESTORE_NO]);
```

dragonmux:
There's only one item that we're concerned is a little off still, highlighted in the review comment above, but other than that, this looks ready to merge - good work!

mean00:
oops,
Should be okay now (still very hard to test for me)
Thank you

dragonmux (approved):
LGTM 👍🏼  Thank you for the contribution! Merging

PR #1602, opened 2023-08-13
riscv: Check if hw reset worked, if not do a DM reset

mean00:
## Detailed description

If the board does not have a hw reset pin (or  misconfigured or does not work), and without RV_TOPT_INHIBIT_NRST,
the riscv_reset relies entirely on the hw reset and ignores errors.

As a result the board can be not reset and behave weirdly.

The MR tries to detect the HW reset failed (riscv_dm_poll_state timeout) and if so, tries to do a system reset through DM.

With that MR i can use the "r" gdb command to restart the system properly without explicitly disabling the HW reset
(i dont have it on my board)

### DIFF
--- a/src/target/riscv_debug.c
+++ b/src/target/riscv_debug.c
@@ -848,6 +848,7 @@ static void riscv_reset(target_s *const target)
 	}
 	/* Acknowledge the reset */
 	riscv_dm_write(hart->dbg_module, RV_DM_CONTROL, hart->hartsel | RV_DM_CTRL_HART_ACK_RESET);
+	riscv_halt_request(target);
 	target_check_error(target);
 }
 
--- a/src/target/riscv_debug.c
+++ b/src/target/riscv_debug.c
@@ -827,17 +827,20 @@ static target_halt_reason_e riscv_halt_poll(target_s *const target, target_addr_
 static void riscv_reset(target_s *const target)
 {
 	riscv_hart_s *const hart = riscv_hart_struct(target);
+	bool has_reset = false;
 	/* If the target does not have the nRST pin inhibited, use that to initiate reset */
 	if (!(target->target_options & RV_TOPT_INHIBIT_NRST)) {
 		platform_nrst_set_val(true);
-		riscv_dm_poll_state(hart->dbg_module, RV_DM_STAT_ALL_RESET);
+		has_reset = riscv_dm_poll_state(hart->dbg_module, RV_DM_STAT_ALL_RESET);
 		platform_nrst_set_val(false);
 		/* In theory we're done at this point and no debug state was perturbed */
-	} else {
-		/*
-		 * Otherwise, if nRST is not usable, use instead reset via dmcontrol. In this case,
-		 * when reset is requested, use the ndmreset bit to perform a system reset
-		 */
+	}
+
+	/*
+	 * Otherwise, if nRST is not usable (or failed), use instead reset via dmcontrol. In this case,
+	 * when reset is requested, use the ndmreset bit to perform a system reset
+	 */
+	if (!has_reset) {
 		riscv_dm_write(hart->dbg_module, RV_DM_CONTROL, hart->hartsel | RV_DM_CTRL_SYSTEM_RESET);
 		riscv_dm_poll_state(hart->dbg_module, RV_DM_STAT_ALL_RESET);
 		/* Complete the reset by resetting ndmreset */


mean00:
Might be better to wait a bit
The MR in itself works but there may be side effects, looking into it

mean00:
Ok, the issue is that the hart was not stopped at reset exit.  So it refused some commands (WRONG STATE)
With the 2nd commit, when being reset by DM, the core is halted.
That way is works fine for me, but i'm not convinced this is consistent with the overall expected behaviour.
Help ?

(if it looks okay, i'll fix the offset comment and squash it into a single commit)
Thank you

src/target/riscv_debug.c
```diff
@@ -827,24 +827,28 @@ static target_halt_reason_e riscv_halt_poll(target_s *const target, target_addr_
...
-		riscv_dm_write(hart->dbg_module, RV_DM_CONTROL, hart->hartsel | RV_DM_CTRL_SYSTEM_RESET);
+	}
+
+	/*
+		* Otherwise, if nRST is not usable (or if it failed), use instead reset via dmcontrol. In this case,
+		* when reset is requested, use the ndmreset bit to perform a system reset
+		*/
+
+	if (!has_reset) {
+		riscv_dm_write(hart->dbg_module, RV_DM_CONTROL, hart->hartsel | RV_DM_CTRL_SYSTEM_RESET | RV_DM_CTRL_HALT_REQ);
```
dragonmux, line 845: Rather than or'ing in RV_DM_CTRL_HALT_REQ and hoping that by the end of this stretch, halt has been actioned - might it not be better to call riscv_halt_request() as that function ensures the halt completed and the request flag gets cleared so we don't wind up in a continued bad state when poking the DM?

Otherwise, this is a great catch on your part - our test targets all had the reset signal available and usable so the else code path was untested in this regard.

src/target/riscv_debug.c
```diff
@@ -827,24 +827,28 @@ static target_halt_reason_e riscv_halt_poll(target_s *const target, target_addr_
...
-		 * Otherwise, if nRST is not usable, use instead reset via dmcontrol. In this case,
-		 * when reset is requested, use the ndmreset bit to perform a system reset
-		 */
-		riscv_dm_write(hart->dbg_module, RV_DM_CONTROL, hart->hartsel | RV_DM_CTRL_SYSTEM_RESET);
+	}
+
+	/*
+		* Otherwise, if nRST is not usable (or if it failed), use instead reset via dmcontrol. In this case,
+		* when reset is requested, use the ndmreset bit to perform a system reset
+		*/
```
dragonmux, line 842: This comment needs slightly reformatting - your editor removed the spaces that indent the *'s on the continuation lines instead of the tabs, hence the odd appearance.

src/target/riscv_debug.c
mean00: Using riscv_halt_request does not completely work for me.
I still have to have this at the end  for it to work
 `riscv_dm_write( hart->dbg_module, RV_DM_CONTROL, hart->hartsel | RV_DM_CTRL_HART_ACK_RESET | RV_DM_CTRL_HALT_REQ);`
As a result i dont know if that part is still of interest ?

dragonmux: Does it work properly if a riscv_halt_request() call is made after the write to ACK reset (with the halt request being set and made through the other DM writes)?

mean00: I did a quick check and this seems to work.
Just adding the riscv_halt_request at the end seems to be enough.
No need to play with HALT_REQ. 
That looks much simpler, but the code may execute a little bit before we stop the hart (?)
```

	if (!has_reset) {		
		riscv_dm_write(hart->dbg_module, RV_DM_CONTROL, hart->hartsel | RV_DM_CTRL_SYSTEM_RESET);
		riscv_dm_poll_state(hart->dbg_module, RV_DM_STAT_ALL_RESET);
		/* Complete the reset by resetting ndmreset */
		riscv_dm_write(hart->dbg_module, RV_DM_CONTROL, hart->hartsel);
	}

	/* Acknowledge the reset */
	riscv_dm_write(hart->dbg_module, RV_DM_CONTROL, hart->hartsel | RV_DM_CTRL_HART_ACK_RESET);
	riscv_halt_request(target);
	target_check_error(target);
```

dragonmux: That problem of execution before the halt request gets ack'd in riscv_halt_request() is why we were thinking this would be in combination with doing `| RV_DM_CTRL_HALT_REQ` in the preceding writes - the reason for calling riscv_halt_request() at the end is to get it to ack the halt request and make sure halt succeeded before continuing on. Agreed about it being much simpler in result though

src/target/riscv_debug.c
mean00: indeed,  i tried to fix that, it looks ok here now (?)

src/target/riscv_debug.c
mean00: I re-set the 1st commit in accordance to the above
It's a manual process, i need 4 or 5 other patches to be able to use the CH32V3xx board, it looks alright.

mean00:
I've pushed the 2nd version
Thank you

src/target/riscv_debug.c
dragonmux: The whitespace has been entirely converted to spaces - the indentation should be tabs (1x) followed by a single space to align the *'s in the comment block.

dragonmux: image
vs
image

dragonmux:
With the outstanding indentation issue fixed, we're happy to merge this - nice work :)

mean00:
I think this is correct now, "set noet" finally did the trick 
Thank you

dragonmux (approved):

PR #1603, opened 2023-08-16
Fix: warning pedantic range expr

perigoso:
## Detailed description

This one was missed from the initial batch

Fix (some) warnings for `-Wpedantic`

See #1590 for context

### DIFF
--- a/src/target/adiv5.c
+++ b/src/target/adiv5.c
@@ -279,32 +279,36 @@ static const char *adiv5_arm_ap_type_string(const uint8_t ap_type, const uint8_t
 	 * table C1-2 "AP Identification types for an AP designed by Arm"
 	 */
 
-	switch (ap_type) {
-	case 0x1U:
-		return "AHB3-AP";
-	case 0x2U:
-		return "APB2/3-AP";
-	/* 0x3 is not defined */
-	case 0x4U:
-		return "AXI3/4-AP";
-	case 0x5U:
-		return "AHB5-AP";
-	case 0x6U:
-		return "APB4/5-AP";
-	case 0x7U:
-		return "AXI5-AP";
-	case 0x8U:
-		return "AHB5-AP";
-	case 0U:
-		/* Type 0 APs are determined by the class code */
-		if (ap_class == 0U)
-			return "JTAG-AP";
-		if (ap_class == 1U)
-			return "COM-AP";
-	/* fall through */
-	default:
-		return "Unknown";
+	/* All types except 0 are only valid for ap_class == 0x8 */
+	if (ap_class == 0x8U || ap_type == 0U) {
+		switch (ap_type) {
+		case 0U:
+			/* Type 0 APs are determined by the class code */
+			if (ap_class == 0U)
+				return "JTAG-AP";
+			if (ap_class == 1U)
+				return "COM-AP";
+			break;
+		case 0x1U:
+			return "AHB3-AP";
+		case 0x2U:
+			return "APB2/3-AP";
+		/* 0x3 is not defined */
+		case 0x4U:
+			return "AXI3/4-AP";
+		case 0x5U:
+			return "AHB5-AP";
+		case 0x6U:
+			return "APB4/5-AP";
+		case 0x7U:
+			return "AXI5-AP";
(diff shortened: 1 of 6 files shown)


src/target/adiv5.c
```diff
@@ -94,16 +94,22 @@ typedef enum cid_class {
...
-	[cidc_dc] = "Debug component",                            /* 0x9 */
-	[0xa] = "R",                                              /* 0xa */
-	[cidc_ptb] = "Peripheral Test Block",                     /* 0xb */
-	[0xc] = "R",                                              /* 0xc */
-	[cidc_dess] = "OptimoDE Data Engine SubSystem component", /* 0xd */
-	[cidc_gipc] = "Generic IP component",                     /* 0xe */
-	[cidc_sys] = "Non STD System component",                  /* 0xf */
+	[cidc_gvc] = "Generic verification component",            /* 0x00 */
+	[cidc_romtab] = "ROM Table",                              /* 0x01 */
+	[0x2] = "R",                                              /* 0x02 */
```
dragonmux, line 99: Do we know from the spec or the architecture manuals etc what these "R" entries should be? Or even, should they be valid strings rather than NULL for now to make it more obvious when the code hits one of these unknown/reserved CID values? The comment talks about saving space but having the table store null pointers would save more and be more obvious if we see one in the wild.

dragonmux (changes requested):
This is looking good. There's one item that we have queries on that this touches but with that resolved, think this is good for merge.

Might we suggest, given this touches the Cortex-A and Cortex-M core support, that you give feature/arm-cortex-refactoring a look as that changes a fair whack to do with the Cortex code, as does feature/cortex-r-support which is built on top of the previously mentioned branch

src/target/adiv5.c
perigoso: In looking into this I refactored `adiv5_ap_type`, which seemed to have a mistake, the function could look up from indexes 0 to 7, but only 6 entries existed

dragonmux: The structure of adiv5_ap_type() was built around this table from the ADIv5 spec:
image

This is why the table and function was laid out how it was.

perigoso: Ah! I looked for that table but it's not in the ADIv5.2 spec I was looking at, or at least not under that name (or maybe I just missed it)

But there was in fact a problem (not the missing `> 0` check), the table was doing a simple lookup for `type -1` but the type `0x3` does not exist so the table was missing one empty entry, every type after `0x3` would have been showing the incorrect string, and type `8` would be out of bounds

I corrected the values in my commit

dragonmux: Oh, great catch! that apparently completely blind-sided us. We'll re-review with that in mind

src/target/adiv5.c
```diff
@@ -289,32 +272,62 @@ static const struct {
...
+static const char *adiv5_arm_ap_type_string(const uint8_t ap_type, const uint8_t ap_class)
 {
-	/* Type 0 APs are determined by the class code */
-	if (ap_type == 0U) {
+	/*
+	 * Values taken from ADIv5 spec §C1.3 pg146.
+	 * table C1-2 "AP Identification types for an AP designed by Arm"
+	 */
+
+	switch (ap_type) {
```
dragonmux, line 282: This is looking much better. Given the `ap_class` is supposed to be 8 per the table for this first block of `ap_type`'s, it is probably a good idea to pivot this slightly to first checking `ap_class` then switching on `ap_type` (and yes, this was a flaw in our original logic too), but we'd be happy merging this as-is if you don't want to go through that to make this exactingly correct.

dragonmux (commented):
Only have one comment that, once answered, leaves us happy to merge this. Great work! We look forward to finding a resolution to the `enum` problem too in a follow-up PR

src/target/adiv5.c
perigoso: Good catch, added the additional validation

dragonmux (approved):
LGTM, we'll merge this once we've given it a test against some local parts to make sure things seem sane.

dragonmux:
Looks good, seems to work well. Merging!

PR #1610, opened 2023-08-24
Fix: cmsis_dap: error when writing long buffers to DAP (>59 bytes)

projectgus:
## Detailed description

Without this fix, dap_mem_write() accounts for write buffer header size incorrectly when calculating `blocks_per_transfer`. Long transfers (via HID) fail with error "Attempted to make over-long request of 66 bytes, max length is 65".

Sample stack trace from where this error is triggered:

<details><summary>Click for a a sample stack trace where this error is triggered.</summary>

```
0  dbg_dap_cmd_hid (request_data=request_data@entry=0x7fffffffc103 "\006", request_length=request_length@entry=65, 
    response_data=response_data@entry=0x7fffffffc077 "", response_length=65) at platforms/hosted/cmsis_dap.c:383
#1  0x000055555557930d in dap_run_cmd_raw (response_length=3, response_data=0x7fffffffc100 "\377\377\377\006", request_length=65, 
    request_data=0x7fffffffc103 "\006") at platforms/hosted/cmsis_dap.c:447
#2  dap_run_cmd (request_data=0x7fffffffc103, request_length=request_length@entry=65, response_data=response_data@entry=0x7fffffffc100, 
    response_length=response_length@entry=3) at platforms/hosted/cmsis_dap.c:469
#3  0x000055555557a322 in perform_dap_transfer_block_write (target_dp=0x5555555cd260, reg=reg@entry=13 '\r', 
    block_count=block_count@entry=15, blocks=blocks@entry=0x7fffffffc568) at platforms/hosted/dap_command.c:196
#4  0x0000555555579aac in dap_write_block (target_ap=target_ap@entry=0x5555555cc570, dest=dest@entry=536870916, src=<optimized out>, 
    src@entry=0x5555555e0100, len=len@entry=60, align=align@entry=ALIGN_WORD) at platforms/hosted/dap.c:299
#5  0x0000555555578761 in dap_mem_write (align=ALIGN_WORD, len=256, src=0x5555555e0100, dest=536870916, ap=0x5555555cc570)
    at platforms/hosted/cmsis_dap.c:540
#6  dap_mem_write (ap=0x5555555cc570, dest=536870916, src=0x5555555e0100, len=256, align=ALIGN_WORD) at platforms/hosted/cmsis_dap.c:513
#7  0x00005555555773fa in target_mem_write (t=t@entry=0x5555555d7a50, dest=<optimized out>, src=src@entry=0x5555555e0100, 
    len=len@entry=256) at target/target.c:282
#8  0x000055555556cab1 in rp_spi_xfer (target=0x5555555d7a50, command=<optimized out>, address=0, wr_buf=0x5555555e0100 "", wr_len=256, 
    rd_buf=rd_buf@entry=0x0, rd_len=0) at target/rp.c:459
#9  0x000055555556cb6e in rp_spi_write (target=<optimized out>, command=<optimized out>, address=<optimized out>, buffer=<optimized out>, 
    length=<optimized out>) at target/rp.c:550
#10 0x000055555556fd45 in bmp_spi_flash_write (flash=0x5555555cd7c0, dest=<optimized out>, src=0x5555555e0100, length=4096)
    at target/spi.c:198
#11 0x0000555555577c5c in flash_buffered_flush (flash=0x5555555cd7c0) at target/target_flash.c:206
#12 flash_buffered_flush (flash=0x5555555cd7c0) at target/target_flash.c:191
#13 0x0000555555577f44 in flash_buffered_write (len=64, src=0x5555555a34b5 <pbuf+917>, dest=268439552, flash=0x5555555cd7c0)
    at target/target_flash.c:224
#14 target_flash_write (target=0x5555555d7a50, dest=268438656, src=0x5555555a3135 <pbuf+21>, len=len@entry=960)
    at target/target_flash.c:288
#15 0x0000555555564030 in handle_v_packet (plen=1024, packet=0x5555555a3120 <pbuf> "vFlashWrite:10000c80:\233\030\006\232bD\232B")
    at gdb_main.c:678
#16 gdb_main_loop (tc=tc@entry=0x5555555a23e0 <gdb_controller>, 
    pbuf=pbuf@entry=0x5555555a3120 <pbuf> "vFlashWrite:10000c80:\233\030\006\232bD\232B", pbuf_size=pbuf_size@entry=1024, 
    size=<optimized out>, in_syscall=in_syscall@entry=false) at gdb_main.c:358
#17 0x0000555555564156 in gdb_main (pbuf=pbuf@entry=0x5555555a3120 <pbuf> "vFlashWrite:10000c80:\233\030\006\232bD\232B", 
    pbuf_size=pbuf_size@entry=1024, size=<optimized out>) at gdb_main.c:735
#18 0x000055555555b842 in bmp_poll_loop () at main.c:69
#19 main (argc=<optimized out>, argv=<optimized out>) at main.c:85
```

(Note the calling rp.c target here is different code to "main", but the trigger is just calling `target_mem_write` with a large-ish length value.)

</details>

Includes some refactoring to try and wrap these header sizes in constants.

Thanks for all your work on the Black Magic project! :)

### DIFF
--- a/src/platforms/hosted/cmsis_dap.c
+++ b/src/platforms/hosted/cmsis_dap.c
@@ -127,6 +127,21 @@ static size_t mbslen(const char *str)
 	return result;
 }
 
+// Return maximum length in bytes that can be sent in the 'data' payload of a
+// DAP transfer, given the interface type and (provided) DAP command header
+// size.
+static inline size_t dap_max_transfer_data(size_t command_header_len)
+{
+	const size_t result = report_size - command_header_len;
+
+	// Allow for an additional byte of payload overhead
+	// when sending data in HID Report payloads
+	if (type == CMSIS_TYPE_HID)
+		return result - 1U;
+
+	return result;
+}
+
 #ifdef __linux__
 static void dap_hid_print_permissions_for(const hid_device_info_s *const dev)
 {
@@ -379,6 +394,7 @@ void dap_exit_function(void)
 ssize_t dbg_dap_cmd_hid(const uint8_t *const request_data, const size_t request_length, uint8_t *const response_data,
 	const size_t response_length)
 {
+	// Need room to prepend HID Report ID byte
 	if (request_length + 1U > report_size) {
 		DEBUG_ERROR(
 			"Attempted to make over-long request of %zu bytes, max length is %zu\n", request_length + 1U, report_size);
@@ -386,7 +402,7 @@ ssize_t dbg_dap_cmd_hid(const uint8_t *const request_data, const size_t request_
 	}
 
 	memset(buffer + request_length + 1U, 0xff, report_size - (request_length + 1U));
-	buffer[0] = 0x00; // Report ID??
+	buffer[0] = 0x00; // Report ID
 	memcpy(buffer + 1, request_data, request_length);
 
 	const int result = hid_write(handle, buffer, report_size);
@@ -484,7 +500,7 @@ static void dap_mem_read(adiv5_access_port_s *ap, void *dest, uint32_t src, size
 		return;
 	}
 	/* Otherwise proceed blockwise */
-	const size_t blocks_per_transfer = (report_size - 4U) >> 2U;
+	const size_t blocks_per_transfer = dap_max_transfer_data(DAP_CMD_BLOCK_READ_HDR_LEN) >> 2U;
 	uint8_t *const data = (uint8_t *)dest;
 	for (size_t offset = 0; offset < len;) {
 		/* Setup AP_TAR every loop as failing to do so results in it wrapping */
@@ -521,7 +537,7 @@ static void dap_mem_write(adiv5_access_port_s *ap, uint32_t dest, const void *sr
 		return;
 	}
 	/* Otherwise proceed blockwise */
-	const size_t blocks_per_transfer = (report_size - 4U) >> 2U;
+	const size_t blocks_per_transfer = dap_max_transfer_data(DAP_CMD_BLOCK_WRITE_HDR_LEN) >> 2U;
 	const uint8_t *const data = (const uint8_t *)src;
 	for (size_t offset = 0; offset < len;) {
 		/* Setup AP_TAR every loop as failing to do so results in it wrapping */
--- a/src/platforms/hosted/dap_command.c
+++ b/src/platforms/hosted/dap_command.c
@@ -193,7 +193,7 @@ bool perform_dap_transfer_block_write(
 
 	dap_transfer_block_response_write_s response;
 	/* Run the request having set up the request buffer */
-	if (!dap_run_cmd(&request, 5U + (block_count * 4U), &response, sizeof(response)))
+	if (!dap_run_cmd(&request, DAP_CMD_BLOCK_WRITE_HDR_LEN + (block_count * 4U), &response, sizeof(response)))
 		return false;
 
 	/* Check the response over */
--- a/src/platforms/hosted/dap_command.h
+++ b/src/platforms/hosted/dap_command.h
@@ -117,11 +117,15 @@ typedef struct dap_transfer_block_request_write {
 	uint8_t data[256][4];
 } dap_transfer_block_request_write_s;
 
+#define DAP_CMD_BLOCK_WRITE_HDR_LEN offsetof(dap_transfer_block_request_write_s, data)
+
 typedef struct dap_transfer_block_response_read {
 	uint8_t count[2];
 	uint8_t status;
 	uint8_t data[256][4];
 } dap_transfer_block_response_read_s;
+
+#define DAP_CMD_BLOCK_READ_HDR_LEN offsetof(dap_transfer_block_response_read_s, data)
 
 typedef struct dap_transfer_block_response_write {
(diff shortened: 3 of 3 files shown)


src/platforms/hosted/cmsis_dap.c
```diff
@@ -484,7 +484,7 @@ static void dap_mem_read(adiv5_access_port_s *ap, void *dest, uint32_t src, size
 		return;
 	}
 	/* Otherwise proceed blockwise */
-	const size_t blocks_per_transfer = (report_size - 4U) >> 2U;
+	const size_t blocks_per_transfer = (report_size - 1 - DAP_CMD_BLOCK_READ_HDR_LEN) >> 2U;
```
dragonmux, line 487: Why the `- 1` (which should have a U on it anyway to stop the signed-unsigned conversion) here? We agree with the macros, their introduction and value, but we don't see why you're subtracting 1 here and what that's hoping to accomplish (same for the one below)

src/platforms/hosted/dap.h
```diff
@@ -66,6 +66,9 @@ typedef enum dap_led_type {
 
 #define DAP_QUIRK_NO_JTAG_MUTLI_TAP (1U << 0U)
 
+#define DAP_CMD_BLOCK_WRITE_HDR_LEN 5U // Size of block request write header
```
dragonmux, line 69: Rather than introducing these in the dap.h header, given they're internal to the DAP implementation would it not be better to put them in dap_command.h? That also prevents the cross-linking of the headers

dragonmux (changes requested):
There are a couple of things we've spotted in review that need answering at least - with those taken care of, we'll be happy to merge this and fix the oops made.

src/platforms/hosted/dap.h
projectgus: That does seem like a much better idea, can define them using `offsetof()` as well and remove the static asserts!

I think I got confused here about the layering (which layer is meant to depend on which). Will change.

src/platforms/hosted/cmsis_dap.c
projectgus: Good question! `dbg_dap_cmd_hid` needs one byte for the HID Report ID, so actually this byte is only needed in HID mode not bulk mode.

I'll make this self-documenting and update the PR.

src/platforms/hosted/dap.h
projectgus: Done!

src/platforms/hosted/cmsis_dap.c
projectgus: @dragonmux PTAL. This makes some behaviour changes that I'm not able to test here (the maximum transfer length for Bulk type CMSIS adapters is now one byte longer.)

I am starting to wonder if it's too fiddly. Another way to do it would be `#define DAP_CMD_BLOCK_HDR_OVERHEAD 6U` and apply this offset everywhere - worst case, some transfers are a couple of bytes shorter than the hardware would have accepted. Let me know what you think :)

src/platforms/hosted/cmsis_dap.c
```diff
@@ -127,6 +127,22 @@ static size_t mbslen(const char *str)
...
+// Return maximum length in bytes that can be sent in the 'data' payload of a
+// DAP transfer, given the interface type and (provided) DAP command header
+// size.
+static inline size_t dap_max_transfer_data(size_t command_header_len)
+{
+	size_t result = report_size - command_header_len;
+
+	// Allow for an additional byte of payload overhead
+	// when sending data in HID Report payloads
+	if (type == CMSIS_TYPE_HID) {
```
dragonmux, line 139: The braces here are superfluous so please remove them. `result` could also be made const if the body of this statement was changed to `return result - 1U;`. Looks good otherwise

src/platforms/hosted/cmsis_dap.c
dragonmux: We're easily able to test CMSIS-DAP v2 via our ORBTrace unit, so this is fine - if this works on v1 devices (HID), then we'll do the testing leg for this.

The LPC4370 and ORBTrace makes for a great test case for protocol changes like these :)

src/platforms/hosted/cmsis_dap.c
projectgus: Sure, done!

dragonmux (approved):
LGTM. We'll get the v2 (bulk) version tested as soon as possible and get this merged once we have - assuming nothing falls out from the tests. Nice work, thank you for the contribution!

src/platforms/hosted/cmsis_dap.c
projectgus: Yes, I've been testing with a DAP-Link so CMSIS-DAP V1. The write length bug I've been triggering by running the changes in #1611. I haven't got a way to max out the read response length but it doesn't change for HID in this PR. 

ORBTrace looks like very nice hardware, but the next debugger I buy will probably be BMP :grin:

(I'm waiting for Adafruit to restock them, as the shipping to Australia is pretty pricey otherwise.)

dragonmux:
This all seems to work correctly on ORBTrace, so we're going to merge this.

PR #1609, opened 2023-08-23
WIP: flashstub for RP2040 flashing

vesim987:
## Detailed description

TODO

## Your checklist for this pull request

* [x] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md)
* [x] I've read the [guidelines for contributing](https://github.com/blackmagic-debug/blackmagic/blob/main/CONTRIBUTING.md) to this repository
* [] It builds for hardware native (`make PROBE_HOST=native`)
* [ ] It builds as BMDA (`make PROBE_HOST=hosted`)
* [ ] I've tested it to the best of my ability
* [ ] My commit messages provide a useful short description of what the commits do

PR #1449, opened 2023-04-11
Fix: RP2040 Flash reliability

dragonmux:
## Detailed description

Performing ROM calls on the RP2040 is not completely reliable as if the device starts off in a bad state, the calls will simply fail with no real recourse. Instead of trying to hit our head on that, and noting that the ROM call code was consuming a fair whack of Flash on BMP, we've opted to rewrite the SPI Flash erase and programming routines using the same mechanism we're using to directly talk with the SPI controller as used for SFDP readout.

This PR is the culmination of that work. Tested working on both a ROM v0 and ROM v2 RP2040.

This PR does not address the secondary bug of `kill` in GDB after `load` on the RP2040 causing the target to spin off to 0xfffffffe and trap rather than execute the newly loaded program. We will address that in a follow-up PR

## Your checklist for this pull request

* [x] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md)
* [x] I've read the [guidelines for contributing](https://github.com/blackmagic-debug/blackmagic/blob/main/CONTRIBUTING.md) to this repository
* [x] It builds for hardware native (`make PROBE_HOST=native`)
* [x] It builds as BMDA (`make PROBE_HOST=hosted`)
* [x] I've tested it to the best of my ability
* [x] My commit messages provide a useful short description of what the commits do

## Closing issues

Fixes #1364

Issue #1364, opened 2023-01-22
BMP hangs after load on raspberry pico


ptillemans:
When running the master branch when I load the program after attach to the first target (mon s shows 3 targets so basic comms works) it hangs on:

```
Reading symbols from blink.elf...
(gdb) tar ext /dev/ttyACM0
Remote debugging using /dev/ttyACM0
(gdb) mon s
Available Targets:
No. Att Driver
 1      Raspberry RP2040 M0+
 2      Raspberry RP2040 M0+
 3      Raspberry RP2040 Rescue(Attach to reset!) 
(gdb) attach 1
Attaching to program: /home/pti/sandbox/pico-examples/build/blink/blink.elf, Remote target
0x10000c0a in sleep_until ()
(gdb) load
Loading section .boot2, size 0x100 lma 0x10000000
```

then it hangs....

When I revert to the version with tag v1.8.2 it succeeds just fine:

```
Reading symbols from blink.elf...
(gdb) tar ext /dev/ttyACM0
Remote debugging using /dev/ttyACM0
(gdb) mon s
Available Targets:
No. Att Driver
 1      Raspberry RP2040 M0+
 2      Raspberry RP2040 M0+
 3      Raspberry RP2040 Rescue(Attach to reset!) 
(gdb) attach 1
Attaching to program: /home/pti/sandbox/pico-examples/build/blink/blink.elf, Remote target
0x10000c0a in sleep_until ()
(gdb) load
Loading section .boot2, size 0x100 lma 0x10000000
Loading section .text, size 0x1d80 lma 0x10000100
Loading section .rodata, size 0xf8 lma 0x10001e80
Loading section .binary_info, size 0x20 lma 0x10001f78
Loading section .data, size 0x59c lma 0x10001f98
Start address 0x100001e8, load size 9524
Transfer rate: 27 KB/sec, 732 bytes/write.
(gdb) run
The program being debugged has been started already.
Start it from the beginning? (y or n) y
Starting program: /home/pti/sandbox/pico-examples/build/blink/blink.elf 
```

This is on a bluepill compiled with

```
make PROBE_HOST=swlink
```

(I ordered a real BMP to compare but in the meantime I can proceed with using the v1.8.2 version)

dragonmux:
Couple of questions to help us determine if this is a regression, etc:

[...]

PR #1611, opened 2023-08-24
WIP/Experiment: rp: Use DMA controller for faster flash operations

projectgus:
(There must be something in the air, I started fiddling with this a couple of days ago and now @vesim987 has a stub loader in #1609. :).)

## Background

I got a bit obsessed trying to find ways to improve the performance impact from the RP2040 flash fixes applied in #1449 (fixing #1364). 

I originally tried writing a stub loader to read/write DR0, however I couldn't make it particularly fast. I think this is because the SPI flash abstraction is low-level so it runs a lot of small operations (Write Enable, Read Status, etc). Running the stub loader requires pushing/popping the full CPU register state each time. Splitting out large and small SPI commands, so only large ones use the stub, would probably help but will increase code size... I'm interested to see what @vesim987 comes up with, though, maybe moving all of the SPI register operations into the stub will be faster than I imagined. :)

## Detailed description

This draft PR is an experiment in using the RP2040's DMA Controller to manage SPI flash operations, instead:

1. Write data to target RAM (similar to when using a stub).
2. Configure DMA controller to send them to the SPI flash and read any results.
3. Wait for DMA transfers to complete.
4. Read back any read data from target RAM.

I hoped that this would need fewer SWD operations, and both "large" and "small" commands could use the same code path.

It's still not as fast as I'd like it to be, though... :thinking: 

## Performance

| Commit | stlink target | hosted target (+ DAP-Link adapter) | 
| ---- | ---- | ---- |
| current main | 2KB/s | 1KB/s |
| pre-#1449 | 49KB/s | |
| this PR | 18KB/s | 10KB/s |
| this PR + `mon freq 60000000` | 26KB/s |

(Benchmarked by running `load` in gdb with a 300KB-ish Micropython rp2 ELF file, looking at the speed reported by gdb.)

## Code size impact

At the moment this patch adds 192 bytes to the `stlink` binary size. This at least keeps it smaller than pre-#1449 (I think about 500 bytes saved there). But some things are not yet implemented...

## TODOs

* [ ] Save and restore the DMA controller register state.
* [ ] Do we need to save and restore the working area RAM? I noticed the EFM32 and LMI target code doesn't do this, but wanted to ask.
* [ ] Clean up a bunch of commented out debugging code.
* [ ] For some weird reason, SFDP reads return all 00s with this branch. I can't work out why, other read commands seem to work. Have ordered an RP2040 board with a bigger flash chip (SOIC-8) so I can easily attach a logic analyzer clip and see exactly what is going on.
* [ ] The RP2040 datasheet implies that the hardware can manage the CS pin, which would remove two SWD operations if so. (Maybe not in bidirectional mode though, idk.)

### DIFF
--- a/src/target/rp.c
+++ b/src/target/rp.c
@@ -95,7 +95,10 @@
 #define RP_SSI_TXFLR                           (RP_SSI_BASE_ADDR + 0x20U)
 #define RP_SSI_RXFLR                           (RP_SSI_BASE_ADDR + 0x24U)
 #define RP_SSI_SR                              (RP_SSI_BASE_ADDR + 0x28U)
+#define RP_SSI_RISR                            (RP_SSI_BASE_ADDR + 0x34U)
 #define RP_SSI_ICR                             (RP_SSI_BASE_ADDR + 0x48U)
+#define RP_SSI_DMACR                           (RP_SSI_BASE_ADDR + 0x4CU)
+#define RP_SSI_DMATDLR                         (RP_SSI_BASE_ADDR + 0x50U)
 #define RP_SSI_DR0                             (RP_SSI_BASE_ADDR + 0x60U)
 #define RP_SSI_XIP_SPI_CTRL0                   (RP_SSI_BASE_ADDR + 0xf4U)
 #define RP_SSI_CTRL0_FRF_MASK                  0x00600000U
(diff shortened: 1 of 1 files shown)


projectgus:
I had another thought, that maybe the better way to do this is to split the `cortexm_run_stub` function into `start stub` and `finish stub` functions, so the stub starts running on CPU when the flash session starts and doesn't stop until the flash session is being cleaned up.

Then it can spin on a "control" memory word (or register) that communicates "write N bytes from the buffer, then read N bytes into the buffer" each time. This potentially gets the per-transfer overhead down to just the length of the bytes transferred plus the control word... :thinking:

(The other thought I had is that perhaps at this point the performance bottleneck is doing individual sector erases, so implementing block erase might be a big boost.)

dragonmux:
Given we hit 12KB/s from BMDA with vesim's PR now it's been straightened out and that's what's landed on `main`, we're going to respectfully close this PR - if you can show this approach can actually be faster using less Flash then you'd be welcome to open a new PR to replace the speed up approach we've just landed.

projectgus:
No problems, thanks @dragonmux . Sorry I never got back to this draft.

dragonmux:
No worries, thank you for proposing a very interesting and novel approach to the problem and taking the time to share 🙂

PR #1612, opened 2023-08-24
samx5x: don't assume that flash block size == flash region lock size

Qyriad:
This fixes flashing on samx5x chips with larger amounts of flash (e.g. 512KiB).

## Detailed description

Currently, we lock and unlock each flash region multiple times when the flash block size differs from the size of a protected flash region — i.e. the size of flash encoded by a single bit in the RUNLOCK register.

However, the RUNLOCK register also reads as "locked" while a write is still in progress, and we check that before each subsequent write. When we also manually lock after each block, and the block is smaller than a protected region, we misinterpret our own lock as an indication that the write is still in progress, and error out.

### DIFF
--- a/src/target/samx5x.c
+++ b/src/target/samx5x.c
@@ -474,11 +474,16 @@ static bool samx5x_flash_erase(target_flash_s *f, target_addr_t addr, size_t len
 		return false;
 	}
 
+	bool is_first_section = true;
+
 	for (size_t offset = 0; offset < len; offset += f->blocksize) {
 		target_mem_write32(t, SAMX5X_NVMC_ADDRESS, addr + offset);
 
-		/* Unlock */
-		samx5x_unlock_current_address(t);
+		/* If we're about to touch a new flash region, unlock it. */
+		if (is_first_section || (offset % lock_region_size) == 0) {
+			samx5x_unlock_current_address(t);
+			is_first_section = false;
+		}
 
 		/* Issue the erase command */
 		target_mem_write32(t, SAMX5X_NVMC_CTRLB, SAMX5X_CTRLB_CMD_KEY | SAMX5X_CTRLB_CMD_ERASEBLOCK);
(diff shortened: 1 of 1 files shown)


src/target/samx5x.c
```diff
@@ -474,11 +474,16 @@ static bool samx5x_flash_erase(target_flash_s *f, target_addr_t addr, size_t len
...
 
+	bool is_first_section = true;
+
 	for (size_t offset = 0; offset < len; offset += f->blocksize) {
 		target_mem_write32(t, SAMX5X_NVMC_ADDRESS, addr + offset);
 
-		/* Unlock */
-		samx5x_unlock_current_address(t);
+		/* If we're about to touch a new flash region, unlock it. */
+		if (is_first_section || ((offset % lock_region_size) == 0)) {
```
dragonmux, line 483: The parens around the second half of the check are not required, please drop them like as in:
```suggestion
		if (is_first_section || (offset % lock_region_size) == 0) {
```

They suppress the GCC warning for accidental assignment which we want here.

src/target/samx5x.c
```diff
@@ -496,8 +501,11 @@ static bool samx5x_flash_erase(target_flash_s *f, target_addr_t addr, size_t len
 			return false;
 		}
 
-		/* Lock */
-		samx5x_lock_current_address(t);
+		/* If we've just finished writing to a flash region, lock it. */
+		size_t next_offset = offset + f->blocksize;
```
dragonmux, line 505: This can be `const`, let's make it `const`.

src/target/samx5x.c
```diff
@@ -496,8 +501,11 @@ static bool samx5x_flash_erase(target_flash_s *f, target_addr_t addr, size_t len
...
+		if ((next_offset % lock_region_size) == 0) {
```
dragonmux, line 506: The braces here are superfluous, please remove them.

dragonmux (changes requested):
Nicely caught and figured out! There are a couple of stylistic changes to be made please, and then we'll be happy to merge this.

ktemkin:
Feel free to make whatever stylistic changes you want :) 

We won't have time to sit down with @Qyriad and update this for a bit. [Already had to debug my debugger; now I need to get back to debugging my app. ;)]

dragonmux:
Ok, thank you - we'll get on that shortly then so we can get this merged :)

dragonmux (approved):
We'll merge this as soon as the builds complete - we kept the changes strictly to the review items

PR #1614, opened 2023-08-27
Fixes for CMSIS_DAP

UweBonnes:
main with MCU-Link V3.108 with a DPV2 target prints tons of error message. MCU-Link V2.250 does not find aDPV2 target

-  Fix decoding of the MCU-Link V2.250 version string
- Provide code for waking up from dormant state for CMSIS-DAP < V1.2
- Do not use dap_write_reg_no_check if adapter does provides swd_sequence
- Configure for DAP_SWD_FAULT_NO_DATA_PHASE
-  Handle turnaround with dap_swd_seq_in and dap_swd_seq_out_parity for >= V1.2 devices
- Use raw sequences in multidrop_scan when transaction will fail for most tries or transaction does not send acknowledge by design

### DIFF
--- a/src/platforms/hosted/cmsis_dap.c
+++ b/src/platforms/hosted/cmsis_dap.c
@@ -475,7 +475,7 @@ static ssize_t dap_run_cmd_raw(const uint8_t *const request_data, const size_t r
 	const size_t result = (size_t)response;
 
 	DEBUG_WIRE("response: ");
-	for (size_t i = 0; i < result; i++)
+	for (size_t i = 0; i < response_length + 1; i++)
 		DEBUG_WIRE("%02x ", data[i]);
 	DEBUG_WIRE("\n");
 
--- a/src/platforms/hosted/cmsis_dap.c
+++ b/src/platforms/hosted/cmsis_dap.c
@@ -325,6 +325,13 @@ dap_version_s dap_adaptor_version(const dap_info_e version_kind)
 	if (!end)
 		return version;
 	version.minor = minor;
+
+	if (minor > 9 ) {
+        /* Version 1.1.0 is encoded as 1.10 on MCU-Link firmware V2.250 */
+		version.minor = minor / 10;
+		version.revision = minor % 10;
+		return version;
+	}
 	/* Check if it's worth trying to convert anything more */
 	if ((size_t)(end - version_str) >= version_length || end[0] != '.')
 		return version;
--- a/src/platforms/hosted/dap_swd.c
+++ b/src/platforms/hosted/dap_swd.c
@@ -67,7 +67,7 @@ bool dap_swd_init(adiv5_debug_port_s *target_dp)
 	/* Mark that we're going into SWD mode and configure the CMSIS-DAP adaptor accordingly */
 	dap_disconnect();
 	dap_mode = DAP_CAP_SWD;
-	dap_swd_configure(DAP_SWD_TURNAROUND_1_CYCLE, DAP_SWD_FAULT_ALWAYS_DATA_PHASE);
+	dap_swd_configure(DAP_SWD_TURNAROUND_1_CYCLE, DAP_SWD_FAULT_NO_DATA_PHASE);
 	dap_connect();
 	dap_reset_link(target_dp);
 
--- a/src/platforms/hosted/dap_swd.c
+++ b/src/platforms/hosted/dap_swd.c
@@ -50,6 +50,7 @@ typedef enum dap_swd_fault_cfg {
 static uint32_t dap_swd_seq_in(size_t clock_cycles);
 static bool dap_swd_seq_in_parity(uint32_t *result, size_t clock_cycles);
 static void dap_swd_seq_out(uint32_t tms_states, size_t clock_cycles);
+static void dap_swj_seq_out(uint32_t tms_states, size_t clock_cycles);
 static void dap_swd_seq_out_parity(uint32_t tms_states, size_t clock_cycles);
 
 static bool dap_swd_configure(dap_swd_turnaround_cycles_e turnaround, dap_swd_fault_cfg_e fault_cfg);
@@ -73,14 +74,16 @@ bool dap_swd_init(adiv5_debug_port_s *target_dp)
 	/* Set up the underlying SWD functions using the implementation below */
 	swd_proc.seq_in = dap_swd_seq_in;
 	swd_proc.seq_in_parity = dap_swd_seq_in_parity;
-	swd_proc.seq_out = dap_swd_seq_out;
 	swd_proc.seq_out_parity = dap_swd_seq_out_parity;
 
 	/* If we have SWD sequences available, make use of them */
-	if (dap_has_swd_sequence)
+	if (dap_has_swd_sequence) {
+		swd_proc.seq_out = dap_swd_seq_out;
 		target_dp->dp_low_write = dap_write_reg_no_check;
-	else
+	} else {
+		swd_proc.seq_out = dap_swj_seq_out;
 		target_dp->dp_low_write = NULL;
+	}
 	/* Set up the accelerated SWD functions for basic target operations */
 	target_dp->dp_read = dap_dp_read_reg;
 	target_dp->error = dap_swd_clear_error;
@@ -117,6 +120,15 @@ static void dap_swd_seq_out(const uint32_t tms_states, const size_t clock_cycles
 	/* And perform it */
 	if (!perform_dap_swd_sequences(&sequence, 1U))
 		DEBUG_ERROR("dap_swd_seq_out failed\n");
+}
+
+/*Fallback when SWD sequence to implemented*/
+static void dap_swj_seq_out(const uint32_t tms_states, const size_t clock_cycles)
+{
+	const uint8_t data[4] = {(tms_states >> 0) & 0xff, (tms_states >> 8) & 0xff,
+		(tms_states >> 16) & 0xff, (tms_states >> 24) & 0xff};
+	if (!perform_dap_swj_sequence(clock_cycles, data))
+		DEBUG_ERROR("dap_swj_seq_out failed\n");
 }
 
 static void dap_swd_seq_out_parity(const uint32_t tms_states, const size_t clock_cycles)
--- a/src/platforms/hosted/dap_swd.c
+++ b/src/platforms/hosted/dap_swd.c
@@ -248,7 +248,7 @@ static uint32_t dap_swd_clear_error(adiv5_debug_port_s *const target_dp, const b
 		 * into the expected state.
 		 */
 		dap_line_reset();
-		if (target_dp->version >= 2)
+		if (dap_has_swd_sequence && target_dp->version >= 2)
 			dap_write_reg_no_check(ADIV5_DP_TARGETSEL, target_dp->targetsel);
 		dap_read_reg(target_dp, ADIV5_DP_DPIDR);
 		/* Exception here is unexpected, so do not catch */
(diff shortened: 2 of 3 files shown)


src/platforms/hosted/dap_swd.c
```diff
@@ -119,59 +128,94 @@ static void dap_swd_seq_out(const uint32_t tms_states, const size_t clock_cycles
 		DEBUG_ERROR("dap_swd_seq_out failed\n");
 }
 
+/*Fallback when SWD sequence to implemented*/
+static void dap_swj_seq_out(const uint32_t tms_states, const size_t clock_cycles)
+{
+	const uint8_t data[4] = {(tms_states >> 0) & 0xff, (tms_states >> 8) & 0xff,
```
dragonmux, line 134: Don't do this manually, use the write_le4() function from buffer_utils.h (already included above).

src/platforms/hosted/dap_swd.c
```diff
@@ -236,7 +280,7 @@ static uint32_t dap_swd_clear_error(adiv5_debug_port_s *const target_dp, const b
 		 * into the expected state.
 		 */
 		dap_line_reset();
-		if (target_dp->version >= 2)
+		if (dap_has_swd_sequence && target_dp->version >= 2)
```
dragonmux, line 283: This cannot be guarded like this. Failure to execute the TARGETSEL phase on a DPv2 target *is an error*. Please fix dap_write_reg_no_check() instead.

src/platforms/hosted/dap_swd.c
```diff
@@ -119,59 +128,94 @@ static void dap_swd_seq_out(const uint32_t tms_states, const size_t clock_cycles
...
 	uint32_t data = 0;
 	for (size_t offset = 0; offset < clock_cycles; offset += 8U)
-		data |= sequence.data[offset >> 3U] << offset;
+		data |= sequences[0].data[offset >> 3U] << offset;
 	*result = data;
 	uint8_t parity = __builtin_parity(data) & 1U;
-	parity ^= sequence.data[4] & 1U;
-	return !parity;
+	parity ^= sequences[0].data[4] & 1U;
+	return parity;
```
dragonmux, line 218: The ! on the partiy check was correct, this will now fail parity checks incorrectly.

src/platforms/hosted/dap_swd.c
```diff
@@ -119,59 +128,94 @@ static void dap_swd_seq_out(const uint32_t tms_states, const size_t clock_cycles
...
+			{0},
+		},
 	};
+	swdio_status = SWDIO_STATUS_DRIVE;
 	/* And perform it */
-	if (!perform_dap_swd_sequences(&sequence, 1U)) {
+	if (!perform_dap_swd_sequences(sequences, 2U)) {
 		DEBUG_ERROR("dap_swd_seq_in_parity failed\n");
-		return false;
+		return true;
```
dragonmux, line 209: This should return false for the parity check failing. Please don't invert the logic of the returns on this function. It doesn't work this way.

src/target/adiv5_swd.c
```diff
@@ -344,8 +347,12 @@ uint32_t firmware_swdp_error(adiv5_debug_port_s *dp, const bool protocol_recover
 		 * into the expected state.
 		 */
 		swd_line_reset_sequence(true);
-		if (dp->version >= 2U)
-			firmware_dp_low_write(ADIV5_DP_TARGETSEL, dp->targetsel);
+		dp->fault = 0;
```
dragonmux, line 350: This change is not required. This code already works and the fault state properly reset below. Please revert this change.

src/target/adiv5_swd.c
```diff
@@ -291,18 +291,21 @@ void adiv5_swd_multidrop_scan(adiv5_debug_port_s *const dp, const uint32_t targe
...
+		swd_proc.seq_out(make_packet_request(ADIV5_LOW_WRITE, ADIV5_DP_TARGETSEL), 8U);
+		swd_proc.seq_in(3);
+		swd_proc.seq_out_parity(target, 32U);
+		swd_proc.seq_out(make_packet_request(ADIV5_LOW_READ, ADIV5_DP_DPIDR), 8U);
+		uint32_t ack = swd_proc.seq_in(3);
+		/* DPIDR may not issue WAIT! */
+		if ((ack & 0x7) !=SWDP_ACK_OK)
+			continue; /* No DP here, next instance */
+		uint32_t dpidr;
+		if (swd_proc.seq_in_parity(&dpidr, 32U) || (dpidr == 0))
```
dragonmux, line 307: The inner parens here are extraneous, please drop them so we get the GCC warning about accidental assignment. Please suffix all integers here with U as they're unsigned constants, so make them unsigned.

As much as we admire this attempt to fix the "SWD access resulted in no response" outputs on scan, this whole change feels very misguided, and especially as this will increase Flash footprint and is way harder to read and reason about. It's not a problem that there's a line of "SWD access resulted in no response" that gets emitted as it's being truthful about what's going on under the hood. Unless you can strongly justify this change *and include proper comments in the code about what it's doing*, we're going to ask you to revert this.

dragonmux (changes requested):
There are several serious problems with this PR that need addressing - see below. We appreciate the work that's been put into figuring this out, but it is important that it's done correctly, and not just as a hot fix for your current problem.

We will test all of this against ORBTrace, DragonProbe and a couple of other CMSIS-DAP implementations + several problem targets that this is liable to re-break.

src/platforms/hosted/cmsis_dap.c
```diff
@@ -325,6 +325,13 @@ dap_version_s dap_adaptor_version(const dap_info_e version_kind)
 	if (!end)
 		return version;
 	version.minor = minor;
+
+	if (minor > 9 ) {
```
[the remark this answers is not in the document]
UweBonnes: Minor > 9 is already a strong guard.  Look at https://arm-software.github.io/CMSIS_5/DAP/html/dap_revisionHistory.html. No minor is over 3. There are more boards, at least from NXP like the MIXRT boards with 1.10 as version. We only should reconsider when  we see an official version > 9.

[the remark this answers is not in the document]
dragonmux: We should build things to be right into the future, not patch things as they come along. Please guard this on the major version being 1 and being a MCU-Link adaptor.

src/platforms/hosted/cmsis_dap.c
```diff
@@ -468,7 +475,7 @@ static ssize_t dap_run_cmd_raw(const uint8_t *const request_data, const size_t r
 	const size_t result = (size_t)response;
 
 	DEBUG_WIRE("response: ");
-	for (size_t i = 0; i < result; i++)
+	for (size_t i = 0; i < response_length + 1; i++)
```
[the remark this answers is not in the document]
UweBonnes: result before was const size_t result = (size_t)response or 64.  Before the whole buffer was printed, now only the return code and the resulting data is printed.

src/platforms/hosted/dap_swd.c
UweBonnes: Will do.

src/platforms/hosted/cmsis_dap.c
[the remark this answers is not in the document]
dragonmux: The whole point is we want the full used portion of the buffer printed as this is useful for debugging CMSIS-DAP issues and has been used multiple times since its introduction to solve problems in both adaptors and BMDA. `result = (size_t)response` does not result in 64 under all CMSIS-DAP implementations and results in the properly sized used portion of the buffer being used on CMSIS-DAP v2 adaptors and properly implemented v1 adaptors.

Even then, `response_length` is the total length of the response buffer, adding one is strictly an error and will result in reading a byte past the end - ASAN should have told you this.

src/platforms/hosted/dap_swd.c
```diff
@@ -66,21 +73,23 @@ bool dap_swd_init(adiv5_debug_port_s *target_dp)
 	/* Mark that we're going into SWD mode and configure the CMSIS-DAP adaptor accordingly */
 	dap_disconnect();
 	dap_mode = DAP_CAP_SWD;
-	dap_swd_configure(DAP_SWD_TURNAROUND_1_CYCLE, DAP_SWD_FAULT_ALWAYS_DATA_PHASE);
+	dap_swd_configure(DAP_SWD_TURNAROUND_1_CYCLE, DAP_SWD_FAULT_NO_DATA_PHASE);
```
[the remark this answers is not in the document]
dragonmux: To explain ourself a little better here - this has nothing to do with STICKYORUN which has nothing to do with this. This has everything to do with how ADIv5.2 requires we handle the no-response error condition:

image

Strictly confirming targets /require/ we run the full data phase, and this isn't optional. That is why we set up to always run the data phase.

src/platforms/hosted/dap_swd.c
UweBonnes: It is only an error if TARGETSEL is changed from 0 to anything other than 0. It is set up to 0 already with the line reset. For an adapter < V1.2 we never get to a target not 0.

dragonmux: BMDA always sets TARGETSEL to a non-zero value for DPv2 targets, and this isn't dependant on the adaptor. It's dependant on the target. TARGETSEL is also non-zero out of reset for DPv2 targets but must also be set to the selector for the DP we want to talk to, otherwise this will incorrectly switch DPs mid communication when doing error recovery. It's straight-up incorrect to guard this like you have. Please fix `dap_write_reg_no_check()` to work with < v1.2 adaptors otherwise this logic will be incorrect and do weird (bad) things.

src/platforms/hosted/dap_swd.c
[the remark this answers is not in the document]
UweBonnes: Please show the case where we do the data phase even when we have no acknowlege but we expect an ack. For writing ADIV5_DP_TARGETSEL we do not expect an ack, as no ack is given by the specs.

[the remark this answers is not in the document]
dragonmux: RP2040, LPC4370, Tiva-C, i.MXRT, STM32H5xx - these all exhibit no-response behaviour when upset during memory IO by doing it to invalid memory addresses, or when noise on the line makes them enter a protocol error state after probe but prior to attach. All 5 of these will throw their toys with this change.

src/target/adiv5_swd.c
UweBonnes: As pointed out before, write to ADIV5_DP_TARGETSEL by design does not give an ack. On cmsis_dap < v1.2 by using a normal register access, we provoke an error that we than must recover from. Using a raw packet access without checking ack does not provoke the error.

dragonmux: This code isn't even used on CMSIS-DAP - that's `dap_swd_clear_error()` that is. It's wrong to be modifying the firmware to fix that as this doesn't change the code path in question.

src/platforms/hosted/dap_swd.c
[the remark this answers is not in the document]
dragonmux: After further conversation on the BMD channel with other contributors, we've come to the conclusion that switching to `DAP_SWD_FAULT_NO_DATA_PHASE` is the correct behaviour *but* also requires the introduction of a quirk and handling in `dap_dispatch_status()` to be able to be correct due to adaptors being out there that fail to generate no-response data phases. That solves this thread.

dragonmux:
As this PR has been unanswered for more than a year, various fixes and changes it proposed have wound up worked into the code base anyway. The only issue that this PR raises that is now not fixed as of #2035 is the CMSIS-DAP v1.x support issue. This will be addressed in a future release via fallbacks.

As a result, we are opting to close this PR as notionally fixed. If you wish to contribute the older firmware fix then please open a new PR targeting specifically that. Everything else here should be sorted and working.

PR #1615, opened 2023-08-27
Feature: Soft start for target power on BMP

### LINKED ISSUE #1567: Ramp tPWR using PWM
Problem: In some cases the Black Magic Probe can experience a brownout when enabling tPWR with a target connected. The fast transient can drag the voltage regulator down enough to reset the microcontroller.

Solution: Implement PWM ramp on the tPWR enable pin to mitigate the sharp transient.

dragonmux:
## Detailed description

In this PR we address an issue with target power bring-up that has come about on BMP v2.3b hardware. During bring-up with targets that have a heavy capacitive load, the power supply on BMP winds up dropping below the brown-out threshold and on recovering, shoots way up to 4.5V before coming back down into regulation.

This results in dangerous conditions for the processor that could lead to damage, and for the target to which we're attached. To fix this, this PR implements power-on soft start using PWM to gradually bring the target's power rail up and charge the capacitors without causing problems on our own power supply. This is accomplished with a high frequency (1.125MHz) pulse train and 32 steps from "off" to fully "on".

## Closing issues

Fixes #1567

### DIFF
--- a/src/platforms/native/platform.c
+++ b/src/platforms/native/platform.c
@@ -37,12 +37,15 @@
 #include <libopencm3/stm32/adc.h>
 #include <libopencm3/stm32/spi.h>
 #include <libopencm3/stm32/flash.h>
+#include <libopencm3/stm32/timer.h>
 
 static void adc_init(void);
 static void setup_vbus_irq(void);
 
 /* This is defined by the linker script */
 extern char vector_table;
+
+#define TPWR_SOFT_START_STEPS 64U
 
 /*
  * Starting with hardware version 4 we are storing the hardware version in the
@@ -149,8 +152,10 @@ void platform_init(void)
 	rcc_periph_clock_enable(RCC_USB);
 	rcc_periph_clock_enable(RCC_GPIOA);
 	rcc_periph_clock_enable(RCC_GPIOB);
-	if (hwversion >= 6)
+	if (hwversion >= 6) {
 		rcc_periph_clock_enable(RCC_GPIOC);
+		rcc_periph_clock_enable(RCC_TIM1);
+	}
 	rcc_periph_clock_enable(RCC_AFIO);
 	rcc_periph_clock_enable(RCC_CRC);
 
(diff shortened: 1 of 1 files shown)


esden:
I just tested this patch. I can confirm that it solves the brownout issue. But for some completely unexplained reason after applying this patch the AUX UART LED is not working any more. O_O? (I know this code does not seem to touch anything related to that ...)

esden:
Ok, it seems to be an errata issue on the GD32F103 :( ... if we want to use PB2 (LED) the PB1 (PWR_BR) has to be  Out_PP, Analog, Input_floating. It can't be set to AF_PP. :( 

See the [GD32 compatibility summary](http://easyelectronics.ru/img/ARM_kurs/STvsGD/Migration/compatibility%20sumup%20between%20GD32%20and%20STM32_V2.0.pdf) section 3.7.6.

image

This means we can only set it to be a pwm when we switch TPWR on. After that we need to set the GPIO mode to a PP or OpenDrain mode so that the UART activity LED can be used.

dragonmux:
Ah, nice catch on the errata. We'll rework the patch accordingly and include comments now we know it works more confidently.

esden:
Fantastic! Thank you very much! :)

dragonmux:
The change set has been amended to fix that errata. There was an additional resource usage conflict noted involving the timer used and Trace/SWO. We will patch for that next and then mark this PR as ready for review.

esden:
Together with @dragonmux we have amended the patch a bit more. Now we are using PushPull on the PWR_BR pin which does not leave the mosfet open for as long as driving it with OpenDrain. This results in keeping the sag of the power supply above 3V using our default 1Bitsy power load.

F9C47B73-EC89-4569-81F4-DB52954E4476_1_102_o
* The yellow trace is PWR_BR (control line of the mosfet)
* The magenta trace is the 3.3V power supply
* The blue trace is nothing, just ignore. :P

esden (approved):
LGTM

Issue #1566, opened 2023-07-28
Blink an LED until USB is enumerated


esden:
If we blink an LED until USB becomes enumerated it is clear that there is a USB connectivity issue. A common user problem is the use of charge only USB cables, that lack data lines.

It probably makes most sense to blink the RED led, possibly using the morse subsystem?

krogozinski:
@esden @dragonmux I'm keen to work on this as a first issue, if no objections from you

dragonmux:
Certainly, please do! Our mental plan for this was to start by poking in the SysTick handler and accessing the USB stack state to see if a connection has been established in whatever way we can without modifying libopencm3 if possible.

esden:
No objections. :) If you end up modifying locm3 for this, please submit your changes up stream: https://github.com/libopencm3/libopencm3 and to our fork: https://github.com/blackmagic-debug/libopencm3

ALTracer:
That is a useful feature.
I had a go at this as well. I had in mind non-blocking platform_timeout and some (most) code in super-loop, without intervening in other source files much.

Since there already are two usb_set_config callbacks (for serial and for DFU), I used `usb_get_config()` of usb_serial to check for PC presence. Alternatively, there is an unused `(void)value` in `dfu_set_config()` callback. Or you could register your 3rd callback, library ABI statically reserves 4 slots. I think it's unlikely for BMF to become a multi-config composite device on f103 platforms with usbd and 8 flexible endpoints, but it may be necessary for f4 platforms with dwc_otg and 4 pairs of endpoints to actually support traceswo capture. That is a whole different story.

I started by appending the top-most super-loop with a gpio_toggle+timeout reload on timeout expiry (for non-hosted).
However, there are problems with that, this code wasn't reached. Several routines behind `bmp_poll_loop()`, `gdb_getpacket()/gdb_if_getchar()` and `gdb_if_update_buf()` in gdb_if.c seem to busy-wait on `usb_get_config()` returning non-zero, so the super-loop actually does not run to completion for me to be able to do that. I tried altering gdb_if logic to avoid that, but it started returning \x04 symbols up its stack and other meaningless activity. `bool gdb_serial_dtr` is initialized as true which doesn't help.
[...]

PR #1616, opened 2023-08-31
Feature: Add LED feedback for USB enumeration

krogozinski:
## Detailed description

* Addresses #1566 (WIP).
* Monitors USB configuration state in `usb_config_morse_msg_update()` within the systick handler.
* Outputs a message using the morse subsystem (error LED) if USB is not configured.
* No patches to libopencm3 required.

Tested in the following scenarios:
1. USB host data lines disabled and then enabled
- Plug in to host USB hub with data lines disabled
  ```bash
  $ echo 1 | sudo tee /sys/bus/usb/devices/3-1.4/port/disable
   ```
  -  Result- error LED outputs morse message
- Enable data lines
  ```bash
  $ echo 0 | sudo tee /sys/bus/usb/devices/3-1.4/port/disable
  ```
  - Result- error LED turns off
2. Unbind and rebind host driver
- Plug in to USB hub with data lines enabled
  - Result- error LED is off
- Unbind host driver
  ```bash
  $ echo '3-1.4' | sudo tee /sys/bus/usb/drivers/usb/unbind
  ```
  -  Result- error LED outputs morse message
- Rebind host driver
  ```bash
  $ echo '3-1.4' | sudo tee /sys/bus/usb/drivers/usb/bind
  ```
  - Result- error LED turns off

### DIFF
--- a/src/platforms/common/stm32/timing_stm32.c
+++ b/src/platforms/common/stm32/timing_stm32.c
@@ -21,6 +21,7 @@
 #include "general.h"
 #include "platform.h"
 #include "morse.h"
+#include "usb.h"
 
 #include <libopencm3/cm3/systick.h>
 #include <libopencm3/cm3/nvic.h>
@@ -43,6 +44,18 @@ static uint8_t monitor_ticks = 0;
  */
 #define ADC_VREFINT_MIN 1404U
 #endif
+
+static void usb_config_morse_msg_update(void)
+{
+	if (usb_config_is_updated()) {
+		if (usb_config == 0)
+			morse("NO USB HOST.", true);
+		else
+			morse(NULL, false);
+
+		usb_config_clear_updated();
+	}
+}
 
 void platform_timing_init(void)
 {
@@ -71,6 +84,7 @@ void sys_tick_handler(void)
 	if (morse_tick >= MORSECNT) {
 		if (running_status)
 			gpio_toggle(LED_PORT, LED_IDLE_RUN);
+		usb_config_morse_msg_update();
 		SET_ERROR_STATE(morse_update());
 		morse_tick = 0;
 	} else
--- a/src/platforms/common/usb.c
+++ b/src/platforms/common/usb.c
@@ -40,6 +40,15 @@ static uint8_t usbd_control_buffer[512];
  * to fit your EP0 transactions.
  */
 
+static bool usb_config_updated = true;
+
+static void usb_config_set_updated(usbd_device *const dev, const uint16_t value)
+{
+	(void)dev;
+	(void)value;
+	usb_config_updated = true;
+}
+
 void blackmagic_usb_init(void)
 {
 	read_serial_number();
@@ -51,6 +60,7 @@ void blackmagic_usb_init(void)
 	microsoft_os_register_descriptor_sets(usbdev, microsoft_os_descriptor_sets, DESCRIPTOR_SETS);
 	usbd_register_set_config_callback(usbdev, usb_serial_set_config);
 	usbd_register_set_config_callback(usbdev, dfu_set_config);
+	usbd_register_set_config_callback(usbdev, usb_config_set_updated);
 
 	nvic_set_priority(USB_IRQ, IRQ_PRI_USB);
 	nvic_enable_irq(USB_IRQ);
@@ -64,4 +74,14 @@ uint16_t usb_get_config(void)
 void USB_ISR(void)
 {
 	usbd_poll(usbdev);
+}
+
+bool usb_config_is_updated(void)
+{
+	return usb_config_updated;
+}
+
+void usb_config_clear_updated(void)
+{
+	usb_config_updated = false;
 }
--- a/src/platforms/common/usb.h
+++ b/src/platforms/common/usb.h
@@ -58,4 +58,10 @@ void blackmagic_usb_init(void);
 /* Returns current usb configuration, or 0 if not configured. */
 uint16_t usb_get_config(void);
 
+/* Returns true if usb config has been updated. */
+bool usb_config_is_updated(void);
+
+/* Clears usb config updated flag. */
+void usb_config_clear_updated(void);
+
 #endif /* PLATFORMS_COMMON_USB_H */


src/platforms/common/usb.c
```diff
@@ -29,7 +29,7 @@
 #include "serialno.h"
 
 usbd_device *usbdev = NULL;
-uint16_t usb_config;
+uint16_t usb_config = 0;
```
perigoso, line 32: As per the standard objects with static storage are initialized to zero, no need to explicitly provide an initializer

src/platforms/native/platform.c
```diff
@@ -485,3 +487,17 @@ static void setup_vbus_irq(void)
 
 	exti15_10_isr();
 }
+
+static void wait_for_usb_host_config_request(void)
+{
+	// Toggle idle/running LED while waiting for USB serial or
+	// DFU set configuration requests
+	while (usb_get_config() == 0)
```
perigoso, line 495: I think this should be handled within the super-loop, there must be some state variable we can check to see if we have a connection an set/reset an error code with the morse subsystem somehow

dragonmux: Slightly more the point while keeping another future feature in mind - is there a way this could be done non-blocking via the SysTick handler so the firmware can do other things while the condition is true?

This would entirely block the ability for the firmware to respond to an auto-programming request which would mean rewriting it in v2.0 just to get asynchronous/non-blocking behaviour, so we'd like to explore that now instead.

src/platforms/native/platform.c
```diff
@@ -485,3 +487,17 @@ static void setup_vbus_irq(void)
...
+	{
+		// TODO use platform_timeout and poll for expiry to toggle LED
+		platform_delay(300);
+		SET_ERROR_STATE(true);
+		platform_delay(300);
+		SET_ERROR_STATE(false);
+	}
+}
```
dragonmux, line 503: Please add a new line at the end of the file, and run clang-format to format the loop to the project's code style.

krogozinski: Got it, thanks

src/platforms/native/platform.c
krogozinski: Understood, can move to SysTick handler and make non-blocking. Latest commit uses the morse subsystem but still needs to be moved to SysTick handler.

Regarding a method of checking connection state, I wanted to use an already existing API function and not intrude into how USB is configured. I have not found an API function that explicitly refers to the state of the USB connection in my exploration of the code base so far.

There are options to register callbacks for different USB events. I'm not a USB expert, but I believe the first step in an enumeration sequence is the host sending the "reset" signal. There is the option to register a callback for this using `usbd_register_reset_callback()` in `usbd.h` in libopencm3.

Similarly, a USB device will enter "suspend" mode after not receiving a start-of-frame signal for three consecutive frames. Once again, a callback can be registered using `usbd_register_suspend_callback()`. Resumption of USB comms is possible with a "resume" event. Again, a callback can be registered with `usbd_register_resume_callback()`.

I'd appreciate your feedback and I'm happy to implement a "USB connection state". Otherwise, if there is already another more suitable API function or state variable that I'm not aware of, we can simply use it instead.

src/platforms/common/usb.c
krogozinski: Understood, reverted.

src/platforms/native/platform.c
perigoso: Being in suspend mode also means we are supposed to consume very little power, so i don't think ot would be appropriate here

krogozinski: Yes, the spec suggests device current shouldn't exceed 50mA in this state. Without looking further right now, I assume the BMP reports to the host that it does not implement this functionality, and therefore the host does not request this mode.

Another possibility would be to monitor the start-of-frame count and assert a USB disconnection after an elapsed time with a static SOF count.

The original issue requests that an LED blinks until the BMP is enumerated. Checking `usb_get_config()` addresses this and works for the described use case of connecting the BMP with power-only USB cable. I'm happy to expand on this to present a "USB connection state" if you think it will be useful in the future.

dragonmux: We actually don't signal either way to the host beyond saying we're bus powered:
```
    bmAttributes         0x80
      (Bus Powered)
```

Personally, we're fine with checking if the probe is in the configured state or not as the way this is checked as this should cover USB reset events among others and requires the USB enumeration process to not only have completed but for the host to have driven the probe into the configured state and therefore have bound drivers to the interfaces it can.

krogozinski:
I've added a usb config callback `usb_config_set_updated()` in `blackmagic_usb_init()`, which sets a flag `usb_config_updated`.

This flag is monitored in a new subroutine `usb_config_morse_msg_update()` in the systick handler. The usb configuration is checked, morse message is set accordingly, and the flag is cleared.

The addition of the `usb_config_updated` flag permits the morse message to only be written when a usb config update occurs. This minimises the chance of messages from other subsystems being overwritten.

I'm not satisfied with `usb_config_morse_msg_update()` living in `timing_stm32.c`, although I'm not sure where is the best place. I see there is other functionality around voltage monitoring in the systick handler that may also be applicable.

src/platforms/common/stm32/timing_stm32.c
```diff
@@ -44,6 +45,19 @@ static uint8_t monitor_ticks = 0;
...
 #endif
 
+static void usb_config_morse_msg_update(void)
+{
+	if (usb_config_is_updated()) {
+		if (usb_config == 0) {
+			morse("NO USB HOST.", true);
+		} else {
+			morse(NULL, false);
+		}
```
dragonmux, line 55: The braces on this if-else block are unnecessary, please drop them.

dragonmux (changes requested):
Patch seems reasonable - not the approach we were originally thinking, but certainly serviceable. There's just the one style item to resolve that review picked up, and with that taken care of we'll work on testing this with the intention on merging it for v1.10 release

src/platforms/common/stm32/timing_stm32.c
krogozinski: Thanks for feedback, and apologies for wasting your time with formatting issues. I've made the correction in my branch.

I see clang-format v14.0 has a [style option](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#removebracesllvm) that can address this.

dragonmux: It's quite ok - we need to finish writing an updated style guide, so that's on us as much as anything.

While clang-format has an option, do note the rather important warning notices on that option. The some of the bad cases indicated would actually occur in BMD which is why we have chosen not to use the option. We find it easy enough to check and indicate the problem via review rather than trying to enforce it using clang-format.

dragonmux (approved):
LGTM, approved for merge pending some testing to make sure things are working as intended with this patch. Nice work and thank you for the contribution!

dragonmux:
This indeed seems to work very nicely, merging!

PR #1618, opened 2023-09-07
misc: consolidate alignment macros into single header out of adiv5

perigoso:
## Detailed description

This is a cleanup of alignment type and helper macros. the align type was moved out of the adiv5 specific files to a general header, along with it's related macros, some of which were being defined in multiple places, an additional macro was added for the common pattern of getting the minimum alignment for two given addresses.

### DIFF
--- a/src/include/align.h
+++ b/src/include/align.h
@@ -1,0 +1,48 @@
+/*
+ * This file is part of the Black Magic Debug project.
+ *
+ * Copyright (C) 2023 1BitSquared <info@1bitsquared.com>
+ * Written by Rafael Silva <perigoso@riseup.net>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its
+ *    contributors may be used to endorse or promote products derived from
+ *    this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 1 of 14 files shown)


src/include/align.h
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+#ifndef INCLUDE_ALIGN_H
+#define INCLUDE_ALIGN_H
+
+typedef enum align {
+	ALIGN_BYTE = 0,     /* 8 bit alignment */
+	ALIGN_HALFWORD = 1, /* 16 bit alignment */
+	ALIGN_WORD = 2,     /* 32 bit alignment */
+	ALIGN_DWORD = 3     /* 64 bit alignment */
```
dragonmux, line 40: We've been thinking on this a whole lot and.. as much as we appreciate that you've opted to keep the naming the same to keep the patch minimal, think we want to suggest a rename here:
`ALIGN_8BIT`, `ALIGN_16BIT`, etc, as the terms used here are super architecturally specific and opaque, reducing accessibility of the nomenclature.

What are your thoughts and would you agree?

src/target/adiv5.c
```diff
@@ -1079,10 +1077,10 @@ const void *adiv5_pack_data(const uint32_t dest, const void *const src, uint32_t
 	return (const uint8_t *)src + (1 << align);
 }
 
-void advi5_mem_read_bytes(adiv5_access_port_s *ap, void *dest, uint32_t src, size_t len)
+void advi5_mem_read_bytes(adiv5_access_port_s *const ap, void *dest, uint32_t src, size_t len)
```
dragonmux, line 1080: Might as well `const` the other parameters as much as possible if doing this.

dragonmux (changes requested):
This is looking great and provides some much needed de-duplication and definition rationalisation - thank you!

There are just a couple of items which we've seen in review to answer/change and then we can get this merged head of the v1.10 release

src/include/align.h
perigoso: Funny you suggest this because its something I really wanted to do and just opted out of doing to keep the patch less intrusive and opinionated xd

I'll do that in a bit then

src/target/adiv5.c
perigoso: I think this slipped because I was commiting a  couple concurrent change sets in parallel, I'll do that

perigoso: Nevermind, looks like I didn't miss anything, those parameters are modified in the function body

dragonmux (approved):
LGTM, just need to consider now whether to merge this in v1.10 or v2.0, but we think this is decently low risk so should go in v1.10

dragonmux:
Having re-checked over the PR, we have decided to merge it optimistically.

PR #1619, opened 2023-09-08
Feature: mass erase target API

perigoso:
## Detailed description
  
This PR is a refactor of the target mass erase API, it can be summed up in:
- deeper integration with target flash API
    Target is placed in "flash mode" (enter/exit flash mode routine) before a mass erase, the flash API is used when a dedicated mass erase routine is not provided
- all targets now support mass erase by default
   when a specialized mass erase routine is not provided the mass erase consists of erasing all the target flash' block by block via the target flash API
- flash API routine added for mass erase of individual flash', used in place of block by block erase when provided and appropriate, useful for example on targets with multiple flash banks, though none were added as part of this PR

This is not yet thoroughly tested, and due to the lack of targets at my disposal, it likely won't be adequately tested by just me.

### DIFF
--- a/src/target/lpc546xx.c
+++ b/src/target/lpc546xx.c
@@ -182,6 +182,8 @@ static void lpc546xx_reset_attach(target_s *t)
 
 static bool lpc546xx_cmd_erase_sector(target_s *t, int argc, const char **argv)
 {
+	tc_printf(t, "This command is deprecated in favor of erase_range and may be removed in the future\n");
+
 	if (argc > 1) {
 		uint32_t sector_addr = strtoul(argv[1], NULL, 0);
 		sector_addr *= t->flash->blocksize;
--- a/src/target/lpc546xx.c
+++ b/src/target/lpc546xx.c
@@ -61,7 +61,7 @@ static bool lpc546xx_cmd_write_sector(target_s *t, int argc, const char **argv);
 static void lpc546xx_reset_attach(target_s *t);
 static bool lpc546xx_flash_init(target_s *t);
 static bool lpc546xx_flash_erase(target_flash_s *f, target_addr_t addr, size_t len);
-static bool lpc546xx_mass_erase(target_s *t);
+static bool lpc546xx_mass_erase(target_s *t, platform_timeout_s *print_progess);
 static void lpc546xx_wdt_set_period(target_s *t);
 static void lpc546xx_wdt_kick(target_s *t);
 
@@ -182,8 +182,10 @@ static void lpc546xx_reset_attach(target_s *t)
 	cortexm_attach(t);
 }
 
-static bool lpc546xx_mass_erase(target_s *t)
+static bool lpc546xx_mass_erase(target_s *t, platform_timeout_s *const print_progess)
 {
+	(void)print_progess;
+
 	const int result = lpc546xx_flash_erase(t->flash, t->flash->start, t->flash->length);
 	if (result != 0)
 		tc_printf(t, "Error erasing flash: %d\n", result);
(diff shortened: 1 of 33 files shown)


perigoso:
@dragonmux Managed to rebase this, sorry for the MIA...

There have been quite a lot of changes since this was first submitted, and though I got things to build and it appears most changes should not affect it I could not meaningfully test things, so here be dragons.

src/target/lpc546xx.c
```diff
@@ -182,20 +180,13 @@ static void lpc546xx_reset_attach(target_s *t)
...
-{
-	const int result = lpc546xx_flash_erase(t->flash, t->flash->start, t->flash->length);
-	if (result != 0)
-		tc_printf(t, "Error erasing flash: %d\n", result);
-	return result == 0;
-}
-
 static bool lpc546xx_cmd_erase_sector(target_s *t, int argc, const char **argv)
 {
+	/* FIXME: is this command really needed? feels like it should at best be a generic target command */
```
dragonmux, line 185: It is probably entirely superseded by the `mon erase_range` command, which is considerably more flexible

src/target/target_flash.c
```diff
@@ -174,6 +185,65 @@ bool target_flash_erase(target_s *target, target_addr_t addr, size_t len)
...
+	if (!target_enter_flash_mode(target))
+		return false;
+
+	/* Setup progress printout */
+	platform_timeout_s print_progess;
+	platform_timeout_set(&print_progess, 500U);
+
+	bool result = false;
+	if (target->mass_erase) {
+		DEBUG_INFO("Running specialized target mass erase\n");
```
dragonmux, line 210: Given this appears to be for target debugging, perhaps let's drop it to the `DEBUG_TARGET()` log level? A user running BMD under `-v 1` to get diagnostics on information like the ROM tables for their part and general information possibly don't want diagnostics about how their part is to be mass erased, that feels more like if a user wants diagnostics on how BMD is driving their target.

src/target/target_flash.c
```diff
@@ -154,18 +154,29 @@ bool target_flash_erase(target_s *target, target_addr_t addr, size_t len)
...
 		}
 
+		/* Align the start address to the erase block size */
 		const target_addr_t local_start_addr = addr & ~(flash->blocksize - 1U);
-		const target_addr_t local_end_addr = local_start_addr + flash->blocksize;
 
-		if (!flash_prepare(flash, FLASH_OPERATION_ERASE))
+		/* Check if we can use mass erase */
+		const bool can_use_mass_erase = flash->mass_erase != NULL && local_start_addr == flash->start &&
+			(addr + len) >= (flash->start + flash->length);
```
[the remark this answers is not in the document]
perigoso: You are correct, good catch

src/target/target_flash.c
```diff
@@ -174,6 +185,65 @@ bool target_flash_erase(target_s *target, target_addr_t addr, size_t len)
 	return result;
 }
 
+static inline bool flash_manual_mass_erase(target_flash_s *const flash, platform_timeout_s *const print_progess)
+{
+	for (target_addr_t addr = flash->start; addr < (flash->start + flash->length); addr += flash->blocksize) {
```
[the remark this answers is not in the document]
perigoso: You are correct, good catch

src/target/lpc546xx.c
perigoso: Hm, I think this was a note to myself to investigate further, shouldn't have made it through to the PR

Any actions to take here?

dragonmux: Think having it display a message saying to use `mon erase_range` instead, marking it deprecated, would be a good choice - that then opens up the possibility of removing it in a next release. Possibly included in this behaviour change would be having it call the erase range machinery or making the command a no-op.

perigoso: Added a deprecation message, did not change anything with the actual command because it boils down to a `target_flash_erase` api call, which is in essence the same as erase_range already

dragonmux (approved):
All LGTM, merging. Thank you for the contribution and this huge clean-up!

PR #1623, opened 2023-09-13
Fix: honor capability bits of legacy J-Link V5 probes

ALTracer:
## Detailed description

Add support for older J-Link V5 adapters to BMDA. 
OEM considers these "legacy" https://wiki.segger.com/J-Link_BASE_V5_or_lower , in fact everything pre-V11.
IMO these are still okay for raw sequences and "accelerated bit-banging" of conventional good old JTAG, fixed-direction bus transceivers included.

* This does not look like a new feature.
* Current Blackmagic Hosted App crashes V5 hardware (2008 firmware) adapters (but works with JLink V8 hw 2014 fw)
* This pull request adds a few checks regarding "multiple interfaces" capability bits to keep BMDA+JLink V5 always in JTAG mode.

I tested this on V5.4 and V5.3 yellow "Kickstart" adapters from 2006 and now they at least work. Also tested on V8 (same AT91SAM7S64, more bus transceivers for bidirectional SWD), no regressions, both SWD & JTAG are accessible and working.
Code loosely follows similar checks from openocd+libjaylink, published UM08001 and aims to be a follow-up to #1548. 
The firmware in question does not support 0xc7 "EMU_CMD_SELECT_IF" commands at all. I can remove irrelevant parts of in-code comments on request.

* [x] It builds for hardware native (`make PROBE_HOST=native`) -- not applicable

## Closing issues
I would open an issue about this back in August, but instead I provide a suggested fix here.

### DIFF
--- a/src/platforms/hosted/jlink.c
+++ b/src/platforms/hosted/jlink.c
@@ -332,6 +332,10 @@ static inline bool jlink_interface_available(const uint8_t interface)
 
 static uint8_t jlink_selected_interface(void)
 {
+	/* V5.4 does only JTAG and hangs on 0xc7 commands */
+	if (!(jlink.capabilities & JLINK_CAPABILITY_INTERFACES))
+		return JLINK_INTERFACE_JTAG;
+
 	uint8_t buffer[4U];
 	if (!jlink_simple_request_8(JLINK_CMD_INTERFACE_GET, JLINK_INTERFACE_GET_CURRENT, buffer, sizeof(buffer)))
 		return UINT8_MAX; /* Invalid interface, max value is 31 */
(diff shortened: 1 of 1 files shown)


src/platforms/hosted/jlink.c
```diff
@@ -364,11 +368,20 @@ bool jlink_select_interface(const uint8_t interface)
...
 	uint8_t buffer[4U];
-	if (!jlink_simple_request_8(JLINK_CMD_INTERFACE_GET, JLINK_INTERFACE_GET_AVAILABLE, buffer, sizeof(buffer)))
-		return false;
+	/*
+	 * HW V5.40 IAR-KS 2006 hangs on 0xc7 commands (0xff and 0xfe),
+	 * and is known to not implement SWD or anything else besides JTAG.
+	 * Check the dedicated capability bit
+	 */
+	if (!(jlink.capabilities & JLINK_CAPABILITY_INTERFACES))
+		jlink.available_interfaces = JLINK_INTERFACE_AVAILABLE(JLINK_INTERFACE_JTAG);
```
perigoso, line 377: I would do an early return here instead of branching with the else for the remainder

perigoso (approved):
LGTM

src/platforms/hosted/jlink.c
dragonmux: Any early return would bypass the selected interface logic below and the resulting print-out of the interfaces. We think that the logic in this PR is fine based on that reading.

dragonmux (approved):
Patch looks good. Dependant on perigoso's response to the review comment on the `jlink_select_interface()` change, LGTM to merge.

src/platforms/hosted/jlink.c
perigoso: My bad, you're right. I didn't parse what i was reading correctly, quirks of the GH mobile interface...

LGTM

PR #1625, opened 2023-09-18
rtt: fix hang when aux serial not read

### LINKED ISSUE #1488: RTT output hangs the BMP when not read by host
There seems to be a problem with RTT is the host is not reading the data out.

I have a target image that is running RTT and generates a fair amount of output. If I am running a monitor program on the RTT serial port, the output works well and there is no problem. Additionally if I do not enable RTT in the BMP, there is no problem. However, in the case where RTT is enabled but not being read on the host system, the BMP firmware hangs and becomes unresponsive.

This issues is caused because the first UART packet is loaded into the USB endpoint buffer and the buffer is never taken by the host. When the next rtt polling occurs, the rtt data is read but the rtt_write function gets stuck in the while loop attempting to write the packet to the endpoint buffer.  As the buffer has not been (and likely won't be) consumed by the host, this loop can never succeed and the device is hung. Further, because the BMP is stuck in this loop, it is unable to service the requests required to open the rtt serial port leaving the only recovery method to be a BMP power cycle.

koendv:
## Detailed description

This PR silently drops rtt output if the aux serial is unresponsive for 10ms.
Fixes #1488 RTT output hangs the BMP when not read by host.
Tested on blackpill-f411.

A more general solution would be to append the output of usb serial, traceswo decode, rtt, and redirect_stdout to a single circular buffer, and use the usb tx interrupt to trigger usb writes.

## Closing issues

fixes #1488

### DIFF
--- a/src/platforms/common/stm32/rtt_if.c
+++ b/src/platforms/common/stm32/rtt_if.c
@@ -121,8 +121,11 @@ uint32_t rtt_write(const char *buf, uint32_t len)
 	if (len != 0 && usbdev && usb_get_config() && gdb_serial_get_dtr()) {
 		for (uint32_t p = 0; p < len; p += CDCACM_PACKET_SIZE) {
 			uint32_t plen = MIN(CDCACM_PACKET_SIZE, len - p);
-			while (usbd_ep_write_packet(usbdev, CDCACM_UART_ENDPOINT, buf + p, plen) <= 0)
-				continue;
+			uint32_t start_ms = platform_time_ms();
+			while (usbd_ep_write_packet(usbdev, CDCACM_UART_ENDPOINT, buf + p, plen) <= 0) {
+				if (platform_time_ms() - start_ms >= 25)
+					return 0; /* drop silently */
+			}
 		}
 		/* flush 64-byte packet on full-speed */
 		if (CDCACM_PACKET_SIZE == 64 && (len % CDCACM_PACKET_SIZE) == 0)


src/platforms/common/stm32/rtt_if.c
```diff
@@ -121,8 +121,11 @@ uint32_t rtt_write(const char *buf, uint32_t len)
 	if (len != 0 && usbdev && usb_get_config() && gdb_serial_get_dtr()) {
 		for (uint32_t p = 0; p < len; p += CDCACM_PACKET_SIZE) {
 			uint32_t plen = MIN(CDCACM_PACKET_SIZE, len - p);
-			while (usbd_ep_write_packet(usbdev, CDCACM_UART_ENDPOINT, buf + p, plen) <= 0)
-				continue;
+			uint32_t start_ms = platform_time_ms();
+			while (usbd_ep_write_packet(usbdev, CDCACM_UART_ENDPOINT, buf + p, plen) <= 0) {
+				if (platform_time_ms() - start_ms >= 10)
```
dragonmux, line 126: Does it make sense to wait a little longer - say, 25 or 50ms - in case the host happens to be busy and a little slow? We'd be concerned with 10ms not being long enough and causing the possibility of dropping data wrongly just because things got a little busy on the USB link.

dragonmux (commented):
We've got one query, but other than that this looks good. Once the query's been answered and any required actions resulting taken, we'll happily approve this for merge.

koendv:
OK. Let's set the aux serial write timeout to 25ms.
I prefer not to set the timeout too high, else if rtt is active and the usb aux serial is not being read, bmp might spend too much time busy-waiting.

dragonmux (approved):
Sounds good, and LGTM. Merging once the builds finish.

PR #1626, opened 2023-09-19
Fix: Cortex-A support

dragonmux:
## Detailed description

In this PR we set out to complete the replacement of the `apb_read()` and `apb_write()` functions in the Cortex-A code using the newer `cortex_dbg_read32()` and `cortex_dbg_write32()` functions as the former functions make gross assumptions about the state and configuration of the AP that no longer hold.

This had resulted in scan-then-crash issues on targets and would have caused issues on any part that has both a Cortex-M and Cortex-A core on the same Access Port as the Cortex-M code would have caused the same state inconsistency problem.

This also reduces the Flash footprint for the Cortex-A support by ~200 bytes.

### DIFF
--- a/src/target/cortexa.c
+++ b/src/target/cortexa.c
@@ -82,10 +82,10 @@ typedef struct cortexa_priv {
 } cortexa_priv_s;
 
 #define CORTEXAR_DBG_IDR   0x000U
-#define CORTEXAR_DBG_DTRTX 0x080U
+#define CORTEXAR_DBG_DTRTX 0x080U /* DBGDTRRXext */
 #define CORTEXAR_DBG_ITR   0x084U
 #define CORTEXAR_DBG_DSCR  0x088U
-#define CORTEXAR_DBG_DTRRX 0x08cU
+#define CORTEXAR_DBG_DTRRX 0x08cU /* DBGDTRTXext */
 #define CORTEXAR_DBG_DRCR  0x090U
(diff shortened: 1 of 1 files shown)


esden (approved):
LGTM, thanks! :)

src/target/cortexa.c
```diff
@@ -546,10 +570,10 @@ bool cortexa_attach(target_s *target)
 	target_check_error(target);
 
 	/* Enable halting debug mode */
-	uint32_t dbgdscr = apb_read(target, DBGDSCR);
-	dbgdscr |= DBGDSCR_HDBGEN | DBGDSCR_ITREN;
-	dbgdscr = (dbgdscr & ~DBGDSCR_EXTDCCMODE_MASK) | DBGDSCR_EXTDCCMODE_STALL;
-	apb_write(target, DBGDSCR, dbgdscr);
+	uint32_t dbgdscr = cortex_dbg_read32(target, CORTEXAR_DBG_DSCR);
+	dbgdscr |= CORTEXAR_DBG_DSCR_ITR_ENABLE | CORTEXAR_DBG_DSCR_ITR_ENABLE;
```
ALTracer, line 574: to be fixed in PR1628, should be `CORTEXAR_DBG_DSCR_HALT_DBG_ENABLE | CORTEXAR_DBG_DSCR_ITR_ENABLE`

PR #1628, opened 2023-09-22
Fix: Avoid a hang in scanning Cortex-A of STM32MP15x

ALTracer:
## Detailed description
BMD at current `main` busy-loops (hangs, effective DoS) if user attempts to scan a STM32MP15x device with any of supported adapters, requiring BMDA restart or probe reboot. This behaviour was enabled recently by Cortex-A related PRs. Previously BMD did not attempt to attach, halt, or otherwise touch CA7 during scan/probe.

* This is more a fix than a feature, but introduces registers not touched previously.
* The pull request solves an existing problem of BMD hang upon trying to even scan STM32MP15x SoC.
* To solve this, the pull request implements checking and clearing OSLock in ARMv7-A Debug registers.

It took three consecutive days of log analysis from different adapters and gdbserver software with help and suggestions from Dragonmux to come up with some working fix.
Our working hypothesis is that even in Engineering Boot, the primary bootloader (BootROM) enables OSLock, which per Arch.ref.man. prohibits external debugger access to some registers. OpenOCD clears it, so when working through a kind of adapter both servers speak, I could observe BMDA working fine with this target *after* running OOCD once and *before* power-cycling DUT (there is no good SRST/Reset button solution without dropping Vdd+Vddcore, see errata).

I tried coding a few different things across cortexa.c and most of them ended up here. Namely, the OSLock check, the HALT_DBG_ENABLE + ITR_ENABLE typo fix, the unused LAR/DSCCR/DSMCR clears, macro definitons for these, and last but not least, my helper routine for printing human-readable names for set bits of a register to save developers from the hassle of ~~using a decoder ring~~ keeping A.R.M. and calculator open for understanding some hex dumps.

Note that because the problem happens in cortexa_probe, earlier than cortexa_attach, I had to move HDBGen+ITRen writes up there. This is less clean because the user is not guaranteed to attach (so BMD then runs detach cleanup), but I submit just the bug fix for now.

This patchset as it stands was successfully tested in operation by me against STM32MP157D-DK1 with its onboard STLink/V2-1 SWD-only, but also against custom hardware connected to {STLink/V2 standalone; JLink V8; CMSIS-DAP v1/v2 free-dap on rp2040; finally a BMP(-compatible) blackpill-f411ce}. On all debuggers the hang is not seen anymore, in GDB attach and run/continue works, proper general register values are shown. On BMP there's no hang anymore in *_scan which required pressing the probe's reset button.

I understand this patchset may be not as clean and subject to renames and nitpicks, but a) I needed to fix in VCS a known good working solution; b) this problem blocks my PR1546, in which CM4 target support code would not be reachable anymore. I may agree to rebase away all changes except OSLock which is the core change, and/or reformat the DSCR-touching stuff. 

Users of Cortex-A targets are welcome to test this patchset on their setups in case I introduce a regression for them by fixing the regression I face.

## Closing issues

Not quite fixes but helps #1546.

### DIFF
--- a/src/target/cortexa.c
+++ b/src/target/cortexa.c
@@ -330,6 +330,7 @@ static size_t create_tdesc_cortex_a(char *buffer, size_t max_len)
 	return (size_t)total;
 }
 
+#ifdef ENABLE_DEBUG
 typedef struct bitfield_entry {
 	char *desc;
 	uint8_t bitnum;
@@ -353,14 +354,13 @@ static const bitfields_lut_s cortexa_dbg_dscr_lut[] = {
 	{"RXfull", 30U},
 };
 
-static void helper_print_bitfields(const uint32_t val, const bitfields_lut_s lut[], const size_t array_length)
+static void helper_print_bitfields(const uint32_t val, const bitfields_lut_s *lut, const size_t array_length)
 {
 	for (size_t i = 0; i < array_length; i++) {
-		if (val & (1U << lut[i].bitnum)) {
+		if (val & (1U << lut[i].bitnum))
 			DEBUG_TARGET("%s ", lut[i].desc);
-		}
 	}
-};
+}
 
 static void cortexa_decode_bitfields(const uint32_t reg, const uint32_t val)
 {
@@ -377,6 +377,13 @@ static void cortexa_decode_bitfields(const uint32_t reg, const uint32_t val)
 
 	DEBUG_TARGET("\n");
 }
+#else
+static void cortexa_decode_bitfields(const uint32_t reg, const uint32_t val)
+{
+	(void)reg;
+	(void)val;
+}
+#endif
 
 static void cortexar_run_insn(target_s *const target, const uint32_t insn)
 {
@@ -579,9 +586,9 @@ bool cortexa_probe(adiv5_access_port_s *ap, target_addr_t base_address)
 #endif
 
 	uint32_t dbg_osreg = cortex_dbg_read32(target, CORTEXAR_DBG_OSLSR);
-	DEBUG_INFO("%s: DBGOSLSR = 0x%08X\n", __func__, dbg_osreg);
+	DEBUG_INFO("%s: DBGOSLSR = 0x%08" PRIx32 "\n", __func__, dbg_osreg);
 	/* Is OS Lock implemented? */
-	if (((dbg_osreg & DBGOSLSR_OSLM) == DBGOSLSR_OSLM0) || ((dbg_osreg & DBGOSLSR_OSLM) == DBGOSLSR_OSLM1)) {
+	if ((dbg_osreg & DBGOSLSR_OSLM) == DBGOSLSR_OSLM0 || (dbg_osreg & DBGOSLSR_OSLM) == DBGOSLSR_OSLM1) {
 		/* Is OS Lock set? */
 		if (dbg_osreg & DBGOSLSR_OSLK) {
 			DEBUG_WARN("%s: OSLock set! Trying to unlock\n", __func__);
--- a/src/target/cortexa.c
+++ b/src/target/cortexa.c
@@ -508,6 +508,8 @@ bool cortexa_probe(adiv5_access_port_s *ap, target_addr_t base_address)
 	target->halt_poll = cortexa_halt_poll;
 	target->halt_resume = cortexa_halt_resume;
 
+#if 0 /* Trying to halt STM32MP15x Cortex-A cores during probe locks up BMDA/BMF! */
+
 	/* Try to halt the target core */
 	target_halt_request(target);
 	platform_timeout_s timeout;
@@ -518,6 +520,7 @@ bool cortexa_probe(adiv5_access_port_s *ap, target_addr_t base_address)
 	/* If we did not succeed, we must abort at this point. */
 	if (reason == TARGET_HALT_FAULT || reason == TARGET_HALT_ERROR)
 		return false;
+#endif
 
 	cortex_read_cpuid(target);
 	/* The format of the debug identification register is described in DDI0406C §C11.11.15 pg2217 */
--- a/src/target/cortexa.c
+++ b/src/target/cortexa.c
@@ -837,6 +837,11 @@ static target_halt_reason_e cortexa_halt_poll(target_s *t, target_addr_t *watch)
 void cortexa_halt_resume(target_s *t, bool step)
 {
 	cortexa_priv_s *priv = t->priv;
+
+	uint32_t dbgdscr = cortex_dbg_read32(t, CORTEXAR_DBG_DSCR);
+	if (!(dbgdscr & CORTEXAR_DBG_DSCR_HALTED)) /* Not halted */
+		return;
+
 	/* Set breakpoint comparator for single stepping if needed */
 	if (step) {
 		uint32_t addr = priv->reg_cache.r[15];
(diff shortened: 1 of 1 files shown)


src/target/cortexa.c
```diff
@@ -91,6 +91,19 @@ typedef struct cortexa_priv {
...
+#define CORTEXAR_DBG_OSLAR 0x300U
+#define CORTEXAR_DBG_OSLSR 0x304U
+#define CORTEXAR_DBG_LAR   0xfb0U /* Lock Access */
+#define CORTEXAR_DBG_LSR   0xfb4U /* Lock Status */
+
+#define DBGOSLSR_OSLM0 (1U << 0U)
+#define DBGOSLSR_OSLK  (1U << 1U)
+#define DBGOSLSR_NTT   (1U << 2U)
+#define DBGOSLSR_OSLM1 (1U << 3U)
+#define DBGOSLSR_OSLM  (DBGOSLSR_OSLM0 | DBGOSLSR_OSLM1)
```
dragonmux, line 105: These should be prefixed with `CORTEXAR_` and the `_` between `DBG` and the register name preserved, but given how much still needs rewriting in this file, we're going to let that slide if you don't feel like fixing it - this is just a note so it's not forgotten.

src/target/cortexa.c
```diff
@@ -317,6 +330,54 @@ static size_t create_tdesc_cortex_a(char *buffer, size_t max_len)
...
+	{"MDBGen", 15U},
+	{"InstrCompl_l", 24U},
+	{"PipeAdv", 25U},
+	{"TXfull_l", 26U},
+	{"RXfull_l", 27U},
+	{"TXfull", 29U},
+	{"RXfull", 30U},
+};
+
+static void helper_print_bitfields(const uint32_t val, const bitfields_lut_s lut[], const size_t array_length)
```
dragonmux, line 356: `const bitfields_lut_s lut[]` -> `const bitfields_lut_s *lut` please because C arrays do not pass like this so this code is unintentionally lying. It's thanks to how C arrays decay to pointers.

src/target/cortexa.c
```diff
@@ -317,6 +330,54 @@ static size_t create_tdesc_cortex_a(char *buffer, size_t max_len)
...
+{
+	for (size_t i = 0; i < array_length; i++) {
+		if (val & (1U << lut[i].bitnum)) {
+			DEBUG_TARGET("%s ", lut[i].desc);
+		}
+	}
+};
```
dragonmux, line 363: Extra `;` here - please remove

src/target/cortexa.c
```diff
@@ -317,6 +330,54 @@ static size_t create_tdesc_cortex_a(char *buffer, size_t max_len)
 	return (size_t)total;
 }
 
+typedef struct bitfield_entry {
```
dragonmux, line 333: Given how much extra Flash this'll use, could this please be topped and tailed in `#ifdef ENABLE_DEBUG` (including down to the end of `cortexa_decode_bitfields()`) please?

src/target/cortexa.c
```diff
@@ -508,6 +569,45 @@ bool cortexa_probe(adiv5_access_port_s *ap, target_addr_t base_address)
...
+	/* Cache write-through */
+	cortex_dbg_write32(target, CORTEXAR_DBG_DSCCR, 0U);
+	/* Disable TLB lookup and refill/eviction */
+	cortex_dbg_write32(target, CORTEXAR_DBG_DSMCR, 0U);
+#endif
+
+	uint32_t dbg_osreg = cortex_dbg_read32(target, CORTEXAR_DBG_OSLSR);
+	DEBUG_INFO("%s: DBGOSLSR = 0x%08X\n", __func__, dbg_osreg);
+	/* Is OS Lock implemented? */
+	if (((dbg_osreg & DBGOSLSR_OSLM) == DBGOSLSR_OSLM0) || ((dbg_osreg & DBGOSLSR_OSLM) == DBGOSLSR_OSLM1)) {
```
dragonmux, line 584: There's an extra pair of parens here each side of the `||`, please remove them as in:
```c
if ((dbg_osreg & DBGOSLSR_OSLM) == DBGOSLSR_OSLM0 || (dbg_osreg & DBGOSLSR_OSLM) == DBGOSLSR_OSLM1) {
```

dragonmux (changes requested):
There's a couple of items to address, but this is looking good. Nice work figuring out how to fix what was going so wrong!

src/target/cortexa.c
ALTracer: The PR uses just extra 64..128 bytes on different all_platforms binaries, and +6 KiB on hosted. I usually trust the toolchain to drop unused functions and sections. But sure, I can hide the helper before compile-time.

src/target/cortexa.c
ALTracer: I understand the item but other bitfields just 20 lines down are written that way, and I complied with that style to not have to change all of it.

src/target/cortexa.c
ALTracer: Changed. I assumed I could use the ARRAY_LENGTH macro inside that function, but apparently passing arrays in C loses information about them. For now that's handled in the caller which passes the size alongside the array/pointer argument.

src/target/cortexa.c
ALTracer: Yes, that's totally extra. Maybe I put that there by inertia after fixing struct and array declarations/definitions.
I also took the liberty of preemptively dropping the `for` & `if` curly-braces enclosing the single statement.

src/target/cortexa.c
ALTracer: Removed. I was confused by oocd code from which I took that check.
Also I replaced that `%08X` by `%08PRIx32` just above so the ENABLE_DEBUG=1 firmware builds don't fail.

src/target/cortexa.c
dragonmux: Those other fields are going to be changed in the rewrite, follow the style set in cortexr.c from the Cortex-R support feature branch please as that is the style this code will be switching to.

src/target/cortexa.c
dragonmux: Indeed, this is tied up with the same reason writing it as an array is an unintentional lie - the declaration becomes `const bitfields_lut_s *lut` before even the function declaration has finished being parsed. This means that by the time things like ARRAY_LENGTH() get to see it, you wind up asking for the length of a pointer to a struct divided by the length of that struct, rather than anything to do with arrays.

src/target/cortexa.c
ALTracer: I can rename the 5 macros in anticipation of code style switch iff it is planned to happen before v1.10 release (and I don't know the plan).
nTT is unused, should I drop that bit?

dragonmux: It will happen after the release, but not long after - it would be better if it's done here to reduce churn and preserve attribution as much as possible.

Regarding the nTT bit, please keep it as that should become used when we extend things out to support 64-bit processors.

src/target/cortexa.c
dragonmux: You are correct on the `if`, but consider the `for` to be a multi-line block and so need braces - it's technically correct as-is, but the style used elsewhere is for such a block to be written as
```c
for (;;) {
    if (cond)
        action;
}
```

ALTracer: Partially backed out and force-pushed to comply with this.

src/target/cortexa.c
ALTracer: Rename complete.
I don't honestly like the OSLM bitmask and the (know broken into 2 lines by clang-format) check down on L591. Why did they do it that way and can I just use individual bits in the logic condition, dropping the bitmask altogether? OSLM bits [3,0] value 0b11 is reserved anyways.

dragonmux: Yeah, we couldn't figure out why that happened or what thinking was involved there either - it just makes everything a bit more difficult around those bits. We can try to refine this in v2.0 as we'd like to get this merged and the main STM32MP15 PR too - that also gives more time to do research and figure out the best approach here.

dragonmux (approved):
LGTM at least for v1.10 - as said in one of the review threads, we can refine this further as part of bringing the Cortex-A and Cortex-R support together in v2.0, so "good enough that it works" is fine for now. Merging!

PR #1629, opened 2023-09-25
Fix: Remote protocol target communications initialisation

dragonmux:
## Detailed description

This PR seeks to address a crash-inducing bug reported by ALTracer on Discord. Under certain conditions during failed communications with a target over the remote protocol, `dp->error()` is called and the remote protocol layer in the firmware was improperly initialising the faked DP structure it uses, resulting in an attempted call through a NULL function pointer.

This patch addresses this problem for both JTAG and SWD over remote protocol + documents that structure a bit to help prevent future problems.

Tested by ALTracer against the STM32MP157 with which this bug was found and triggered.

This bug has existed in the firmware since about 3 years ago and does affect v1.7, v1.8 and v1.9 - we may wish to backport it at least to v1.9 because of this.

### DIFF
--- a/src/remote.c
+++ b/src/remote.c
@@ -187,7 +187,9 @@ static void remote_packet_process_jtag(const char *const packet, const size_t pa
 {
 	switch (packet[1]) {
 	case REMOTE_INIT: /* JS = initialise ============================= */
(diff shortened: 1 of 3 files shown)


esden (approved):
LGTM

PR #1630, opened 2023-09-27
Fix: Cortex-A halt handling of OSLock becoming set between resumption and the next halt

dragonmux:
## Detailed description

In #1628 handling for the Cortex-A/R register for "OSLock" was introduced into `cortexa_probe()`, but this handling doesn't quite go far enough as it's possible for software running on the core to re-set this bit between the core being resumed and its next halt.

This PR address that by moving that logic into cortexa_halt_poll() so it gets applied every halt so we reduce the chances of getting stuck unable to set ITRen and so unable to launch instructions on the core that enable us to do debugging.

### DIFF
--- a/src/target/cortexa.c
+++ b/src/target/cortexa.c
@@ -548,6 +548,25 @@ const char *cortexa_regs_description(target_s *t)
 	return description;
 }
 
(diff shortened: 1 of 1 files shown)


esden (approved):
LGTM

PR #1640, opened 2023-10-04
Fix: swd command message swd naming

perigoso:
## Detailed description

The purpose of this PR is mainly to update the SWD naming (from SW-DP to SWD ) in the SWD scan commands, since the command itself has already been renamed, it also reflects the rename of the swd scan command in the documentation, and fixes a typo and some comment farming

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -56,7 +56,7 @@ static bool cmd_version(target_s *t, int argc, const char **argv);
 static bool cmd_help(target_s *t, int argc, const char **argv);
 
 static bool cmd_jtag_scan(target_s *target, int argc, const char **argv);
-static bool cmd_swd_scan(target_s *t, int argc, const char **argv);
+static bool cmd_swd_scan(target_s *target, int argc, const char **argv);
 static bool cmd_auto_scan(target_s *t, int argc, const char **argv);
 static bool cmd_frequency(target_s *t, int argc, const char **argv);
 static bool cmd_targets(target_s *t, int argc, const char **argv);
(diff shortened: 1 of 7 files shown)


src/command.c
```diff
@@ -177,7 +177,7 @@ bool cmd_version(target_s *t, int argc, const char **argv)
 #else
 	gdb_out(BOARD_IDENT);
 	gdb_outf(", Hardware Version %d\n", platform_hwversion());
-	gdb_out("Copyright (C) 2022 Black Magic Debug Project\n");
+	gdb_out("Copyright (C) 2023 Black Magic Debug Project\n");
```
dragonmux, line 180: Particularly good catch - though this should probably state `2010-2023` or so as the project's been around way longer and one is supposed to declare a year range

src/command.c
```diff
@@ -177,7 +177,7 @@ bool cmd_version(target_s *t, int argc, const char **argv)
 #else
```
dragonmux, line 177: The copyright and license notices should probably get copied up into the BMDA block (it needs a major rework anyway, but that's another problem)

src/target/adiv5_swd.c
```diff
@@ -109,15 +110,18 @@ static void jtag_to_swd_sequence()
...
 	/* SWD interface must be in reset state */
 	swd_line_reset_sequence(false);
 
 	/* Send the 16-bit JTAG-to-SWD select sequence on SWDIOTMS */
 	swd_proc.seq_out(ADIV5_JTAG_TO_SWD_SELECT_SEQUENCE, 16U);
 
-	/* This ensures that if SWJ-DP was already in SWD operation before sending the select sequence, the interface enters reset state */
+	/*
+	 * This ensures that if a SWD/JTAG DP was already in SWD operation before sending the select sequence,
+	/* the interface enters reset state
```
dragonmux, line 123: the / on the front here needs removing

dragonmux (changes requested):
Looking pretty good - there are a couple of tweaks it could do with but they aren't big. Let's try to land this for v1.10.

src/command.c
perigoso: Yes, I should've gone for the proper fix instead of the quick one :) done

src/target/adiv5_swd.c
perigoso: ups, good catch

dragonmux (approved):
LGTM, thank you for the quick response. Merging!

Issue #1516, opened 2023-06-17
blackpill-f4: bootloader LED issues


lenvm:
I am encountering two issues with the LED_BOOTLOADER on the blackpill-f4 platforms.

#### Description
The bootloader LED only works when it is assigned to PC13. If LED_BOOTLOADER is set to for instance PC15, as in the most recent commit (df014ff), the bootloader LED does not work.
#### Details
- I have tried assigning LED_BOOTLOADER to PC14, PC15 and PA1, but all result in the same: the LED of the respective pin does not turn on. Simultaneously, the LED_IDLE_RUN stops working if LED_BOOTLOADER is not assigned to PC13.
- I have also tried to remove all references to LED_BOOTLOADER from the code. In this case LED_IDLE_RUN works correctly on PC14, PC15 and PA1. LED_IDLE_RUN does not work when it is assigned to PC13.

The easy solution would be to swap the pin assignment of LED_BOOTLOADER and LED_IDLE_RUN, as then the LEDs are working. I would however prefer not to swap these pins, and understand what is the root cause of this issue.

#### Description
The bootloader LED works inverted to my expectation.
#### Details
- I would expect the bootloader LED to be on when the boot loader is active, and to be off when the boot loader is inactive. Instead, the bootloader LED is off when the boot loader is active, and the LED is on when the bootloader is not active.

I would appreciate your input to find out what could be the cause of and a fix for both of these issues.

ALTracer:
Platforms like `stlink` and `swlink` toggle their respective `led_bootloader`
https://github.com/blackmagic-debug/blackmagic/blob/b2156f9ee8e1245dc1749ce598c48224b71cb611/src/platforms/stlink/usbdfu.c#L85-L96
 (or outright GPIO_C13) in usbdfu.c via `systick_handler`, 
https://github.com/blackmagic-debug/blackmagic/blob/b2156f9ee8e1245dc1749ce598c48224b71cb611/src/platforms/swlink/usbdfu.c#L92-L99
and I would like to port that behaviour to blackpill-f4 platform as well. Something as simple as
```diff
diff --git a/src/platforms/common/blackpill-f4/usbdfu.c b/src/platforms/common/blackpill-f4/usbdfu.c
index a116e41e..0cc6a521 100644
--- a/src/platforms/common/blackpill-f4/usbdfu.c
+++ b/src/platforms/common/blackpill-f4/usbdfu.c
@@ -51,10 +51,16 @@ int main(void)
 
[...]

PR #1645, opened 2023-10-17
Fix: platforms/blackpill-f4: invert state of SET_IDLE_STATE

lenvm:
## Detailed description

This pull request fixes the issue that LED_IDLE_RUN is not working on the blackpill-f4 platforms when it is assigned to PC13, the onboard LED. The issue is caused by the need to set PC13 to low to turn the LED on, rather than to set PC13 to high to turn the LED on.
The root cause of this is the schematic of how the LED is connected to PC13:
3.3v -> R 1k -> Blue LED -> PC13 ([reference](https://stm32-base.org/assets/pdf/boards/original-schematic-STM32F411CEU6_WeAct_Black_Pill_V2.0.pdf)).

This pull request fixes the issue described, by inverting the state used by for SET_IDLE_STATE. This results in setting PC13 to low when the state is true, and PC13 to high when the state is false. This is exactly the behavior needed to turn the LED on when the state is true.

## Closing issues
This issue is a partial fix to #1516

### DIFF
--- a/src/platforms/common/blackpill-f4/blackpill-f4.h
+++ b/src/platforms/common/blackpill-f4/blackpill-f4.h
@@ -230,9 +230,13 @@
 	{                             \
 		running_status = (state); \
 	}
-#define SET_IDLE_STATE(state)                        \
-	{                                                \
-		gpio_set_val(LED_PORT, LED_IDLE_RUN, state); \
+/*
+ * The state of LED_IDLE_RUN is inverted, as the led used for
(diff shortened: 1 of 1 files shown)


lenvm:
@dragonmux, as this is a fix for an existing issue with the blackpill-f4 platforms, can it also be included in the v1.10.0 release?

dragonmux (approved):
Good catch! Given how simple this fix is, and how low risk, we're going to merge this as part of v1.10.0.

PR #1651, opened 2023-10-20
misc: add nix-direnv integration

wucke13:
## Detailed description

Quality of life improvement for [nix-direnv](https://github.com/nix-community/nix-direnv) users:
When such a users cd's into the blackmagic repo dir, direnv let's them know that a development environment is available. The user can opt-in to using it, which immediately imports the devshell that is described in the `shell.nix` already present in the repo. In contrast to just running `nix-shell` from the repo root, the direnv integration defaults to using the currently installed nixpkgs checkout, instead of picking the pinpointed version from the repo. Given the generic nature of the `shell.nix`, this is likely not less robust than the pinpointed version from `shell.nix`, and it saves a lot of time upon first execution, as no completely unrelated nixpkgs checkout has to be fetched.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use nix --arg pkgs 'import <nixpkgs> { }'
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,5 @@ src/artifacts/
 src/ftd2xx.dll
 
 scripts/swolisten
+
+.direnv/


dragonmux (approved):
LGTM, we'll merge this at the start of v2.0 (we are currently doing release candidates for v1.10.0 with the intent to push out a complete release very soon).

We don't use Nix so unless someone chimes in with any complaint or reason why we should not merge this as-is, that is what we will do.

PR #1652, opened 2023-10-21
Add debug and flash support for S32K344

via:
## Detailed description

This adds the ability to attach to an NXP S32K344 and basic erase/flash support. The bulk of the changes are in two parts:
- First, adiv5.c is updated with an explicit overriden discovery sequence for APs.  This chip will hard fault if we probe for one that does not exist, and there is an additional step to enable debug access.
- Second, implement a s32k3xx skeleton, and add probe and flash support for the s32k344

I have mostly tested this by some regular use as a debugger, and to test the flash I generated a 4MB (-48KB to avoid the protected region at the end) random binary file, converted to an elf, and repeatedly erased, flashed, and compare-section-ed.


src/target/adiv5.c
```diff
@@ -945,6 +1002,14 @@ void adiv5_dp_init(adiv5_debug_port_s *const dp)
 	/* Probe for APs on this DP */
 	size_t invalid_aps = 0;
 	dp->refcnt++;
+
+	if (dp->target_designer_code == JEP106_MANUFACTURER_FREESCALE)
```
dragonmux, line 1006: This outer `if` needs its braces added as the inner `if` being a multi-line block makes this a multi-line block too (technically this is valid C, but we generally consider this confusing  and try to avoid doing it).

src/target/s32k3xx.c
```diff
@@ -0,0 +1,271 @@
...
+}
+
+static bool s32k3xx_flash_write(target_flash_s *flash, target_addr_t dest, const void *src, size_t len)
+{
+	assert(len == flash->writesize);
+
+	const uint32_t *const s_data = src;
+	s32k3xx_flash_prepare(flash);
+	target_mem_write32(flash->t, PFCPGM_PEADR_L, dest);
+	for (uint32_t i = 0; i < len; i += 4) {
```
dragonmux, line 258: This loop counter should use `size_t` for its type to match `len` so we don't accidentally generate integer promotions that would turn `i` into a signed value

dragonmux (changes requested):
Doing a review, we've only spotted two small mistakes that are easily fixed.

This looks great, thank you for the contribution and your hard work implementing this! (We've been talking on Discord chatting back and forth on approaches and how best to support this part.)

We know that this is overflowing Flash right now which is why the firmware builds are failing, and that's well noted as a problem to solve for v2.0 via the new build system.

dragonmux (dismissed):

src/target/s32k3xx.c
```diff
@@ -0,0 +1,271 @@
+/*
+ * This file is part of the Black Magic Debug project.
+ *
+ * Copyright (C) 2015  Black Sphere Technologies Ltd.
+ * Written by Gareth McMullin <gareth@blacksphere.co.nz>
```
dragonmux, line 5: These should have their attribution updated to have the right copyright year, you as the author, and 1b² as the copyright holder.

dragonmux (dismissed):

dragonmux:
We know we approved this PR but now the Meson build system has landed, this needs some updating.

If you're able to do the necessary changes to add an enable macro to the Makefile and Meson build systems and gate this properly then please do so, if not we'll be happy to do the necessary changes ourself with your permission. Just let us know what path you'd like to pursue here.

dragonmux (approved):
All builds are passing, and the Meson build system change looks good 👍🏼 LGTM, merging after a little light testing as you've said this works on BMDA to us.

PR #1654, opened 2023-10-22
platforms/stlink: add support for forcing ST-Link v2 clone detection

silbe:
One of the ST-Link v2 clones I have (labelled "MINI ST-Link v2", "A" on one side of the PCB and "QYF-0685" on the other) gets detected as hardware version 1 rather than as a v2 clone (hardware version 257). It has PC13 pulled low via a 4.7kΩ resistor and PB11 connected to PB10 rather than pulled up. PA15 seems to be unconnected.

It's hard to tell how other clones are wired and if PB10 is connected to something else on them. The safest choice is to provide a compile-time option to force "detection" of a clone (so we use PB6 for nRST).

  * nop, neither does `main` (flash size overflow, known issue)

### DIFF
--- a/src/platforms/stlink/Makefile.inc
+++ b/src/platforms/stlink/Makefile.inc
@@ -35,6 +35,10 @@ ifeq ($(BLUEPILL), 1)
 CFLAGS += -DBLUEPILL=1
 endif
 
+ifeq ($(STLINK_FORCE_CLONE), 1)
+CFLAGS += -DSTLINK_FORCE_CLONE=1
+endif
+
 VPATH += platforms/common/stm32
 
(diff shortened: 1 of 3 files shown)


dragonmux (approved):
LGTM, merging as this is gated by a macro and fixes a bug.

In regards to the firmware not fitting on native - please update the compiler you're using. We don't really support using ARM's GCC 10 any more due to it doing a worse job on code generation, ARM's GCC 12 (release 12.2.Rel1) does a much better job and produces a binary that does fit the Flash available.

silbe:
Thanks for review and merging! Good to know about GCC 12 producing better code.

PR #1656, opened 2023-10-28
Feature: remove deprecated tooling

dragonmux:
## Detailed description

This PR does some maintenance on the repository, removing old tooling and scripts that are unused or marked deprecated. This is to lighten the maintenance burden on the project and because many of the tools here have since been replaced with better alternatives or made entirely obsolete.

If it turns out that we're removing something someone does actually use that has no good replacement, then it should probably get implemented into bmputil and formalised in that repository.


esden (approved):
Thank you. This is the step is as we discussed for 2.0. For most of the functionality we now use bmputil, where we do not we use other tools. Stubs are not something we used in a while. AFAIK

PR #1664, opened 2023-10-31
Feature: Cortex handling unification

dragonmux:
## Detailed description

This PR we unify the Cortex-A and Cortex-R support and perform some related cleanup to the Cortex-M support.

This results in a stable extensible basis for supporting Cortex-A/R parts using part-specific probe routines just like in the Cortex-M support, which solves several major headaches in supporting new parts in the Cortex-A side of things.

There will be two follow-up PRs that add proper support for the STM32MP15's Cortex-A cores, and the Zynq-7000 series. There are still some devtype issues in the ADIv5 component LUT but these should now be easily solved.

NB: This does not provide support for 64-bit Cortex devices, support for such devices needs some rework of the ADIv5 implementation and the implementation of 32- and 64-bit target address types w/ a union for holding an either address value. This is known about and in the pipeline but not handled here.


esden (approved):
LGTM

PR #1673, opened 2023-11-05
Fix: libusb overflow with JLink

ALTracer:
## Detailed description

* Trying to use BMDA with JLink adapters may raise a LIBUSB_ERROR_OVERFLOW under specific conditions.
* This pull request solves this problem by properly anticipating full-length responses in USB read calls.

RM08001 states that the IO TRANSACTION (JTAG3) command should return TDO (or SWDIO) data according to requested nBytes length, like JTAG2, and one more byte for OK status from adapter (or nonzero error). Requesting that in two libusb calls may not work on [my] setup and render BMDA inoperable with this adapter type.

Reading some libusb docs I figured that we should optimistically request a read for length of entire possible packet, then check whether we got handed that 1 byte less (apparently some firmwares give it in a separate bulk packet), and only then do a second read for it. The 1028-byte stack buffer of hosted (4+2*512 for header, tms & tdi) allows for receiving (512+1), but a dedicated smaller buffer could also suffice.

~~I consider this an intermittent bug because I cannot reproduce it after rebooting the machine. Detected against a STLinkReflashed adapter after a week of uptime on an Intel Pentium N6000 host, that is with Jasper Lake PCH USB3 HCI.~~
Once you let **J-Link Commander software** interact with their adapter firmware, it changes the responses' framing. I suppose anything else using `libjlinkarm.so`  also will, like RTTViewer and GDBServer. I reproduced the bug on two different machines running Linux, and on three different adapters (V8 from 2015 and Reflash from 2017). V5 from 2008 are not affected and always split the status byte. I could NOT reproduce the bug on Windows 10, and it required swapping drivers between proprietary and WinUSB a lot. Still, 
I believe fixing BMDA for developers toggling between JLink gdbserver and BMDA on Linux machines (this is possible in Eclipse CDT and derived IDEs) is useful, and might even reduce the number of libusb calls. I tried V7.88k, V6.98, V6.34, V5.12h.

Below that change I also tacked two other commits. One avoids discarding the copyright string returned by adapter in firmware version behind a nul byte (which is treated like a terminator by printf), another queries any extended capabilities of adapters, when supported. These are not strictly required to fix the problem I faced.

* [x] It builds for hardware native (`make PROBE_HOST=native`) -- not applicable

## Closing issues
None reported yet.

### DIFF
--- a/src/platforms/hosted/jlink.c
+++ b/src/platforms/hosted/jlink.c
@@ -46,7 +46,7 @@
 typedef struct jlink {
 	char fw_version[256U];         /* Firmware version string */
 	uint32_t hw_version;           /* Hardware version */
-	uint32_t capabilities;         /* Bitfield of supported capabilities */
+	uint32_t capabilities[4];      /* Bitfield of supported capabilities */
 	uint32_t available_interfaces; /* Bitfield of available interfaces */
 
 	struct jlink_interface_frequency {
@@ -302,7 +302,7 @@ static bool jlink_get_version(void)
 	DEBUG_INFO("Firmware version: %s\n", jlink.fw_version);
 
 	/* Read the hardware version if supported */
-	if (jlink.capabilities & JLINK_CAPABILITY_HARDWARE_VERSION) {
+	if (jlink.capabilities[0] & JLINK_CAPABILITY_HARDWARE_VERSION) {
 		if (!jlink_simple_query(JLINK_CMD_INFO_GET_HARDWARE_VERSION, buffer, 4U))
 			return false;
 
@@ -323,12 +323,12 @@ static bool jlink_get_extended_capabilities(void)
 	if (!jlink_simple_query(JLINK_CMD_INFO_GET_PROBE_EXTENDED_CAPABILITIES, buffer, sizeof(buffer)))
 		return false;
 
-	uint32_t ext_caps[4];
-	for (size_t i = 0; i < 4; i++)
-		ext_caps[i] = read_le4(buffer, i * 4U);
-
-	DEBUG_INFO("Extended capabilities: 0x%08" PRIx32, ext_caps[0]);
-	DEBUG_INFO(" 0x%08" PRIx32 " 0x%08" PRIx32 " 0x%08" PRIx32 "\n", ext_caps[1], ext_caps[2], ext_caps[3]);
+	DEBUG_INFO("Extended capabilities:");
+	for (size_t i = 0; i < 4; i++) {
+		jlink.capabilities[i] = read_le4(buffer, i * 4U);
+		DEBUG_INFO(" 0x%08" PRIx32, jlink.capabilities[i]);
+	}
+	DEBUG_INFO("\n");
 
 	return true;
 }
@@ -339,11 +339,11 @@ static bool jlink_get_capabilities(void)
 	if (!jlink_simple_query(JLINK_CMD_INFO_GET_PROBE_CAPABILITIES, buffer, sizeof(buffer)))
 		return false;
 
-	jlink.capabilities = read_le4(buffer, 0);
-	if (jlink.capabilities & JLINK_CAPABILITY_EXTENDED_CAPABILITIES)
+	jlink.capabilities[0] = read_le4(buffer, 0);
+	if (jlink.capabilities[0] & JLINK_CAPABILITY_EXTENDED_CAPABILITIES)
 		return jlink_get_extended_capabilities();
 
-	DEBUG_INFO("Capabilities: 0x%08" PRIx32 "\n", jlink.capabilities);
+	DEBUG_INFO("Capabilities: 0x%08" PRIx32 "\n", jlink.capabilities[0]);
 	return true;
 }
 
@@ -355,7 +355,7 @@ static inline bool jlink_interface_available(const uint8_t interface)
 static uint8_t jlink_selected_interface(void)
 {
 	/* V5.4 does only JTAG and hangs on 0xc7 commands */
-	if (!(jlink.capabilities & JLINK_CAPABILITY_INTERFACES))
+	if (!(jlink.capabilities[0] & JLINK_CAPABILITY_INTERFACES))
 		return JLINK_INTERFACE_JTAG;
 
 	uint8_t buffer[4U];
@@ -395,7 +395,7 @@ static bool jlink_get_interfaces(void)
 	 * and is known to not implement SWD or anything else besides JTAG.
 	 * Check the dedicated capability bit
 	 */
-	if (!(jlink.capabilities & JLINK_CAPABILITY_INTERFACES))
+	if (!(jlink.capabilities[0] & JLINK_CAPABILITY_INTERFACES))
 		jlink.available_interfaces = JLINK_INTERFACE_AVAILABLE(JLINK_INTERFACE_JTAG);
 	else {
 		if (!jlink_simple_request_8(JLINK_CMD_INTERFACE_GET, JLINK_INTERFACE_GET_AVAILABLE, buffer, sizeof(buffer)))
@@ -423,7 +423,7 @@ static bool jlink_get_interfaces(void)
 
 static bool jlink_get_interface_frequency(const uint8_t interface)
 {
-	if (!(jlink.capabilities & JLINK_CAPABILITY_INTERFACE_FREQUENCY)) {
+	if (!(jlink.capabilities[0] & JLINK_CAPABILITY_INTERFACE_FREQUENCY)) {
 		DEBUG_WARN("J-Link interface frequency commands are not available\n");
 		return false;
 	}
@@ -494,7 +494,7 @@ static bool jlink_get_interface_frequency(const uint8_t interface)
 
 static bool jlink_set_interface_frequency(const uint8_t interface, const uint32_t frequency)
 {
-	if (!(jlink.capabilities & JLINK_CAPABILITY_INTERFACE_FREQUENCY)) {
+	if (!(jlink.capabilities[0] & JLINK_CAPABILITY_INTERFACE_FREQUENCY)) {
 		DEBUG_WARN("J-Link interface frequency commands are not available\n");
 		return false;
 	}
@@ -566,8 +566,8 @@ static uint16_t jlink_target_voltage(void)
 
 static bool jlink_kickstart_power(void)
 {
-	if (!(jlink.capabilities & JLINK_CAPABILITY_POWER_STATE)) {
-		if (jlink.capabilities & JLINK_CAPABILITY_KICKSTART_POWER)
+	if (!(jlink.capabilities[0] & JLINK_CAPABILITY_POWER_STATE)) {
+		if (jlink.capabilities[0] & JLINK_CAPABILITY_KICKSTART_POWER)
 			DEBUG_ERROR("J-Link power state command is not available, but kickstart power is, this is unexpected\n");
 		return false;
 	}
@@ -588,7 +588,7 @@ static bool jlink_set_kickstart_power(const bool enable)
 	 * Exposed on pin 19 of the J-Link 20 pin connector
 	 */
 
-	if (!(jlink.capabilities & JLINK_CAPABILITY_KICKSTART_POWER))
+	if (!(jlink.capabilities[0] & JLINK_CAPABILITY_KICKSTART_POWER))
 		return false;
 
 	return jlink_simple_request_8(JLINK_CMD_POWER_SET_KICKSTART, enable ? JLINK_POWER_KICKSTART_ENABLE : 0, NULL, 0);
--- a/src/platforms/hosted/jlink.c
+++ b/src/platforms/hosted/jlink.c
@@ -137,10 +137,17 @@ bool jlink_transfer(const uint16_t clock_cycles, const uint8_t *const tms, const
 	/* Copy in the TDI values to transmit (if present) */
 	if (tdi)
 		memcpy(buffer + 4U + byte_count, tdi, byte_count);
-	/* Send the resulting transaction and try to read back the response data */
-	if (bmda_usb_transfer(bmda_probe_info.usb_link, buffer, sizeof(*header) + (byte_count * 2U), buffer, byte_count,
-			JLINK_USB_TIMEOUT) < 0 ||
-		/* Try to read back the transaction return code */
+	/*
+	 * Send the resulting transaction and try to read back the response data (including the response code first try),
+	 * because V8 and newer adapters, once touched by libjlinkarm.so (via Commander or else) on Linux hosts,
+	 * start returning the transaction status code *in the same packet*.
+	 */
+	const ssize_t bytes_received = bmda_usb_transfer(bmda_probe_info.usb_link, buffer,
+		sizeof(*header) + (byte_count * 2U), buffer, byte_count + 1U, JLINK_USB_TIMEOUT);
+	if (bytes_received < 0)
+		return false;
+	/* If the first read didn't return the transaction return code, try to read it back separately */
+	if (bytes_received < (ssize_t)byte_count + 1 &&
 		bmda_usb_transfer(bmda_probe_info.usb_link, NULL, 0, buffer + byte_count, 1U, JLINK_USB_TIMEOUT) < 0)
 		return false;
 	/* Copy out the response into the TDO buffer (if present) */
--- a/src/platforms/hosted/jlink.c
+++ b/src/platforms/hosted/jlink.c
@@ -295,9 +295,16 @@ static bool jlink_get_version(void)
 	if (version_length > sizeof(jlink.fw_version))
 		return false;
 
-	/* Read vesion string directly into jlink.version */
+	/* Read version string directly into jlink.version */
 	bmda_usb_transfer(bmda_probe_info.usb_link, NULL, 0, jlink.fw_version, version_length, JLINK_USB_TIMEOUT);
-	jlink.fw_version[version_length - 1U] = '\0'; /* Ensure null termination */
+	/* Ensure null termination */
+	char *const null_termination = jlink.fw_version + version_length - 1U;
+	*null_termination = '\0';
+
+	/* Replace NULL separating version and copyright string, if it exists */
+	char *const null_separator = strchr(jlink.fw_version, '\0');
+	if (null_separator != NULL && null_separator != null_termination)
+		*null_separator = '\n';
 
 	DEBUG_INFO("Firmware version: %s\n", jlink.fw_version);
 
--- a/src/platforms/hosted/jlink.c
+++ b/src/platforms/hosted/jlink.c
@@ -317,6 +317,22 @@ static bool jlink_get_version(void)
 	return true;
 }
 
+static bool jlink_get_extended_capabilities(void)
+{
+	uint8_t buffer[32U];
+	if (!jlink_simple_query(JLINK_CMD_INFO_GET_PROBE_EXTENDED_CAPABILITIES, buffer, sizeof(buffer)))
+		return false;
+
+	uint32_t ext_caps[4];
+	for (size_t i = 0; i < 4; i++)
+		ext_caps[i] = read_le4(buffer, i * 4U);
+
+	DEBUG_INFO("Extended capabilities: 0x%08" PRIx32, ext_caps[0]);
+	DEBUG_INFO(" 0x%08" PRIx32 " 0x%08" PRIx32 " 0x%08" PRIx32 "\n", ext_caps[1], ext_caps[2], ext_caps[3]);
+
+	return true;
+}
+
 static bool jlink_get_capabilities(void)
 {
 	uint8_t buffer[4U];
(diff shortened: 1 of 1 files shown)


ALTracer:
Log from latest main:
```
$ ./src/blackmagic -tv 1 -f 2000k
Black Magic Debug App v1.10.0-181-ga9697dd9
Using 1366:0105 000772256119 SEGGER
 J-Link ---
Capabilities: 0x98ea5a33
Firmware version: J-Link STLink V21 compiled Aug 12 2019 10:29:20
Hardware version: J-Link v1.0.0
Available interfaces: 
        1: SWD* 
Running in Test Mode
Target voltage:  3.300V Volt
SWD interface frequency:
        Base frequency: 72000000Hz
        Minimum divisor: 18
Speed set to 2.000MHz for SWD
bmda_usb_transfer: Receiving response from adaptor failed (-8): LIBUSB_ERROR_OVERFLOW
jlink_swd_seq_out failed
bmda_usb_transfer: Receiving response from adaptor failed (-8): LIBUSB_ERROR_OVERFLOW
jlink_swd_seq_out failed
```
Log with this patchset:
```
./src/blackmagic -tv 1 -f 2000k
Black Magic Debug App v1.10.0-184-gee8cb420
Using 1366:0105 000772256119 SEGGER
 J-Link ---
Capabilities: 0x98ea5a33
Extended capabilities: 0x98ea5a33 0x52b17c0c 0x00000021 0x00000000
Firmware version: J-Link STLink V21 compiled Aug 12 2019 10:29:20
Copyright 2016-2019 SEGGER: www.segger.com
Hardware version: J-Link v1.0.0
Available interfaces: 
        1: SWD* 
Running in Test Mode
Target voltage:  3.300V Volt
SWD interface frequency:
        Base frequency: 72000000Hz
        Minimum divisor: 18
Speed set to 2.000MHz for SWD
DP DPIDR 0x0bc11477 (v1 MINDP rev0) designer 0x43b partno 0xbc
AP   0: IDR=04770031 CFG=00000000 BASE=f0000003 CSW=83000040 (AHB3-AP var3 rev0)
Halt via DHCSR(01030003): success after 4ms
ROM: Table BASE=0xf0000000 SYSMEM=0x00000001, Manufacturer 020 Partno 460
ROM: Table BASE=0xe00ff000 SYSMEM=0x00000001, Manufacturer 43b Partno 4c0
 0 0xe000e000: Generic IP component - Cortex-M0 SCS (System Control Space) (PIDR = 0x00000004000bb008 DEVTYPE = 0x00 ARCHID = 0x0000)
 -> cortexm_probe
CPUID 0x410cc601 (M0+ var 0 rev 1)
ID Code: 00000000
 1 0xe0001000: Generic IP component - Cortex-M0 DWT (Data Watchpoint and Trace) (PIDR = 0x00000004000bb00a DEVTYPE = 0x00 ARCHID = 0x0000)
 2 0xe0002000: Generic IP component - Cortex-M0 BPU (Breakpoint Unit) (PIDR = 0x00000004000bb00b DEVTYPE = 0x00 ARCHID = 0x0000)
 ROM: Table END
1 Entry 0x200002 -> Not present
2 Entry 0x10000002 -> Not present
3 Entry 0x10000002 -> Not present
ROM: Table END
***  1   STM32G07/8 M0+
RAM   Start: 0x20000000 length = 0x9000
Flash Start: 0x08000000 length = 0x20000 blocksize 0x800
Flash Start: 0x1fff7000 length = 0x400 blocksize 0x8
```
This is Nucleo-G071RB. Hosted RTT performs at only 6700 chars/s.

ALTracer:
Verbose log from latest main. Each timeout error takes 5 seconds. Could also crank LIBUSB_DEBUG to 4.
```
$ ./src/blackmagic -tv 61
Black Magic Debug App v1.10.0-181-ga9697dd9
Using 1366:0105 000772256119 SEGGER
 J-Link ---
 request: e8
response: 33 5a ea 98
Capabilities: 0x98ea5a33
 request: 01
response: 70 00
response: 4a 2d 4c 69 6e 6b 20 53 54 4c 69 6e 6b 20 56 32 31 20 63 6f 6d 70 69 6c 65 64 20 41 75 67 20 31 ...
Firmware version: J-Link STLink V21 compiled Aug 12 2019 10:29:20
 request: f0
response: 10 27 00 00
Hardware version: J-Link v1.0.0
 request: c7 ff
response: 02 00 00 00
 request: c7 fe
response: 01 00 00 00
Available interfaces:
        1: SWD*
 request: dd
Running in Test Mode
 request: 07
response: e4 0c 00 80 40 00 01 ff
Target voltage:  3.300V Volt
 request: c7 fe
response: 01 00 00 00
 request: c7 fe
response: 01 00 00 00
 request: c0
response: 00 a2 4a 04 12 00
SWD interface frequency:
        Base frequency: 72000000Hz
        Minimum divisor: 18
 request: 05 a0 0f
Speed set to 4.000MHz for SWD
-> jlink_swd_init(0)
 request: c7 fe
response: 01 00 00 00
jlink_swd_seq_out 32 clock_cycles: ffffffff
 request: cf 00 20 00 ff ff ff ff ff ff ff ff
bmda_usb_transfer: Receiving response from adaptor failed (-8): LIBUSB_ERROR_OVERFLOW
jlink_swd_seq_out failed
jlink_swd_seq_out 28 clock_cycles: 0fffffff
 request: cf 00 1c 00 ff ff ff ff ff ff ff 0f
bmda_usb_transfer: Receiving response from adaptor failed (-8): LIBUSB_ERROR_OVERFLOW
jlink_swd_seq_out failed
jlink_swd_seq_out 32 clock_cycles: 6209f392
 request: cf 00 20 00 ff ff ff ff 92 f3 09 62
bmda_usb_transfer: Receiving response from adaptor failed (-7): LIBUSB_ERROR_TIMEOUT
jlink_swd_seq_out failed
jlink_swd_seq_out 32 clock_cycles: 86852d95
 request: cf 00 20 00 ff ff ff ff 95 2d 85 86
bmda_usb_transfer: Receiving response from adaptor failed (-8): LIBUSB_ERROR_OVERFLOW
jlink_swd_seq_out failed
jlink_swd_seq_out 32 clock_cycles: e3ddafe9
 request: cf 00 20 00 ff ff ff ff e9 af dd e3
bmda_usb_transfer: Receiving response from adaptor failed (-7): LIBUSB_ERROR_TIMEOUT
jlink_swd_seq_out failed
jlink_swd_seq_out 32 clock_cycles: 19bc0ea2
 request: cf 00 20 00 ff ff ff ff a2 0e bc 19
bmda_usb_transfer: Receiving response from adaptor failed (-8): LIBUSB_ERROR_OVERFLOW
jlink_swd_seq_out failed
jlink_swd_seq_out 12 clock_cycles: 000001a0
 request: cf 00 0c 00 ff ff a0 01
bmda_usb_transfer: Receiving response from adaptor failed (-7): LIBUSB_ERROR_TIMEOUT
jlink_swd_seq_out failed
jlink_swd_seq_out 32 clock_cycles: ffffffff
 request: cf 00 20 00 ff ff ff ff ff ff ff ff
bmda_usb_transfer: Receiving response from adaptor failed (-8): LIBUSB_ERROR_OVERFLOW
jlink_swd_seq_out failed
jlink_swd_seq_out 32 clock_cycles: 0fffffff
 request: cf 00 20 00 ff ff ff ff ff ff ff 0f
bmda_usb_transfer: Receiving response from adaptor failed (-7): LIBUSB_ERROR_TIMEOUT
jlink_swd_seq_out failed
jlink_adiv5_raw_access: Attempting access to addr 0000
 request: cf 00 0b 00 ff f0 a5 00
bmda_usb_transfer: Receiving response from adaptor failed (-8): LIBUSB_ERROR_OVERFLOW
jlink_swd_seq_out 32 clock_cycles: ffffffff
 request: cf 00 20 00 ff ff ff ff ff ff ff ff
bmda_usb_transfer: Receiving response from adaptor failed (-7): LIBUSB_ERROR_TIMEOUT
jlink_swd_seq_out failed
jlink_swd_seq_out 28 clock_cycles: 0fffffff
 request: cf 00 1c 00 ff ff ff ff ff ff ff 0f
bmda_usb_transfer: Receiving response from adaptor failed (-8): LIBUSB_ERROR_OVERFLOW
jlink_swd_seq_out failed
jlink_swd_seq_out 16 clock_cycles: 0000e79e
 request: cf 00 10 00 ff ff 9e e7
bmda_usb_transfer: Receiving response from adaptor failed (-7): LIBUSB_ERROR_TIMEOUT
jlink_swd_seq_out failed
jlink_swd_seq_out 32 clock_cycles: ffffffff
 request: cf 00 20 00 ff ff ff ff ff ff ff ff
bmda_usb_transfer: Receiving response from adaptor failed (-8): LIBUSB_ERROR_OVERFLOW
jlink_swd_seq_out failed
jlink_swd_seq_out 32 clock_cycles: 0fffffff
 request: cf 00 20 00 ff ff ff ff ff ff ff 0f
bmda_usb_transfer: Receiving response from adaptor failed (-7): LIBUSB_ERROR_TIMEOUT
jlink_swd_seq_out failed
jlink_adiv5_raw_access: Attempting access to addr 0000
 request: cf 00 0b 00 ff f0 a5 00
bmda_usb_transfer: Receiving response from adaptor failed (-8): LIBUSB_ERROR_OVERFLOW
No usable DP found
No target found
[255] $
```

src/platforms/hosted/jlink.c
```diff
@@ -295,10 +298,15 @@ static bool jlink_get_version(void)
...
 
-	/* Read vesion string directly into jlink.version */
+	/* Read version string directly into jlink.version */
 	bmda_usb_transfer(bmda_probe_info.usb_link, NULL, 0, jlink.fw_version, version_length, JLINK_USB_TIMEOUT);
 	jlink.fw_version[version_length - 1U] = '\0'; /* Ensure null termination */
 
+	/* Replace first NUL separating platform and copyright string */
+	char *firstnul = strchr(jlink.fw_version, 0);
+	if (firstnul != NULL)
+		*firstnul = '\n';
```
perigoso, line 308: I missed the fact the version string contained the copyright a swell the first time, how about that, took me a second to understand what this was doing :sweat_smile:...

There's a subtle edge case here, that I think we should guard for. If a J-Link probe, be it via a firmware bug, or just inconsistent implementation, reports a firmware version that does not have a null as a separator, or termination, this code will replace the null termination we just ensured before it, which could lead to a bad time (from the `strchr` reference: _"The terminating null character is considered to be a part of the string and can be found when searching for '\0'."_ )
We can't just move this code to before the "Ensure null termination" because on this particular edge case, the lack of null termination would be bad news for `strchr` too, so I think our best bet would be to add a check to make sure we don't replace it after

```suggestion
	/* Ensure null termination */
	char *const null_termination = jlink.fw_version + version_length - 1U;
	*null_termination = '\0';

	/* Replace NULL separating version and copyright string, if it exists */
	char *const null_separator = strchr(jlink.fw_version, '\0');
	if (null_separator != NULL && null_separator < null_termination)
		*null_separator = '\n';
```

src/platforms/hosted/jlink.c
```diff
@@ -317,6 +325,24 @@ static bool jlink_get_version(void)
...
 }
 
+static bool jlink_get_extended_capabilities(void)
+{
+	uint8_t buffer[32U];
+
+	if (!jlink_simple_query(JLINK_CMD_INFO_GET_PROBE_EXTENDED_CAPABILITIES, buffer, sizeof(buffer)))
+		return false;
+
+	uint32_t ext_caps[4];
```
perigoso, line 335: Let's store this in the `jlink_s` struct already, even if we don't need it right now. Make the `jlink_s->capabilities` bitfield an array of 4 `uint32_t`, the first 32 bits should be shared between the regular capabilities and extended capabilities responses

src/platforms/hosted/jlink.c
```diff
@@ -317,6 +325,24 @@ static bool jlink_get_version(void)
...
+	ext_caps[0] = read_le4(buffer, 0);
+	ext_caps[1] = read_le4(buffer, 4);
+	ext_caps[2] = read_le4(buffer, 8);
+	ext_caps[3] = read_le4(buffer, 12);
```
perigoso, line 339:
```suggestion
    for (size_t i = 0; i < 4; i++)
		ext_caps[i] = read_le4(buffer, i * 4U);
```

src/platforms/hosted/jlink.c
```diff
@@ -326,6 +352,9 @@ static bool jlink_get_capabilities(void)
 	jlink.capabilities = read_le4(buffer, 0);
 	DEBUG_INFO("Capabilities: 0x%08" PRIx32 "\n", jlink.capabilities);
 
+	if (jlink.capabilities & JLINK_CAPABILITY_EXTENDED_CAPABILITIES)
+		jlink_get_extended_capabilities();
```
perigoso, line 356: Instead of printing the capabilities here, and then duplicating the information printing the extended capabilities, move this print to after the extended capabilities check, and do an early return there

Also, don't forget to return the result, since we are delegating to `jlink_get_extended_capabilities()` we should propagate the result

```suggestion
	if (jlink.capabilities & JLINK_CAPABILITY_EXTENDED_CAPABILITIES)
		return jlink_get_extended_capabilities();
	
	DEBUG_INFO("Capabilities: 0x%08" PRIx32 "\n", jlink.capabilities);
```

perigoso (changes requested):
Looks good, good catch :smiley:! I have just a couple notes

src/platforms/hosted/jlink.c
ALTracer: Yes, string content uncertainty also occured to me. If this were C++, I could just ask to print a std::string of the size reported by JLink firmware, but alas. In BMDA logging is hidden behind macros, so it's not like I can call `write()` for fixed bytes and then it dumps everything with literal NULs to stdout.
Your logic looks clean enough, will apply. Maybe I forgot that I wanted to check where that nul byte found by strchr is located, and didn't want to perform pointer comparison, as it may be unspecified.

src/platforms/hosted/jlink.c
ALTracer: Okay, I'll put an array there. Reading RM08001 seems to suggest that `extended_capabilities[0]` is an exact copy of `capabilities` value save for most significant bit (31), which indicates there ARE extended capabilities.
Am I to correct the 9 references to that single uint32_t to refer to `capabilities[0]` now? It would be easier (on the existing code) to store the remainder of ext_caps in a separate uint32_t[3] array.

src/platforms/hosted/jlink.c
ALTracer: Agreed. I wrote it that way to see both on BMDA verbose stdout. Since extended capabilities is strictly a superset of capabilities, then it being printed should suppress printing ordinary ones. I understand that this code needs to handle both firmware types.

src/platforms/hosted/jlink.c
ALTracer: Applied and changed the debug printout to use this loop as well.

ALTracer:
Thank you very much @perigoso for coming up and reviewing this. Applied the review items and rebased on main.
By the way, can you test any of this for regressions on your adapters? I wouldn't mean to break any existing setups, rather the opposite.

src/platforms/hosted/jlink.c
perigoso: Yes, currently there is no check for any extended capability.
I'm not against keeping the extended capabilities sepperate

dragonmux: Given the docs say they should be identical save for that MSb, we'd prefer them to use the same storage, and extend the storage, rather than using new storage. It'll keep the J-Link structure simpler.

src/platforms/hosted/jlink.c
```diff
@@ -295,14 +298,21 @@ static bool jlink_get_version(void)
...
+	/* Read version string directly into jlink.version */
 	bmda_usb_transfer(bmda_probe_info.usb_link, NULL, 0, jlink.fw_version, version_length, JLINK_USB_TIMEOUT);
-	jlink.fw_version[version_length - 1U] = '\0'; /* Ensure null termination */
+	/* Ensure null termination */
+	char *const null_termination = jlink.fw_version + version_length - 1U;
+	*null_termination = '\0';
+
+	/* Replace NULL separating version and copyright string, if it exists */
+	char *const null_separator = strchr(jlink.fw_version, '\0');
+	if (null_separator != NULL && null_separator < null_termination)
```
dragonmux, line 309: /technically/ not well defined to use `<` on pointers, please prefer `!=` as in:
```suggestion
	if (null_separator != NULL && null_separator != null_termination)
```

src/platforms/hosted/jlink.c
```diff
@@ -317,15 +327,34 @@ static bool jlink_get_version(void)
 	return true;
 }
 
+static bool jlink_get_extended_capabilities(void)
+{
+	uint8_t buffer[32U];
+
```
dragonmux, line 333: Please remove this excess new line to better match the style of similar functions in the file

dragonmux (commented):
There are two small items from us, see the review below, and then we're happy to merge this.

Tested against our LPC Link-2 running the J-Link "OB" firmware, and it works great.

```
❯ src/blackmagic -s 000611000000 -tjv 5
Black Magic Debug App v1.10.0-186-gfe848f23
Using 1366:0101 000611000000 SEGGER
 J-Link ---
Extended capabilities: 0xb9ff7bbf 0x00003c0d 0x00000000 0x00000000
Firmware version: J-Link LPC-Link 2 compiled May 31 2013 17:31:46
Copyright 2003-2013 SEGGER: www.segger.com
Hardware version: LPC-Link2 v1.0.0
Available interfaces:
        0: JTAG*
        1: SWD
Running in Test Mode
Target voltage:  3.300V Volt
JTAG interface frequency:
        Base frequency: 48000000Hz
        Minimum divisor: 8
Speed set to 4.000MHz for JTAG
```

src/platforms/hosted/jlink.c
ALTracer: These are two pointers into the same `jlink.fw_version[]` which is a contiguous char array, and comparing them like this worked (on glibc anyway). We are not asking to compare pointers to different objects. But *technically* replacing this with a not-equal check yields the same behaviour for this exact firmware type and fw_version binary content.

dragonmux: It's not about what appears to work, it's about what is actually defined by the C standard and what isn't. Pointer comparison is only completely safe when done as `==` and `!=`. It is *implementation defined* what happens otherwise for `<`, and that means compiler-dependant. If we were comparing `>` then it's covered by the defined semantics, although best practice says to avoid doing this too as it's bugprone.

See the [cppreference page on C comparison operators for more](https://en.cppreference.com/w/c/language/operator_comparison)

ALTracer: Sorry, I don't follow. From the same cppreference page:
> if two pointers point to different elements of the same array, the one pointing at the element with the larger index compares greater.

I do NOT read it as "comparing for less-than is UB, always use greater-than and bend over backwards by swapping arguments". Also see C11 6.5.8 Relational operators from $60 pdf or http://port70.net/%7Ensz/c/c11/n1570.html#6.5.8
>  If the objects pointed to are members of the same aggregate object, pointers to structure members declared later compare greater than pointers to members declared earlier in the structure, and pointers to array elements with larger subscript values compare greater than pointers to elements of the same array with lower subscript values.

And especially

> Each of the operators < (less than), > (greater than), <= (less than or equal to), and >= (greater than or equal to) shall yield 1 if the specified relation is true and 0 if it is false. The result has type int.

I spent an hour writing a test snippet so people can check more easily. Generated assembly is not identical, but runtime behaviour is. https://godbolt.org/z/n5WEon9sM
BMDA is supported AFAIK on MSYS2, Linux, Darwin, but isn't on 16-bit segmented memory arches.

Looking at URB responses in Wireshark I see a 112-byte response with ~16 NULL bytes at the tail of version string, probably because their firmware simply returns then entire chunk from .rodata (or configuration space) of fixed length. I don't want to chase these NULL bytes, just the one separating version from copyright (if it always exists).

dragonmux: What you're observing is what we mentioned about "implementation defined what happens". Note that the text on pointers never talks about `<`, only `>` - you are inferring that defining greater than defines less than, but this is not in the plain text of the standard and while that is how many implementations do define it, they don't have to.

Please just use `!=` - it is fully defined for pointers for both C and C++ and will be safer into the future of the code base. This is what the recommended best practice is, this is the relation that is defined on std::span<> and C++ iterator comparisons too. `<` is not as is `>` because they are unable to be fully well defined as a relation on pointers. There is good reason for this. We are trying in this case to avoid someone in the future editing this code and invalidating the precondition that the pointers have be in the same buffer so we can avoid all the pitfalls of the inequality comparison operators.

`strchr()` won't go past the first `\0` so that's not a problem - for all J-Link firmware that's working the way you've observed (that 112-byte response), `null_separator != null_termination` will always be true - but more importantly, will also always be well defined.

dragonmux (approved):
It all LGTM now. We'll get this merged - thanks for the contribution!

src/platforms/hosted/jlink.c
ALTracer: Thank you for the detailed explanation. Will prefer pointer equality checks where possible over less-than/greater-than. I wonder if compilers have a diagnostic for that.

PR #1676, opened 2023-11-06
Fix: -Wformat in gdbserver (gdb_main, command, gdb_hostio)

### LINKED ISSUE #1675: Wformat (part 2) in gdbserver core and target drivers
As you all know, C compiler diagnostics are optional because C.
There are three more functions in BMD core which accept format-strings and print messages via GDB remote protocol. Namely, `gdb_putpacket_f()`, `gdb_outf()`
https://github.com/blackmagic-debug/blackmagic/blob/d0c8bb7c84a96bd743d0b104610896c53c7fd893/src/include/gdb_packet.h#L42-L48
and `tc_printf()` -- this one is more special. 
https://github.com/blackmagic-debug/blackmagic/blob/d0c8bb7c84a96bd743d0b104610896c53c7fd893/src/target/target.c#L563
It resolves via a bunch of indirection (`target`->`target_controller_s`->`printf()` to `gdb_target_printf()`, which is a shim around `gdb_voutf()`. 
https://github.com/blackmagic-debug/blackmagic/blob/d0c8bb7c84a96bd743d0b104610896c53c7fd893/src/gdb_main.c#L109
But it's used in most of target drivers a whopping ... *checks* `grep -rn tc_printf\( src/ | wc -l` ... 217 times. `gdb_voutf()` itself is referenced only 4 times, and non-varargs `gdb_outf()` is referenced 54 times.

I discovered this while trying to add a formatting GDB async notification printer. Luckily, when I labeled it, GCC helpfully warned me that I want mismatching format-string specifiers, but only when I build for `native` and not for `hosted` -- this in particular stems from the fact that `target_addr_t` is `uint32_t` (hence either `long unsigned int` or `unsigned int`), but `0x%08X` is always `unsigned int`.

**I propose labeling these three functions with `__attribute__((format(printf, 1, 2)))` (or 2, 3 in case of tc_printf) so that new contributions get checked automatically by CI.**

Usage of debug logging infrastructure (upgraded in PR1452, fixed for Windows BMDA in PR1541) is already covered by this mechanism, and diagnostic-triggering code fixed in #1592. I'm glad that much is done and dealt with.

After applying this patch and trying both `native` and `hosted` builds I see a lot of warnings. Some of them can be mitigated by breaking format-strings into multiple pieces and swapping `%0x08X` for `%0x08" PRIx32 "`. I don't have the perl skills to mass-replace like that. Also I didn't post it as a PR because CI will fail too early (it doesn't `make -k`), and I would have to fix all the warnings myself on that branch. So I'm leaving it to more dedicated contributors.
* hosted: 6 groups of warnings (cmd_heapinfo, hostio_* hex, efm32_cmd_serial, exec_q_crc, stm32h7_crc, stm32lx_cmd_option)
* blackpill-f411ce: many more (+efm32_probe, efm32_cmd_efm_info, gdb_poll_target, cortexm_hostio_request, lpc546xx_cmd_read_partid, nrf51_cmd_read_*, sam3x cmd_gpnvm, samd/x5x _cmd_serial, stm32h7/h5/mp15 _uid, stm32f1/l4/lx _cmd_option, stm32lx_cmd_eeprom, stm32g0_display_registers, cmd_rtt)
* native: mostly same. 32 functions total.

ALTracer:
## Detailed description

* This is not a functional feature; rather, an attempt to reduce unexpected behaviour, especially in BMF.
* The problem is many printf-like calls in BMD core code are not checked by compiler for their correctness.
* This pull request solves this problem by marking the printf-like functions with format attributes. Where applicable, this PR also tries to fix appearing warnings.

I deal here only with `gdb_voutf()` and `gdb_putpacket_f()`, as their direct usage (with dubious specificators, specifically) is less numerous. Let someone else deal with all the `tc_printf()`.

I am build-testing the changes for `native` on Linux amd64, and for `swlink` which is armv7-m. Runtime testing pending -- there is no code coverage tooling, so I have to reason about which aspects need testing, like GDB remote monitor commands for rtt options and semihosting heapinfo.

## Closing issues

Partially fixes #1675. See that for a more detailed description.

### DIFF
--- a/src/gdb_hostio.c
+++ b/src/gdb_hostio.c
@@ -90,7 +90,7 @@ static int hostio_get_response(target_controller_s *const tc)
 /* Interface to host system calls */
 int hostio_open(target_controller_s *tc, target_addr_t path, size_t path_len, target_open_flags_e flags, mode_t mode)
 {
-	gdb_putpacket_f("Fopen,%08X/%X,%08X,%08X", path, path_len, flags, mode);
+	gdb_putpacket_f("Fopen,%08" PRIX32 "/%08" PRIX32 ",%08X,%08" PRIX32, path, (uint32_t)path_len, flags, mode);
 	return hostio_get_response(tc);
 }
 
@@ -102,49 +102,50 @@ int hostio_close(target_controller_s *tc, int fd)
 
 int hostio_read(target_controller_s *tc, int fd, target_addr_t buf, unsigned int count)
 {
-	gdb_putpacket_f("Fread,%08X,%08X,%08X", fd, buf, count);
+	gdb_putpacket_f("Fread,%08X,%08" PRIX32 ",%08X", fd, buf, count);
 	return hostio_get_response(tc);
 }
 
 int hostio_write(target_controller_s *tc, int fd, target_addr_t buf, unsigned int count)
 {
-	gdb_putpacket_f("Fwrite,%08X,%08X,%08X", fd, buf, count);
+	gdb_putpacket_f("Fwrite,%08X,%08" PRIX32 ",%08X", fd, buf, count);
 	return hostio_get_response(tc);
 }
 
 long hostio_lseek(target_controller_s *tc, int fd, long offset, target_seek_flag_e flag)
 {
-	gdb_putpacket_f("Flseek,%08X,%08X,%08X", fd, offset, flag);
+	gdb_putpacket_f("Flseek,%08X,%08lX,%08X", fd, offset, flag);
 	return hostio_get_response(tc);
 }
 
 int hostio_rename(target_controller_s *tc, target_addr_t oldpath, size_t old_len, target_addr_t newpath, size_t new_len)
 {
-	gdb_putpacket_f("Frename,%08X/%X,%08X/%X", oldpath, old_len, newpath, new_len);
+	gdb_putpacket_f("Frename,%08" PRIX32 "/%08" PRIX32 ",%08" PRIX32 "/%08" PRIX32, oldpath, (uint32_t)old_len, newpath,
+		(uint32_t)new_len);
 	return hostio_get_response(tc);
 }
 
 int hostio_unlink(target_controller_s *tc, target_addr_t path, size_t path_len)
 {
-	gdb_putpacket_f("Funlink,%08X/%X", path, path_len);
+	gdb_putpacket_f("Funlink,%08" PRIX32 "/%08" PRIX32, path, (uint32_t)path_len);
 	return hostio_get_response(tc);
 }
 
 int hostio_stat(target_controller_s *tc, target_addr_t path, size_t path_len, target_addr_t buf)
 {
-	gdb_putpacket_f("Fstat,%08X/%X,%08X", path, path_len, buf);
+	gdb_putpacket_f("Fstat,%08" PRIX32 "/%08" PRIX32 ",%08" PRIX32, path, (uint32_t)path_len, buf);
 	return hostio_get_response(tc);
 }
 
 int hostio_fstat(target_controller_s *tc, int fd, target_addr_t buf)
 {
-	gdb_putpacket_f("Ffstat,%X,%08X", fd, buf);
+	gdb_putpacket_f("Ffstat,%X,%08" PRIX32, fd, buf);
 	return hostio_get_response(tc);
 }
 
 int hostio_gettimeofday(target_controller_s *tc, target_addr_t tv, target_addr_t tz)
 {
-	gdb_putpacket_f("Fgettimeofday,%08X,%08X", tv, tz);
+	gdb_putpacket_f("Fgettimeofday,%08" PRIX32 ",%08" PRIX32, tv, tz);
 	return hostio_get_response(tc);
 }
 
@@ -156,6 +157,6 @@ int hostio_isatty(target_controller_s *tc, int fd)
 
 int hostio_system(target_controller_s *tc, target_addr_t cmd, size_t cmd_len)
 {
-	gdb_putpacket_f("Fsystem,%08X/%X", cmd, cmd_len);
+	gdb_putpacket_f("Fsystem,%08" PRIX32 "/%08" PRIX32, cmd, (uint32_t)cmd_len);
 	return hostio_get_response(tc);
 }
--- a/src/gdb_hostio.c
+++ b/src/gdb_hostio.c
@@ -28,20 +28,20 @@ int hostio_reply(target_controller_s *const tc, char *const pbuf, const int len)
 {
 	(void)len;
 
-	/* 
+	/*
 	 * File-I/O Remote Protocol Extension
 	 * See https://sourceware.org/gdb/onlinedocs/gdb/Protocol-Basics.html#Protocol-Basics
-	 * 
+	 *
 	 * This handles the F Reply Packet, sent by GDB after handling the File-I/O Request Packet.
-	 * 
+	 *
 	 * The F reply packet consists of the following:
-	 * 
+	 *
 	 * - retcode, the return code of the system call as hexadecimal value.
 	 * - errno, the errno set by the call, in protocol-specific representation.
 	 * 		Can be omitted if the call was successful.
 	 * - Ctrl-C flag, sent only if user requested a break.
 	 * 		In this case, errno must be sent as well, even if the call was successful.
-	 * 		The Ctrl-C flag itself consists of the character ‘C’: 
+	 * 		The Ctrl-C flag itself consists of the character ‘C’:
 	 */
 
 	const bool retcode_is_negative = pbuf[1U] == '-';
@@ -52,7 +52,7 @@ int hostio_reply(target_controller_s *const tc, char *const pbuf, const int len)
 	const int items = sscanf(pbuf + (retcode_is_negative ? 2U : 1U), "%x,%x,%c", &retcode, &errno_, &ctrl_c_flag);
 
 	if (items < 1) {
-		/* 
+		/*
 		 * Something went wrong with the sscanf or the packet format, avoid UB
 		 * FIXME: how do we properly handle this?
 		 */
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -496,7 +496,7 @@ static void exec_q_crc(const char *packet, const size_t length)
 		if (!generic_crc32(cur_target, &crc, addr, addr_length))
 			gdb_putpacketz("E03");
 		else
-			gdb_putpacket_f("C%lx", crc);
+			gdb_putpacket_f("C%" PRIx32, crc);
 	}
 }
 
@@ -846,7 +846,7 @@ void gdb_poll_target(void)
 		gdb_putpacket_f("T%02X", GDB_SIGINT);
 		break;
 	case TARGET_HALT_WATCHPOINT:
-		gdb_putpacket_f("T%02Xwatch:%08X;", GDB_SIGTRAP, watch);
+		gdb_putpacket_f("T%02Xwatch:%08" PRIX32 ";", GDB_SIGTRAP, watch);
 		break;
 	case TARGET_HALT_FAULT:
 		gdb_putpacket_f("T%02X", GDB_SIGSEGV);
--- a/src/include/gdb_packet.h
+++ b/src/include/gdb_packet.h
@@ -39,7 +39,7 @@ size_t gdb_getpacket(char *packet, size_t size);
 void gdb_putpacket(const char *packet, size_t size);
 void gdb_putpacket2(const char *packet1, size_t size1, const char *packet2, size_t size2);
 #define gdb_putpacketz(packet) gdb_putpacket((packet), strlen(packet))
-void gdb_putpacket_f(const char *packet, ...);
+void gdb_putpacket_f(const char *packet, ...) __attribute__((format(printf, 1, 2)));
 void gdb_put_notification(const char *packet, size_t size);
 #define gdb_put_notificationz(packet) gdb_put_notification((packet), strlen(packet))
 
--- a/src/include/gdb_packet.h
+++ b/src/include/gdb_packet.h
@@ -45,6 +45,6 @@ void gdb_put_notification(const char *packet, size_t size);
 
 void gdb_out(const char *buf);
 void gdb_voutf(const char *fmt, va_list);
-void gdb_outf(const char *fmt, ...);
+void gdb_outf(const char *fmt, ...) __attribute__((format(printf, 1, 2)));
 
 #endif /* INCLUDE_GDB_PACKET_H */


src/target/target.c
```diff
@@ -688,3 +689,4 @@ int tc_system(target_s *t, target_addr_t cmd, size_t cmdlen)
 	}
 	return t->tc->system(t->tc, cmd, cmdlen);
 }
+#endif // PC_HOSTED
```
dragonmux, line 692: While we probably should introduce a scheme like this onto the end of `#endif` statements, we don't currently use one, so please drop it, or at least use `/* */` comment style to match the header include guards

src/gdb_hostio.c
```diff
@@ -146,6 +148,7 @@ int hostio_isatty(target_controller_s *tc, int fd)
 
 int hostio_system(target_controller_s *tc, target_addr_t cmd, size_t cmd_len)
 {
-	gdb_putpacket_f("Fsystem,%08X/%X", cmd, cmd_len);
+	gdb_putpacket_f("Fsystem,%08" PRIx32 "/%zX", cmd, cmd_len);
 	return hostio_get_response(tc);
 }
+#endif // PC_HOSTED
```
dragonmux, line 154: Same comment about this as the other

src/gdb_main.c
```diff
@@ -128,12 +129,17 @@ target_controller_s gdb_controller = {
...
 	.isatty = hostio_isatty,
 	.system = hostio_system,
+#endif
 };
 
 /* execute gdb remote command stored in 'pbuf'. returns immediately, no busy waiting. */
 
 int gdb_main_loop(target_controller_s *tc, char *pbuf, size_t pbuf_size, size_t size, bool in_syscall)
 {
+#if PC_HOSTED == 1
```
dragonmux, line 139: Are these actually unused under BMDA? We're surprised it they are that it had been building without errors.

src/gdb_hostio.c
```diff
@@ -80,7 +82,7 @@ static int hostio_get_response(target_controller_s *const tc)
 /* Interface to host system calls */
 int hostio_open(target_controller_s *tc, target_addr_t path, size_t path_len, target_open_flags_e flags, mode_t mode)
 {
-	gdb_putpacket_f("Fopen,%08X/%X,%08X,%08X", path, path_len, flags, mode);
+	gdb_putpacket_f("Fopen,%08" PRIx32 "/%zX,%08X,%08" PRIx32, path, path_len, flags, mode);
```
dragonmux, line 85: If these %z's are in the firmware, this will break things. Needs to use `PRIX32` and a cast of the number to uint32_t - eg, `(uint32_t)path_len`

Note that PRIx32 makes the results lower-case hex, while PRIX32 upper-case. We should maintain the case the original formatting string aimed for.

src/command.c
```diff
@@ -679,8 +679,9 @@ static bool cmd_heapinfo(target_s *t, int argc, const char **argv)
 		target_addr_t heap_limit = strtoul(argv[2], NULL, 16);
 		target_addr_t stack_base = strtoul(argv[3], NULL, 16);
 		target_addr_t stack_limit = strtoul(argv[4], NULL, 16);
-		gdb_outf("heapinfo heap_base: %p heap_limit: %p stack_base: %p stack_limit: %p\n", heap_base, heap_limit,
-			stack_base, stack_limit);
+		gdb_outf("heapinfo heap_base: 0x%08" PRIx32 " heap_limit: 0x%08" PRIx32 " stack_base: 0x%08" PRIx32
```
dragonmux, line 682: To save a tiny bit of space here, and because the radix is quite obvious we think.. perhaps drop the `0x` suffix? We'd be interested in your thoughts there.

dragonmux (changes requested):
There are a couple of items picked up in review, but with them taken care of this is looking excellent and we'll be happy to merge it. Once it is, we'll take a look at the tc_printf() side of this as that'd be a nice thing to get ironed out

src/gdb_hostio.c
ALTracer: I am actually chasing a semihosting hostio bug in the firmware. I should probably post my findings in a separate issue.
For File-I/O F-packets I'm not sure which width should the arguments be in -- target's or gdb host's? (probably not gdb server's which is BMF here) If size_t represents 64-bit ints here, then I reckon PRIX64 will be safer.

src/command.c
ALTracer: I don't use it, this is Koen's code, and it comes without help/usage in-tree, but out-of-tree examples suggest `monitor heapinfo 0x1 0x2 0x3 0x4` to which BMD answers similarly. There is no easy way to read back existing heapinfo values because of encapsulation. %p corresponded to 0x%#x (lx).

dragonmux: Right, fair - just think dropping the `0x`'s in the output could be more elegant here while limiting the firmware bloat; but this works as is, so

src/gdb_hostio.c
dragonmux: newlib-nano does not understand 64-bit format specifiers, it's one of the limitations it has.. that's why we're suggesting going to a 32-bit number unless you want to go through the pain we did for the PIDR in adiv5.c.

`path` is an address on the target, `path_len` is the length of that buffer (for all currently supported targets, that limits it to 32-bit, but obviously that changes once we introduce proper 64-bit target support), and flags + mode are the standard integral `open()` values, so also 32-bit there.

That means that all parameters here are described in terms of the target's address space.

src/gdb_main.c
ALTracer: Yes, once I hid the `hostio_reply()` reference on F-packet case, gcc-12 threw a couple -Wunused
```
gdb_main.c: In function ‘gdb_main_loop’:
gdb_main.c:137:40: error: unused parameter ‘tc’ [-Werror=unused-parameter]
  137 | int gdb_main_loop(target_controller_s *tc, char *pbuf, size_t pbuf_size, size_t size, bool in_syscall)
      |                   ~~~~~~~~~~~~~~~~~~~~~^~
gdb_main.c:137:92: error: unused parameter ‘in_syscall’ [-Werror=unused-parameter]
  137 | int gdb_main_loop(target_controller_s *tc, char *pbuf, size_t pbuf_size, size_t size, bool in_syscall)
      |                                                                                            ^
```
Of course I built BMDA and ran a semihosting testbench sketch against it. It does not need gdb transport hostio, and uses direct host syscalls instead. I believe it was dead code, but I don't have code coverage reports. It can be measured with gcov if need be.

src/target/target.c
ALTracer: Changed style as requested. I only try to mark `#endif` statements far below their corresponding `#if/#ifdef`s (more than 40 lines below), or in complicated nesting cases.

src/gdb_main.c
dragonmux: Ah, perfect. That works for us.

src/target/target.c
dragonmux: Aye, we agree something needs to be done as there are some pretty large gaps in existing code between the preprocessor conditional and the closing `#endif` in some places.

If you want to go through all the code adding these in the style of `#endif /* PC_HOSTED */` in a separate PR, being this would be consistent with the style used for the include guards in the headers, this would be a welcome addition to the code base. Consistency is the key there.

src/gdb_hostio.c
ALTracer: Understood about the casts, even though I don't expect Aarch32 applications to use paths longer that 65535 symbols even. Amended the commit.
After reordering the hostio-hiding commit behind this, I saw a different solution of keeping `%08X` and introducing `%08lX`, which does not litter the format-strings with `PRIX32` all over the place, hurting readability (including reported compiler warning messages). Should I attempt this?
Will 64-bit targets require 64-bit args and `%016lX` in BMD to support hostio for them, and does mainline GDB support it?

dragonmux: No, we are not guaranteed in any way what `uint32_t` maps to - so if we build this with, say, a RISC-V compiler for a potential future version of BMP, we would be sunk again. The `PRI*` macros exist to solve this problem, so please use that solution.

The hope is that we'll be able to rip the format strings out entirely in a future PR where we replace them with direct number formatting instead - eg:

```c
gdb_putpacket_str_part("Fopen,");
gdb_putpacket_upper_hex(path, 8);
gdb_putpacket_char_part('/');
gdb_putpacket_upper_hex(path_len, 8);
gdb_putpacket_char_part(',');
gdb_putpacket_upper_hex(flags, 8);
gdb_putpacket_char_part(',');
gdb_putpacket_upper_hex(mode, 8);
gdb_putpacket_end();
```

If we can get that right we may well be able to save a significant amount of Flash space and all the format string headaches - but this is a ways off for now.

src/command.c
ALTracer: If any software consumers of BMD remote monitor messages exist and use `scanf()` to parse the result, then `man 3 scanf` states that it should consume %p, %x and %X similarly, and discard the optional `0x` or `0X` prefix. I haven't noticed anything like this in the desktop software stack from compuphase. Not sure about Eclipse and other IDEs / graphical debugging frontends with semihosting support.
Wrt people reading the results I suggest just inviting @koendv for review, if you're after these 20 bytes of BMP flash space (concise lowercase %p versus equivalent 0x%08lx as PRIx32, versus prefix-less %x or %X, four times). Excluding the duplicate 54-byte usage message and referring to monitor help is more effective.

src/gdb_hostio.c
ALTracer: Okay. Just saying Farpatch port already includes esp32c3 and allows using BMP on RV32IMC. Not sure about any 64-bit adapters.
The idea with ditching printf and replacing it with fputs/utoa/ftoa is nice, I'd like to follow progess on that (for F4 BMPBootloader reasons).

dragonmux: We aren't referring to being able to use BMD with an ESP32-C3 or other RISC-V targets, but rather *running on RISC-V*. Farpatch actually provides another counter-example to switching to `%08lX` - Xtensa architecture ESP32s also have a different definition of `uint32_t` to ARM.

Wasn't going to go with `fputs()` and those other functions as such, but rather using a direct formatted approach like we use in substrate::console_t which winds up producing far better code generation. But yes, that sort of thing.

src/command.c
dragonmux: We would certainly hope people aren't screen scraping monitor output like that - if we worried about that, we'd never be able to change or fix messages produced by the project.

That said, we cannot use %p here because it is not defined for the conversion in question - converting an integer - only for pointers. See the [cppreference page on fprintf()](https://en.cppreference.com/w/c/io/fprintf) for more there. The compiller should (with gdb_outf() defined as having a format string for its first argument) be warning on this.

We'll table trying to claw back a little space for the moment though with just taking the like-output-for-like-output change and we can try and improve the Flash usage of this command at a later date.

koendv:
The purpose of "mon heapinfo" is setting values for newlib/libc/sys/arm/crt0.S to pick up.
I think few use this. Maybe we can simply drop the command, and see if anyone shouts?

If changing to %x saves a few bytes, fine. If wishing to save more bytes, perhaps:
```
        } else
-               gdb_outf("heapinfo heap_base heap_limit stack_base stack_limit\n");
+               gdb_outf("?\n");
        return true;
 }
```

dragonmux:
Ahh, that's good to know why it exists - perhaps that could be added as a comment in the code documenting its purpose and function + what it interfaces with? It's not currently obvious or noted anywhere in the source.

ALTracer:
> perhaps that could be added as a comment in the code documenting its purpose

@dragonmux @koendv As enabling attribute format for `gdb_voutf` triggers diagnostics on `cmd_heapinfo`, maybe it's better to create a new separate smaller PR for heapinfo changes? And make it a prerequisite for this PR (which is not urgent), so that builds keep passing.
That will need to contain "command: Convert format-string in heapinfo" change and optionally usage message changes, other code-annotating comments **related to semihosting/heapinfo**.

> I think few use this. Maybe we can simply drop the command, and see if anyone shouts?

No need to drop existing tested proved working functionality.

It can also decouple me from the authors who touched that code, if Koen is okay with dedicating some time to fixing printf-correctness in heapinfo (and *probably* RTT).
To observe `-Werror` build failure logs, cherry-pick 1st/4th commit and build with or without `ENABLE_RTT=1`.

koendv:
heapinfo can be used to quickly test a system with different heap and stack values, to see how much heap and stack is needed.
The mechanism is like this:
- user sets up values for heap and stack using "mon heapinfo"
- when the target boots, crt0.S does a heapinfo semihosting call to get these values for heap and stack. See newlib/libc/sys/arm/crt0.S comment: /*  Issue Angel SWI to read stack info.  */
- if the system crashes you need to increase heap or stack

The code in arm/crt0.S is conditionally compiled #ifdef ARM_RDI_MONITOR.

I think the "heapinfo" command is a good candidate for adding to the 2.0 build system. How about "#ifdef ENABLE_HEAPINFO", with default of 0?

koendv:
> It can also decouple me from the authors who touched that code, if Koen is okay with dedicating some time to fixing printf-correctness in heapinfo (and _probably_ RTT).
> To observe `-Werror` build failure logs, cherry-pick 1st/4th commit and build with or without `ENABLE_RTT=1`.

Have taken a look, will make PR.

dragonmux:
> I think the "heapinfo" command is a good candidate for adding to the 2.0 build system. How about "#ifdef ENABLE_HEAPINFO", with default of 0?

That'd be a KConfig style option and we're not yet sure what the macro format for those is going to look like exactly - the ENABLE_ macros we introduced are placeholders to make things work in the current build system.

We're not certain turning this command off in some builds is going to be a good idea over just.. documenting what it exists for in the source. Perhaps though - needs some careful thought and consultation with Esden before we can give a firm direction.

koendv:
ok. done, 60 bytes saved in heapinfo, 66 bytes saved in rtt.

ALTracer:
> Have taken a look, will make PR.

Thank you very much. Rebased this PR branch on main (by cherry-picking), omitted / rebased away the two rtt/heapinfo draft commits as dealt separately in dedicated PRs.

src/gdb_main.c
koendv: ok. BMF needs gdb_hostio to e.g. read a file on the host. BMDA can read files itself, as it's running on the host. So no hostio fields needed.

src/gdb_hostio.c
```diff
@@ -24,24 +24,26 @@
 #include "gdb_hostio.h"
 #include "gdb_packet.h"
 
+#if PC_HOSTED == 0
```
dragonmux, line 27: Just realised re-reading this part of the patch that this means the whole file is gated behind this check if we're building BMDA or not - perhaps rather than topping and tailing the file, we can modify the build system instead to omit this file under BMDA?

ALTracer: I decided to drop that commit and leave it be in favor of some dedicated PR. The 7 warnings for `hosted` disappear anyway after brushing up gdb_hostio.c for PRIX32.

src/gdb_main.c
```diff
@@ -846,7 +846,7 @@ void gdb_poll_target(void)
 		gdb_putpacket_f("T%02X", GDB_SIGINT);
 		break;
 	case TARGET_HALT_WATCHPOINT:
-		gdb_putpacket_f("T%02Xwatch:%08X;", GDB_SIGTRAP, watch);
+		gdb_putpacket_f("T%02Xwatch:%08" PRIx32 ";", GDB_SIGTRAP, watch);
```
dragonmux, line 849: This needs to use `PRIX32` as otherwise some GDBs will fail to properly parse the response.

dragonmux (commented):
Looks like all the big review items have been solved. We spotted one mistake in this round, a simple `x` vs `X` issue in gdb_main.c, but with that fixed we think this is ready for merge.

src/gdb_main.c
ALTracer: Applied, thanks. Do you know when this notification is triggered (to check/reproduce gdb failing to parse the response)?

ALTracer:
Even though I briefly tested these two PRs combined, I strongly recommend merging #1686 first, so that PRIx32 impact can be tested/measured on an actually working in-BMP implementation.

src/gdb_main.c
dragonmux: Whenever you set a watchpoint and it's hit. Something like `awatch *0x<addr>` on an address you know is going to be hit, and then `c` in GDB

dragonmux (approved):
LGTM, we'll merge this after #1686 as suggested as we agree on that order of operations point for testing.

ALTracer:
Rebased away the FIXME comments I left in headers; and rephrased the latest commit message to stay under 70 characters limit of most Git UIs. I don't follow the 50/72 rule (yet), but especially GitHub web interface likes to mangle long subjects.
Cheers to the next developer doing the `tc_printf()` refactor.

dragonmux (approved):
Excellent, we'll get this merged then. Thank you for all your hard work and the contribution!

PR #1679, opened 2023-11-09
Fix GCC warnings

elagil:
## Detailed description

I build with GCC's `-Werror`, so I needed to fix a few warnings.

* Assert that constants are `#define`d, before checking their value
* Fixes for function declarations not being prototypes (missing `void` argument)
* Added a missing `NO_LIBOPENCM3` check, when aiming to disable compilation of the library

### DIFF
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -41,7 +41,7 @@
 #include "rtt.h"
 #endif
 
-#if ADVERTISE_NOACKMODE == 1
+#if defined(ADVERTISE_NOACKMODE) && ADVERTISE_NOACKMODE == 1
 /*
  * This lets GDB know that the probe supports ‘QStartNoAckMode’
  * and prefers to operate in no-acknowledgment mode
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -41,7 +41,7 @@
 #include "rtt.h"
 #endif
 
-#if defined(ADVERTISE_NOACKMODE) && ADVERTISE_NOACKMODE == 1
+#if ADVERTISE_NOACKMODE == 1
 /*
  * This lets GDB know that the probe supports ‘QStartNoAckMode’
  * and prefers to operate in no-acknowledgment mode
--- a/src/include/general.h
+++ b/src/include/general.h
@@ -81,7 +81,12 @@
 #define PRINT_NOOP(...) \
 	do {                \
 	} while (false)
-#if defined(ENABLE_DEBUG)
+
+#ifndef ENABLE_DEBUG
+#define ENABLE_DEBUG 0
+#endif
+
+#if ENABLE_DEBUG == 1
 #define DEBUG_ERROR(...) PLATFORM_PRINTF(__VA_ARGS__)
 #define DEBUG_WARN(...)  PLATFORM_PRINTF(__VA_ARGS__)
 #define DEBUG_INFO(...)  PLATFORM_PRINTF(__VA_ARGS__)
--- a/src/platforms/common/usb_serial.c
+++ b/src/platforms/common/usb_serial.c
@@ -77,7 +77,7 @@ static void debug_serial_receive_callback(usbd_device *dev, uint8_t ep);
 static bool debug_serial_send_complete = true;
 #endif
 
-#if (ENABLE_DEBUG == 1) && defined(PLATFORM_HAS_DEBUG)
+#if ENABLE_DEBUG == 1 && defined(PLATFORM_HAS_DEBUG)
 /*
  * This call initialises "SemiHosting", only we then do our own SVC interrupt things to
  * route all output through to the debug USB serial interface if debug_bmp is true.
@@ -214,7 +214,7 @@ void usb_serial_set_config(usbd_device *dev, uint16_t value)
 	usb_serial_set_state(dev, GDB_IF_NO, CDCACM_GDB_ENDPOINT + 1U);
 	usb_serial_set_state(dev, UART_IF_NO, CDCACM_UART_ENDPOINT + 1U);
 
-#if (ENABLE_DEBUG == 1) && defined(PLATFORM_HAS_DEBUG)
+#if ENABLE_DEBUG == 1 && defined(PLATFORM_HAS_DEBUG)
 	initialise_monitor_handles();
 #endif
 }
@@ -249,7 +249,7 @@ uint32_t debug_serial_fifo_send(const char *const fifo, const uint32_t fifo_begi
 	return fifo_begin;
 }
 
-#if (ENABLE_DEBUG == 1) && defined(PLATFORM_HAS_DEBUG)
+#if ENABLE_DEBUG == 1 && defined(PLATFORM_HAS_DEBUG)
 static bool debug_serial_fifo_buffer_empty(void)
 {
 	return debug_serial_debug_write_index == debug_serial_debug_read_index;
@@ -270,17 +270,17 @@ static void debug_serial_send_data(void)
 	 * If fifo empty, nothing further to do. */
 	if (usb_get_config() != 1 ||
 		(aux_serial_receive_buffer_empty()
-#if (ENABLE_DEBUG == 1) && defined(PLATFORM_HAS_DEBUG)
+#if ENABLE_DEBUG == 1 && defined(PLATFORM_HAS_DEBUG)
 			&& debug_serial_fifo_buffer_empty()
 #endif
 				)) {
-#if (ENABLE_DEBUG == 1) && defined(PLATFORM_HAS_DEBUG)
+#if ENABLE_DEBUG == 1 && defined(PLATFORM_HAS_DEBUG)
 		debug_serial_debug_read_index = debug_serial_debug_write_index;
 #endif
 		aux_serial_drain_receive_buffer();
 		debug_serial_send_complete = true;
 	} else {
-#if (ENABLE_DEBUG == 1) && defined(PLATFORM_HAS_DEBUG)
+#if ENABLE_DEBUG == 1 && defined(PLATFORM_HAS_DEBUG)
 		debug_serial_debug_read_index = debug_serial_fifo_send(
 			debug_serial_debug_buffer, debug_serial_debug_read_index, debug_serial_debug_write_index);
 #endif
--- a/src/platforms/common/usb_serial.c
+++ b/src/platforms/common/usb_serial.c
@@ -40,7 +40,7 @@
  *
  */
 
-#ifndef ENABLE_DEBUG
+#if ENABLE_DEBUG != 1
 #include <sys/stat.h>
 #include <string.h>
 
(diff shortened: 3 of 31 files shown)


dragonmux:
We're unsure which GCC you're building with and what extra options you might be specifying, but if you could please let us know that extra information it would help us review this proposed change set. That's because even with GCC 13 on our Arch systems, we do not get any warnings (turned into errors) so this raises some interesting questions.

Aside from that, please could you rework your commit messages to comply with the contribution guidelines.

We'll endeavour to do a proper review as soon as we can once we have the requested extra information.

elagil:
I changed the commit messages to start with `misc:`. 

You are right, I should have specified the build settings. I am not using the BMP build system directly, but my own `Makefile` with some of the strictest possible GCC settings, namely:
`-Wall -Wextra -Wundef -Wstrict-prototypes -Werror`

I am on `arm-none-eabi-gcc (Arm GNU Toolchain 13.2.rel1 (Build arm-13.7)) 13.2.1 20231009`

Indeed, it is not `-Werror` causing this, but `-Wundef` and `-Wstrict-prototypes`. Most of the existing codebase passes and adheres to these standards.

dragonmux:
[`-Wundef`](https://docs.jade.fyi/gnu/gcc/gcc.html#index-Wundef) only impacts preprocessor macros and them defaulting to 0 is both intended and correct in this code base.

Having it as a `-Wno-error=`'d warning is fine, but we're not sure, at least in the current build system, it makes sense to 'fix' with code-side work over trying to get `-Wconversion -Wsign-conversion -Warith-conversion` enabled which would allow us to crack down on undefined behaviour invocation and logic bugs.

We'll have a think about that particular warning and the resulting code changes and get back to you though as it might be some of the changes actually migrate to the build system (setting `-DENABLE_DEBUG=0` on the compiler command line, for example).

We would recommend adding to your build options `-Wpedantic` in conjunction with `-std=c11` (note, not gnu11, that does something different and is not completely compatible with the codebase).

This all links up with issue #1590 which gives an overview of the progress that's being made toward a much stricter set of warnings (and because of -Werror, errors) being enabled to cover the multitude of crimes and bugs the compiler is able to pick up on and tell us about.

elagil:
Since you mention `ENABLE_DEBUG`, its use is rather inconsistent.

In some places, it is checked for with `#if defined(ENABLE_DEBUG)`, while in others just with `#if ENABLE_DEBUG`, or even `#if ENABLE_DEBUG == 1`.

dragonmux:
Aye, the result of many hands over many years all doing things slightly differently. In principal, `#if ENABLE_DEBUG` means the same thing as `#if ENABLE_DEBUG == 1`, and it's this style of check for the macro that is preferred in all new code. It's on the list of things to sort out to standardise to this.

elagil:
As I understood it:

* `#if ENABLE_DEBUG == 1` is the preferred way of checking for debug mode
* `-Wundef` is not a desired option
* `-Wstrict-prototypes` is not of interest? You mostly use strict prototypes, there are only very few instances of not using them. Having that consistent might make sense.

I can adjust this PR accordingly, and refactor the use of `ENABLE_DEBUG` across the codebase.

dragonmux:
> As I understood it:
> 
> * `#if ENABLE_DEBUG == 1` is the preferred way of checking for debug mode

Correct
> 
> * `-Wundef` is not a desired option

We need to think on it, it might be that actually it makes a lot of sense to stop the current ENABLE_DEBUG mess from reoccurring - but it's late here, so we want to do that thinking after we've had sleep.
> 
> * `-Wstrict-prototypes` is not of interest? You mostly use strict prototypes, there are only very few instances of not using them. Having that consistent might make sense.

It is of interest, and that's actually why we didn't say anything about it - there's nothing contentious there.
> 
> 
> I can adjust this PR accordingly, and refactor the use of `ENABLE_DEBUG` across the codebase.

If you could particularly deal with the ENABLE_DEBUG consistency, that would be a much appreciated contribution in addition to the strict prototypes bit. We'll ping a message on this PR in the morning once we've had a deep think on `-Wundef` and had a chat with Esden to see what direction he wants to take there.

elagil:
Great, looking forward to it!

dragonmux:
Sorry it's taken a bit more than one day to get back to you on this. Having thought on the subject of -Wundef, what we're going to suggest is that: macro usage triggering the warning should be fixed, but by fixing the build system to be more consistent rather than fixing the code to check for the macros being defined. Our thinking is this: If the macros in question are always defined to either 0 or 1, it makes the checks they're involved in not only simpler but more robust for contributors to get right. It allows the project to be perfectly consistent.

Our other ask is that the -Wundef related changes go in a separate commit to the strict prototypes fixes - this way if something is wrong or has to be revisited we can more effectively `git blame` and `git revert` those specific changes. This isn't to say that it'll be required, but more that if it is, it will be better to have such a split.

elagil:
No worries. I will make the changes in the next few days, so they can be reviewed.

dragonmux:
Checking back in with this as it's been a couple of weeks - it'd be nice to get this merged as we've hit some of the warnings this fixes over on the Meson feature PR.

elagil:
Ok, I have not gotten around to it yet. I will prioritize it.

src/target/adiv5.c
```diff
@@ -476,7 +476,7 @@ static uint32_t cortexm_initial_halt(adiv5_access_port_s *ap)
  */
 static bool cortexm_prepare(adiv5_access_port_s *ap)
 {
-#if PC_HOSTED == 1 || ENABLE_DEBUG == 1
+#if (defined(PC_HOSTED) && PC_HOSTED == 1) || (ENABLE_DEBUG == 1)
```
dragonmux, line 479: PC_HOSTED should always be defined - the main Makefile should be ensuring it's 0 for non-BMDA platforms. Same comment about parens.

src/gdb_main.c
```diff
@@ -41,7 +41,7 @@
 #include "rtt.h"
 #endif
 
-#if ADVERTISE_NOACKMODE == 1
+#if defined(ADVERTISE_NOACKMODE) && ADVERTISE_NOACKMODE == 1
```
dragonmux, line 44: We should probably pass `-DADVERTISE_NOACKMODE=0` in the build system rather than do this - that'll make it more directly comparable to what we're doing in the Meson build system to land, and more consistent to the other enable macros.

src/include/general.h
```diff
@@ -76,7 +76,12 @@
 #define PRINT_NOOP(...) \
 	do {                \
 	} while (false)
-#if defined(ENABLE_DEBUG)
+
+#ifndef ENABLE_DEBUG
+#define ENABLE_DEBUG 0
```
dragonmux, line 81: It would be nice if we could pass this via the build system when it shouldn't be true.

src/target/imxrt.c
```diff
@@ -218,7 +218,7 @@ bool imxrt_probe(target_s *const target)
 	snprintf(priv->name, IMXRT_NAME_MAX_LENGTH, "i.MXRT%u", priv->chip_id);
 	target->driver = priv->name;
 
-#if defined(ENABLE_DEBUG) && (PC_HOSTED == 1 || defined(ESP_LOGD))
+#if (ENABLE_DEBUG == 1) && (PC_HOSTED == 1 || defined(ESP_LOGD))
```
dragonmux, line 221: Same comment about parens

dragonmux (changes requested):
This is looking considerably better. While there are several review items, they're mostly targeting the same one issue (excessive use of parens where they can be dropped and avoided) that should be quite an easy (fast) fix. Once these have been addressed, we'll be happy to merge.

src/include/general.h
elagil: You can, see https://github.com/blackmagic-debug/blackmagic/pull/1679/files#diff-3e2513390df543315686d7c85bd901ca9256268970032298815d2f893a9f0685L24-R31

Just set `ENABLE_DEBUG != 1`

This is just a safe guard.

src/platforms/common/usb_serial.c
```diff
@@ -77,7 +77,7 @@ static void debug_serial_receive_callback(usbd_device *dev, uint8_t ep);
 static bool debug_serial_send_complete = true;
 #endif
 
-#if defined(ENABLE_DEBUG) && defined(PLATFORM_HAS_DEBUG)
+#if (ENABLE_DEBUG == 1) && defined(PLATFORM_HAS_DEBUG)
```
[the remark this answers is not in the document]
elagil: I prefer parantheses for explicit operator precedence (as suggested by MISRA), but I will change it.

elagil:
I made these adjustments: https://github.com/blackmagic-debug/blackmagic/pull/1679/files/c7ee563515f65c4403f425fe0c56f2e0516f85f4..2d6f65516f26a3630bdb8e1514444d36b0e92f4d

src/platforms/common/usb_serial.c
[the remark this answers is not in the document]
dragonmux: Ok, we're going to call MISRA out as doing a bad here: The compiler, outside of preprocessor macros, has a warning for accidental assignment, which the parens entirely disable. This has lead to several bugs in BMD that we had to fix that we only found by *removing* the parens to make the compiler chirp - ie, the extra parens are bugprone, turning off the only help you can get from the compiler.

It's a little less bad in the preprocessor situation in that it just muddies the expression needlessly (the preprocessor can't assign in this context), but it's inconsistent with the rest of the code base and comes from a place of worrying about && and || precedence vs equality (equality always binds tighter, making it a non-problem) while disregarding the side effects of the policy (inhibited -Wparentheses).

This is perhaps something to keep in mind as putting in the parens around a complete expression - ie `(a == b && b == c)` is fine and doesn't inhibit this warning - it's specifically around the comparisons that it's a problem.

src/include/general.h
dragonmux: Think it's entirely safe to drop this then in favour of the build system - but we'll keep it in for this PR so that, eg, Farpatch can catch up to the build system changes here.

dragonmux (approved):
LGTM, merging once the CI builds complete. Thank you for the contribution! It's greatly appreciated 😄

elagil:
Great, thanks for merging! 🥳

src/platforms/common/usb_serial.c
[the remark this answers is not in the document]
elagil: I see, interesting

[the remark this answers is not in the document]
dragonmux: We have to admit that we've seen that and whenever it's been done we find it leads to very difficult to read code as all the conditionals are backwards, making one have to do serious mental gymnastics, particularly for inequalities - we honestly favour writing things the normal way around and putting `-Wparentheses` on, banning all assignment in comparisons in the process by virtue of the compiler enforcing it.

We feel it leads to more natural code (a better expression of intent) while keeping the likelihood of introducing the class of bugs MISRA is trying to fix there, as minimal as MISRA might - hence our favouring this other approach. (You might have noticed, we strongly favour having the tools help us rather than trying to write code as if the tools can't without help and it's beholden on the programmer instead, because the programmer can and will make mistakes)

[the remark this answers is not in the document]
elagil: Yes, I agree that it looks quite strange with the conditionals in that order. Using `-Wparantheses` is certainly a better solution to the problem.

Regarding the use of tools and avoiding programmer errors, we enforce MISRA rules by means of something like `cppcheck`, but it is better if the compiler can do it by itself.

PR #1681, opened 2023-11-14
command: Convert format-string in heapinfo

koendv:
## Detailed description

This PR 
- replaces the "%p" format with "%x", keeping existing functionality, but saving a few bytes flash.
- documents how to use "mon heapinfo" 
- see #1676

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -21,7 +21,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-/* 
+/*
  * This file implements a basic command interpreter for GDB 'monitor' commands.
  */
 
@@ -112,7 +112,7 @@ const command_s cmd_list[] = {
 	{"traceswo", cmd_traceswo, "Start trace capture, Manchester mode: [decode [CHANNEL_NR ...]]"},
 #endif
 #endif
-	{"heapinfo", cmd_heapinfo, "Set semihosting heapinfo: HEAPINFO HEAP_BASE HEAP_LIMIT STACK_BASE STACK_LIMIT"},
+	{"heapinfo", cmd_heapinfo, "Set semihosting heapinfo: HEAP_BASE HEAP_LIMIT STACK_BASE STACK_LIMIT"},
 #if defined(PLATFORM_HAS_DEBUG) && PC_HOSTED == 0
 	{"debug_bmp", cmd_debug_bmp, "Output BMP \"debug\" strings to the second vcom: [enable|disable]"},
 #endif
@@ -670,6 +670,14 @@ static bool cmd_shutdown_bmda(target_s *t, int argc, const char **argv)
 }
 #endif
 
+/*
+ * Heapinfo allows passing up to four uint32_t from host to target.
+ * Heapinfo can be used to quickly test a system with different heap and stack values, to see how much heap and stack is needed.
+ * - User sets up values for heap and stack using "mon heapinfo"
+ * - When the target boots, crt0.S does a heapinfo semihosting call to get these values for heap and stack.
+ * - If the target system crashes, increase heap or stack
+ * See newlib/libc/sys/arm/crt0.S "Issue Angel SWI to read stack info"
+ */
 static bool cmd_heapinfo(target_s *t, int argc, const char **argv)
 {
 	if (t == NULL)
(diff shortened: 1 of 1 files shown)


src/command.c
```diff
@@ -670,6 +670,14 @@ static bool cmd_shutdown_bmda(target_s *t, int argc, const char **argv)
 }
 #endif
 
+/*
+ heapinfo can be used to quickly test a system with different heap and stack values, to see how much heap and stack is needed.
+ - user sets up values for heap and stack using "mon heapinfo"
+ - when the target boots, crt0.S does a heapinfo semihosting call to get these values for heap and stack.
+   See newlib/libc/sys/arm/crt0.S "Issue Angel SWI to read stack info"
+ - if the system crashes increase heap or stack
+ */
```
dragonmux, line 679: Please make this comment match the code base multi-line comment style:
```
/*
 * Line 1
 * Line 2
 * - Bullet 1
 * - Bullet 2
 */
```

Please also use proper start-of-bullet and sentence capitalisation.

Looks good otherwise.

src/command.c
```diff
@@ -679,10 +687,11 @@ static bool cmd_heapinfo(target_s *t, int argc, const char **argv)
...
 		target_addr_t stack_limit = strtoul(argv[4], NULL, 16);
-		gdb_outf("heapinfo heap_base: %p heap_limit: %p stack_base: %p stack_limit: %p\n", heap_base, heap_limit,
-			stack_base, stack_limit);
+		gdb_outf("heap_base: %08" PRIx32 " heap_limit: %08" PRIx32 " stack_base: %08" PRIx32 " stack_limit: %08" PRIx32
+				 "\n",
+			heap_base, heap_limit, stack_base, stack_limit);
 		target_set_heapinfo(t, heap_base, heap_limit, stack_base, stack_limit);
 	} else
-		gdb_outf("heapinfo heap_base heap_limit stack_base stack_limit\n");
+		gdb_outf("?\n");
```
dragonmux, line 695: Rather than doing this which is.. cryptic.. at best and a bad change, please make this help string match the command usage string above - string folding will mean this uses no more Flash than the original string while also providing useful help to the user.

src/command.c
```diff
@@ -670,6 +670,14 @@ static bool cmd_shutdown_bmda(target_s *t, int argc, const char **argv)
...
 #endif
 
+/*
+ heapinfo can be used to quickly test a system with different heap and stack values, to see how much heap and stack is needed.
+ - user sets up values for heap and stack using "mon heapinfo"
+ - when the target boots, crt0.S does a heapinfo semihosting call to get these values for heap and stack.
+   See newlib/libc/sys/arm/crt0.S "Issue Angel SWI to read stack info"
+ - if the system crashes increase heap or stack
+ */
+
```
dragonmux, line 680: Please drop the extra new line so InteliSense and other tools see the comment as documentation for the function.

dragonmux (changes requested):
There are a couple of relatively small (easily solved) issues we've spotted in review, but otherwise this looks good as a change. We'll be glad to get this merged once the review items have been addressed as amendments to the existing commit (not new commits).

src/command.c
koendv: OK.
Every command has a one-line help message stored in cmd->help.
Maybe some bytes could be saved if a command, instead of printing a help message and exiting, could signal back: print the one-line help message for my command and exit.

src/command.c
```diff
@@ -670,6 +670,14 @@ static bool cmd_shutdown_bmda(target_s *t, int argc, const char **argv)
 }
 #endif
 
+/*
+ * Heapinfo allows passing up to four int32_t from host to target.
```
dragonmux, line 674: Reading this more closely, is it not 4 uint32_t's? They're not allowed to be negative, and they're stored firmware-side in target_addr_t which is an alias of uint32_t at this time.

src/command.c
dragonmux: That could work for a simple version of the implementation - though the parens around the argc check need dropping and you'd need to use something other than `bool` for `retval` as `bool` is not allowed to have an inspectable numerical value other than 0 and not-0 w/ not-0 being coerced to 1 when converted back to integer.

ALTracer: I like the idea, but this will require changing the `bool (*cmd_handler_fn)(target_s *target, int argc, const char **argv)` function-signature and consequently all the commands just so they can return 2. I'm more inclined towards using identical message strings. See PR1522 for an example of gcc/ld.bfd folding many identical constant string literals.

src/command.c
```diff
@@ -679,8 +687,9 @@ static bool cmd_heapinfo(target_s *t, int argc, const char **argv)
...
 		target_addr_t stack_base = strtoul(argv[3], NULL, 16);
 		target_addr_t stack_limit = strtoul(argv[4], NULL, 16);
-		gdb_outf("heapinfo heap_base: %p heap_limit: %p stack_base: %p stack_limit: %p\n", heap_base, heap_limit,
-			stack_base, stack_limit);
+		gdb_outf("heap_base: %08" PRIx32 " heap_limit: %08" PRIx32 " stack_base: %08" PRIx32 " stack_limit: %08" PRIx32
+				 "\n",
+			heap_base, heap_limit, stack_base, stack_limit);
 		target_set_heapinfo(t, heap_base, heap_limit, stack_base, stack_limit);
 	} else
 		gdb_outf("heapinfo heap_base heap_limit stack_base stack_limit\n");
```
ALTracer, line 695:
```suggestion
		gdb_outf("%s\n", "Set semihosting heapinfo: HEAP_BASE HEAP_LIMIT STACK_BASE STACK_LIMIT");
```
Such that we reuse the *Usage*/help message but still can stick a newline after it. Maybe even `"Usage: %s\n"`.

koendv: > Such that we reuse the _Usage_/help message but still can stick a newline after it.

That's a good one. Maybe even pass the help string to the command?

```
--- a/src/command.c
+++ b/src/command.c
@@ -155,7 +155,7 @@ int command_process(target_s *const t, char *const cmd_buffer)
                 * So 'mon ver' will match 'monitor version'
                 */
                if ((argc == 0) || !strncmp(argv[0], cmd->cmd, strlen(argv[0])))
-                       return !cmd->handler(t, argc, argv);
+                       return !cmd->handler(t, argc, argv, cmd->help);
        }
 
        if (!t)
```

dragonmux (approved):
The current patch as it stands looks good to us - addressing all the review points. We're going to merge it as it stands to clear the way for #1676.

Further improvements to the command system to allow either the commands to take their help strings as a parameter, or for them to return if they want their usage displayed can be worked out in a separate PR that addresses the problem across all commands. Until a particular direction is agreed, may we suggest making an issue ticket for that so we can put all the conversation for it in one spot?

PR #1683, opened 2023-11-14
Feature: RISC-V System Bus support

dragonmux:
## Detailed description

In this PR we implement support for the RISC-V debug 'System Bus' (sysbus) component, as not all targets have to nor do implement the memory I/O abstract command. This complicates things but is required to support, eg, the ESP32-C3.

To accomplish this, we have to deal with sysbus having made all I/O widths optional to support while also needing to be able to do any arbitrarily sized and aligned read/write. The one benefit is that, depending on implementation, sysbus may be used without halting the Hart, unlike the abstract command interface which requires it halted (read: due to it injecting an instruction stream into the core to run the read/write, though this is strictly an implementation detail).

Unfortunately, this puts the burden of performing the I/O correctly on us, but we've attempted to minimise the logic and complications caused by this and document everything thoroughly. This logic has been tested on an ESP32-C3 and works.

### DIFF
--- a/src/target/riscv32.c
+++ b/src/target/riscv32.c
@@ -217,7 +217,7 @@ uint32_t riscv32_pack_data(const void *const src, const uint8_t access_width)
 static void riscv32_abstract_mem_read(
 	riscv_hart_s *const hart, void *const dest, const target_addr_t src, const size_t len)
 {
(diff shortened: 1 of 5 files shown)


esden (approved):
LGTM

PR #1151, opened 2022-07-30
Added support for AT32F403A/407/415.

grumat:
The following PR adds support for ArteryTek ATF32F403A/407/415 family of MCUs based on Cortex M4 and peripherals compatible to STM32F103 family.
Some of the elements of this family does support an _external Flash_ feature, which would require customized flash routines. The current solution will impose flash limit to 256KB, which are accessible using the STM32F103 compatible interface.
The PR was tested with the [WeActStudio Blackpill](https://github.com/WeActStudio/WeActStudio.BlackPill) based on AT32F403ACGU. English datasheet for these parts can be obtained on the [ArteryTek website](https://www.arterytek.com/en/product/AT32F403A.jsp).
A simple Blink Led program was used to test using VisualGDB and Visual Studio 2022. The GDB interaction log is on the attachment: [stm32_02-Debug.log](https://github.com/blackmagic-debug/blackmagic/files/9227040/stm32_02-Debug.log)

PR #1685, opened 2023-11-16
Fix: AT32F403A/407 dual bank flash support

ALTracer:
## Detailed description

* Support for Artery AT32F403A "Mainstream" devices introduced in #1151 declares only the common/minimum 256 KiB of flash in first bank.
* This PR increases the writable flash size to full 512 KiB in first bank and introduces logic to enable directing erases/writes to second 512 KiB bank, using the existing XL-density part_id checks of stm32f1 flash driver API.

I tested the idea by dropping the dual-bank 0x430U check altogether, and an AT32F403ACGU7 target (WeAct.BlackPill) gets written with my test payload with a long zeroed .rodata array, which passes the check in-payload, mass erase also works (checked with gdb x/16zw 0x08080000 as there is no blank-check routine in BMD). There is no obvious reason to me why at least the first 512 KiB bank wasn't declared with full size.

A related change includes updating no-op aliases for `at32f40x_probe()` I renamed from `at32f4xx_probe()`, and for new `at32f43x_probe()`, such that the at32f43x.c translation unit can be dropped from Makefile target driver sources list.

## Closing issues

Closes my #1542 point 1.


dragonmux (approved):
LGTM, we'll run some tests on STM32F1's - see if we can find one on our desk that uses the dual bank 0x430U check - and verify that this all till works after. After checking and assuming nothing is broken, we'll get this merged.

ALTracer:
Actually hold on, nothing in at32f40_probe/_detect writes the target->part_id, which is why my testing `hosted` build failed to erase and write into the second bank. I'll need to add that first. Not marking as WIP/Draft either.
W.r.t. regress-testing XL-density dual-bank stm32f1 targets -- yes please, check on some if you have them reachable (I don't seem to have any, just the Medium Density). I only delegated/extracted the check, but it doesn't lift the requirements for testing on live devices.

dragonmux:
Is this ready for re-review?

dragonmux:
We were unsure as you hadn't let us know to stay the request to hold. We'll get this review in a bit then 🙂

dragonmux (approved):
This all looks good to us now - so, based on the eager merge policy in the face of us not having any XL parts to test with, we're going to merge this and see what people report back with it on `main`.

PR #1686, opened 2023-11-19
Fix: Semihosting in BMP

### LINKED ISSUE #1678: Semihosting broken in v1.10
I was testing the latest release v1.10 to see if https://github.com/blackmagic-debug/blackmagic/issues/970 is resolved there. While doing so, I discovered that the semihosting support appears to be broken there. Testing has been conducted with arm-none-eabi GDB versions 10.3 and 12.3 with identical results. The target MCU is STM32F446 connected via SWD. The BMP hardware is the native probe.

First, I was able to successfully load and run ordinary firmware that does not use semihosting. The whole stack operated as intended without noticeable issues.

When I attempted to run a different firmware image that does make use of semihosting, I discovered that some of the semihosting calls were not being executed by the host and that at least some attempts to execute a semihosting call caused the debugger to stall code execution for somewhere between a few hundred ms up to seconds. Shortly after starting code execution, GDB prints the following message and then stops completely:

```
warning: Invalid remote reply:
```

Sic! There is nothing after the colon. Pressing Ctrl+C results in "Disconnected from target." being printed.

Enabling remote protocol debugging revealed the following:

```gdb
(gdb) cont  # At this point the program is in the reset state.
Continuing.
[remote] Sending packet: $c#63
[remote] Received Ack
[remote] wait: enter
[remote] wait: exit
[remote] wait: enter
  [remote] Packet received: Funlink,080559D4/11
  [remote] Sending packet: $Hg0#df
  [remote] Received Ack
  [remote] Packet received: OK
  [remote] Sending packet: $m80559d4,11#9e
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] Sending packet: $m80559d4,11#9e
  [remote] Received Ack
  [remote] Packet received: <REDACTED, the value is correct>
  [remote] Sending packet: $F-1,2#02
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] Sending packet: $F-1,2#02
  [remote] Received Ack
[remote] wait: exit
[remote] wait: enter
  [remote] Packet received: Fwrite,00000002,080575DC,00000007
  [remote] Sending packet: $m80575dc,7#a0
  [remote] Received Ack
  [remote] Packet received: 4153534552543A
ASSERT:  [remote] Sending packet: $F7#7d
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] Sending packet: $F7#7d
  [remote] Received Ack
[remote] wait: exit
[remote] wait: enter
  [remote] Packet received: Fwrite,00000002,08054A70,00000029
  [remote] Sending packet: $m8054a70,29#9d
  [remote] Received Ack
  [remote] Packet received: <REDACTED, the value is correct>
[remote] Sending packet: $F29#b1
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] Sending packet: $F29#b1
  [remote] Received Ack
[remote] wait: exit
[remote] wait: enter
  [remote] Packet received: Fwrite,00000002,08054F20,00000001
  [remote] Sending packet: $m8054f20,1#63
  [remote] Received Ack
  [remote] Packet received: 3A
[remote] Sending packet: $F1#77
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] Sending packet: $F1#77
  [remote] Received Ack
[remote] wait: exit
[remote] wait: enter
  [remote] Packet received: Fwrite,00000002,20007795,00000003
  [remote] Sending packet: $m20007795,3#6a
  [remote] Received Ack
  [remote] Packet received: 323234
[remote] Sending packet: $F3#79
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] Sending packet: $F3#79
  [remote] Received Ack
[remote] wait: exit
[remote] wait: enter
  [remote] Packet received: Fwrite,00000002,080574FC,00000002
  [remote] Sending packet: $m80574fc,2#9c
  [remote] Received Ack
  [remote] Packet received: 0D0A

  [remote] Sending packet: $F2#78
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] putpkt_binary: Packet instead of Ack, ignoring it
  [remote] Sending packet: $F2#78
  [remote] Received Ack
[remote] wait: exit
[remote] wait: enter
  [remote] Packet received: 
warning: Invalid remote reply: 
[remote] wait: exit
[remote] pass_ctrlc: enter
  [remote] interrupt: enter
  [remote] interrupt: exit
[remote] pass_ctrlc: exit
[remote] pass_ctrlc: enter
[remote] pass_ctrlc: exit
Disconnected from target.
```

For comparison, here is the log captured with BMP firmware v1.9.1 with the same firmare using GDB v10.3 (not possible to test it with GDB v12.3 because of https://github.com/blackmagic-debug/blackmagic/issues/929), where everything is functioning correctly:

```gdb
(gdb) cont
Continuing.
Sending packet: $c#63...Ack
Packet received: Funlink,080559D4/11
Sending packet: $Hg0#df...Ack
Packet received: OK
Sending packet: $m80559d4,11#9e...Ack
Packet received: <REDACTED>
Sending packet: $F-1,2#02...Ack
Packet received: Fwrite,00000002,200077F4,00000017
Sending packet: $m200077f4,17#cb...Ack
Packet received: <REDACTED>
Started at 0.000088766
Sending packet: $F17#ae...Ack
Packet received: Fopen,08055A7C/14,00000602,000001A4
Sending packet: $m8055a7c,14#cb...Ack
Packet received: <REDACTED>
Sending packet: $F3#79...Ack
Packet received: Fwrite,00000003,08055A90,00000211
Sending packet: $m8055a90,200#c7...Ack
```

The connection sequence I used is as follows:

```
tar ext <PORT>
monitor connect_rst enable
monitor auto_scan
attach 1
load
```

ALTracer:
## Detailed description

* The problem is BMP's implementation of semihosting via GDB File-I/O (not BMDA) has been broken since #1284. Some attempt was made in #1432 to make it at least not crash the adapter.
* The PR solves the problem by introducing a second loop to service GDB packets before/other than F-reply.

I based the WIP tree on PR1432 single commit v1.9.0-435-g906f8878 backported to just past PR1284 merge, v1.9.0-55-g5cf0dd27. I had to enable DEBUG_GDB, DEBUG_GDB_WIRE and drop a bunch of drivers to fit into `swlink`, because my adapter of choice (blackpill-f411ce) did not exist until approximately v1.9.0-700 in usable form. The logs from BMP internals were **critical** for understanding how RSP packets go missing.

Current patch adds a `while(true)` loop at stack depth of 9 functions to allow for parsing the 1-4 packets GDB has to emit before finishing the semihosting File-I/O syscall and returning its result code. AFAIK this has to be avoided in out-of-tree projects like Farpatch which will reboot due to watchdog timeouts from not going through the superloop often enough. But this variant seems to work, and you all can decide how to improve it, because it provides some starting/reference point.

The less related change replaces the `syscall 0lx1 (800289c 0 3 20000054)` prints 
with more readable `syscall     SYS_OPEN (800289c 0 3 20000054)`. Since v1.10.0 release I don't care about flash space much; it can be dropped in case of backporting, but it shouldn't affect release firmware because it's in DEBUG_INFO (and I could gate it on ENABLE_DEBUG for 332 bytes).

Tested on `swlink` attached to Nucleo-G071RB target running SemihostingTest.ino over SWD (with its onboard JLinkReflash idling), using GDB-8/10/12 (NuttX, ST, GNU-RM).

* [x] It builds as BMDA (`make PROBE_HOST=hosted`) *and should not affect it*

## Closing issues

Fixes #1678.

### DIFF
--- a/src/gdb_hostio.c
+++ b/src/gdb_hostio.c
@@ -73,8 +73,18 @@ int hostio_reply(target_controller_s *const tc, char *const pbuf, const int len)
 static int hostio_get_response(target_controller_s *const tc)
 {
 	char *const packet_buffer = gdb_packet_buffer();
-	const size_t size = gdb_getpacket(packet_buffer, GDB_PACKET_BUFFER_SIZE);
-	return gdb_main_loop(tc, packet_buffer, GDB_PACKET_BUFFER_SIZE, size, true);
+	/* Still have to service normal 'X'/'m'-packets */
+	while (true) {
+		/* Get back the next packet to process and have the main loop handle it */
+		const size_t size = gdb_getpacket(packet_buffer, GDB_PACKET_BUFFER_SIZE);
+		/* If this was an escape packet (or gdb_if reports link closed), fail the call */
+		if (size == 1U && packet_buffer[0] == '\x04')
+			return -1;
+		const int result = gdb_main_loop(tc, packet_buffer, GDB_PACKET_BUFFER_SIZE, size, true);
+		/* If this was an F-packet, we're done */
+		if (packet_buffer[0] == 'F')
+			return result;
+	}
 }
 
 /* Interface to host system calls */
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -117,6 +117,42 @@ typedef struct cortexm_priv {
 	uint32_t demcr;
 } cortexm_priv_s;
 
+#ifdef ENABLE_DEBUG
+const char *const semihosting_names[] = {
+	"",
+	"SYS_OPEN",
+	"SYS_CLOSE",
+	"SYS_WRITEC",
+	"SYS_WRITE0",
+	"SYS_WRITE",
+	"SYS_READ",
+	"SYS_READC",
+	"SYS_ISERROR",
+	"SYS_ISTTY",
+	"SYS_SEEK",
+	"0x0b",
+	"SYS_FLEN",
+	"SYS_TMPNAM",
+	"SYS_REMOVE",
+	"SYS_RENAME",
+	"SYS_CLOCK",
+	"SYS_TIME",
+	"SYS_SYSTEM",
+	"SYS_ERRNO",
+	"0x14",
+	"SYS_GET_CMDLINE",
+	"SYS_HEAPINFO",
+	"0x17",
+	[SEMIHOSTING_SYS_EXIT] = "SYS_EXIT",
+	/* 7 reserved */
+	[SEMIHOSTING_SYS_EXIT_EXTENDED] = "SYS_EXIT_EXTENDED",
+	/* 15 reserved */
+	[SEMIHOSTING_SYS_ELAPSED] = "SYS_ELAPSED",
+	[SEMIHOSTING_SYS_TICKFREQ] = "SYS_TICKFREQ",
+	"",
+};
+#endif
+
 /* Register number tables */
 static const uint32_t regnum_cortex_m[CORTEXM_GENERAL_REG_COUNT] = {
 	0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, /* standard r0-r15 */
@@ -1434,8 +1470,16 @@ static int cortexm_hostio_request(target_s *target)
 		target_mem_read(target, params, arm_regs[1], sizeof(params));
 	int32_t ret = 0;
 
-	DEBUG_INFO("syscall 0" PRIx32 "%" PRIx32 " (%" PRIx32 " %" PRIx32 " %" PRIx32 " %" PRIx32 ")\n", syscall, params[0],
-		params[1], params[2], params[3]);
+#ifdef ENABLE_DEBUG
+	const char *syscall_descr = NULL;
+	if (syscall < ARRAY_LENGTH(semihosting_names))
+		syscall_descr = semihosting_names[syscall];
+	if (syscall_descr == NULL)
+		syscall_descr = "";
+
+	DEBUG_INFO("syscall %12s (%" PRIx32 " %" PRIx32 " %" PRIx32 " %" PRIx32 ")\n", syscall_descr, params[0], params[1],
+		params[2], params[3]);
+#endif
 	switch (syscall) {
 #if PC_HOSTED == 1
 


src/gdb_hostio.c
```diff
@@ -73,8 +73,16 @@ int hostio_reply(target_controller_s *const tc, char *const pbuf, const int len)
...
+	/* Still have to service normal 'X'/'m'-packets */
+	while (true) {
+		const size_t size = gdb_getpacket(packet_buffer, GDB_PACKET_BUFFER_SIZE);
+		/* GDB will need us to proxy access the inferior's memory */
+		if (packet_buffer[0] != 'F')
+			gdb_main_loop(tc, packet_buffer, GDB_PACKET_BUFFER_SIZE, size, false);
+		else
+			/* GDB finished proxied inferior memory manipulations and returns syscall exitcode */
+			return gdb_main_loop(tc, packet_buffer, GDB_PACKET_BUFFER_SIZE, size, true);
+	};
```
dragonmux, line 85: Given the error handling in gdb_main_loop() and the insensitivity to the in_syscall parameter for the entire rest of the handling switch-case (it only impacts `F` packets), this feels like it would be better pivoted slightly. It also needs to handle the BMDA shutdown command otherwise that will cause an infinite loop here:

```suggestion
	while (true) {
		/* Get back the next packet to process and have the main loop handle it */
		const size_t size = gdb_getpacket(packet_buffer, GDB_PACKET_BUFFER_SIZE);
#if PC_HOSTED == 1
		/* If the packet was a escape packet and we're supposed to be shutting down, exit */
		if (size == 1U && packet_buffer[0] == '\x04' && shutdown_bmda)
			return 0;
#endif
		const int result = gdb_main_loop(tc, packet_buffer, GDB_PACKET_BUFFER_SIZE, size, true);
		/* If this was an F-packet, we're done */
		if (packet_buffer[0] == 'F')
			return result;
	}
```

Note that you have a trailing `;` on the loop which we're surprised is not linting.

dragonmux (changes requested):
The new/extra debug info for syscalls looks quite handy - nice addition. There's a small problem your patch for hostio_reply() introduces, for which we have a suggested fix, but this is looking good overall and explains in the code comments why what we did before was broken (subtly) which is excellent.

It's looking good and with the review item addressed, we'll be happy to merge this.

src/gdb_hostio.c
ALTracer: Adjusted to suggested-changes. I disagree on PC_HOSTED because this code is unreachable under BMDA (it never goes here but uses desktop OS syscalls directly).

dragonmux: Ah, right, because it handles those requests an entirely different way - sure, you can drop that chunk of the suggestion then.

ALTracer: Though I wonder if Ctrl+D or \x04 event handling should be added here in the rare race condition event that the user presses this in the short timeframe of a syscall in flight (or does gdb guard against this? returning -1 and getting us back to first main loop)

dragonmux: Unsure - it's entirely possible that GDB just slams the link closed, at which point all gdb_getpacket() will ever return are injected `'\x04'` packets making this forever busy loop - which is the essence of the worry we had in the case of `mon shutdown` too for BMDA. It's a narrow window when this could happen, but perhaps err on the side of caution?

dragonmux:
When we merge this we'll backport it to v1.10 as part of the first point release for that firmware - seeing how broken this is in that release.

It would be helpful if @xobs could please give this patch a spin and let us know how it interacts on Farpatch. We are inclined to merge this fix and then any further work that's needed to make Farpatch happy as a second pass.

xobs:
No issues on Farpatch -- still works just fine, though I've never used semihosting.

src/gdb_hostio.c
ALTracer: My logic is as follows: in case gdb_getpacket() indeed returns a genuine (or injected by DTR-low/link loss) `\x04`, let's fail the semihosting call with a -1, make BMP resume the target and give it a chance to see this. Then the very next gdb_main_loop() called from bmp_poll_loop() will trigger on another `\x04` and detach from the target.
The other option is not to do this and let gdb_main_loop() detach in my face, but then I'd need to `if(!cur_target) return 0;`

dragonmux: That logic seems sound to us 👍🏼 let's do that then.

dragonmux (approved):
LGTM, merging and backporting. Thanks for the contribution and all your hard work making it!

ALTracer:
> No issues on Farpatch -- still works just fine, though I've never used semihosting.

The loop is only reachable from (most of) semihosting calls, and it's harder for me to check that it won't cause issue for your platform/RTOS watchdog. If you don't have your own firmware with semihosting, but you have some targets compatible with Arduino cores like stm32duino, then you could just spin Koen's SemihostingTest.ino like I usually do. Although you're welcome to write a simpler testcase, just verify it accesses desktop GDB (some calls like errno and iserror are handled purely inside BMP)

PR #1687, opened 2023-11-22
send initial usb cdc notification from the correct endpoint

brainsmoke:
Hi,

I was looking into usb-cdc-acm implementations to understand the required descriptors/endpoints when I found something that's probably a bug:

It looks like usb cdc acm notifications are sent from the data endpoint in `usb_serial_set_config()`
instead of the notifications endpoint. The function `usb_serial_set_state()` is called twice more in this file, but with notification endpoints as argument.

https://github.com/blackmagic-debug/blackmagic/blob/main/src/platforms/common/usb_serial.c#L106
https://github.com/blackmagic-debug/blackmagic/blob/main/src/platforms/common/usb_serial.c#L144

`make PROBE_HOST=native` fails for me for both pre & post fix, due to firmware size overruns, but I've tested an stlink / blue pill build which works.

I was expecting the wrongly sent data to come out of ttyACM[01] without the fix, but that doesn't seem to be the case.
I guess Linux only sends serial data that arrives after you open the device and I probably wasn't fast enough to win this race.

* [*] It builds as BMDA (`make PROBE_HOST=hosted`)

 * I used HOSTED_BMP_ONLY=1 to avoid installing dependencies, but the change is in the firmware anyway

### DIFF
--- a/src/platforms/common/usb_serial.c
+++ b/src/platforms/common/usb_serial.c
@@ -211,8 +211,8 @@ void usb_serial_set_config(usbd_device *dev, uint16_t value)
 	/* Notify the host that DCD is asserted.
 	 * Allows the use of /dev/tty* devices on *BSD/MacOS
 	 */
-	usb_serial_set_state(dev, GDB_IF_NO, CDCACM_GDB_ENDPOINT);
-	usb_serial_set_state(dev, UART_IF_NO, CDCACM_UART_ENDPOINT);
+	usb_serial_set_state(dev, GDB_IF_NO, CDCACM_GDB_ENDPOINT + 1U);
+	usb_serial_set_state(dev, UART_IF_NO, CDCACM_UART_ENDPOINT + 1U);
 
 #if defined(ENABLE_DEBUG) && defined(PLATFORM_HAS_DEBUG)
 	initialise_monitor_handles();


ALTracer:
Hi. My review with no weight is this:
1. minicom on Linux re-opens ttyBmpTarg on any BMP rebooting, so you can try that to "win the race" condition, if Wireshark+usbmon doesn't cut it;
2. This is some old code regarding desktop OS compatibility. Can you support this with docs, or, better yet, test with MacOS and FreeBSD hosts?
3. I have a non-PR branch `endpoints` in my fork, where I started replacing and documenting related stuff to fight with dwc2 of stm32f4 probes. https://github.com/ALTracer/blackmagic/commit/09bd734dd8b237b9838bd4f73e83d3fffd82f908

dragonmux (commented):
We can confirm that this is indeed a bug (our mistake, in fact) and solves the "where are the improper notification packets coming from?!" question that had been asked recently.

Please update your commit message to adhere to the contribution guidelines and then, once we've been able to confirm the firmware works correctly on a macOS machine, we'll approve this and get it merged.

brainsmoke:
I've changed the commit message; looks like @ALTracer 's refactor with explicit constants for endpoints probably would've prevented this :-)

dragonmux (approved):
LGTM now, merging. Thank you for the contribution!

We'll also backport this to the v1.10 and v1.9 branches too.

PR #1536, opened 2023-07-12
Fix: Increase SysTick frequency to 1kHz

ALTracer:
## Detailed description

* No new features.
* The pull request solves the problem of relatively slow RTT streaming speed of BMF.
* The pull request solves the problem by making it possible to poll faster than every 10ms. This requires raising the SysTick interrupt frequency from historically 100Hz to 1000Hz.

## Your checklist for this pull request

* [x] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md)
* [x] I've read the [guidelines for contributing](https://github.com/blackmagic-debug/blackmagic/blob/main/CONTRIBUTING.md) to this repository
* [x] It builds for hardware native (`make PROBE_HOST=native`)
* [x] It builds as BMDA (`make PROBE_HOST=hosted`)
* [x] I've tested it to the best of my ability
* [x] My commit messages provide a useful short description of what the commits do

## Closing issues

No known to me issues solved by this PR.

## Testing

I conducted tests of RTT streaming from @koendv 's Arduino sketch running on NUCLEO-G071RB (and some tests were repeated on a WeAct Studio BluePIll-Plus with genuine stm32f103cbt6). Here are the results in chars/s as reported by DUT measuring a hardware timer prescaled to 1ms ticks:
* `blackpill-f411ce` before: 34000..51000, after: 87000..102400;
* `swlink` before: 32000..34000, after: 78000;
* `f072rb` before: untested, after: 42000. This is on a 32F072B-DISCO just to check on the slowest (48MHz) supported platform.

If dropping the PC_HOSTED separation is wrong, I will rebase it out. For reference `hosted` uses `usleep(ms * 1000U)` or `Sleep(ms)` for delays but `gettimeofday()` for timestamping.

PR #1688, opened 2023-11-25
Increase performance of SWJ-DP bitbanging

ALTracer:
## Detailed description

* This is a refactoring of core bitbanging routines.
* The existing problem was IMO suboptimal efficiency of swdptap, jtagtap functions.
* This PR solves this problem by extracting and improving the busy-looping delay; dropping double STM32F4 GPIO_BSRR writes; and rebalancing some of the no_delay routines to adjust for changes.

Tested on `blackpill-f411ce` platform connected to a stm32f103cb DUT. Firmware does not crash on exceptions from spurious ACKs. The 24MHz fx2la does not let me verify setup-and-hold timing compliancy, but the target silicon replies okay. I could retest with other targets accessible to me, possibly with low HCLKs. I was focusing ot stm32f4 platforms, but lm4f `launchpad-icdi` may be affected, too.
LINERESET waveform improved from 2.988 MHz to ~8.000 MHz, that is `swd_seq_out_no_delay()`.
BMDA `-r -f 8M` was 22.5 and 41 KiB/s in JTAG and SWD mode, becomes 28 and 44 KiB/s respectively (but that bottleneck is different).

The PR aims to:
* increase raw SWD performance of capable platforms (stm32f4), and by extension JTAG performance;
* improve granularity of selecting max frequency (thanks to a quicker busy-delay loop); and fix said frequency calculations;
* rebalance edge-case timings where needed;
* optionally reduce AP WAIT polling spam (this makes logs and wire capture snapshots cleaner, but may hurt performance).

Important mentions: 
GPIO_BSRR hack was introduced in https://github.com/blackmagic-debug/blackmagic/commit/f1ea5ed8f9f6dc1f006185d051da682bed63c6ac ; the reasoning behind it is not clear to me, but we can either set some default divider to target lower (2 MHz) frequency, and BMDA already asks for some 3 or 4 MHz. I wired everything on a breadboard with flying wires and a couple ground connections (no ribbon cables).
`swdptap.c` has `optimize(3)` since https://github.com/blackmagic-debug/blackmagic/pull/669/commits/4698a26bab59236efc0217068b49ae734853fd86 (but not `jtagtap.c`). -O3 increases flash footprint and may alter the timings.

## Closing issues
No known issues w.r.t. SWJ speed.


src/platforms/common/jtagtap.c
```diff
@@ -31,11 +31,13 @@
 jtag_proc_s jtag_proc;
 
 static void jtagtap_reset(void);
-static void jtagtap_tms_seq(uint32_t tms_states, size_t ticks);
-static void jtagtap_tdi_tdo_seq(uint8_t *data_out, bool final_tms, const uint8_t *data_in, size_t clock_cycles);
-static void jtagtap_tdi_seq(bool final_tms, const uint8_t *data_in, size_t clock_cycles);
-static bool jtagtap_next(bool tms, bool tdi);
-static void jtagtap_cycle(bool tms, bool tdi, size_t clock_cycles);
+static inline void platform_delay_busy(const uint32_t loops) __attribute__((always_inline));
+static void jtagtap_tms_seq(uint32_t tms_states, size_t ticks) __attribute__((optimize(3)));
```
dragonmux, line 35: We'll have to do some captures and testing with this attribute to make sure it's not shooting the timings in the foot - as you note in the PR description. We do note that this particular function signature still has the old `ticks` nomenclature in it, so please switch that to `clock_cycles`.

src/platforms/common/stm32/timing_stm32.c
```diff
@@ -194,8 +204,26 @@ uint32_t platform_max_frequency_get(void)
...
-	result /= USED_SWD_CYCLES + CYCLES_PER_CNT * target_clk_divider;
+	result /= USED_SWD_CYCLES + CYCLES_PER_CNT * target_clk_divider * 2U;
 	return result;
 #endif
 }
+
+/* Busy-looping delay for GPIO bitbanging operations. SUBS+BNE.N take 4 cycles. */
+void platform_delay_busy(const uint32_t loops)
+{
+	/* Avoid using `volatile` variables which incur stack accesses */
```
dragonmux, line 218: The [cppreference page on C's `volatile`](https://en.cppreference.com/w/c/language/volatile) does not mention anything about forced stack usage and we've never seen it imply any such - are we sure that is what the compiler's doing here? Either way, we would strongly prefer you to use a for loop not a do-while here. This is a for loop in disguise.

The disappearance of the loop when using `continue;` without `volatile` can be entirely explained not so much by DCE (it's the result, not the cause) but by there being no observable side effects thus allowing the compiler to assume that nothing of value is done by the loop. This is the reason for using `volatile` in the jtagtap code. Perhaps we can use `volatile register` to force the compiler to consider the operation to have observable side-effects but also keep the loop counter register-bound?

src/platforms/common/swdptap.c
```diff
@@ -53,6 +54,20 @@ void swdptap_init(void)
...
+/* Busy-looping delay snippet for GPIO bitbanging (rely on inlining) */
+static inline void platform_delay_busy(const uint32_t loops)
+{
+#if 1
+	register uint32_t i = loops;
+	do {
+		__asm__("nop");
+	} while (--i > 0U);
+#else
+	for (register uint32_t counter = loops; counter > 0; --counter)
```
dragonmux, line 66: Same comments as in jtagtap.c

src/platforms/common/swdptap.c
```diff
@@ -108,8 +121,9 @@ static uint32_t swdptap_seq_in_no_delay(const size_t clock_cycles)
 	for (size_t cycle = 0; cycle < clock_cycles; ++cycle) {
 		gpio_clear(SWCLK_PORT, SWCLK_PIN);
 		value |= gpio_get(SWDIO_IN_PORT, SWDIO_IN_PIN) ? 1U << cycle : 0U;
+		__asm__("" ::: "memory");
```
dragonmux, line 124: Please insert a note here about this being a reordering barrier - same as in other places the trick is used.

src/platforms/common/swdptap.c
```diff
@@ -182,13 +199,16 @@ static void swdptap_seq_out(const uint32_t tms_states, const size_t clock_cycles
 
 static void swdptap_seq_out_parity(const uint32_t tms_states, const size_t clock_cycles)
 {
-	int parity = __builtin_popcount(tms_states);
+	const int parity = __builtin_popcount(tms_states);
```
dragonmux, line 202: As this is only used in the one place and is bitwise masked, perhaps lets make this `const uint8_t parity` and put the bit mask on the end, lifting it from the `gpio_set_val()` call below? Failing that, we should make it `uint32_t` so we don't do a signed+bitwise oops below.

dragonmux (changes requested):
This is looking provisionally very good and we see a lot of promise here - we'll have to do a pile of testing to make sure this doesn't regress the setup-and-hold timings, but mostly the notes and comments in this review are about form not function. The whole `volatile` vs `register` thing in particular raises some interesting questions to explore.

src/platforms/common/stm32/timing_stm32.c
ALTracer: Indeed `volatile register` was the first thing I tried, only to be yelled at by the compiler to swap qualifiers... and identical codegen. I specifically wanted to avoid extraneous ldr/str of loop counters, so I monitored Puncover and objdump -CSd. Maybe Compiler Explorer can shed some light on why this codegens this way.

dragonmux: Oh how interesting - ok, that'll be an interesting session in CE then as yes, that might well provide some insight into how and why the compiler's choosing to do what it does.

ALTracer: Probably qualifying a variable as `volatile` forces the compiler to assume that it should be observable elsewhere, and hence needs to be pinned to SRAM (if only on the stack). This directly contradicts with the intent of register-allocated counters (which debuggers don't display due to "optimized away" but you can always dump the gp_regs). Something "dead store elimination".
And the for-loop form is longer and has an extra jump, I didn't subjectively like it; objectively it has to fit into Flash prefetch buffer which is a few (4) 64-bit lines on F103 (?) to not incur wait-states of 72MHz MCU against 24MHz flash on random jumps.

dragonmux: From our perspective re the for loop part, the for loop form should code gen to the same or very near the same as the do-while (perhaps needs the condition expressed as `--counter > 0` and `--counter` in the update step removed to do it) while being a much clearer expression of your intent. Think this is very much going to be "take it to Compiler Explorer and tinker for an hour" to find the right thing to do and the right way to express this.

koendv:
With #1688 and #1690 applied, rtt has a measured speed of 114.292 chars/s if target is stm32f102 target, 121.297 chars/s if target is stm32f411.
The documentation UsingRTT.md mentions a speed of 50.000 chars/s, perhaps update this number if PR is merged.

src/platforms/common/swdptap.c
ALTracer: Zooming out: do we actually care about 2..32 number of bits, or just the SWD parity per IHI0031 B4.1.6? The same 4698a26ba introduced the two builtin calls, which was a better idea than spinning the parity manually. And I see an opportunity to eliminate either `__popcountsi2()` or `__paritysi2()`, currently both are used, only once. They are 32+40 bytes, (36+52 depending on platform Thumb2 availability; popcount is bigger). GCC sees we don't care in seq_out and substitutes parity in place of popcount (because of the LSBit masking), but doesn't see in seq_in and leaves popcount alone.

src/platforms/common/jtagtap.c
ALTracer: 750 kHz of tdi_tdo_seq on `native` was unremarkable, I hope performance improves -- please note, during capture/analysis, whether a given hardware relies on ancient FLITF Prefetch (2x64-bit lines) or runs from "0-wait-state" shadow SRAM.
Updated `jtagtap_tms_seq()` declaration and definition.

src/platforms/common/stm32/timing_stm32.c
ALTracer: Your suggestion of `--counter > 0` gets the codegen I want and the C idiomatic form you look to preserve, thanks.
Updated the three places accordingly.
m3/m4/m7 still get the SUBS+BNE.N, m0 (`f072`) gets `subs; cmp; bne.n` so its divisor timings will be off.

src/platforms/common/jtagtap.c
```diff
@@ -51,8 +53,12 @@ void jtagtap_init(void)
 	jtag_proc.tap_idle_cycles = 1;
 
 	/* Ensure we're in JTAG mode */
+#if 1
+	jtagtap_cycle(true, false, 51U);
```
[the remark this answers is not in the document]
ALTracer: Updated as such. Note that it would be `ADIV5_SWD_TO_JTAG_SELECT_SEQUENCE` if the unit included adiv5.h -- this magic number also occurs in hosted/ftdi_jtag. That one will use `_next()`.
I assume the `jtagtap_cycle()` method is a somewhat recent addition -- it's not used widely across the codebase, if only in BMPremote. I can even eliminate it as dead code for ADIv5 and RISC-V DTM targets, breaking the v0/v2 remote contract; unsure about AVR PDI. It's also too fast in comparison to other methods.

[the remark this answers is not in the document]
dragonmux: As we've got a `#define` for it in that header, include the header - that's fine. We should look to reduce the number of places this constant is defined down to the one, documented, and global place ideally.

Correct, `jtagtap_cycle()` was introduced in v1.8 and the BMP remote protocol implementation may need tweaking to provide a jtagtap_next()-based fallback, but we're principally fine with that as a pivot as this is exactly the kind of thing it was introduced for from what we can tell from Uwe's PR and commit for it.

src/platforms/common/swdptap.c
ALTracer: Thread resolved by merging (and rebasing over) PR1700.

src/platforms/common/jtagtap.c
```diff
@@ -61,31 +67,47 @@ static void jtagtap_reset(void)
...
+/* Busy-looping delay snippet for GPIO bitbanging (rely on inlining) */
+static inline void platform_delay_busy(const uint32_t loops)
+{
+#if 1
+	register uint32_t i = loops;
+	do {
+		__asm__("nop");
+	} while (--i > 0U);
+#else
+	for (register uint32_t counter = loops; counter > 0; --counter)
```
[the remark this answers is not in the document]
ALTracer: Updated delay loops to `--counter > 0` form.

src/platforms/common/jtagtap.c
```diff
@@ -61,31 +67,47 @@ static void jtagtap_reset(void)
...
+	platform_delay_busy(target_clk_divider);
 	return result != 0;
 }
 
+static bool jtagtap_next_no_delay() __attribute__((optimize(3)));
+
 static bool jtagtap_next_no_delay()
 {
 	gpio_set(TCK_PORT, TCK_PIN);
+	__asm__("nop" ::: "memory");
```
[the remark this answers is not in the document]
ALTracer: Annotated for the readers. This routine is too fast in comparison to others, because it has nothing to do besides toggle the jTCK pin, and it certainly does not move data into (or out of) BMP. I may even want to add more nops or something similar to tone down this one.

src/platforms/common/swdptap.c
ALTracer: Updated the delay loop to `--counter > 0U;)` form.

src/platforms/common/swdptap.c
ALTracer: Added a comment. Both `swdptap_seq_in_no_delay()` and `swdptap_seq_out_no_delay()` need careful consideration in code influencing bitbanged timings, including effect of `gpio_set()/clear()` or `gpio_set_val()` and `gpio_get()`.
I would like to see this entire routine fit into ART cache of STM32F4, or at least not jump around thanks to all the `gpio_*` bits inlined.

ALTracer:
> With #1688 and #1690 applied, on an stm32f411 platform, rtt has a measured speed of 114.292 chars/s if target is stm32f102, 121.297 chars/s if target is stm32f411. The documentation UsingRTT.md mentions a speed of 50.000 chars/s, perhaps update this number if PR is merged.

Thanks; the documented speed will certainly need updating, more so because of #1536.
I repeated the tests myself on `blackpill-f411ce` platform (96 MHz) and a gd32f103cb DUT clocked to 72 MHz; then 96, and 64, and even 48 MHz by providing a modified `SystemClockConfig()`. At default poll interval settings of `256 8 10` it is ~97000 chars/s, at `50 8 10` it is as you say, ~120000 chars/s, but altering them to `monitor rtt poll 25 5 10` yields ~167000-17000 chars/s. This means the bottleneck is not in DUT downclocked to 48MHz, but rather in the BMD/libopencm3 code. `orbtop` could provide more detailed data.
BMDA, for the record, manages only ~27000 chars/s using BMPremote with this platform (and non-1-byte read libc calls).

src/platforms/common/jtagtap.c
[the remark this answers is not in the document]
dragonmux: We went checking and the remote protocol has this fallback - apparently we thought about this when rewriting BMDA's implementation. We should verify fallbacks for the other adaptors are present, but as most of them outright replace this init code we don't see that being a problem as such - though it is a target for potential future dedupe in the backends.

tlyu:
Thanks for working on this! I tried it out a bit, and it seems to sometimes set `target_clk_divider` to zero, causing underflow and very long delays. This would be on platforms that set `BITBANG_CALIBRATED_FREQS`. On the other platforms, there are added checks that prevent that.

At least the clock speeds seem more consistent across different routines.

There are lots of remaining oddities about the timing that I see on the logic analyzer, but I'm still investigating those.

tlyu:
The main improvement is making the parity read/write functions have a consistent pulse width compared to the non-parity ones, at least in the case that does delays.

For the non-delay case, there seems to be increased asymmetry in the clock waveform, possibly due to replacing the NOP instructions with barriers. Some of these barriers are redundant, because the GPIO accesses are already volatile.

Also, I only have a 24MHz logic analyzer, so my ability to analyze differences in the no-delay case are going to be confounded by inadequate sample rate.

PR #1694, opened 2023-11-30
Fix xtensa formatting

xobs:
## Detailed description

Debug print statements were added that cause build breakage on ESP32 Xtensa targets.

### DIFF
--- a/src/target/cortexar.c
+++ b/src/target/cortexar.c
@@ -964,16 +964,15 @@ static void cortexr_mem_handle_fault(
 	const cortexar_priv_s *const priv = (cortexar_priv_s *)target->priv;
 	/* If we suffered a fault of some kind, grab the reason and restore DFSR/DFAR */
 	if (priv->core_status & CORTEXAR_STATUS_DATA_FAULT) {
-#if ENABLE_DEBUG == 1
+#ifndef DEBUG_WARN_IS_NOOP
 		const uint32_t fault_status = cortexar_coproc_read(target, CORTEXAR_DFSR);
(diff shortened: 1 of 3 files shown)


src/target/cortexar.c
```diff
@@ -967,13 +967,12 @@ static void cortexr_mem_handle_fault(
 #if ENABLE_DEBUG == 1
 		const uint32_t fault_status = cortexar_coproc_read(target, CORTEXAR_DFSR);
 		const uint32_t fault_addr = cortexar_coproc_read(target, CORTEXAR_DFAR);
+		DEBUG_WARN("%s: Failed at 0x%08" PRIx32 " (%08" PRIx32 ")\n", func, fault_addr, fault_status);
```
dragonmux, line 970: Rather than moving this, it should be gated by `#ifndef DEBUG_WARN_IS_NOOP` in place of `#if ENABLE_DEBUG == 1` - this will stop the variable definitions doing the wrong thing.

dragonmux (changes requested):
There's one small tweak we spotted in review but otherwise this looks great. With that one item addressed, we're happy to merge this.

src/target/cortexar.c
dragonmux: We guess it's fine to have the DEBUG_WARN() inside the guard block, just.. not ideal as we should really complete the IO operations before reporting the failure. That said, this nominally works so we'll go with it and merge.

dragonmux (approved):
LGTM 👍🏼, merging. Thank you for the contribution!

PR #1696, opened 2023-12-09
Fix building `hosted` under FreeBSD

ALTracer:
## Detailed description

* Not a feature.
* The problem is BMDA fails to build from source (FTBFS) on FreeBSD because of no `<alloca.h>` and missing hidapi-hidraw.
* PR solves this by purging all (3) of alloca.h references in light of #1661 and guarding the remaining inclusion by OS-specific macro. And adding a proper/available hidapi library name.

Tested on NomadBSD 13.2-RELEASE and a bunch of cheap adapters I have. `HOSTED_BMP_ONLY=0`. CMSIS-DAP (Picoprobe, free-dap) are detected, opened in v2/bulk mode and result in successful scan of an empty chain (nothing wired up). BMP-compatibles work only after indicating a `--device /dev/ttyU0` and under root privileges (there is no `/dev/serial/by-id`); including scanning SWD and attaching to a Cortex-M target. JLink and STLink/v2 result in a crash in `libusb.so.3.0.0` due to write into zero page on first bulk write transaction. `ASAN=1` also works btw. FTDI MPSSE not tested. Noting that `devel/openocd-0.11.0` works on the same machine with the same JLink/STLink without crashing, and `devel/gdb` looks like built with multiarch support.

Building: `sudo pkg install libftdi1 hidapi` and then `gmake -C src/ PROBE_HOST=hosted HOSTED_BMP_ONLY=0 CC=gcc12 -j6` because Makefiles of this project are incompatible with default `bmake` to which `make` aliases.
This PR does not cover building BMF and libopencm3 under *BSD. I could not install https://www.freshports.org/devel/gcc-arm-embedded/ due to symlink collisions, but python 3.9 is present (for GENHDR) and there may not be other problems to that.

## Closing issues
@DanielO may try using BMDA on his FreeBSD machine now, in addition to the Orbuculum suite (see https://github.com/orbcode/orbuculum/pull/134)

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -46,12 +46,6 @@
 #include "traceswo.h"
 #endif
 
-#if defined(_WIN32)
-#include <malloc.h>
-#else
-#include <alloca.h>
-#endif
-
 static bool cmd_version(target_s *t, int argc, const char **argv);
 static bool cmd_help(target_s *t, int argc, const char **argv);
 
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -55,11 +55,6 @@
 #define GDB_QSUPPORTED_NOACKMODE
 #endif
 
-#if defined(_WIN32)
-#include <malloc.h>
-#else
-#include <alloca.h>
-#endif
 #include <stdlib.h>
 
 typedef enum gdb_signal {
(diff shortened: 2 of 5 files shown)


dragonmux (approved):
Seems ok to us - we'll get this merged after testing the result locally. @xobs could you please check if possible that this still builds under your MSVC setup?

xobs:
Absolutely no problems under my weird MSVC testcase.

PR #1699, opened 2023-12-16
Fix: Remote protocol UBSAN error

dragonmux:
## Detailed description

This PR comes after a UBSAN issue was found by ALTracer on Discord which involves a missing widening cast needed to legalise the packet data in shifting in the BMP remote protocol v0 JTAG implementation.

This is because C defines left shifting by more than a type's width to be undefined behaviour so we have to pre-widen the data to resolve this.

### DIFF
--- a/src/platforms/hosted/remote/protocol_v0_jtag.c
+++ b/src/platforms/hosted/remote/protocol_v0_jtag.c
@@ -85,7 +85,7 @@ void remote_v0_jtag_tdi_tdo_seq(uint8_t *data_out, bool final_tms, const uint8_t
 		const size_t bytes = (chunk_length + 7U) >> 3U;
 		if (data_in) {
 			for (size_t idx = 0; idx < bytes; ++idx)
-				packet_data_in |= data_in[offset + idx] << (idx * 8U);
+				packet_data_in |= (uint32_t)data_in[offset + idx] << (idx * 8U);
 		}
 		/*
 		 * Build the remote protocol message to send, and send it.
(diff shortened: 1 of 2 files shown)


ALTracer:
By adding this to toolchain flags
```diff
diff --git a/src/platforms/hosted/Makefile.inc b/src/platforms/hosted/Makefile.inc
index 9a3c4b6..d4816e4 100644
--- a/src/platforms/hosted/Makefile.inc
+++ b/src/platforms/hosted/Makefile.inc
@@ -40,6 +40,12 @@ ifeq ($(ASAN), 1)
     OPT_FLAGS += -g
 endif
 
+ifeq ($(UBSAN),1)
+    CFLAGS += -fsanitize=undefined
+    LDFLAGS += -fsanitize=undefined
+    OPT_FLAGS += -g
+endif
+
 HIDAPILIB = hidapi
 ifneq (,$(findstring linux,$(SYS)))
     SRC += serial_unix.c
```
and rebuilding both on FreeBSD and on Linux, I indeed got two runtime errors from BMDA when operating a BMP-compatible.
`target/adiv5.c:376:24: runtime error: left shift of 177 by 24 places cannot be represented in type 'int'`
`protocol_v0_jtag.c:88:45: runtime error: left shift of 255 by 24 places cannot be represented in type 'int'`
I'll leave this here without suggesting adding a UBSAN makeflag like ASAN, because apparently BMDA 2.0 on Meson will handle it differently. And doing UBSan in firmware is much more complicated.

dragonmux:
Can you confirm that this PR correctly fixes that, or does it still produce those runtime errors?

ALTracer:
This PR removes the errors, that is, they appear with v1.10.0-260-ga42f91a8 (main~2) but disappear with v1.10.0-262-g84b064c3 (fix/remote-protocol-ubsan-error). Runtime-tested on Mint21.2 with gcc-12, too little reason to go test on FreeBSD again, should be the same. I had to modify the working copy to alter CFLAGS/LDFLAGS as stated previously.

Sorry for ambiguous wording. Consider the previous comment as a copy of the original bugreport (and a rant of no UBSan in BMDA v1.10). Thanks for mitigating this. If you have specific tests in mind before merging, post instructions here.

esden (approved):
LGTM

PR #1700, opened 2023-12-17
Fix: extract method for SWD parity calculation

ALTracer:
## Detailed description

This PR is aimed at two problems:
* Problem 1: swdptap of firmware references both popcount and parity builtins to do the same task (of calculating SWD parity), and `__popcountsi2(): 40 bytes` is more expensive. Solution 1: create a shim/dispatcher in maths_utils, name it `swd_odd_parity` and call *that* from swdptap (the gpio-bitbanging one). This leads to cleaner code at that callsite, and less flash space occupied
* Problem 2: *_swd.c `hosted` adapter driver backends also want to calculate parity, and also do that via both popcount and parity builtins. MSVC does not provide these builtins, only the `__popcount` for x86-64 optional instruction. Solution 2: incorporate this _MSC-specific builtin as well into the shim, and tweak all said `hosted` parity calculations to call the shim.

This PR is triggered from the discussion of #1688 two weeks ago.
I decided to name the function `swd_odd_parity()` for where it is going to be used. If future JTAG-PDI code also would want to use it (and attribute alias does not cut it), the reviewer/implementer of that can suggest a better name and I'll mass-rename the touched callsites (with a force-push).
Notice that I drop `maths_utils.h` from `general.h` and include it where needed by this PR. The point on making the shim `static inline` in the header is a valid discussion topic.

## Closing issues
May help fix building BMDA under MSVC with `HOSTED_BMP_ONLY=0` (caveat libusb/hidapi).

### DIFF
--- a/src/include/maths_utils.h
+++ b/src/include/maths_utils.h
@@ -37,5 +37,6 @@
 #include <stdint.h>
 
 uint8_t ulog2(uint32_t value);
+uint8_t calculate_odd_parity(uint32_t value);
 
 #endif /* INCLUDE_MATHS_UTILS_H */
--- a/src/maths_utils.c
+++ b/src/maths_utils.c
@@ -74,3 +74,21 @@ uint8_t ulog2(uint32_t value)
 	return (sizeof(uint8_t) * 8U) - result;
 #endif
 }
+
+uint8_t calculate_odd_parity(const uint32_t value)
+{
+#if defined(__GNUC__)
+	/* Ask for the libgcc impl */
+	return __builtin_parity(value);
+#elif defined(_MSC_VER)
+	/* Ask for a CPU insn */
+	return __popcount(value) & 1U;
+#else
+	/* Generic impl */
+	uint8_t result = 0;
+	for (uint32_t bit = 0; bit < 32; ++bit)
+		result ^= (value >> bit) & 1U;
+
+	return result;
+#endif
+}
--- a/src/platforms/common/swdptap.c
+++ b/src/platforms/common/swdptap.c
@@ -130,8 +130,8 @@ static bool swdptap_seq_in_parity(uint32_t *ret, size_t clock_cycles)
 	for (volatile uint32_t counter = target_clk_divider + 1; counter > 0; --counter)
 		continue;
 
-	size_t parity = __builtin_popcount(result);
-	parity += gpio_get(SWDIO_IN_PORT, SWDIO_IN_PIN) ? 1U : 0U;
+	const bool parity = __builtin_parity(result);
+	const bool bit = gpio_get(SWDIO_IN_PORT, SWDIO_IN_PIN);
 
 	gpio_set(SWCLK_PORT, SWCLK_PIN);
 	for (volatile uint32_t counter = target_clk_divider + 1; counter > 0; --counter)
@@ -140,7 +140,7 @@ static bool swdptap_seq_in_parity(uint32_t *ret, size_t clock_cycles)
 	*ret = result;
 	/* Terminate the read cycle now */
 	swdptap_turnaround(SWDIO_STATUS_DRIVE);
-	return parity & 1U;
+	return parity != bit;
 }
 
 static void swdptap_seq_out_clk_delay(uint32_t tms_states, size_t clock_cycles) __attribute__((optimize(3)));
@@ -182,9 +182,9 @@ static void swdptap_seq_out(const uint32_t tms_states, const size_t clock_cycles
 
 static void swdptap_seq_out_parity(const uint32_t tms_states, const size_t clock_cycles)
 {
-	int parity = __builtin_popcount(tms_states);
+	const bool parity = __builtin_parity(tms_states);
 	swdptap_seq_out(tms_states, clock_cycles);
-	gpio_set_val(SWDIO_PORT, SWDIO_PIN, parity & 1U);
+	gpio_set_val(SWDIO_PORT, SWDIO_PIN, parity);
 	for (volatile uint32_t counter = target_clk_divider + 1; counter > 0; --counter)
 		continue;
 	gpio_set(SWCLK_PORT, SWCLK_PIN);
--- a/src/platforms/common/swdptap.c
+++ b/src/platforms/common/swdptap.c
@@ -24,6 +24,7 @@
 #include "platform.h"
 #include "timing.h"
 #include "swd.h"
+#include "maths_utils.h"
 
 #if !defined(SWDIO_IN_PORT)
 #define SWDIO_IN_PORT SWDIO_PORT
@@ -130,7 +131,7 @@ static bool swdptap_seq_in_parity(uint32_t *ret, size_t clock_cycles)
 	for (volatile uint32_t counter = target_clk_divider + 1; counter > 0; --counter)
 		continue;
 
-	const bool parity = __builtin_parity(result);
+	const bool parity = calculate_odd_parity(result);
 	const bool bit = gpio_get(SWDIO_IN_PORT, SWDIO_IN_PIN);
 
 	gpio_set(SWCLK_PORT, SWCLK_PIN);
@@ -182,7 +183,7 @@ static void swdptap_seq_out(const uint32_t tms_states, const size_t clock_cycles
 
 static void swdptap_seq_out_parity(const uint32_t tms_states, const size_t clock_cycles)
 {
-	const bool parity = __builtin_parity(tms_states);
+	const bool parity = calculate_odd_parity(tms_states);
 	swdptap_seq_out(tms_states, clock_cycles);
 	gpio_set_val(SWDIO_PORT, SWDIO_PIN, parity);
 	for (volatile uint32_t counter = target_clk_divider + 1; counter > 0; --counter)
--- a/src/platforms/hosted/ftdi_swd.c
+++ b/src/platforms/hosted/ftdi_swd.c
@@ -32,6 +32,7 @@
 #include <ftdi.h>
 #include "ftdi_bmp.h"
 #include "buffer_utils.h"
+#include "maths_utils.h"
 
 typedef enum swdio_status {
 	SWDIO_STATUS_DRIVE,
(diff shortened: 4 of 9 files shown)


src/platforms/hosted/ftdi_swd.c
```diff
@@ -237,7 +238,8 @@ static bool ftdi_swd_seq_in_parity_mpsse(uint32_t *const result, const size_t cl
 	uint8_t data_out[5U] = {0};
 	ftdi_jtag_tdi_tdo_seq(data_out, false, NULL, clock_cycles + 1U);
 	const uint32_t data = read_le4(data_out, 0);
-	uint8_t parity = __builtin_parity(data & ((UINT64_C(1) << clock_cycles) - 1U));
+	const uint64_t bitmask = (UINT64_C(1) << clock_cycles) - 1U;
```
dragonmux, line 241: Given the check in `ftdi_swd_seq_in_parity` makes it impossible for clock_cycles to exceed 32, this can be made a uint32_t and the `UINT64_C(1)` switched to `1U`. If this causes a build error, cast `clock_cycles` to inhibit that.

src/platforms/common/swdptap.c
```diff
@@ -140,7 +141,7 @@ static bool swdptap_seq_in_parity(uint32_t *ret, size_t clock_cycles)
 	*ret = result;
 	/* Terminate the read cycle now */
 	swdptap_turnaround(SWDIO_STATUS_DRIVE);
-	return parity & 1U;
+	return parity != bit;
```
dragonmux, line 144: Rather than using `!=`, what does the codegen look like for `^`? If memory serves, the original `+=` pattern (which, granted, should have been `^=`) produces slightly fewer instructions than the comparison, so we'd be interested to see what different operations for calculating if the parity matches or doesn't does to this

src/maths_utils.c
```diff
@@ -74,3 +74,21 @@ uint8_t ulog2(uint32_t value)
...
+	/* Ask for the libgcc impl */
+	return __builtin_parity(value);
+#elif defined(_MSC_VER)
+	/* Ask for a CPU insn */
+	return __popcount(value) & 1U;
+#else
+	/* Generic impl */
+	uint32_t result = 0;
+	for (uint32_t bit = 0; bit < 32; ++bit)
+		result ^= value & (1U << bit);
```
dragonmux, line 90: This needs to be `result ^= (value >> bit) & 1U;`, otherwise it expresses a very slow bitwise copy of `value` to `result` as `^=` is bitwise, not logical.

src/include/maths_utils.h
```diff
@@ -37,5 +37,6 @@
 #include <stdint.h>
 
 uint8_t ulog2(uint32_t value);
+uint32_t swd_odd_parity(const uint32_t value);
```
dragonmux, line 40: The `const` here can be dropped (clang-tidy should be linting about this) from here (but not the definition in maths_utils.c).

Perhaps lets use `calc_odd_parity` or `calculate_odd_parity` as a more generic term for what this is doing as it's non-specific to SWD and that'll keep things open in the nomenclature for the JTAG-PDI support which will definitely benefit from this?

src/maths_utils.c
```diff
@@ -74,3 +74,21 @@ uint8_t ulog2(uint32_t value)
...
+{
+#if defined(__GNUC__)
+	/* Ask for the libgcc impl */
+	return __builtin_parity(value);
+#elif defined(_MSC_VER)
+	/* Ask for a CPU insn */
+	return __popcount(value) & 1U;
+#else
+	/* Generic impl */
+	uint32_t result = 0;
```
dragonmux, line 88: With the change below, this can become a `uint8_t` or a `bool` - your choice - and the result type changed to match.

dragonmux (changes requested):
It's definitely looking good - nice job cleaning things up and rationalising them with this change!

We've spotted a couple of functional issues in review, noted below, and have a couple of name suggestions to pick from for what to name this new function, but we're happy once those are addressed to get this merged.

It's definitely a shame we don't have a nice way of building a sim suite or test suite for the maths functions so we can have in-repo proofs of their correctness, but that's an improvement to strive for in a future PR.

src/maths_utils.c
ALTracer: Most hosted code uses `uint8_t` for packets represented as arrays. I'm more inclined to swap the `bool` in common/swdptap.c to uint8_t, or even `uint32_t`, which is faster (no bitmasking in codegen).
The original libgcc2 `__builtin_parity()` and `__builtin_popcount()` take `unsigned int` and return `int`, which is why I picked `uint32_t` to be transparent enough in the shim case. MSVC docs say `unsigned int __popcnt(unsigned int value);`, supporting this. In the "generic impl" case (on exotic toolchains?) I think we still should aim to return the same "int". Of course, if BMD only requires a 1 or a 0 from it, then it can be demoted to `uint8_t` everywhere, unless the hosted code already does this.

src/maths_utils.c
ALTracer: Hate bitwise loops. Fixed.

src/maths_utils.c
dragonmux: The function can only produce the values '0' and '1', so hence suggesting using `uint8_t` or `bool` - given that point about uint8_t for packets and the fact this get used to initialise values in those packets, it supports going to `uint8_t` and doing all the type truncation here - this will eliminate a pile of `-Wconversion`/`-Warith-conversion` warnings and we don't think will negatively impact codegen.

src/include/maths_utils.h
ALTracer: `calculate_odd_parity` is good, but too verbose -- all math functions "calculate". To help IntelliSense and likes, I suggest `parity_odd` to put the context first. I wonder if `parity_even` exists elsewhere.
The `const`, to my understanding, should be visible from headers so that the compiler can optimize other translation units to not duplicate arguments into r0 (or whatever the AAPCS first argument register is). Or are we only using it for guarantees on pointers?

dragonmux: `parity_odd` reads.. strangely.. to us, due to us referencing here calculating the parity value for an odd parity scheme - it's technically correct, but it's got the same odd lilt as Picard saying "Tea, Earl Grey, Hot, Black" rather than "a hot cup of black Earl Grey tea".

The `const` on a value in a C function declaration (prototype) does nothing, and is purely noise - this is what the clang-tidy lint is about. It only has meaning in the function definition.

In regards pointers, `const char *` is saying the thing the pointer points to is const, but `const char *const` doesn't do anything extra in the context of a decl. That's why we omit on the parameter side of the * in the decl.

src/platforms/common/swdptap.c
ALTracer: Comparing two bools here can be done with `!=` or `^` -- both compile to `eor.w r0, r1, r2;` right before function epilogue `add sp, #16; pop {r4, r5, r6, pc}` or `ldmia.w sp!, {clobbers, pc}`.
This block is cherry-picked from PR1688 for types and const.

dragonmux: Oh, perfect, in that case.. let's stick with the `!=` check as it's more obvious what it does - a better expression of intent. Glad the compiler's seeing sense and doing the good optimisation there.

src/include/maths_utils.h
ALTracer: Renamed `swd_odd_parity` to `calculate_odd_parity` globally.
Removed `const` from uint32_t in this header.

src/maths_utils.c
ALTracer: Adjusted to `uint8_t`.

ALTracer:
Requesting next round of review.
Fixup squashing order: keep first 5 commits, incorporate last 4 commits into them: 7-8 into 1st, 6/9 into 4-5. On the other hand, I can replay this manually to make it clean.

src/platforms/hosted/jlink_swd.c
```diff
@@ -213,7 +214,7 @@ static uint32_t jlink_adiv5_raw_read_no_check(const uint16_t addr)
 	/* Extract the data phase and return it if the transaction suceeded */
 	const uint32_t data = read_le4(response, 0);
 	DEBUG_PROBE("jlink_adiv5_raw_read_no_check %04x -> %08" PRIx32 " %s\n", addr, data,
-		__builtin_parity(data) ^ response[4] ? "ERR" : "OK");
+		calculate_odd_parity(data) ^ response[4] ? "ERR" : "OK");
```
dragonmux, line 217: This comparison should probably be converted to `!=` from `^` for clarity now

dragonmux (commented):
This all looks good to us now - the one review note here is an optional change. Please replay the history to make it clean and we'll get this merged.

ALTracer:
Merged fixups, propagated the rename, build-tested intermediate commits with a custom pre-commit extended hook, and rebased to latest main.
Think that's ready for final live testing for SWD parity regressions and any remaining review that the individual commits adhere to expectations. Please try it on a Tigard or something else FTDI.
There's one last `__builtin_ctz` remaining in ftdi.c:`ftdi_swd_init()`, and I have some ideas on how to go around it, but maybe not in this PR -- and I'd like MSVC build logs w.r.t. that.

dragonmux:
Most excellent! Regarding MSVC, once the new Meson support lands it should become considerably easier to generate full BMDA builds under MSVC and get logs from them.

dragonmux (dismissed):
LGTM, approved pending testing with our Tigard. We'll merge (assuming it all works properly, but we believe it should) once that's been tested.

src/platforms/hosted/ftdi_swd.c
dragonmux: Turns out this was wrong, we'd put that `UINT64_C()` in because the bit mask uses a 64-bit intermediary - the switch to `uint32_t` is fine, but `1U << 32U` is generated when `clock_cycles` is 32 which has a 64-bit result and invokes undefined behaviour if attempted in 32-bit space. Found while testing on our Tigard.

dragonmux (approved):

dragonmux:
We'll get this merged now it's been tested and the one bug found and fixed, good work! 🙂

PR #1704, opened 2023-12-18
rtt: Add gdb monitor rtt down command

HrMitrev:
## Detailed description

Just for convenience I wanted to be able to
push strings to the rtt down buffer without
the need to change the active window from
the one where the gdb session is running

## Closing issues
None


src/command.c
```diff
@@ -564,7 +564,19 @@ static bool cmd_rtt(target_s *t, int argc, const char **argv)
 			if (rtt_ident[i] == '_')
 				rtt_ident[i] = ' ';
 		}
-	} else if (argc == 2 && strncmp(argv[1], "ram", command_len) == 0)
+	} else if (argc == 3 && strncmp(argv[1], "down", command_len) == 0) {
```
dragonmux, line 567: Perhaps instead of "down" lets use "send" - "down" doesn't make much grammatical or linguistical sense to us as a native English speaker and leaves us wondering what this really does. `mon rtt send <stuff>` makes much more sense - you're sending <stuff> to the target.

src/command.c
```diff
@@ -564,7 +564,19 @@ static bool cmd_rtt(target_s *t, int argc, const char **argv)
...
+		/*
+		 * here we bypass the if statement to save flash
+		 * as it's not critical to check if the rtt block is found
+		 */
+		if(rtt_found || true) {
```
dragonmux, line 572: Either include the check, or entirely eliminate it please. Likewise, proper capitalisation - "Here" in the above comment, "The" in the gdb_out() below. Please drop the braces on the else, they're unnecessary.

src/command.c
```diff
@@ -564,7 +564,19 @@ static bool cmd_rtt(target_s *t, int argc, const char **argv)
...
+			uint16_t len = strlen(argv[2]);
```
dragonmux, line 573: `strlen()` returns a size_t, please do not truncate it to a uint16_t.

src/include/rtt.h
```diff
@@ -60,5 +60,6 @@ typedef struct rtt_channel {
 extern rtt_channel_s rtt_channel[MAX_RTT_CHAN];
 
 void poll_rtt(target_s *cur_target);
+void rtt_load_recv_buf(const char *data_buf, uint16_t len);
```
dragonmux, line 63: Likewise to the `strlen()` comment, `size_t` please.

src/platforms/common/stm32/rtt_if.c
```diff
@@ -58,6 +58,17 @@ inline static bool recv_set_nak()
 	return recv_bytes_free() < 2U * CDCACM_PACKET_SIZE;
 }
 
+void rtt_load_recv_buf(const char *data_buf, uint16_t len)
+{
+	/* copy data to recv_buf */
+	for (int i = 0; i < len; i++) {
```
dragonmux, line 64: This isn't actually your fault, but the use of raw `int` introduces a signed-unsigned conversion and a silent error here. Please switch this to use `size_t` or at least `uint16_t` to match the `len` argument.

Similarly with the proper capitalisation of comments. We know that one's also not your fault here, this code has just not yet been cleaned up and the original author was distinctly lazy about it.

HrMitrev:
Hi @koendv 🙂

Since you're the original author of the RTT functionality, could you take a quick look?

The main idea is to start the program with `run &` when using gdb and then I want to have .gdbinit function which sends SIGINT (`interrupt`) then `monitor rtt down 'string'` then `continue &`

Ideally it would be best if I there is a way to interact with the BMP probe from gdb without interrupting the debug session, but I'm not sure how and if it is even possible. Sure I can directly use `ttyBmpTarg` to send data to the target, but just for convenience I don't want to change the active window to the terminal where I view the log, send the string and then go back to the terminal where the gdb session is running

dragonmux (changes requested):
There are a couple of linguistic items we caught in review and some type issues. Please see below for more on those. Please also take care of the clang-format lints the CI picked up.

We're distinctly unsure whether to take this proposed change or not, being how easy it is to use a second console with a serial program to interact with the RTT stream. That said, we can also sort of see the sense of this command too. We'll see how bad the Flash usage hit is after the adjustments requested and have a chat with Esden.

src/command.c
HrMitrev: If I include the check, the error message adds to the .text segment, so I just print something generic like "error"? I think it is good to have this in place just in case and also it is OK to bypass, bacause it is not at all critical. So if someone is browsing the code can re-enable it.

dragonmux: Given what the message is trying to express, perhaps `gdb_print("rtt not found\n");`? It expresses the same information without using quite so much Flash - we'd really prefer this and the check to be enabled, or just entirely eliminate the check. The only reason the `|| true` is not causing additional Flash usage is because of the Dead-Code Elimination pass and we do not want to be relying on that, it's implementation-defined behaviour.

koendv:
On Mon, 18 Dec 2023 09:06:36 -0800
Hristo Mitrev ***@***.***> wrote:

> Hi @koendv 🙂
> 
> Since you're the original author of the RTT functionality, could you
> take a quick look?
> 
Hi Hristo.

This also works:
^C
Program received signal SIGINT, Interrupt.
0x08000528 in ?? ()
(gdb) python rtt = open("/dev/ttyBmpTarg", "w")
(gdb) python rtt.write("hello, world")
(gdb) python rtt.flush()
(gdb) c
Continuing.

I'm wondering:
If you send data, maybe you'll want to wait for some kind of reply.
Like the linux "expect" command, or python pexpect.

koen

HrMitrev:
> Hi Hristo. This also works: ^C Program received signal SIGINT, Interrupt. 0x08000528 in ?? () (gdb) python rtt = open("/dev/ttyBmpTarg", "w") (gdb) python rtt.write("hello, world") (gdb) python rtt.flush() (gdb) c Continuing. I'm wondering: If you send data, maybe you'll want to wait for some kind of reply. Like the linux "expect" command, or python pexpect. koen

Hi @koendv,

This is a great suggestion, but I am afraid that my use-case is a bit different.
I always have /dev/ttyBmpTarg open in another console because I need to save the logs to a file and also look at them in the meantime. My main idea was to skip the need of switching from gdb to the log console in order to send commands. Since I have to have /dev/ttyBmpTarg open in another window I thought that this gdb mon command will be helpful.

PS: it took me a while to realize that there are 2 separate rtt_if.c files for stm32 and launchpad-icdi. I think I fixed the problem and it seems that all the tests now pass. I also rebased to the latest commit in main

src/command.c
koendv, line 1: Spelling error, line 100. "poll [MAXMS MINMS MAXERR]]"},

koendv:
I think an rtt "send" command might be useful as a first step towards making testbenches in gdb/python.

src/command.c
```diff
@@ -96,8 +96,8 @@ const command_s cmd_list[] = {
 #endif
 #ifdef ENABLE_RTT
 	{"rtt", cmd_rtt,
-		"[enable|disable|status|channel [0..15 ...]|ident [STR]|cblock|ram [RAM_START RAM_END]|poll [MAXMS MINMS "
-		"MAXERR]]"},
+		"[enable|disable|status|channel [0..15 ...]|ident [STR]|send [STR]|cblock|ram [RAM_START RAM_END]|"
+		"poll [MAXMS MMINMS AXERR]]"},
```
ALTracer:
```suggestion
		"poll [MAXMS MINMS MAXERR]]"},
```
Is this with the `pre-commit` hook installed and clang-format working?

src/command.c
HrMitrev: I think 0022d0bae7802a97eccfd077722007669712c5f9 fixes it

HrMitrev:
Hi Again!
If I remember correctly, the overhead of this feature was 72 more bytes of ROM.

Since I proposed this enhancement I also realized that it will be helpful when I implement a make-shift password that will be used to lock some debug features of the target. If I implement such password it will be tedious to enter it every time I flash my target. Having this feature will help me to automate this. (I thought of this password because I give debug versions of my firmware to testing colleagues who provide me with logs, but the problem is that I also implement "scanf actions" that can alter the internal state of the program and lead to false positive bug reports)

One more proposal from my side will be to add `ifdef` guards and make this optional if we think that it will add too much overhead.

Cheers!

PR #1708, opened 2024-01-02
Fix: qCRC performance uplift

ALTracer:
## Detailed description

* On the scale between "fix" and "feature" this is closer to "fix".
* The existing problem is suboptimal performance of `(gdb) compare-sections` because of extra ADIv5 MEM-AP setup overhead.
* This PR solves that by increasing the stack buffer inside `generic_crc32()` from 128 to 1024 bytes.

There is an intentionally separate commit which adds DEBUG_INFO logs displaying "KiB/s" metric to BMDA, and copies that (ms and KiB/s) to the implementation for STM32 CRC IP enabled platforms. I avoid this calculation for small sections/regions (abundant in some firmware projects) for integer precision and overhead reasons.

On my quest to increasing runtime performance such that BMP-compatibles are more usable (Quality of Life for developers) against targets with Big Flash and with RTT, following (and testing) the SWD improvements, I decided to use some sort of benchmark internal to BMD (to not have to wire up and look at an LA or MSO). Uploads to reprogrammable flash bottleneck in said flash erase/write times, BMDA communication bottlenecks in either USB FS link, libusb or host syscalls, and there's no dedicated iperf3/speedtest between BMDA and BMPremote (using [un]hexify). But the GDB RSP `qCRC` packet it uses to check for section mismatch is pretty much ideal because a) the packets sent over RSP are pretty small, for just addresses and checksum values; b) STM32 CRC handles most of the calculations; c) the only remaining part is ADIv5 SWD (or JTAG) acceleration driver and gpio-bitbanging driver.

Dragonmux highlighted that ADIv5 MEM-AP has a feature of Target Address Register autoincrement (for up to 10 bit, so 1024-byte long and aligned regions) configured in CSW and checked in driver methods. I tried increasing said buffer to 4096 bytes first, on `blackpill-f411ce` as the development kit with bigger SRAM and CPU speed, and also increasing BMPremote v3 packet lengths from 1024 to 4096 bytes as well. This did not display additional runtime benefit, rather, it may not fly on smaller adapters.

The change 128->1024 is tested by me on both `blackpill-f411ce` and `swlink` adapters with ENABLE_DEBUG=1, and these logs indicate approximately +15% `(gdb) compare-sections` speed boost both on `no_delay` and `1 delay` TAP frequency settings (compared to BMF built without the stack buffer change). 

```
SWD no_delay: len = 163252: 160 KiB/s,        996 ms
SWD  1 delay: len = 163252: 105 KiB/s,        1505 ms
SWD  2 delay: len = 163252: 64 KiB/s, 2489 ms
before (128). After (1024):
SWD no_delay: len = 163252: 183 KiB/s,        867 ms
SWD  1 delay: len = 163252: 115 KiB/s,        1380 ms
SWD  2 delay: len = 163252: 72 KiB/s, 2210 ms
```

I reason it may be useful, and somewhat not expensive, because there are a few more places in BMF with stack buffers of similar size (memory-map.xml generator and samx5x nvm page code come to mind), which are not run simultaneously with this. Pinning the buffer to `static` storage duration in .bss, even if it guarantees allocation at link-time, is worse, because it is not used during >90% of the BMF runtime/lifetime. Repurposing the gdb packet buffer is another idea, but it may be more difficult in execution (or same for target flash buffers). I cannot in good conscience make it a dynamic buffer while BMF relies on newlib's `-lnosys` provided optimistic `_sbrk` -- it does not allow catching failed mallocs or stack/heap overflows.

Value of 128 looks somewhat arbitrary to me -- I couldn't find arguments supporting it in git history commit messages, or GH issues and PRs. It was introduced a long time ago in PR #108 and then copied in https://github.com/blackmagic-debug/blackmagic/commit/72790893aea9887e78d5b57b95ae2e791dc57f62 in 2016. Value of 4096 (0x1000) was introduced in PR #922 in https://github.com/blackmagic-debug/blackmagic/commit/e58b7d623b4c8b4ec9317fa1bfdd27f6189f012c for BMDA along with other changes catering to H7 target usecase.

PR should not have significant impact on Flash size.

### DIFF
--- a/src/crc32.c
+++ b/src/crc32.c
@@ -136,11 +136,14 @@ bool generic_crc32(target_s *const target, uint32_t *const result, const uint32_
 	*result = crc;
 	return true;
 }
+
+__attribute__((alias("generic_crc32"))) bool bmd_crc32(
+	target_s *const target, uint32_t *const result, const uint32_t base, const size_t len);
 #else
 #include <libopencm3/stm32/crc.h>
 #include "buffer_utils.h"
 
-bool generic_crc32(target_s *const target, uint32_t *const result, const uint32_t base, const size_t len)
+bool stm32_crc32(target_s *const target, uint32_t *const result, const uint32_t base, const size_t len)
 {
 	uint8_t bytes[128U];
 
@@ -185,4 +188,7 @@ bool generic_crc32(target_s *const target, uint32_t *const result, const uint32_
 	*result = crc;
 	return true;
 }
+
+__attribute__((alias("stm32_crc32"))) bool bmd_crc32(
+	target_s *const target, uint32_t *const result, const uint32_t base, const size_t len);
 #endif
--- a/src/crc32.c
+++ b/src/crc32.c
@@ -113,7 +113,7 @@ bool generic_crc32(target_s *const target, uint32_t *const result, const uint32_
 	uint8_t bytes[128U];
 #endif
 
-#if ENABLE_DEBUG == 1
+#ifndef DEBUG_INFO_IS_NOOP
 	const uint32_t start_time = platform_time_ms();
 #endif
 	uint32_t last_time = platform_time_ms();
@@ -132,7 +132,9 @@ bool generic_crc32(target_s *const target, uint32_t *const result, const uint32_
 		for (size_t i = 0; i < read_len; i++)
 			crc = crc32_calc(crc, bytes[i]);
 	}
-	DEBUG_WARN("%" PRIu32 " ms\n", platform_time_ms() - start_time);
+#ifndef DEBUG_INFO_IS_NOOP
+	DEBUG_INFO("%s: %" PRIu32 " ms\n", __func__, platform_time_ms() - start_time);
+#endif
 	*result = crc;
 	return true;
 }
@@ -149,6 +151,9 @@ bool stm32_crc32(target_s *const target, uint32_t *const result, const uint32_t 
 
 	CRC_CR |= CRC_CR_RESET;
 
+#ifndef DEBUG_INFO_IS_NOOP
+	const uint32_t start_time = platform_time_ms();
+#endif
 	uint32_t last_time = platform_time_ms();
 	const size_t adjusted_len = len & ~3U;
 	for (size_t offset = 0; offset < adjusted_len; offset += sizeof(bytes)) {
@@ -185,6 +190,9 @@ bool stm32_crc32(target_s *const target, uint32_t *const result, const uint32_t 
 			}
 		}
 	}
+#ifndef DEBUG_INFO_IS_NOOP
+	DEBUG_INFO("%s: %" PRIu32 " ms\n", __func__, platform_time_ms() - start_time);
+#endif
 	*result = crc;
 	return true;
 }
--- a/src/crc32.c
+++ b/src/crc32.c
@@ -133,7 +133,15 @@ bool generic_crc32(target_s *const target, uint32_t *const result, const uint32_
 			crc = crc32_calc(crc, bytes[i]);
 	}
 #ifndef DEBUG_INFO_IS_NOOP
-	DEBUG_INFO("%s: %" PRIu32 " ms\n", __func__, platform_time_ms() - start_time);
+	/* "generic_crc32: 08000110+75272 -> 1353ms, 54 KiB/s" */
+	const uint32_t end_time = platform_time_ms();
+	const uint32_t time_elapsed = end_time - start_time;
+	DEBUG_INFO("%s: %08" PRIx32 "+%" PRIu32 " -> %" PRIu32 "ms", __func__, base, (uint32_t)len, time_elapsed);
+	if (len >= 512U) {
+		const uint32_t speed = len * 1000U / time_elapsed / 1024U;
+		DEBUG_INFO(", %" PRIu32 " KiB/s", speed);
+	}
+	DEBUG_INFO("\n");
 #endif
 	*result = crc;
 	return true;
@@ -191,7 +199,15 @@ bool stm32_crc32(target_s *const target, uint32_t *const result, const uint32_t 
 		}
 	}
 #ifndef DEBUG_INFO_IS_NOOP
-	DEBUG_INFO("%s: %" PRIu32 " ms\n", __func__, platform_time_ms() - start_time);
+	/* "stm32_crc32: 08000110+75272 -> 237ms, 310 KiB/s" */
+	const uint32_t end_time = platform_time_ms();
+	const uint32_t time_elapsed = end_time - start_time;
+	DEBUG_INFO("%s: %08" PRIx32 "+%" PRIu32 " -> %" PRIu32 "ms", __func__, base, (uint32_t)len, time_elapsed);
+	if (len >= 512U) {
+		const uint32_t speed = len * 1000U / time_elapsed / 1024U;
+		DEBUG_INFO(", %" PRIu32 " KiB/s", speed);
+	}
+	DEBUG_INFO("\n");
 #endif
 	*result = crc;
 	return true;
--- a/src/crc32.c
+++ b/src/crc32.c
@@ -155,7 +155,7 @@ __attribute__((alias("generic_crc32"))) bool bmd_crc32(
 
 bool stm32_crc32(target_s *const target, uint32_t *const result, const uint32_t base, const size_t len)
 {
-	uint8_t bytes[128U];
+	uint8_t bytes[1024U]; /* ADIv5 MEM-AP AutoInc range */
 
 	CRC_CR |= CRC_CR_RESET;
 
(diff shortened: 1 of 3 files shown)


src/crc32.c
```diff
@@ -142,10 +149,13 @@ bool generic_crc32(target_s *const target, uint32_t *const result, const uint32_
 
 bool generic_crc32(target_s *const target, uint32_t *const result, const uint32_t base, const size_t len)
 {
-	uint8_t bytes[128U];
+	uint8_t bytes[1024U]; /* ADIv5 MEM-AP AutoInc range */
 
 	CRC_CR |= CRC_CR_RESET;
 
+#if ENABLE_DEBUG == 1
```
dragonmux, line 156: Might be worth using `#ifndef DEBUG_WARN_IS_NOOP` here given this is specifically for DEBUG_WARN level diagnostics. Looking at the generic implementation of this, the same suggestion applies there.

src/crc32.c
```diff
@@ -132,7 +132,14 @@ bool generic_crc32(target_s *const target, uint32_t *const result, const uint32_
...
 			crc = crc32_calc(crc, bytes[i]);
 	}
-	DEBUG_WARN("%" PRIu32 " ms\n", platform_time_ms() - start_time);
+#if ENABLE_DEBUG == 1
+	const uint32_t end_time = platform_time_ms();
+	if (len > 512U) {
+		const uint32_t speed = len * 1000U / (end_time - start_time) / 1024U;
+		DEBUG_WARN("len = %" PRIu32 ": %" PRIu32 " KiB/s,\t", (uint32_t)len, speed);
+	}
+	DEBUG_WARN("%" PRIu32 " ms\n", end_time - start_time);
```
dragonmux, line 141: Let's prefix this and the one below with some context? Otherwise we still have the same problem as before where "20ms" printed on the console without any further information is just meaningless fluff due to the lack of context.

Perhaps:
```suggestion
	DEBUG_WARN("crc32: %08" PRIx32 "+%" PRIu32 " -> %" PRIu32 " ms", base, (uint32_t)len, end_time - start_time);
	if (len > 512) {
		const uint32_t speed = len * 1000U / (end_time - start_time) / 1024U;
		DEBUG_WARN(" %" PRIu32 " KiB/s", speed);
	}
	DEBUG_WARN("\n");
```

src/crc32.c
```diff
@@ -142,10 +149,13 @@ bool generic_crc32(target_s *const target, uint32_t *const result, const uint32_
 
 bool generic_crc32(target_s *const target, uint32_t *const result, const uint32_t base, const size_t len)
 {
-	uint8_t bytes[128U];
+	uint8_t bytes[1024U]; /* ADIv5 MEM-AP AutoInc range */
```
dragonmux, line 152: Ideally we'd re-use a buffer like the GDB packet buffer here, but we'll go with this for now as "probably safe".

dragonmux (changes requested):
This is looking very promising! We've some tweaks to request on the debug output to make it a little more understandable - to put context on it - but otherwise this all looks good to go.

src/crc32.c
ALTracer: Any loglevel works, as long as it gets compiled into BMP optionally. I'd even say it's DEBUG_INFO. Your refactor on noops will surely be idiomatic to use here, even if I don't get build errors for `all_platforms` for now. I was just extending stuff originally here, I could change DEBUG_WARN to something more appropriate now that it exists.

src/crc32.c
ALTracer: As said, that idea of yours is harder in execution, I'll leave it up to you to try to implement it and reduce stack used. Use my meminfo branch to check clobbered depth (it colors the stack at reset).
It's just I tested swlink and it does not crash when doing this against a simple MCU target, but it may behave differently with multicore and multi-region targets like some MPUs.
And memory-map.xml generator may have to keep its buffer separate from gdb packet buffer to be able to hexify and print it (discussion idea).

src/crc32.c
ALTracer: These random "1ms" lines on BMDA stdout log always looked funny to me. Why not `__func__` for generic_crc32: ? This does not get into release BMF builds (yet), and so 8 more bytes is not a space concern.
Do you approve of the 512 section length lower bound? Also I successfully avoided 64-bit integer division here, and I don't think it will overflow practically (biggest targets have 16 MiB of flash on e.g. RP2040), but precision is limited in a way.

src/crc32.c
dragonmux: Think it is technically either target-level debugging or informational - definitely not a warning. Please update to whichever of DEBUG_INFO or DEBUG_TARGET you feel is more appropriate.

src/crc32.c
dragonmux: Those random lines is exactly what we're aiming to avoid 👍🏼 
We could use `__func__`, but well, writing it this way means it's identical between implementations which we think is a good thing with the context.

We weren't able to fully validate the division, but it seems reasonable and the bound while arbitrary seems fine. As it's intended as a lower bound, then perhaps `>=` instead of just `>`?

ALTracer: The way BMD is organized means both these functions have to have identical names and signatures to link (inverse of ODR). I wonder if using attribute alias would allow renaming the second one to `stm32_hwcrc32()` or something alike.
Noted the `len >= 512U` suggestion. Users could look at gdb console output and be able to match hex section bounds with BMD ms-timing logs for smaller regions.

dragonmux: Attribute alias certainly would allow the function name to reflect what it is properly - that might be a fair idea to use here! Perhaps `bmd_crc32()` for the base function name, `generic_crc32()` for the non-STM32-specific one and `stm32_crc32()` for the specific one, all aliased to `bmd_crc32()`?

ALTracer: Yes, I tried it and that works, that is it gets into the binary and `strings -n8 src/blackmagic.bin` (similar for stripped hosted) only lists the actual name from supposedly string table resources, but never `bmd_crc32`. And there's only a single callsite in gdb_main, so this PR can still be kept small scope.

dragonmux: oh yeah, given those other errors were using the function name, that seems absolutely fine then too to use __func__ and save back a touch of space overall

src/crc32.c
ALTracer: Updated to `DEBUG_INFO` and relevant NOOP macros (not used elsewhere unfortunately). IMO `DEBUG_TARGET` is for target-specific drivers, and this is part of generic BMD gdbserver core.

ALTracer:
Updated log format code, DEBUG_* macro usage, retested BMDA and BMF+debug.
Backpropagated the suggested/refactoring changes to the rest of crc32.c translation unit as separate atomic commits.
Only the last commit is important (affects logic) for release firmware and can be cherry-picked mostly anywhere. The rest will stay on the development/main branch going forward.
Rebased to latest main (past PR1711) to see CI size-diff reports.

src/crc32.c
dragonmux: You are somewhat correct about DEBUG_TARGET - the actual meaning and reason for it's introduction is target-specific/level diagnostics - like figuring out why something is running so slow against the target.. but we're happy in this case to accept the argument that this is informational and so belongs in the DEBUG_INFO level diagnostics :)

src/crc32.c
```diff
@@ -125,27 +125,43 @@ bool generic_crc32(target_s *const target, uint32_t *const result, const uint32_
...
-	DEBUG_WARN("%" PRIu32 " ms\n", platform_time_ms() - start_time);
+#ifndef DEBUG_INFO_IS_NOOP
+	/* "generic_crc32: 08000110+75272 {tab} -> 1353 ms,   54 KiB/s" */
+	const uint32_t end_time = platform_time_ms();
+	const uint32_t time_elapsed = end_time - start_time;
+	DEBUG_INFO("%s: %08" PRIx32 "+%" PRIu32 "\t-> %4" PRIu32 " ms", __func__, base, (uint32_t)len, time_elapsed);
+	if (len >= 512U) {
+		const uint32_t speed = len * 1000U / time_elapsed / 1024U;
+		DEBUG_INFO(", %4" PRIu32 " KiB/s", speed);
+	}
```
dragonmux, line 143: We know why you've added all the whitespace here, but please can we not - this makes it extremely difficult for those of us with line tracking problems to read. It might "look nice" (subjective, no comment here) but that doesn't help us when our eyes keep trying to skip between lines.

```suggestion
	DEBUG_INFO("%s: %08" PRIx32 "+%" PRIu32 " -> %" PRIu32 "ms", __func__, base, (uint32_t)len, time_elapsed);
	if (len >= 512U) {
		const uint32_t speed = len * 1000U / time_elapsed / 1024U;
		DEBUG_INFO(", %" PRIu32 " KiB/s", speed);
	}
```

Would be our strong preference owing to ease of scanning along the line this way without finding we're accidentally reading the next and getting real confused.

dragonmux (changes requested):

src/crc32.c
ALTracer: Reverted/readjusted format-strings to comply. I don't feel strongly about this patch hunk, and I will rarely stare at its output myself once I'm done enough with all the perf dances. Also updated the oneline comment above depicting example output as seen in live testing.

dragonmux (approved):
LGTM, and this all seems straight-forward and reasonable. Merging - thank you for the contribution!

PR #1709, opened 2024-01-02
Feature: Handle GD32F3 XL-density as dual-bank

ALTracer:
## Detailed description

* On the scale between "fix" and "feature" this is closer to "feature" (like new target support).
* The existing problem is the stm32f1.c driver handling `GD32F30x_XD` (and _CL) chips somewhat wrong, relying only on "Memory density" information register.
* This PR solves this problem by adding target-specific knowledge to BMD about bank sizes.

This commit is originally from some time in November. Please note that it may be wrong in a few ways. The idea is there, but the implementation should trigger only for GD32F3 (PR scope) chips. I don't have all the targets to test on them, only gd32f303cc with 256 KiB of Flash (and maybe gd32f1, gd32e5).
Datasheet says Code area (0 wait state) is up to 256 KiB, Data area (lots of hclks to read) is up to 2816 KiB -- but I figure this is a different boundary, meaning probably shadow SRAM. Ref.man. says first 512 KiB are in pages of 2 KiB writable through "bank0", and the rest, if present, is in pages of 4 KiB writable through "bank1".

## Closing issues


dragonmux (dismissed):
We unfortunately haven't got any good way to test this, but the patch looks good and seems reasonable, there are no major style lints beyond some missing `U` on constants (but this file needs going through carefully for this anyway, so no worries there). LGTM, so merging when the builds complete unless you've got a pressing reason we should hold off.

ALTracer:
Please hold, for I didn't prove yet (by reading all the refmans) that this patch hunk has to run for all the targets in `gd32f1_probe()`. I'd like to see a nice table from someone who has worked extensively with them and researched this as such. Also I'm awaiting testing from other contributors soon. Doing this before the merge shouldn't be difficult (to checkout branch of fork).

ALTracer:
Reading through User Manuals for GD32F1, GD32F2, GD32F3 highlights that all of the chips at HD, CL, XD (High and eXtra density, and Connectivity Line) variants have first 512 KiB in 2 KiB pages, then whatever left in 512 to 3072 range in 4 KiB pages. It's only GD32F10x_MD medium density that are 128 KiB in 1 KiB pages, much like the popular F103CB in MD class. So I suppose it is correct to run it like that. I'd also like to bump block_size to 0x800 in these cases. When combined with an ALIGN_32BIT hacky change, this boosts write speed by ~+50% without losing correctness.
GD32E230 are up to 64 KiB in 1 KiB pages. GD32E50x are up to 512 KiB in 8 KiB pages, there's no second bank ever. If you'd like me to update the descriptions in comments, I'm happy to push that commit as well. Previous contributors didn't add much information detail to the codebase when posting code around this patch's context.

dragonmux:
Yeah, that sound like a good plan - please update the comments with your findings and lets sort the documentation aspect out while we're here.

If there are any other tweaks required such as fixing page size definitions, please also apply those as best as you can. We'll hold off merging till we've had your thumbs up that you think you've got it all. No worries if something turns out to be subtly wrong later on a part or another vendor's products - as long as the documentation is as accurate as you can presently make it and documented, that'll hopefully stop the guessing game problem you've noted here.

src/target/stm32f1.c
```diff
@@ -679,6 +682,9 @@ static bool stm32f1_flash_write(target_flash_s *flash, target_addr_t dest, const
 	const size_t offset = stm32f1_bank1_length(dest, len);
 	DEBUG_TARGET("%s: at %08" PRIx32 " for %zu bytes\n", __func__, dest, len);
 
+	/* Allow wider writes on Gigadevices (is there a better way to hook this?) */
```
dragonmux, line 685: This looks like a prime case to introduce a target-specific TOPT (`STM32F1_TOPT_GIGADEVICE`?) which the appropriate GD32* probe routines set

dragonmux (commented):
This all looks good save for the psize strncmp() bit which we've got a suggestion on how to improve - once that's been addressed, we're happy to merge this PR

src/target/stm32f1.c
```diff
@@ -173,10 +176,21 @@ bool gd32f1_probe(target_s *target)
 	const uint16_t flash_size = signature & 0xffffU;
 	const uint16_t ram_size = signature >> 16U;
 
+	/*
+	 * GD32F303x User Manual Rev2.9, 2.3.1 Flash memory architecture:
+	 * HD and <=512 KiB CL devices have only bank0 with 2 KiB sized pages
+	 * XD and  >512 KiB CL devices also have bank1 with 4 KiB sized pages
+	 */
+	if (flash_size > 512U) {
```
tlyu, line 184: I'm a bit concerned that this isn't coordinated with `FLASH_BANK_SPLIT`, and that it might catch other GD32 parts that have more than 512kiB of flash, but don't have split banks.

ALTracer: Are there such targets BMD explicitly supports? I couldn't detect them here. This STM32F1 driver expects a bank split at 512 KiB, and I don't want to refactor it much (because it incurs giant amounts of regress-testing).
GD32F450/470 are handled by STM32F4 flash driver because of their flash controller resembling F2/F4/F7. Even newest GD32H7 have a single bank of 3840 KiB in 4 KiB pages.
I try to be forward-looking and to break ideally zero targets, but looks like this project is contributor-driven w.r.t. less popular targets' driver support. I waited two months between posting Issue 1426 and submitting PR1503. Feel free to mention your target you use now (or will use soon) that needs additional handling in BMD auto-detection logic.

tlyu: Sorry, I was trying to suggest that some slight future-proofing of the code. A little bit of documentation is probably good enough.

It might be helpful to add comments in this area, and around the existing code that deals with `FLASH_BANK_SPLIT` (maybe just the `FLASH_BANK_SPLIT` definition), warning that they have to be manually checked for consistency when that code is modified.

tlyu:
I also confirm that this correctly detects the flash configuration on my GD32F303CGT6, at least according to the manual.

```
(gdb) info mem
Using memory regions provided by the target.
Num Enb Low Addr   High Addr  Attrs
0   y  	0x08000000 0x08080000 flash blocksize 0x800 nocache
1   y  	0x08080000 0x08100000 flash blocksize 0x1000 nocache
2   y  	0x20000000 0x20018000 rw nocache
(gdb) x/z 0xe0042000
0xe0042000:	0x21040430
```

tlyu:
Loading to bank 0 seems to work, using 2kiB erase blocks. Loading to bank 1 fails while attempting to erase:
```
gdb_getpacket: vFlashErase:08082000,00002000
Flash Erase 08082000 00002000
stm32f1_flash_erase: at 08082000
unlock failed, cr: 0x00000080
Erase failed at 8082000
gdb_putpacket: EFF
```

tlyu:
There seems to be an existing bug with unlocking bank 1 ("bank 2" in STM terminology). This seemed to fix it for me:
```diff
--- a/src/target/stm32f1.c
+++ b/src/target/stm32f1.c
@@ -587,7 +587,7 @@ static bool stm32f1_flash_unlock(target_s *target, uint32_t bank_offset)
 {
        target_mem_write32(target, FLASH_KEYR + bank_offset, KEY1);
        target_mem_write32(target, FLASH_KEYR + bank_offset, KEY2);
-       uint32_t ctrl = target_mem_read32(target, FLASH_CR);
+       uint32_t ctrl = target_mem_read32(target, FLASH_CR + bank_offset);
        if (ctrl & FLASH_CR_LOCK)
                DEBUG_ERROR("unlock failed, cr: 0x%08" PRIx32 "\n", ctrl);
        return !(ctrl & FLASH_CR_LOCK);
```

src/target/stm32f1.c
ALTracer: Added a bit per suggestion, unable to test. Now that I think about it, I want Arterytek AT32F403A (and by extension AT32F415) also benefit from 32-bit writes, because they are definitely capable of this per their reference manuals. Can I rename this bit flag to something like `STM32F1_TOPT_32BIT_WRITES`? This sounds more like JLink FW capability bit flags now.

ALTracer:
> There seems to be an existing bug with unlocking bank 1 ("bank 2" in STM terminology). This seemed to fix it for me:

Oh, very nice, thanks for finally testing that my idea is somewhat sane! I'd like to also see whether the performance numbers change, at least the gdb load side reported ones.
A bugfix, even. I agree with your patch. *Wonder if I ported it over to at32f43x.c driver or that's clean? I think I tested it and the relevant function doesn't look buggy*
Do you want me to submit it in this patchseries with Reported/Tested/Co-Authored By, or submit the oneliner yourself? Although I don't see how you'd test it without this PR merged, maybe I'd need to rebase it on top of that.

tlyu:
I think it makes sense to submit mine separately, because I think it affects more than GD32. It's probably isolated enough to merge cleanly, so I can test it that way.

src/target/stm32f1.c
tlyu: Either works for me; i'll leave it up to @dragonmux as the maintainer.

dragonmux: Think we're fine with just adding a comment for now (use `/* XXX: <message> */` to make it stand out as a known issue) and the next person who messes with this code can make it right. If the comment references this thread then it should provide all the context required.

We're working on the assumption that you'd rather not put the work and spoons in right now into doing this "right" rather than "good enough". It probably should do the proper thing for bank splitting though.

We don't mind taking it as is (+ comment) or fixed though - that's up to what you feel comfortable doing.

ALTracer: Expanded the comment a bit to set up expectations / remind of de-facto API.

dragonmux: Thank you, think that resolves this thread for now w/ the expectation that it'll be revisited at some later date to make it more broadly correct.

ALTracer:
Squashed the three PoC commits w.r.t. `ALIGN_32BIT` and TOPT into one. Appended the commentary block in first commit and updated the message to be broader/less misleading. Fast-forwarded a couple times on top of current `main` to record `size-diff`s.
Asking for any final round of review. Perf of reprogramming as tested increases by about 40% (20 KiB/s -> 27 KiB/s for AT32F403A)

PR #1710, opened 2024-01-02
Feature: Increase SRAM layout detail for STM32H7 chip variants

ALTracer:
## Detailed description

* This sounds like a feature (target flavour support).
* The existing problem is STM32H7 driver declaring the memory map of H74x for all chips detected, even if there are ID codes for H72x and H7Bx.
* This PR solves the problem by adding different parts of the memory maps, dispatched on part ID.

Flash cost is around +170 bytes. I am not sure whether contiguous regions composed of different SRAM banks should be added as such (increasing detail, but also BMP flash footprint and memory impact for memory-map.xml) or as single long regions. I tried keeping the common regions (ITCM and DTCM) unified, but a portion of DTCM can be aliased on some flavours (see RM).

I could only test on Nucleo-H743ZI2 and WeAct Studios MiniH723 boards. H72 has smaller regions and total amount of SRAM than H74. I don't have a H7B0 family chip to test on. STM32H745I-DISCO (dual-core?) is accessible but problematic (JTAG doesn't work).
Hence I add the Ref.man. descriptions and links from which the infromation is lifted.

There is a related bug of STM32H743 being detected over JTAG as "STM32MP15x M7" because of IDcode errata of the latter. I may tackle it here (to reduce churn) or in a new PR. SWD works as expected, but BMD may want actual DBGMCU->IDC read outs added to this target (and CM7/CM4 core checks).

## Closing issues

I could've posted an issue about these two things, but I also seem to be able to write a fix, and code contributions are best discussed/reviewed in a PR, not an Issue.

### DIFF
--- a/src/target/stm32h7.c
+++ b/src/target/stm32h7.c
@@ -24,8 +24,14 @@
  * the device, providing the XML memory map and Flash memory programming.
  *
  * References:
+ * RM0399 - STM32H745/755 and STM32H747/757 advanced Arm®-based 32-bit MCUs, Rev. 4
+ *   https://www.st.com/resource/en/reference_manual/rm0399-stm32h745755-and-stm32h747757-advanced-armbased-32bit-mcus-stmicroelectronics.pdf
  * RM0433 - STM32H742, STM32H743/753 and STM32H750 Value line advanced Arm®-based 32-bit MCUs, Rev. 8
  *   https://www.st.com/resource/en/reference_manual/dm00314099-stm32h742-stm32h743-753-and-stm32h750-value-line-advanced-arm-based-32-bit-mcus-stmicroelectronics.pdf
+ * RM0455 - STM32H7A3/7B3 and STM32H7B0 Value line advanced Arm®-based 32-bit MCUs, Rev. 10
+ *   https://www.st.com/resource/en/reference_manual/rm0455-stm32h7a37b3-and-stm32h7b0-value-line-advanced-armbased-32bit-mcus-stmicroelectronics.pdf
+ * RM0468 - STM32H723/733, STM32H725/735 and STM32H730 Value line advanced Arm®-based 32-bit MCUs, Rev. 3
+ *   https://www.st.com/resource/en/reference_manual/rm0468-stm32h723733-stm32h725735-and-stm32h730-value-line-advanced-armbased-32bit-mcus-stmicroelectronics.pdf
  */
 
 /*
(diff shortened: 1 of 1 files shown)


src/target/stm32h7.c
```diff
@@ -207,14 +213,44 @@ bool stm32h7_probe(target_s *target)
...
+	}
+	case ID_STM32H7Bx: {
+		/* Table 6. Memory map and default device memory area attributes RM0455, pg131 */
+#if 1
+		target_add_ram(target, 0x24000000, 0x40000); /* AXI RAM1, 256kiB */
+		target_add_ram(target, 0x24040000, 0x60000); /* AXI RAM2, 384kiB */
+		target_add_ram(target, 0x240a0000, 0x60000); /* AXI RAM3, 384kiB */
+#else
+		target_add_ram(target, 0x24000000, 0x100000); /* AXI RAM123, 1024kiB [256+384+384] contiguous */
+#endif
```
dragonmux, line 245: Given this is contiguous, think it's fine to just do the one `target_add_ram()`, which also saves on space a little. Same for any other contiguous ranges in the surrounding code - RAM isn't handled like Flash, so this is only telling GDB about the valid IO ranges when `inaccessible-by-default` is enabled.

dragonmux (changes requested):
Oooh, this is definitely a nice addition to the H7 support! There's one review item to do with range grouping, and with that addressed, we'll be happy to merge this

src/target/stm32h7.c
dragonmux: Please do, yeah

ALTracer:
Reworked the branch: backpropagated the KiB change, made the central commit cleaner (dropped `#if 1`), grouped all adjacent regions for a resulting flash cost reduced to only +76.
A couple minor issues: 
1. H72x has an option byte for `TCM_AXI_SHARED` which controls how much AXI RAM is split off to appear on ITCM (0-3 bits times 64 KiB granularity), so where do I declare the upper "half" of ITCM, if ever? And it's the only variant to get 2+4 memory ranges, others have 2+3 after aggregating. If the policy is to always present to GDB the maximum ranges (and not default ones), then I'd have to either add another in-ITCM range, or move/copy the ITCM definitions into flavour cases (optimizing for BMP heap usage, each `sizeof(struct target_ram)` is 12 bytes). Or we could present the minimum possible ranges, if accessing out of bounds upsets some AP, but then developers will have to `set mem inaccessible-by-default off` to see application stacks at end of AXI RAM.
2. Option byte support was never added to this driver. Ever since it was contributed back in PR286, the two lines were retained as comments. From what I understand, they are not presented as memory-mapped flash, but only as registers inside FLASH control APB area at `FPEC1_BASE`. This logic existing could allow me to gate/recalculate AXI/ITCM capacities distribution as part of BMP runtime detection. Something to consider in a follow-up PR -- make OB at least readable, if not writeable.

dragonmux:
Reading what you're saying about the H72x ITCM vs AXI RAM mapping - we'd suspect that the region remains valid or at least RaZ (Read as Zero) in the unused version of its mapping (whichever that is currently configured as), so it's fine - logical even - to map it twice, once in each of the two locations. This keeps BMD simpler while giving complete mappings for the user. Even if it bus faults, as you noted, it's really on the user to access the active version of the region - at least with both mappings provided, accesses will just work to the active one so making a nicer user experience.

We'd certainly be interested in support for the option bytes on these parts - perhaps as a secondary PR to improve support for these parts?

dragonmux (approved):
This all LGTM now, so we'll merge this if you feel it's ready now - we'll let you drive when we merge this in case a tweak needs making for the aforementioned AXI RAM vs ITCM mapping thing on the H72x parts.

ALTracer:
> ... the region remains valid or at least RaZ (Read as Zero) in the unused version of its mapping (whichever that is currently configured as), so it's fine - logical even - to map it twice, once in each of the two locations. This keeps BMD simpler while giving complete mappings for the user. Even if it bus faults, as you noted, it's really on the user ...

Cortex-M7(M4) BusFault is no big deal, but an AP fault or busy lockup may be harder to recover from (target lost, system reset). You'll see from user testing whether this bites back. A second reason I restrained from adding both aliases is `(gdb) info mem` then giving more SRAM than is practically present in the device -- these three parts are difficult to count kilobytes of real estate already (read AN5293 for a shorter summary than RM).

Go ahead and merge, I believe, and let the burden of taking up more BMP flash (with ITCM/AXI alias handling) lay on the next contributor 🙂 

> We'd certainly be interested in support for the option bytes on these parts - perhaps as a secondary PR to improve support for these parts?

Not from me -- I personally don't work with H7 parts in any significant extent (especially to require Option Byte manipulation), and there are tools to do so without relying on BMP probe intelligence: ST MaskROM DFU, CubeProgrammer, maybe other gdbservers.

dragonmux:
Ok, sounds good to us - merging then. Thank you for the contribution!

PR #1711, opened 2024-01-03
fix SWD turnaround issues

### LINKED ISSUE #1706: Inconsistent clock signal during SWDIO turnaround
### Summary

Serial wire debug (SWD/SWJ) clock signal generation is inconsistent during the turnaround cycle. The clock low time is to short during the turnaround cycle, which depending on the setup may result in the target's debug module missing the clock cycle entirely.

### Version and platform information

Commit SHA ef32a54d the most recent tag is v1.10.1

gcc-arm-none-eabi-10.3-2021.07 toolchain.

STM32F411 "black pill" board, but I am guessing the issue is hardware agnostic.

10KOhm resistor pull up on the SWDIO line.

### Steps to reproduce

The issue can be observed in all SWD transactions. Running `monitor swd_scan` with a valid debug target connected is enough to observe the issue on the oscilloscope trace.  The issue becomes more apparent at *higher* frequencies since the troubled clock pulse becomes short enough to be missed by the target.

### What is the current *bug* behavior?

Low side of the clock (SWCLK) signal does not adhere to the protocol timing during the turnaround cycle on the SWDIO pin. This causes a missing clock cycle at higher frequencies (see screenshot below).

### What is the expected *correct* behavior?

The clock signal remains consistent throughout the entire serial wire debug transaction. By consistent I mean the clock pulse width (high and low) should be greater or equal 1/(2*f), were f is the configured SWD frequency.

### Relevant logs, screenshots, or inputs

Following is a scope trace showing the missing clock low period during the first turnaround (BMP probe assuming receiver role). Clock frequency is measured at 36.4kHz. `monitor frequency` command says 79207Hz.

Screenshot 2023-12-27 at 6 08 05 PM

All consecutive SWD transactions contain the same issue during the turnaround cycle when the probe is assuming receiver role.

### Possible causes or fixes

The low clock delay is only applied when BMP is turning around to be transmitter. The low clock delay is missing when BMP is becoming receiver.

https://github.com/blackmagic-debug/blackmagic/blob/main/src/platforms/common/swdptap.c#L72

tlyu:
## Detailed description

There were multiple problems with SWD turnaround. In the write-to-read direction, a short clock pulse could be produced (#1706). When terminating a read transaction, the clock wasn't driven low. If a subsequent clock float was requested, as for a detach, the clock would float low, then be driven high when re-enabled, causing an extra rising clock edge.

Additionally, idle cycles weren't being inserted as required after a read transaction.

There's a small chance of a regression with targets that pull the SWD clock high.

## Closing issues

Fixes #1706.

## Details

These are debug logs from BMDA.

Log excerpt from unpatched, attaching after detaching. Note the non-response error after re-enabling the clock.

```
!SI20#
       K2000497C
remote_v0_swd_seq_in_parity 32 clock_cycles: 2000497c OK
Read RDBUFF: 0x2000497c
!GE0#
       K0
gdb_putpacket: OK
gdb_getpacket: vAttach;1
!GE1#
       K0
remote_v0_swd_seq_out 8 clock_cycles: 0000008d
!So088d#
       K0
!Si03#
       K7
remote_v0_swd_seq_in 3 clock_cycles: 00000007
!SI20#
       PFFFFFFFF
remote_v0_swd_seq_in_parity 32 clock_cycles: ffffffff ERR
Read STATUS: 0x00000000
DP Error 0x00000000
ap_mem_write_sized @ e000edf0 len 4, align 4: 03 00 5f a0
Write AP 0 CSW: 0xa3000052
remote_v0_swd_seq_out 8 clock_cycles: 000000b1
!So08b1#
       K0
!Si03#
       K7
remote_v0_swd_seq_in 3 clock_cycles: 00000007
!SI20#
       PFFFFFFFF
remote_v0_swd_seq_in_parity 32 clock_cycles: ffffffff ERR
Recovering and re-trying access
```

Full log from unpatched code:
[unpatched.txt](https://github.com/blackmagic-debug/blackmagic/files/13815259/unpatched.txt)

Log excerpt from patched, with only the `swdptap_turnaround` fix. Note the lack of non-response error after re-enabling the clock.

```
!SI20#
       K1
remote_v0_swd_seq_in_parity 32 clock_cycles: 00000001 OK
Read RDBUFF: 0x00000001
!GE0#
       K0
gdb_putpacket: OK
gdb_getpacket: vAttach;1
!GE1#
       K0
remote_v0_swd_seq_out 8 clock_cycles: 0000008d
!So088d#
       K0
!Si03#
       K1
remote_v0_swd_seq_in 3 clock_cycles: 00000001
!SI20#
       KF0000040
remote_v0_swd_seq_in_parity 32 clock_cycles: f0000040 OK
Read STATUS: 0xf0000040
DP Error 0x00000000
```

Full log from patch with only `swdptap_turnaround` fix:
[ta-no-8idle.txt](https://github.com/blackmagic-debug/blackmagic/files/13815260/ta-no-8idle.txt)

Full log with both commits:
[ta-with-8idle.txt](https://github.com/blackmagic-debug/blackmagic/files/13815262/ta-with-8idle.txt)

### DIFF
--- a/src/platforms/common/swdptap.c
+++ b/src/platforms/common/swdptap.c
@@ -68,17 +68,18 @@ static void swdptap_turnaround(const swdio_status_t dir)
 
 	if (dir == SWDIO_STATUS_FLOAT) {
 		SWDIO_MODE_FLOAT();
-	} else {
+	} else
 		gpio_clear(SWCLK_PORT, SWCLK_PIN);
-		for (volatile uint32_t counter = target_clk_divider + 1; counter > 0; --counter)
-			continue;
-	}
+
+	for (volatile uint32_t counter = target_clk_divider + 1; counter > 0; --counter)
+		continue;
 
 	gpio_set(SWCLK_PORT, SWCLK_PIN);
 	for (volatile uint32_t counter = target_clk_divider + 1; counter > 0; --counter)
 		continue;
 
 	if (dir == SWDIO_STATUS_DRIVE) {
+		gpio_clear(SWCLK_PORT, SWCLK_PIN);
 		SWDIO_MODE_DRIVE();
 	}
 }
--- a/src/target/adiv5_swd.c
+++ b/src/target/adiv5_swd.c
@@ -142,6 +142,7 @@ uint32_t adiv5_swd_read_no_check(const uint16_t addr)
 	const uint8_t res = swd_proc.seq_in(3U);
 	uint32_t data = 0;
 	swd_proc.seq_in_parity(&data, 32U);
+	swd_proc.seq_out(0, 8U);
 	return res == SWDP_ACK_OK ? data : 0;
 }
 
@@ -427,19 +428,20 @@ uint32_t firmware_swdp_low_access(adiv5_debug_port_s *dp, const uint8_t RnW, con
 			DEBUG_ERROR("SWD access resulted in parity error\n");
 			raise_exception(EXCEPTION_ERROR, "SWD parity error");
 		}
-	} else {
+	} else
 		swd_proc.seq_out_parity(value, 32U);
-		/* ARM Debug Interface Architecture Specification ADIv5.0 to ADIv5.2
-		 * tells to clock the data through SW-DP to either :
-		 * - immediate start a new transaction
-		 * - continue to drive idle cycles
-		 * - or clock at least 8 idle cycles
-		 *
-		 * Implement last option to favour correctness over
-		 *   slight speed decrease
-		 */
-		swd_proc.seq_out(0, 8U);
-	}
+
+	/* ARM Debug Interface Architecture Specification ADIv5.0 to ADIv5.2
+	 * tells to clock the data through SW-DP to either :
+	 * - immediate start a new transaction
+	 * - continue to drive idle cycles
+	 * - or clock at least 8 idle cycles
+	 *
+	 * Implement last option to favour correctness over
+	 *   slight speed decrease
+	 */
+	swd_proc.seq_out(0, 8U);
+
 	return response;
 }
 


tlyu:
Target is a GD32F303 on a Keyboardio Model 100 keyboard. Probe is BMP native, hardware v2.3b. There are 10k pull up resistors on all JTAG/SWD lines, except JTCLK/SWDCLK, which has a 10k pull down. (I misread the schematic when reporting this to Discord earlier.) In any case, the user manual for the chip says at reset, JTDI, JTMS/SWDIO, NJTRST are pulled up, JTCK/SWDCLK is pulled down, and JTDO is floating.

src/platforms/common/swdptap.c
```diff
@@ -70,15 +70,16 @@ static void swdptap_turnaround(const swdio_status_t dir)
 		SWDIO_MODE_FLOAT();
 	} else {
 		gpio_clear(SWCLK_PORT, SWCLK_PIN);
-		for (volatile uint32_t counter = target_clk_divider + 1; counter > 0; --counter)
-			continue;
 	}
```
dragonmux, line 73: Please eliminate the braces from the else block (can't for the if due to the macro, but can for the else)

src/platforms/common/swdptap.c
```diff
@@ -70,15 +70,16 @@ static void swdptap_turnaround(const swdio_status_t dir)
...
 
+	for (volatile uint32_t counter = target_clk_divider + 1; counter > 0; --counter)
+		continue;
```
dragonmux, line 76: Think this would read a little better to have a new line after the continue - personally we would move the block up two lines, but it's not a problem if you want space after the if block.. just don't want it cramping the gpio_set and its delay loop.

Our organisation pattern below was to \<change pin state\> immediately followed by \<delay loop\> then a blank new line to give space if it felt appropriate - but to have the pin state change and delay loop be the unit otherwise (two examples of this is are on lines 131 and 138).

src/target/adiv5_swd.c
```diff
@@ -427,6 +428,7 @@ uint32_t firmware_swdp_low_access(adiv5_debug_port_s *dp, const uint8_t RnW, con
 			DEBUG_ERROR("SWD access resulted in parity error\n");
 			raise_exception(EXCEPTION_ERROR, "SWD parity error");
 		}
+		swd_proc.seq_out(0, 8U);
```
dragonmux, line 431: Rather than inserting this here, let's move the comment and call from the else below to just above `return response` (and eliminate the else's braces)? Saves a duplicated call, and puts the explainer back with it

dragonmux (changes requested):
This is looking good - there are a couple of stylistic review notes we've got, but functionally this all looks good - we'll get this merged once the review items have been addressed as we can see it fixes the bug and we're satisfied that it's all functioning properly.

Particularly appreciate this gives an explanation for the attach bug we'd hit near the release of v1.9 and couldn't explain! We knew at the time that protocol recovery was just papering over the real problem, but this explains exactly what the problem actually was and fixes its root cause which is awesome.

tlyu:
Thanks for the review! I addressed your style comments.

dragonmux (approved):
All looks good to us now - good stuff, merging once we've briefly tested this with a couple of targets for good measure. Thank you for the contribution!

tlyu:
I created the patch for the missing turnaround delay before seeing that @stansotn had already made a very similar one. I can add credit to the commit message, if that's desired.

dragonmux:
Yeah, probably a kindness to add a `Co-Authored-By: ` line for them, good thinking!

dragonmux:
Well, we can confirm this all works as intended on our RP2040 and a STM32F411 and the attach protocol recovery trigger is now gone 👍🏼

tlyu:
@stansotn do you consent to be listed as co-author on this commit, and if so, should I use the author name and email address that you used in the commit you linked to #1706? Thanks!

stansotn:
@tlyu 
> do you consent to be listed as co-author on this commit, and if so, should I use the author name and email address that you used in the commit you linked to https://github.com/blackmagic-debug/blackmagic/issues/1706? Thanks!

I just changed the authorship of the commit to my personal email (gmail). I consent to be included as co-author. Thanks for asking!

https://github.com/stansotn/blackmagic/commit/aef2a641b3da3fe20cb78f6a0921f369f18120a9

dragonmux:
Merging now the authorship bit is solved, thank you all for your efforts and work on this!

PR #1713, opened 2024-01-04
Fix: Replace attribute alias in CRC32

ALTracer:
## Detailed description

* Not a new feature.
* The existing problem is MacOS/darwin failing to build from source due to `aliases are not supported on darwin` after #1711
* This PR solves the problem by replacing the mechanism used to keep distinct function names.

During writing PR1711 I applied the `DEBUG_ERROR("%s: ...", __func__, ...)` idiom, removing the `generic_crc32` name message-string from `stm32_crc32()` renamed implementation in the process. To keep the proper function names and avoid complicating the single callsite in `gdb_main.c`, I introduced a GCC-specific extension `__attribute__((alias("")))` to substitute jumps at link time. This does not fly on MacOS (Mach-O limitation vs. ELF), and I'm unable to test for that desktop platform. Note that this is fine on `arm-none-eabi-` adapters and Linux desktop, also MSYS2 Windows desktop (all of which are checked by CI). No comment on *BSD.

Also mark actual impls for `static` linkage so that they get inlined, it saves a jump.

### DIFF
--- a/src/crc32.c
+++ b/src/crc32.c
@@ -113,9 +113,6 @@ static bool generic_crc32(target_s *const target, uint32_t *const result, const 
 	uint8_t bytes[128U];
 #endif
 
-#ifndef DEBUG_INFO_IS_NOOP
-	const uint32_t start_time = platform_time_ms();
-#endif
 	uint32_t last_time = platform_time_ms();
 	for (size_t offset = 0; offset < len; offset += sizeof(bytes)) {
 		const uint32_t actual_time = platform_time_ms();
@@ -132,17 +129,6 @@ static bool generic_crc32(target_s *const target, uint32_t *const result, const 
 		for (size_t i = 0; i < read_len; i++)
 			crc = crc32_calc(crc, bytes[i]);
 	}
-#ifndef DEBUG_INFO_IS_NOOP
-	/* "generic_crc32: 08000110+75272 -> 1353ms, 54 KiB/s" */
-	const uint32_t end_time = platform_time_ms();
-	const uint32_t time_elapsed = end_time - start_time;
-	DEBUG_INFO("%s: %08" PRIx32 "+%" PRIu32 " -> %" PRIu32 "ms", __func__, base, (uint32_t)len, time_elapsed);
-	if (len >= 512U) {
-		const uint32_t speed = len * 1000U / time_elapsed / 1024U;
-		DEBUG_INFO(", %" PRIu32 " KiB/s", speed);
-	}
-	DEBUG_INFO("\n");
-#endif
 	*result = crc;
 	return true;
 }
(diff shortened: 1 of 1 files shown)


src/crc32.c
```diff
@@ -100,7 +100,7 @@ static uint32_t crc32_calc(const uint32_t crc, const uint8_t data)
 	return (crc << 8U) ^ crc32_table[((crc >> 24U) ^ data) & 0xffU];
 }
 
-bool generic_crc32(target_s *const target, uint32_t *const result, const uint32_t base, const size_t len)
+static bool generic_crc32(target_s *const target, uint32_t *const result, const uint32_t base, const size_t len)
```
dragonmux, line 103: If this is marked `inline`, does the compiler elide the call penalty `bmd_crc32()` should introduce? Or perhaps a better question: does adding the keyword make any chance to the resulting binary size? (same for stm32_crc32() below)

dragonmux (commented):
One query before we merge this - once we've got an answer, whatever that may be, we'll be happy to merge this.

src/crc32.c
ALTracer: Just `inline` breaks at link-time (after I un-static `crc32_calc` and `crc32_table` per GCC errors) with `src/crc32.c:220: undefined reference to 'generic_crc32'`. Similar error for `read_be4`, it being `static` also forces `stm32_crc32` impl to be `static` if the latter wanted to be `inline`. And even `static inline bool stm32_crc32()` results in this codegen:
```
0800550c <bmd_crc32>:
 800550c:       e786            b.n     800541c <stm32_crc32>
 800550e:       bf00            nop
```
i.e. `static inline` does not shrink, because it depends on the compiler weighting/decisionmaking. 
Adding `__attribute__((always_inline))` shrinks binary size back by 4, eliminating the cheap branch/jump on `native` and on `hosted`, but I don't trust toolchain-specific attributes to not cause FTBFS again.
At this point I could drop or keep these two lines of changes, it doesn't matter right now (but it could matter in the future when searching for optimization opportunities).

ALTracer:
Now that I think harder about it, and we got a common function in this path anyways (which the compiler doesn't feel like inlining), I got a new idea. Let's embrace it (the `bmd_crc32()` function). I can move the new fully duplicate perf log code into it, making the implementations themselves cleaner/drier and changing their clobbered register allocation / stack spillage. And then any instrumenting stack profilers get a chance at doing something before that big stack allocation (i.e. check stack depth at current moment or into the future, trigger errors etc.)

dragonmux:
Ooh, sure, that works! We were going to suggest experimenting with an approach like in target_probe.c, but we like this idea way better.

src/crc32.c
dragonmux: So it's doing the thing we worried it would.. well, given the new direction this PR's taking, we'll mark this resolved.

dragonmux (commented):
LGTM, don't see anything that needs addressing with this revised approach. Merging once the builds finish.

dragonmux (approved):

dragonmux:
Also, a note to self for the future: Once Meson branch lands, implement macOS CI so we can catch mistakes like this in the future.

PR #1714, opened 2024-01-04
common/swdptap: make SWD timing more consistent

tlyu:
Make SWD timing more consistent: output changes (including floating) take place after the falling edge, and reads take place before the rising edge.

This includes moving a gpio_get to occur after the clock-low delay, which would help reduce incorrect reads from lower-bandwidth target connections.

Eliminate redundant GPIO operations, typically ones that set the clock low, by making the bit string input/output primitives always end by outputting a falling clock edge.

## Detailed description

This makes the bit banging code more consistent and easier to analyze. It also removes some redundant clearing of the clock, typically because most existing big banging functions ended on a falling clock edge, and also began with an explicit clearing of the clock. This is no longer necessary, if we adopt a convention that the "resting" phase of the clock is low.

It also moves a misplaced `gpio_get` from before a clock-low delay to after it, so it comes right before the rising clock edge. This can help avoid incorrect reads from lower-bandwidth target connections. It might be helpful for @stansotn or someone else with such target connections to test it out.

I'll mark it ready for review after testing on BMDA, though I suspect that BMDA might slow things down too enough to make the differences harder to see on the logic analyzer.

Note that adopting an inlined busy-loop function as in #1688 might be needed to gain more significant clock consistency.

## Closing issues

N/A

### DIFF
--- a/src/platforms/common/swdptap.c
+++ b/src/platforms/common/swdptap.c
@@ -101,6 +101,11 @@ static uint32_t swdptap_seq_in_clk_delay(const size_t clock_cycles)
 	uint32_t value = 0;
 	if (!clock_cycles)
 		return 0;
+	/*
+	 * Count down instead of up, because with an up-count, some ARM-GCC
+	 * versions use an explicit CMP, missing the optimization of converting
+	 * to a faster down-count that uses SUBS followed by BCS/BCC.
+	 */
 	for (size_t cycle = clock_cycles; cycle--;) {
 		for (volatile uint32_t counter = target_clk_divider; counter > 0; --counter)
 			continue;
@@ -127,6 +132,11 @@ static uint32_t swdptap_seq_in_no_delay(const size_t clock_cycles)
 	if (!clock_cycles)
 		return 0;
 	uint32_t value = 0;
+	/*
+	 * Count down instead of up, because with an up-count, some ARM-GCC
+	 * versions use an explicit CMP, missing the optimization of converting
+	 * to a faster down-count that uses SUBS followed by BCS/BCC.
+	 */
 	for (size_t cycle = clock_cycles; cycle--;) {
 		/* Reordering barrier */
 		__asm__("" ::: "memory");
@@ -183,6 +193,11 @@ static void swdptap_seq_out_clk_delay(const uint32_t tms_states, const size_t cl
 	bool bit = value & 1U;
 	if (!clock_cycles)
 		return;
+	/*
+	 * Count down instead of up, because with an up-count, some ARM-GCC
+	 * versions use an explicit CMP, missing the optimization of converting
+	 * to a faster down-count that uses SUBS followed by BCS/BCC.
+	 */
 	for (size_t cycle = clock_cycles; cycle--;) {
 		/* Reordering barrier */
 		__asm__("" ::: "memory");
@@ -209,6 +224,11 @@ static void swdptap_seq_out_no_delay(const uint32_t tms_states, const size_t clo
 	bool bit = value & 1U;
 	if (!clock_cycles)
 		return;
+	/*
+	 * Count down instead of up, because with an up-count, some ARM-GCC
+	 * versions use an explicit CMP, missing the optimization of converting
+	 * to a faster down-count that uses SUBS followed by BCS/BCC.
+	 */
 	for (size_t cycle = clock_cycles; cycle--;) {
 		/* Reordering barrier */
 		__asm__("" ::: "memory");
--- a/src/platforms/common/swdptap.c
+++ b/src/platforms/common/swdptap.c
@@ -104,7 +104,7 @@ static uint32_t swdptap_seq_in_clk_delay(const size_t clock_cycles)
 	for (size_t cycle = clock_cycles; cycle--;) {
 		for (volatile uint32_t counter = target_clk_divider; counter > 0; --counter)
 			continue;
-		bool bit = gpio_get(SWDIO_IN_PORT, SWDIO_IN_PIN);
+		const bool bit = gpio_get(SWDIO_IN_PORT, SWDIO_IN_PIN);
 		gpio_set(SWCLK_PORT, SWCLK_PIN);
 		for (volatile uint32_t counter = target_clk_divider; counter > 0; --counter)
 			continue;
@@ -180,7 +180,7 @@ static void swdptap_seq_out_clk_delay(uint32_t tms_states, size_t clock_cycles) 
 static void swdptap_seq_out_clk_delay(const uint32_t tms_states, const size_t clock_cycles)
 {
 	uint32_t value = tms_states;
-	bool bit = value & 1;
+	bool bit = value & 1U;
 	if (!clock_cycles)
 		return;
 	for (size_t cycle = clock_cycles; cycle--;) {
@@ -194,7 +194,7 @@ static void swdptap_seq_out_clk_delay(const uint32_t tms_states, const size_t cl
 			continue;
 		__asm__("nop" ::: "memory");
 		value >>= 1U;
-		bit = value & 1;
+		bit = value & 1U;
 		/* Reordering barrier */
 		__asm__("" ::: "memory");
 		gpio_clear(SWCLK_PORT, SWCLK_PIN);
@@ -206,7 +206,7 @@ static void swdptap_seq_out_no_delay(uint32_t tms_states, size_t clock_cycles) _
 static void swdptap_seq_out_no_delay(const uint32_t tms_states, const size_t clock_cycles)
 {
 	uint32_t value = tms_states;
-	bool bit = value & 1;
+	bool bit = value & 1U;
 	if (!clock_cycles)
 		return;
 	for (size_t cycle = clock_cycles; cycle--;) {
@@ -217,7 +217,7 @@ static void swdptap_seq_out_no_delay(const uint32_t tms_states, const size_t clo
 		gpio_set(SWCLK_PORT, SWCLK_PIN);
 		__asm__("nop" ::: "memory");
 		value >>= 1U;
-		bit = value & 1;
+		bit = value & 1U;
 		/* Reordering barrier */
 		__asm__("" ::: "memory");
 	}
--- a/src/platforms/common/swdptap.c
+++ b/src/platforms/common/swdptap.c
@@ -46,6 +46,18 @@ static bool swdptap_seq_in_parity(uint32_t *ret, size_t clock_cycles) __attribut
 static void swdptap_seq_out(uint32_t tms_states, size_t clock_cycles) __attribute__((optimize(3)));
 static void swdptap_seq_out_parity(uint32_t tms_states, size_t clock_cycles) __attribute__((optimize(3)));
 
+/*
+ * Overall strategy for timing consistency:
+ *
+ * - Each primitive ends with a falling clock edge
+ * - Output is driven after the falling clock edge
+ * - Input is read immediately before the rising clock edge
+ * - Each primitive assumes it was immediately preceded by a falling clock edge
+ *
+ * This increases the chances of meeting setup and hold times when the target
+ * connection is lower bandwidth (with adequately slower clocks configured).
+ */
+
 void swdptap_init(void)
 {
 	swd_proc.seq_in = swdptap_seq_in;
@@ -68,9 +80,7 @@ static void swdptap_turnaround(const swdio_status_t dir)
 
 	if (dir == SWDIO_STATUS_FLOAT) {
 		SWDIO_MODE_FLOAT();
-	} else
-		gpio_clear(SWCLK_PORT, SWCLK_PIN);
-
+	}
 	for (volatile uint32_t counter = target_clk_divider + 1; counter > 0; --counter)
 		continue;
 
@@ -78,8 +88,8 @@ static void swdptap_turnaround(const swdio_status_t dir)
 	for (volatile uint32_t counter = target_clk_divider + 1; counter > 0; --counter)
 		continue;
 
+	gpio_clear(SWCLK_PORT, SWCLK_PIN);
 	if (dir == SWDIO_STATUS_DRIVE) {
-		gpio_clear(SWCLK_PORT, SWCLK_PIN);
 		SWDIO_MODE_DRIVE();
 	}
 }
@@ -90,15 +100,14 @@ static uint32_t swdptap_seq_in_clk_delay(const size_t clock_cycles)
 {
 	uint32_t value = 0;
 	for (size_t cycle = 0; cycle < clock_cycles; ++cycle) {
-		gpio_clear(SWCLK_PORT, SWCLK_PIN);
-		value |= gpio_get(SWDIO_IN_PORT, SWDIO_IN_PIN) ? 1U << cycle : 0U;
 		for (volatile uint32_t counter = target_clk_divider; counter > 0; --counter)
 			continue;
+		value |= gpio_get(SWDIO_IN_PORT, SWDIO_IN_PIN) ? 1U << cycle : 0U;
 		gpio_set(SWCLK_PORT, SWCLK_PIN);
 		for (volatile uint32_t counter = target_clk_divider; counter > 0; --counter)
 			continue;
+		gpio_clear(SWCLK_PORT, SWCLK_PIN);
 	}
-	gpio_clear(SWCLK_PORT, SWCLK_PIN);
 	return value;
 }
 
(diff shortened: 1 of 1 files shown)


src/platforms/common/swdptap.c
```diff
@@ -150,27 +160,26 @@ static void swdptap_seq_out_clk_delay(uint32_t tms_states, size_t clock_cycles)
...
 
 static void swdptap_seq_out_no_delay(uint32_t tms_states, size_t clock_cycles) __attribute__((optimize(3)));
 
 static void swdptap_seq_out_no_delay(const uint32_t tms_states, const size_t clock_cycles)
 {
 	for (size_t cycle = 0; cycle < clock_cycles; ++cycle) {
-		gpio_clear(SWCLK_PORT, SWCLK_PIN);
 		gpio_set_val(SWDIO_PORT, SWDIO_PIN, tms_states & (1 << cycle));
 		gpio_set(SWCLK_PORT, SWCLK_PIN);
+		__asm__("nop");
```
ALTracer, line 180: This 0xbf00 architectural NOP may be consumed by Cortex-M3/M4 in 0 HCLK cycles. Multiple nops can stall the ART, or L1 I-cache, or FLITF prefetch on I-bus with low predictability. If you want to stretch the SWCLK high phase, consider using something different, like a second GPIO write (`gpio_set(SWCLK_PORT, SWCLK_PIN)` again, it should be hot/cached). This use of double writes seems much more justified to me (as opposed to forcing every gpio write be done twice for "faster" stm32f4). Or just "call"/inline my 3(4?)-cycle loop once, whatever is quicker.

tlyu:
I think the cycles saved by this change are small enough that my logic analyzer can't reliably see them, so I'm going to mark this as ready for review.

dragonmux (approved):
From a code perspective this all looks good to us. We'll try to get this tested on our local setup in the next couple of days to see what the before-and-after on the timing and waveforms looks like with a mind to merge this if it all looks fine.

We'll post our testing results as images and traces here to be able to reference back to if needed.

tlyu:
I wonder if it's useful to hang some extra capacitance off of the probe wiring, to slow things down enough to A/B test the change to the bit read timing. (I don't have a MSO or similar at the moment, which would likely help a lot with this kind of test.)

src/platforms/common/swdptap.c
dragonmux: Due to how this is used, this makes a noticeable difference in the bitbanged output, at least on a STM32F103/GD32F103. It may need different handling on a souped up part, but is correct/fine on `native`.

src/platforms/common/swdptap.c
```diff
@@ -108,12 +117,11 @@ static uint32_t swdptap_seq_in_no_delay(const size_t clock_cycles)
 {
 	uint32_t value = 0;
 	for (size_t cycle = 0; cycle < clock_cycles; ++cycle) {
-		gpio_clear(SWCLK_PORT, SWCLK_PIN);
 		value |= gpio_get(SWDIO_IN_PORT, SWDIO_IN_PIN) ? 1U << cycle : 0U;
 		gpio_set(SWCLK_PORT, SWCLK_PIN);
 		__asm__("nop");
```
dragonmux, line 122: This and the `nop` below could maybe do with being duplicated once to improve the resulting waveform, but this is only a suggestion as this does work and produce a valid minimum delay signal.

dragonmux (commented):
We've just given this a test, screenshots of the resulting waveforms to follow, and while the delay routines on minimum delay now have a very slight duty cycle imbalance on SWCLK, the clarity improvements and other changes more than make up for that and it doesn't cause bad behaviour from parts tested. If all involved are happy, we're happy to merge this.

dragonmux:
Here are our results from testing this PR - first is a set of before captures from `main`:

no-delay SWD out:
no-delay SWD out signaling

no-delay SWD in:
no-delay SWD in signaling

minimum delay SWD out:
minimum delay SWD out signaling

minimum delay SWD in:
minimum delay SWD in signaling

Now follow the after captures taken without any extra nop's:

no-delay SWD out:
no-delay SWD out signaling

no-delay SWD in:
no-delay SWD in signaling

minimum delay SWD out:
minimum delay SWD out signaling

minimum delay SWD in:
minimum delay SWD in signaling

And finally, the no-delay routines with one extra nop:

no-delay SWD out:
no-delay SWD out signaling

no-delay SWD in:
no-delay SWD in signaling

tlyu:
Thanks for the review!

I noticed similar changes as what you saw, but was also limited to a logic analyzer with 24MHz sample rate. Many of the asymmetries are on the order of one LA sample, and aren't that consistent. I can take a closer look to see if there are any asymmetries that are somewhat more consistent, and try to adjust the timing on those.

On BMP native clocked at 72MHz (default), it seems like a 1-cycle difference could easily result in an inconsistent 1-sample asymmetry at 24MSamp/sec. It would be nice to get some traces from a faster logic analyzer.

dragonmux:
What we tend to do is assume that the asymmetry errs on the side that appears more commonly - so that final capture is showing an asymmetry ~2/3 of the way to the right between the two seen asymetries - they're both asymetrical on the side of the high time being low; just.. slightly closer to even than the captures showing one high period would otherwise suggest

Agreed, though, that getting some high sample density captures would be great - if we can get our scope set up for it'll we'll try to get some captures with that - but that'll be a few days out.

tlyu:
Force-push is to resolve a merge conflict; I haven't dug into the asymmetries in detail yet.

tlyu:
After poking at this a bit, staring at LA traces and disassembly, most of the calculations that occur inside the loop happen during the clock low phase for output. Input calculations can be moved around a bit. In any case, moving the clock clear to the top of the loop for output makes it easier to rebalance the duty cycle.

(`gpio_set_val` has enough branching that it takes a comparable number of cycles to the loop increment)

tlyu:
Thanks for the review and detailed timing analysis! I've made some adjustments that seem to improve the duty cycle for me, at least as well as I can see at 24MSamp/sec.

dragonmux:
Ah, sorry the ball got dropped a bit on this one - could you rebase this on `main` please and we'll look at getting this merged ASAP as we think it's probably ready to go otherwise.

src/platforms/common/swdptap.c
```diff
@@ -89,31 +99,49 @@ static uint32_t swdptap_seq_in_clk_delay(size_t clock_cycles) __attribute__((opt
...
 	uint32_t value = 0;
-	for (size_t cycle = 0; cycle < clock_cycles; ++cycle) {
-		gpio_clear(SWCLK_PORT, SWCLK_PIN);
-		value |= gpio_get(SWDIO_IN_PORT, SWDIO_IN_PIN) ? 1U << cycle : 0U;
+	if (!clock_cycles)
+		return 0;
+	for (size_t cycle = clock_cycles; cycle--;) {
 		for (volatile uint32_t counter = target_clk_divider; counter > 0; --counter)
 			continue;
+		bool bit = gpio_get(SWDIO_IN_PORT, SWDIO_IN_PIN);
```
dragonmux, line 107: Ideally this would be const, but can be sorted later

src/platforms/common/swdptap.c
```diff
@@ -89,31 +99,49 @@ static uint32_t swdptap_seq_in_clk_delay(size_t clock_cycles) __attribute__((opt
 static uint32_t swdptap_seq_in_clk_delay(const size_t clock_cycles)
 {
 	uint32_t value = 0;
-	for (size_t cycle = 0; cycle < clock_cycles; ++cycle) {
-		gpio_clear(SWCLK_PORT, SWCLK_PIN);
-		value |= gpio_get(SWDIO_IN_PORT, SWDIO_IN_PIN) ? 1U << cycle : 0U;
+	if (!clock_cycles)
+		return 0;
+	for (size_t cycle = clock_cycles; cycle--;) {
```
dragonmux, line 104: We're still not sold on turning this into a count-down loop due to the increased difficulty in reasoning about the termination of the loop - given the shift at the end, and how Thumb works, we'd imagine that testing `if (clock_cycles)` before the right shift and leaving this as a count up loop would result in the same or more efficient code, while more clearly conveying the termination conditions.

src/platforms/common/swdptap.c
```diff
@@ -89,31 +99,49 @@ static uint32_t swdptap_seq_in_clk_delay(size_t clock_cycles) __attribute__((opt
...
-	gpio_clear(SWCLK_PORT, SWCLK_PIN);
+	value >>= (32U - clock_cycles);
 	return value;
 }
 
 static uint32_t swdptap_seq_in_no_delay(size_t clock_cycles) __attribute__((optimize(3)));
 
 static uint32_t swdptap_seq_in_no_delay(const size_t clock_cycles)
 {
+	if (!clock_cycles)
```
dragonmux, line 127: Same comment in this function as the one above about count-down vs count-up loops.

src/platforms/common/swdptap.c
```diff
@@ -132,43 +160,66 @@ static bool swdptap_seq_in_parity(uint32_t *ret, size_t clock_cycles)
...
 static void swdptap_seq_out_clk_delay(const uint32_t tms_states, const size_t clock_cycles)
 {
-	for (size_t cycle = 0; cycle < clock_cycles; ++cycle) {
-		gpio_clear(SWCLK_PORT, SWCLK_PIN);
-		gpio_set_val(SWDIO_PORT, SWDIO_PIN, tms_states & (1U << cycle));
+	uint32_t value = tms_states;
+	bool bit = value & 1;
+	if (!clock_cycles)
+		return;
+	for (size_t cycle = clock_cycles; cycle--;) {
```
dragonmux, line 186: Same comment about count up vs down loops - minus the need for a post-loop check.

src/platforms/common/swdptap.c
```diff
@@ -132,43 +160,66 @@ static bool swdptap_seq_in_parity(uint32_t *ret, size_t clock_cycles)
...
 static void swdptap_seq_out_no_delay(uint32_t tms_states, size_t clock_cycles) __attribute__((optimize(3)));
 
 static void swdptap_seq_out_no_delay(const uint32_t tms_states, const size_t clock_cycles)
 {
-	for (size_t cycle = 0; cycle < clock_cycles; ++cycle) {
+	uint32_t value = tms_states;
+	bool bit = value & 1;
+	if (!clock_cycles)
+		return;
+	for (size_t cycle = clock_cycles; cycle--;) {
```
dragonmux, line 212: Same comment again.

src/platforms/common/swdptap.c
```diff
@@ -132,43 +160,66 @@ static bool swdptap_seq_in_parity(uint32_t *ret, size_t clock_cycles)
...
 
 static void swdptap_seq_out_clk_delay(uint32_t tms_states, size_t clock_cycles) __attribute__((optimize(3)));
 
 static void swdptap_seq_out_clk_delay(const uint32_t tms_states, const size_t clock_cycles)
 {
-	for (size_t cycle = 0; cycle < clock_cycles; ++cycle) {
-		gpio_clear(SWCLK_PORT, SWCLK_PIN);
-		gpio_set_val(SWDIO_PORT, SWDIO_PIN, tms_states & (1U << cycle));
+	uint32_t value = tms_states;
+	bool bit = value & 1;
```
dragonmux, line 183: Almost all the places bit masking is being done are fine, but this and the 3 other `& 1`'s below are missing their `U` suffix to make sure this stays unsigned and doesn't cast. This should be a signed-unsigned conversion warning were `-Wconversion`/`-Warith-conversion` to be enabled.

dragonmux (changes requested):
Apologies this PR slipped away from us.. We've a few review notes from this round.

With the up-vs-down loops, we don't recall if we've already had a conversation about it and what the outcome was, so please forgive us if you've already explained to us before if there's a merit for the proposed change and we've just forgotten it. If there's a clearly demonstrated technical reason for using the count-down style here, please add comments in the code explaining the choice and otherwise ignore those 4 lints.

dragonmux:
We're working to tidy up and merge outstanding good quality PRs at the moment, and notice this one's not seen any movement since our review. If you're happy with us sorting the outstanding items and rebasing this on `main`, then we'll get cracking with that in the next week with the intention of landing this by mid month.

tlyu:
Catching up… I'll try to look at it this week as well. I first have to see if I can even get the tree to build on macOS at the moment (I remember hearing about some Meson problems).

dragonmux:
Sounds good! 🤞🏼 you can figure out the macOS build - there was a set of bug fixes pushed to `main` recently that solved some issues around BMDA on the fruity OS when doing a cross-build to build the firmware. Once rebased on `main`, that should be a reasonable process.

tlyu:
Got Meson to work on macOS. Native doesn't fit into ROM with the default config on `main`.

Still puzzling through at least one merge conflict.

dragonmux:
That sounds like some fairly reasonable progress. Which compiler version are you using for `main` to not be fitting (as far as we know, 12.2.Rel1 builds still fit)?

tlyu:
`arm-none-eabi-gcc (Arm GNU Toolchain 13.2.rel1 (Build arm-13.7)) 13.2.1 20231009`

dragonmux:
Ah, yes, that toolchain has worse codegen than 12.2.Rel1 so that makes sense of that - this is why we've held back on upgrading toolchains for a while now because the project's trying to squeeze every possible bit of space out at the moment..

tlyu:
Force push is to resolve merge conflicts. I'll check the other comments shortly.

src/platforms/common/swdptap.c
tlyu: The left operand is a `uint32_t`. My understanding is that the only way the C arithmetic conversion rules work, the only way that literal un-suffixed integer `1` can convert the left-hand operand to a signed type is if `uint32_t` can't represent all the values of `int`, which is unlikely on a 32-bit platform.

I can add the `U` suffix anyway if that helps suppress warnings or makes it easier to manually analyze.

src/platforms/common/swdptap.c
tlyu: I documented the difference in the commit message "adjust timing". Unfortunately, I didn't note the compiler version involved, which might be relevant.

src/platforms/common/swdptap.c
dragonmux: The way the rules work, mean that if either side doesn't match in signed-ness, both are converted to `int` for the operation - so `& 1` turns `value` -> `int`, and then because the operation clears the sign bit, if set, this invokes UB. All bitwise manipulation of values is best done unsigned to avoid this and all signed-unsigned conversions. The result of `value & 1` is therefore `int` while `value & 1U` is `unsigned`.

Please add the suffix to prevent this error and UB invocation which is what those warnings pick up on and warn about.

src/platforms/common/swdptap.c
dragonmux: Could the reasoning be noted in the code please - if some other PR touches these lines and such, the blame log won't point to your commit, and trying to traverse it just to learn the intent and reasoning is a big non-starter. If it's inline as comments in the code, there's no complex discovery required.

We can then have an experiment in Compiler Explorer with a few different versions of GCC to see if it's compiler version specific and what the impacts are.

src/platforms/common/swdptap.c
tlyu: From C2x N3054

> 6.3.1.8 Usual arithmetic conversions

[...]

> Otherwise, if the operand that has unsigned integer type has rank greater or equal to the rank of the type of the other operand, then the operand with signed integer type is converted to the type of the operand with unsigned integer type.


> Otherwise, if the type of the operand with signed integer type can represent all the values of the type of the operand with unsigned integer type, then the operand with unsigned integer type is converted to the type of the operand with signed integer type.

So the risk of unsigned-to-signed conversion is if `uint32_t` is a lower rank than `int`. I'll add the `U` suffix for future-proofing, though.

src/platforms/common/swdptap.c
tlyu: I wound up making a different set of adjustments that similarly improve symmetry.

src/platforms/common/swdptap.c
dragonmux: In C11 (N1548), the relevant chunk from that starts out with:

> Otherwise, the integer promotions are performed on both operands. Then the
following rules are applied to the promoted operands:

Meaning that before any of those other "Otherwise" clauses occur, integer promotion takes place. This is governed by §6.3.1.1 par 2. Suffixing inhibits all of this completely though.

src/platforms/common/swdptap.c
dragonmux: That works! ^_^

src/platforms/common/swdptap.c
tlyu: Ah, C1x lacks the language about bit-precise types in 6.3.1.1. Anyway, integer promotions happen separately for each relevant operand. I think `uint32_t` won't promote to (signed) `int` unless `int` can represent all the values of `uint32_t`, which it won't on most 32-bit platforms. In the unlikely case that `uint32_t` does promote to (signed) `int`, adding the `U` suffix to the literal will keep both operands unsigned during the usual arithmetic conversions.

dragonmux: Correct, bit-precise was added in 23; and well, that promotion occurring has repeatedly bitten BMD, which is why the campaign to U-suffix all bitwise manipulation as it's fixed all kinds of bugs and miscompilations/misoptimisation allowed by the resulting UB invoked when accidentally doing bad things to the sign bit, accidentally doing `asr` instead of `lsr`, and so on. Suffixing makes the types all agree and inhibits that class of problem for cheap. We'd prefer to over-apply it than under, even if the standard says in this narrow case that it's Probably Ok™.

src/platforms/common/swdptap.c
tlyu: Added comments. Please let me know if you'd prefer that I deduplicate some of them and replace them with references.

dragonmux (approved):
All review items address, so LGTM - merging. Thank you for the contribution!

PR #1715, opened 2024-01-07
Feature: Working logs on `blackpill-f4` via DebugMon

ALTracer:
## Detailed description

* This is a new feature common to other platforms but missing from this one.
* There is an existing problem of no log traces available on `blackpill-f4` family of platforms.
* The PR solves the problem by adding required pieces and fixing linker flags.

How to reproduce the issue: build with `PROBE_HOST=blackpill-f411ce ENABLE_DEBUG=1`, optionally also with `BMP_BOOTLOADER=1`, upgrade/flash to the board, open gdb, ask for `monitor debug_bmp enable` (aka "mon debug en") -- `Target does not support this command.` Even though the binary is 40 KiB bigger.

I've been carrying the patches to use logging on my platform of choice since probably June 2023, which is when I started using the project, and I'm tired of unstashing them on top of every feature branch I'm testing just to see some important `DEBUG_INFO()`s. Both `stlink` and `swlink` work without such changes (but they can't fit both logs and default targets, which is a separate problem solved by an edited Makefile dirtying the working copy).

After another round through reading up on semihosting and debugging down through how it actually works, I now understand how this optional BMP feature was implemented on `native` all the way back in 2016 in PR151 https://github.com/blackmagic-debug/blackmagic/commit/966f360515e387fb79223a75986aa7b79878268a

Note two things: 1) I do not intend to downgrade to newlib-nano in BMPbootloader or in BMF, because it provides a faster memcpy and actual working mallopt+malloc_trim. 2) That means that the bootloader with 12-symbol serialno will no longer fit under 16 KiB, but I have a proposal in a separate PR to deal with it. Platform MaskROM generates a 12-symbol serial number when enumerating on USB, and I would like to enforce feature parity with it.

### DIFF
--- a/src/platforms/common/blackpill-f4/Makefile.inc
+++ b/src/platforms/common/blackpill-f4/Makefile.inc
@@ -10,16 +10,16 @@ CFLAGS +=                           \
 	-mfloat-abi=hard                \
 	-mfpu=fpv4-sp-d16               \
 	-DSTM32F4                       \
+	-DDFU_SERIAL_LENGTH=13          \
 	-I../libopencm3/include         \
 	-Iplatforms/common/stm32        \
 	-Iplatforms/common/blackpill-f4
 
-LDFLAGS_BOOT =              \
+LDFLAGS_BOOT :=              \
 	-lopencm3_stm32f4       \
 	-Wl,-T,$(LINKER_SCRIPT) \
 	-nostartfiles           \
 	-lc                     \
-	-lnosys                 \
 	-Wl,-Map=mapfile        \
 	-mthumb                 \
 	-mcpu=cortex-m4         \
@@ -32,10 +32,8 @@ ifeq ($(BMP_BOOTLOADER), 1)
 $(info  Load address 0x08004000 for BMPBootloader)
 LDFLAGS = $(LDFLAGS_BOOT) -Wl,-Ttext=0x8004000
 CFLAGS += -DAPP_START=0x08004000 -DBMP_BOOTLOADER
-CFLAGS += -DDFU_SERIAL_LENGTH=9
 else
-LDFLAGS += $(LDFLAGS_BOOT)
-CFLAGS += -DDFU_SERIAL_LENGTH=13
+LDFLAGS = $(LDFLAGS_BOOT)
 endif
 
 ifdef ALTERNATIVE_PINOUT
@@ -44,6 +42,12 @@ endif
 
 ifdef SHIELD
 CFLAGS += -DSHIELD=$(SHIELD)
+endif
+
+ifeq ($(ENABLE_DEBUG), 1)
+LDFLAGS += --specs=rdimon.specs
+else
+LDFLAGS += --specs=nosys.specs
 endif
 
 VPATH +=                          \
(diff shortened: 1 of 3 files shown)


dragonmux (approved):
This all LGTM. We haven't the hardware to test it, so if you say this works (and hey, this explains some things in the native platform too that didn't have comments!) we'll trust you on that.

Given your PR #1716 which deals with the bootloader space complaint, we'll provisionally approve this but intend to merge it after that other PR so `main` doesn't wind up broken for an indeterminate amount of time for this platform.

ALTracer:
Eh, I don't feel so confident about the Makefile.inc change. Is --nosys.specs handled right and is it better/proper way than -lnosys? There is another variant of the commit in PR1717 as third in series, but somewhy I though appending --nano.specs was a good idea (not bootloader space constraints but feature/libc/impl parity of firmware with other platforms, or making it fit below 0x08020000), and I didn't rebase it out of there.
Please review through all three PRs as they are related, and I tried to lift / rephrase arguments out of commit messages.

dragonmux:
`-lnosys` is implied by `--specs=nosys.specs`, so that change is definitely fine - and the logic here looks identical, from memory, to the native platform's for the same, so ignoring the difference of not dragging `--specs=nano.specs` in, it reads to us correctly per the stated goals of the PR.

dragonmux:
Now that the other PR is merged, if you want to check over this one still functions then we're happy to merge this and clear the way for #1717

ALTracer:
This PR of two commits provides/enables the mentioned feature. I swapped the commit order because my pre-commit hook indicated a build of one of four combinations failing to link (probably nosys/rdimon conflict)
```sh
make -C src/ clean
make -C src/ PROBE_HOST=blackpill-f411ce CROSS_COMPILE="ccache arm-none-eabi-" -j4 || exit 13
make -C src/ clean
make -C src/ PROBE_HOST=blackpill-f411ce CROSS_COMPILE="ccache arm-none-eabi-" ENABLE_DEBUG=1 -j4 || exit 14
make -C src/ clean
make -C src/ PROBE_HOST=blackpill-f411ce CROSS_COMPILE="ccache arm-none-eabi-" BMP_BOOTLOADER=1 -j4 || exit 15
make -C src/ clean
make -C src/ PROBE_HOST=blackpill-f411ce CROSS_COMPILE="ccache arm-none-eabi-" BMP_BOOTLOADER=1 ENABLE_DEBUG=1 -j4 || exit 16
```
Flashing the build of first commit with debug enabled produces a probe hanging upon first `monitor swdp_scan`, probably because it goes through `DEBUG_INFO()`, `PLATFORM_PRINTF()`, `printf()`, `_swiwrite()`, and a breakpoint (which in absence of `SCS->DHCSR & MON_EN` is equivalent to a HardFault exception, not DebugMon exception), not passing further to `debug_monitor_handler()` or overridden `_write()` into `debug_serial_debug_write()`. By the way, this is the behaviour until now, and coincidentally on meson branches.
Flashing the build of both commits with debug enabled produces a healthy probe printing nothing interesting to ttyBmpTarg unless asked for `monitor debug_bmp enable` and working otherwise.

dragonmux:
That sounds like a fully working patch to us! We'll get this merged then, thank you 🙂

PR #1716, opened 2024-01-07
Feature: Space-efficient 12-symbol serialno generation

ALTracer:
* This can be considered a new feature (because I don't see how it was usable previously)
* There is an existing problem of BMPBootloader with `DFU_SERIAL_LENGTH=13` becoming larger than 16 KiB due to `siprintf` pulled from newlib (not nano).
* The PR solves this problem by replacing the single expensive `sprintf()` call with a few calls of simpler functions and introducing a variant of `utoa()`.

12-symbol serialno consists of three halfwords calculated by pulling 16-bit values from Device Electronic Signature memory. I don't believe `hexify()` can be useful here.

I would not like to choose to mitigate this by downgrading the F4 BMPBootloader to newlib-nano because it would provide a small and inefficient memcpy, which is important for flash writes/upgrade reprogramming speed. And it's not like a bootloader would use *printf anyway.

* [x] It builds for hardware native (`make PROBE_HOST=native`) -- and is not applicable
* [x] It builds as BMDA (`make PROBE_HOST=hosted`) -- and is not applicable

### DIFF
--- a/src/platforms/common/stm32/serialno.c
+++ b/src/platforms/common/stm32/serialno.c
@@ -50,7 +50,9 @@ void read_serial_number(void)
 #elif defined(STM32L0) || defined(STM32F0) || defined(STM32F3)
 	int offset = 5;
 #endif
-	sprintf(serial_no, "%04X%04X%04X", uid[1] + uid[5], uid[0] + uid[4], uid[offset]);
+	utoa_upper(uid[1] + uid[5], serial_no, 16);
+	utoa_upper(uid[0] + uid[4], serial_no + 4, 16);
+	utoa_upper(uid[offset], serial_no + 8, 16);
 #elif DFU_SERIAL_LENGTH == 25
 	const volatile uint32_t *const unique_id_p = (uint32_t *)DESIG_UNIQUE_ID_BASE;
 	uint32_t unique_id = 0;
--- a/src/platforms/common/stm32/serialno.c
+++ b/src/platforms/common/stm32/serialno.c
@@ -2,7 +2,12 @@
  * This file is part of the Black Magic Debug project.
  *
  * Copyright (C) 2015  Black Sphere Technologies Ltd.
+ * Copyright (C) 2017-2021 Uwe Bonnes <bon@elektron.ikp.physik.tu-darmstadt.de>
+ * Copyright (C) 2022-2024 1BitSquared <info@1bitsquared.com>
  * Written by Gareth McMullin <gareth@blacksphere.co.nz>
+ * Modified by Uwe Bonnes <bon@elektron.ikp.physik.tu-darmstadt.de>
+ * Modified by Rachel Mant <git@dragonmux.network>
+ * Modified by ALTracer <tolstov_den@mail.ru>
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
--- a/src/platforms/common/stm32/serialno.c
+++ b/src/platforms/common/stm32/serialno.c
@@ -23,6 +23,8 @@
 
 char serial_no[DFU_SERIAL_LENGTH];
 
+static char *utoa_upper(uint32_t value, char *const str, uint8_t base) __attribute__((unused));
+
 void read_serial_number(void)
 {
 #if DFU_SERIAL_LENGTH == 9
@@ -63,4 +65,47 @@ void read_serial_number(void)
 #warning "Unhandled DFU_SERIAL_LENGTH"
 #endif
 	serial_no[DFU_SERIAL_LENGTH - 1] = '\0';
+}
+
+/*
+ * Converts an unsigned integer value into an equivalent base-N ASCII printable representation.
+ * It takes as arguments (in order): the value to convert,
+ * a pointer to a buffer large enough to store the result,
+ * and the base to perform the conversion to, in between 2 and 36 inclusive.
+ * The result of this function is NULL on error (unsupported base),
+ * or a pointer into the input buffer advanced by the amount written,
+ * not including the nul terminator guaranteed by this function.
+ */
+static char *utoa_upper(uint32_t value, char *const str, const uint8_t base)
+{
+	static const char digits[] = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+	static char buf[32] = {0};
+	size_t total = 0;
+
+	/* Check base is supported. */
+	if (base < 2 || base > 36) {
+		str[0] = '\0';
+		return NULL;
+	}
+
+	/* Push up to 32 symbols (base-2 aka binary, worst case) */
+	for (size_t offset = 0; offset < 32; ++offset) {
+		const div_t result = div(value, base);
+		const uint32_t remainder = result.rem;
+		buf[offset] = digits[remainder];
+		if (result.quot == 0) {
+			total = offset;
+			break;
+		}
+		value = result.quot;
+	}
+
+	/* Reverse local buffer into caller buffer */
+	for (size_t i = 0; i <= total; ++i)
+		str[i] = buf[total - i];
+
+	/* Terminate the result just in case */
+	str[++total] = '\0';
+
+	return str + total;
 }


src/platforms/common/stm32/serialno.c
```diff
@@ -64,3 +72,32 @@ void read_serial_number(void)
 #endif
 	serial_no[DFU_SERIAL_LENGTH - 1] = '\0';
 }
+
+static char *utoa_upper(unsigned value, char *str, int base)
+{
+	const char digits[] = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ";
```
dragonmux, line 78: This must be marked `static` to prevent its pointer being copied to stack every run.

src/platforms/common/stm32/serialno.c
```diff
@@ -64,3 +72,32 @@ void read_serial_number(void)
...
+
+	/* Check base is supported. */
+	if ((base < 2) || (base > 36)) {
```
dragonmux, line 81: Style nitpick, please drop the inner parens - `(base < 2 || base > 36)` - otherwise this turns off the GCC warning about accidental assignment

src/platforms/common/stm32/serialno.c
```diff
@@ -64,3 +72,32 @@ void read_serial_number(void)
...
+		str[0] = '\0';
+		return NULL;
+	}
+
+	/* Convert to string. Digits are in reverse order.  */
+	int l = 0;
```
dragonmux, line 87: `size_t` - it cannot go negative, so let's not represent it with a type that can and induce UB in the process. Likewise, please give this a more descriptive name, not `l`.

src/platforms/common/stm32/serialno.c
```diff
@@ -64,3 +72,32 @@ void read_serial_number(void)
...
+	do {
+		const unsigned remainder = value % base;
+		str[l++] = digits[remainder];
+		value = value / base;
+	} while (value != 0);
```
dragonmux, line 92: This looks like it's begging to be a for loop:
```suggestion
	for (offset = 0; offset < 7U; ++offset) {
		const div_t result = div(value, base);
		const uint32_t remainder = result.rem;
		str[offset] = digits[remainder];
		if (!result.quot)
			break;
		value = result.quot;
	}
```

(This suggestion switches to using [div](https://en.cppreference.com/w/c/numeric/math/div) which should guarantee a fused divide+mod, though it does throw the cat among the pigeons on type correctness and safety because it's specified only for int/long/etc, so if you'd prefer to keep the explicit `/` and `%` to stay unsigned, that's fine with us)

Part of the reason for this switch is so we can enforce bounds on the buffer, improving the safety of this code.

src/platforms/common/stm32/serialno.c
```diff
@@ -64,3 +72,32 @@ void read_serial_number(void)
...
+	str[l] = '\0';
+
+	/* Reverse string.  */
+	for (int j = 0, i = l - 1; j < i; j++, i--) {
```
dragonmux, line 96: Please give this loop counter a more descriptive name, not `i` and `j`.

Right now we cannot easily tell what this is doing and why it might be correct, and because it's using `int` this is subject to overflow UB if something goes wrong.

src/platforms/common/stm32/serialno.c
```diff
@@ -43,7 +45,13 @@ void read_serial_number(void)
 #elif defined(STM32L0) || defined(STM32F0) || defined(STM32F3)
 	int offset = 5;
 #endif
-	sprintf(serial_no, "%04X%04X%04X", uid[1] + uid[5], uid[0] + uid[4], uid[offset]);
+	static char buf[8];
+	utoa_upper(uid[1] + uid[5], buf, 16);
```
dragonmux, line 49: Given how this function works internally, perhaps move the static character buffer into the function, and have it take the serial number buffer at the appropriate offset for the next chunk - that eliminates the `strncpy()` (which itself is a lint item for complex reasons to do with its operation - in this use-case, we'd ask for `memcpy()` instead) and allows this to become something like:
```c
utoa_upper(uid[1] + uid[5], serial_no + 0, 16);
utoa_upper(uid[0] + uid[1], serial_no + 4, 16);
utoa_upper(uid[offset], serial_no + 8, 16);
```

Note that serial_no is already a pointer for the purposes of all this code due to array -> pointer decay, so the `&serial_no[0]` syntax is just convoluted - `serial_no + 0` does exactly the same thing but more obviously.

src/platforms/common/stm32/serialno.c
```diff
@@ -64,3 +72,32 @@ void read_serial_number(void)
 #endif
 	serial_no[DFU_SERIAL_LENGTH - 1] = '\0';
 }
+
+static char *utoa_upper(unsigned value, char *str, int base)
```
dragonmux, line 76: Please let's not use raw types and add some `const` in here - base should be unsigned as it cannot be negative, value should probably be `uint32_t`.

dragonmux (changes requested):
We like the idea of this PR, the execution needs some love though - please see our review notes below.

src/platforms/common/stm32/serialno.c
```diff
@@ -70,3 +70,40 @@ void read_serial_number(void)
 #endif
 	serial_no[DFU_SERIAL_LENGTH - 1] = '\0';
 }
+
+char *__utoa_upper(unsigned value, char *str, int base)
```
dragonmux: Given this is code outside the C runtime and implementation, `__` identifiers are expressly UB and forbidden - so, while we know why it's got those underscores, perhaps drop them in this first commit of the code so we don't accidentally break something later if any of these commits have to be bisected through or reverted?

ALTracer: Absolutely. All commits regarding introducing this function will be squashed and rebased into one. I just wanted to preserve development history so you can judge with having better visibility.

src/platforms/common/stm32/serialno.c
ALTracer: I could look at codegen to see the uldivmod call... it's a `udiv`+`mls` instruction pair. Maybe unsigned was used for indexing reasons, it's not like C has runtime rangechecking. If the number base conversion will run as fine with signed ints, then so be it.

src/platforms/common/stm32/serialno.c
ALTracer: Do you mean the implementation can initially convert into the fixed static buffer (with no malloc, which is good for embedded) and then reverse into callsite-provided nonnul *buf? I like this idea, even if we'd need to put some range bounds guardrail checks (for e.g. asking a base-2 of a uint32_t worst case the buffer has to be char[33])

src/platforms/common/stm32/serialno.c
ALTracer: This loop runs for half the string length iterations and swaps chars in palindrome fashion (in-place, no allocations) -- but with your other suggestion it can be dropped in favor of actually using more memory (for the static writable buffer).

src/platforms/common/stm32/serialno.c
ALTracer: What type should the function return -- a pointer to const char, i.e. `static const char *utoa_upper`? I remember that functions can't return const entities, and the compiler errors out on me trying to make it `char *const`. Other arguments updated. `value` is callee-clobbered, `base` is not -- but I only adjust the definition for elementary types, not the declaration/prototype (callers couldn't care less and will have to pass the element via regs anyways)

src/platforms/common/stm32/serialno.c
ALTracer: Yes, I agree, codegen looks horrible (entire array gets copied via mass ldmia!/stmia! and 4 regs from `.rodata`). Updated to `static`. Do you have any ideas on making it `utoa_lut` and feeding it different `digits[]` arrays to reuse the same algorithm?

src/platforms/common/stm32/serialno.c
ALTracer: Extra parens dropped, even if it's comparison, not assignment.

src/platforms/common/stm32/serialno.c
ALTracer: Renamed to `size_t offset` per next item. This index is not local to the loop, it is used as the upper bound in the next loop (reverser).

src/platforms/common/stm32/serialno.c
dragonmux: Yes, that's exactly what we mean 🙂 think it'll also improve the codegen of the reversal step too.

src/platforms/common/stm32/serialno.c
dragonmux: Because we're not using the return value, we could make it `void`, but given `str` will have to be `const char *`, we can return `char *` still if we're returning `str + offset;` to allow the result to be directly used to tell how many bytes were written and chain.

src/platforms/common/stm32/serialno.c
ALTracer: Yes, this works, and the `div()` function is `sdiv`+`mls` under the cover plus some stack manipulation.
```asm
div():
push	{r4}
sdiv	r4, r1, r2
mls	r1, r2, r4, r1
strd	r4, r1, [r0]
ldr.w	r4, [sp], #4
bx	lr
```
 I had to adjust for the offset being less by one now, but that's next on the menu. Tested live.

src/platforms/common/stm32/serialno.c
dragonmux: It might be better named `length` then as it bounds the length of that secondary iteration and defines how many characters are valid in the intermediary buffers?

src/platforms/common/stm32/serialno.c
ALTracer: Pinned a buffer to the function, which is okay for our purposes (but state is a no-no in libc functions, see Re-entrancy). Could be a VLA, but it's a bootloader, so nope.
Updated callsite per suggestion, eliminating both strncpy and outside buffer, tested live.

src/platforms/common/stm32/serialno.c
ALTracer: Rewritten the reversing part to read simpler, and now it has to do a full copy between different buffers instead of a half-copy in-place. No big deal, it's a short string in DFU descriptors context.

src/platforms/common/stm32/serialno.c
ALTracer: Made `offset` loop-local (for in-register optimization) and introduced `total` instead. Could be `length`.

src/platforms/common/stm32/serialno.c
ALTracer: Applied suggestion, now this function is chainable on consecutive conversions into an auto-advancing buffer pointer.

src/platforms/common/stm32/serialno.c
dragonmux: Unfortunately no particular good ideas spring to mind immediately, but if any do come to mind we'll let you know - for now, we'll resolve this thread as this looks good.

src/platforms/common/stm32/serialno.c
```diff
@@ -64,3 +68,44 @@ void read_serial_number(void)
...
+		return NULL;
+	}
+
+	/* Push up to 32 symbols (base-2 aka binary, worst case) */
+	for (size_t offset = 0; offset < 32; ++offset) {
+		const div_t result = div(value, base);
+		const uint32_t remainder = result.rem;
+		buf[offset] = digits[remainder];
+		if (result.quot == 0) {
+			total = offset;
```
dragonmux, line 97: If this is `offset + 1U`, then the nul terminations become `buf[total]` and `str[total]` and the loop below just `i < total`, - this should be a good optimisation.

src/platforms/common/stm32/serialno.c
dragonmux: `total` or `length` works - it's perhaps a little odd calling it `total` but "total length" makes sense and all that.

src/platforms/common/stm32/serialno.c
dragonmux: To be fair, the buffer could be non-static inside the function which eliminates the statefullness problem - at 33 bytes of stack (32 if we eliminate the extra nul termination reservation), it's not a huge deal.

src/platforms/common/stm32/serialno.c
dragonmux: `base` still needs changing from a raw type and signed (it cannot be negative, it's small - a good choice for `uint8_t` or another limited range unsigned type), but otherwise this looks much better.

src/platforms/common/stm32/serialno.c
```diff
@@ -64,3 +68,44 @@ void read_serial_number(void)
 #endif
 	serial_no[DFU_SERIAL_LENGTH - 1] = '\0';
 }
+
+/** @brief Convert an integer value into printable base-N representation.
```
dragonmux, line 72: Good idea to document what's what, though for consistency with the rest of the code, please let's not use (quasi-?)doxygen syntax. Likewise, `/*` on its own line.

dragonmux (commented):
Most of the review items have been addressed and this is looking better for it - there are a couple of things we've spotted this run through and think once those are addressed this should be about ready for merge if you could squash commits appropriately.

src/platforms/common/stm32/serialno.c
ALTracer: Sorry, my bad at saying char[33] is large. At bootloader scope -- suggestion declined. This firmware currently occupies ... (checks Puncover)
```
prog 	1,036
usbd_control_buffer 	1,024
usbd_dev 	276
if_string 	64
buf 	33
serial_no 	13
magic 	8
...
   8512      84    2404   11000    2af8 src/blackmagic_dfu
```
not much, and using measly 33 (rounded up for any alignment) bytes is insignificant in comparison to dangers of stack allocation. Even for a hypothetical F1 target ram usage of 20 KiB would be ~12%, and this actually targets F4 chips which have more. I would not want people to have doubts that this software component is stable or not. Is this critical?

src/platforms/common/stm32/serialno.c
ALTracer: This is roughly the style of libopencm3 on which this firmware relies on: `/** @brief` and multiline C++ down. I thought to document somehow that this function behaves slightly differently than `utoa()` would. Most of the block can be dropped except `@param[in] str` and return value description.

src/platforms/common/stm32/serialno.c
dragonmux: The problem we foresee is that the component is shared with the firmware were it's eating into the heap availability for something that's used once during startup (during setup of the USB descriptors) when, in both cases, it's called with a shallow call stack and something like 4KiB of stack availability. It's a fair comment to wonder if this might cause stack usage issues though 🤔

src/platforms/common/stm32/serialno.c
ALTracer: Traded three instances of +1 (and `s/<=/</`) with a slightly more convoluted indexing expression in second loop. That's what I tried to avoid. It was mentally easier to me to perceive these +1's as explicitly dealing with the 0-termination.

src/platforms/common/stm32/serialno.c
dragonmux: The suggestion is more to switch the style over to something like you'd find elsewhere in the code base, eg:
```c
/*
 * Converts an unsigned integer value into an equivilent base-N ASCII printable
 * representation. It takes as arguments (in order): the value to convert,
 * a pointer to a buffer large enough to take the result, and the base
 * to perform the conversion in between 2 and 36 (inclusive).
 * The result of this function is NULL on error (unsupported base), or a pointer
 * into the input buffer advanced by the amount written (not including the
 * nul terminator guaranteed by this function)
 */
 ```

File: src/platforms/common/stm32/serialno.c (hunk 9 above)
Comment (dragonmux) [reply src/platforms/common/stm32/serialno.c]: Ah, we hadn't spotted that `total - i` part.. 🤔 that does make it more difficult. This does make another good argument for dispensing with the nul termination in the first buffer though.. as much as we understand why you don't want to do that re debugging.

Comment (ALTracer) [reply src/platforms/common/stm32/serialno.c]: Re dropping nul terminator: I look on this as a string function, it quacks like a string function (inverse of atoi) -- you give it an integer/countable and a buffer, it produces a printable array of symbols aka a message string. So I expect someone to potentially try to misuse it elsewhere, and extra nuls might help protect someone's feet.
Makes me wonder why Dfu St Extensions string looks like a human-readable description of target flash (unrelated, we're dealing with USB serialno).

File: src/platforms/common/stm32/serialno.c (hunk 10 above)
Comment (ALTracer) [reply src/platforms/common/stm32/serialno.c]: Reflowed text per review item, throwing away Doxygen tags intended for IDE and HTML docs. Codespell says there's a stray "desig" in libopencm3 header filename, but nothing more.

File: src/platforms/common/stm32/serialno.c (hunk 9 above)
Comment (dragonmux) [reply src/platforms/common/stm32/serialno.c]: To be clear, we don't mean dropping the final nul termination, we mean dropping the intermediate buffer nul termination - `char buf[32]`, line 102 goes, that kind of thing.

Comment (ALTracer) [reply src/platforms/common/stm32/serialno.c]: Okay, that is minor, I could drop that line, sacrificing some internals' debuggability on multiple calls, but DFU doesn't make any (not that I know of).

Comment (ALTracer) [reply src/platforms/common/stm32/serialno.c:99]: Since that intermediate buffer is private / insignificant, I pulled a sneaky and partially reverted the logic around `total = offset` to mostly avoid any +1/-1. Tested live, the last preincrement is required.

File: src/platforms/common/stm32/serialno.c (hunk 6 above)
Comment (dragonmux) [reply src/platforms/common/stm32/serialno.c]: Given how the code now looks, the main ask here is to make `buf` 32 chars long as you've eliminated the need for the terminator - otherwise we'll take it as-is

File: src/platforms/common/stm32/serialno.c [hunk 11]
```diff
@@ -64,3 +68,46 @@ void read_serial_number(void)
...
+		buf[offset] = digits[remainder];
+		if (result.quot == 0) {
+			total = offset;
+			break;
+		}
+		value = result.quot;
+	}
+
+	/* Reverse local buffer into caller buffer */
+	for (size_t i = 0; i <= total; ++i) {
```
dragonmux, line 106: Braces can now be dropped

ALTracer: Dropped braces for a one-statement for-loop.

src/platforms/common/stm32/serialno.c
ALTracer: Resized to 32, works without changes in an external testcase and live on device (we don't ask to convert to binary anyway... yet)

dragonmux (dismissed):
This all looks good to us now - we're happy to merge this once the copyright and authorship notice issue is fixed as noted on Discord.

dragonmux (approved):

ALTracer:
Squashed intermediate progress into two distinct commits: one adding the function, one adjusting the callsite to use it (so no builds fail). Added a separate commit to refresh copyright info on the file.

Thank you for propelling this PR with your extensive review. We're one step closer to making BMPBootloader/F4 DFU nicer to use (and went through the practice of adding a helper function).

dragonmux:
Merging, thank you for the contribution!

PR #1717, opened 2024-01-07
Feature: Fancy DFU on `blackpill-f4`

### LINKED ISSUE #1516: blackpill-f4: bootloader LED issues

ALTracer:
## Detailed description

* This looks like a feature to end users but is just a collection of tweaks and ported code.
* The existing problem is an inefficient DFU configuration for `blackpill-f4` family of platforms (and probably other STM32F4 boards).
* This PR solves the problem by tightening DFU F4 delays, introducing an LED show/blinky in BMPBootloader (usbdfu), and provisioning it all to fit under 16 KiB for optimal reachable performance.

The first couple commits belong to the serialno PR this PR is based on (or blocks on). I fail to see what mechanism checks that the usbdfu binary fits under the allocated space (8 1-KiB sectors for F1, first 16 KiB sector for F4, etc.), but I noticed that flipping the DFU_SERIAL_LENGTH from 9 to 13 pulled sprintf and bloated the dfu binary past first sector. See that PR for details.

The `dfu_f4.c` changes are required and beneficial for BMPBootloader to work how I tested it to, but can be split into its separate PR if needed. I could submit similar delay tunings for `dfu_f1.c` I am using on F103CB boards.

The last commit is a general platform cleanup, but it can be amended or rebased away depending on situation.

Measurements for reference:
ST MaskROM DFU ~ 7.75 KiB/s (fine for flashing BMPBootloader)
BMPBootloader on `main` ~12 KiB/s (excessive delays)
BMPBootloader on this PR branch ~ 43 KiB/s (up to 54 KiB/s if fitting under 0x08020000 and tuning delays further)
4x reduced upload time (down to ~3 seconds) is worth it for developers reflashing their boards often.

## Closing issues

Closes #1516, or so I suppose (after a possible removal of `LED_BOOTLOADER` for good)

### DIFF
--- a/src/platforms/common/blackpill-f4/blackpill-f4.c
+++ b/src/platforms/common/blackpill-f4/blackpill-f4.c
@@ -45,10 +45,11 @@ volatile uint32_t magic[2] __attribute__((section(".noinit")));
 
 void platform_init(void)
 {
-	/* Enable GPIO peripherals */
+	/* Enable peripherals */
 	rcc_periph_clock_enable(RCC_GPIOA);
 	rcc_periph_clock_enable(RCC_GPIOC);
 	rcc_periph_clock_enable(RCC_GPIOB);
+	rcc_periph_clock_enable(RCC_CRC);
 
 #ifndef BMP_BOOTLOADER
 	/* Blackpill board has a floating button on PA0. Pull it up and use as active-low. */
@@ -73,26 +74,19 @@ void platform_init(void)
 #endif
 	rcc_clock_setup_pll(&rcc_hse_25mhz_3v3[PLATFORM_CLOCK_FREQ]);
 
-	/* Enable peripherals */
-	rcc_periph_clock_enable(RCC_OTGFS);
-	rcc_periph_clock_enable(RCC_CRC);
-
 	/* Set up DM/DP pins. PA9/PA10 are not routed to USB-C. */
 	gpio_mode_setup(GPIOA, GPIO_MODE_AF, GPIO_PUPD_NONE, GPIO11 | GPIO12);
 	gpio_set_af(GPIOA, GPIO_AF10, GPIO11 | GPIO12);
-
-	GPIOA_OSPEEDR &= 0x3c00000cU;
-	GPIOA_OSPEEDR |= 0x28000008U;
 
 	/* Set up TDI, TDO, TCK and TMS pins */
 	gpio_mode_setup(TDI_PORT, GPIO_MODE_OUTPUT, GPIO_PUPD_NONE, TDI_PIN);
 	gpio_mode_setup(TDO_PORT, GPIO_MODE_INPUT, GPIO_PUPD_NONE, TDO_PIN);
 	gpio_mode_setup(TCK_PORT, GPIO_MODE_OUTPUT, GPIO_PUPD_NONE, TCK_PIN);
 	gpio_mode_setup(TMS_PORT, GPIO_MODE_INPUT, GPIO_PUPD_NONE, TMS_PIN);
-	gpio_set_output_options(TDI_PORT, GPIO_OTYPE_PP, GPIO_OSPEED_2MHZ, TDI_PIN);
-	gpio_set_output_options(TDO_PORT, GPIO_OTYPE_PP, GPIO_OSPEED_2MHZ, TDO_PIN);
-	gpio_set_output_options(TCK_PORT, GPIO_OTYPE_PP, GPIO_OSPEED_2MHZ, TCK_PIN);
-	gpio_set_output_options(TMS_PORT, GPIO_OTYPE_PP, GPIO_OSPEED_2MHZ, TMS_PIN);
+	gpio_set_output_options(TDI_PORT, GPIO_OTYPE_PP, GPIO_OSPEED_25MHZ, TDI_PIN);
+	gpio_set_output_options(TDO_PORT, GPIO_OTYPE_PP, GPIO_OSPEED_25MHZ, TDO_PIN);
+	gpio_set_output_options(TCK_PORT, GPIO_OTYPE_PP, GPIO_OSPEED_25MHZ, TCK_PIN);
+	gpio_set_output_options(TMS_PORT, GPIO_OTYPE_PP, GPIO_OSPEED_25MHZ, TMS_PIN);
 
 	/* Set up LED pins */
 	gpio_mode_setup(LED_PORT, GPIO_MODE_OUTPUT, GPIO_PUPD_NONE, LED_IDLE_RUN | LED_ERROR | LED_BOOTLOADER);
@@ -114,6 +108,9 @@ void platform_init(void)
 	/* https://github.com/libopencm3/libopencm3/pull/1256#issuecomment-779424001 */
 	OTG_FS_GCCFG |= OTG_GCCFG_NOVBUSSENS | OTG_GCCFG_PWRDWN;
 	OTG_FS_GCCFG &= ~(OTG_GCCFG_VBUSBSEN | OTG_GCCFG_VBUSASEN);
+
+	/* By default, do not drive the SWD bus too fast. */
+	platform_max_frequency_set(3000000);
 }
 
 void platform_nrst_set_val(bool assert)
--- a/src/platforms/common/blackpill-f4/usbdfu.c
+++ b/src/platforms/common/blackpill-f4/usbdfu.c
@@ -31,6 +31,7 @@
 
 uintptr_t app_address = 0x08004000U;
 volatile uint32_t magic[2] __attribute__((section(".noinit")));
+static uint32_t dfu_activity_counter;
 
 static void sys_tick_init(void);
 
@@ -81,6 +82,12 @@ int main(void)
 
 void dfu_event(void)
 {
+	static bool idle_state = false;
+	/* Ask systick to pause blinking for 1 second */
+	dfu_activity_counter = 10U;
+	/* Toggle-blink it ourself */
+	SET_IDLE_STATE(idle_state);
+	idle_state = !idle_state;
 }
 
 static void sys_tick_init(void)
@@ -98,15 +105,19 @@ static void sys_tick_init(void)
 void sys_tick_handler(void)
 {
 	static uint32_t count = 0U;
+	if (dfu_activity_counter > 0U) {
+		dfu_activity_counter--;
+		return;
+	}
 	switch (count) {
 	case 0U:
-		/* Reload downcounter */
+		/* Reload downcounter and disable LED */
 		count = 10U;
 		SET_IDLE_STATE(false);
 		break;
 	case 1U:
 		count--;
-		/* Blink like a very slow PWM */
+		/* Enable LED for 1/10th of cycle */
 		SET_IDLE_STATE(true);
 		break;
 	default:
--- a/src/platforms/common/blackpill-f4/usbdfu.c
+++ b/src/platforms/common/blackpill-f4/usbdfu.c
@@ -42,6 +42,7 @@ void dfu_detach(void)
 
 int main(void)
 {
+	/* Enable GPIO peripherals */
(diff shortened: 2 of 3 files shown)


dragonmux:
If you could please rebase this PR now the other two precursors are merged, we'll get to try and review it and give it an honest look.

src/platforms/common/blackpill-f4/blackpill-f4.c
```diff
@@ -73,26 +74,19 @@ void platform_init(void)
 #endif
 	rcc_clock_setup_pll(&rcc_hse_25mhz_3v3[PLATFORM_CLOCK_FREQ]);
 
-	/* Enable peripherals */
-	rcc_periph_clock_enable(RCC_OTGFS);
```
dragonmux, line 77: We see this being dropped from here, but not added back elsewhere - why is this correct?

src/platforms/common/blackpill-f4/usbdfu.c
```diff
@@ -75,4 +80,46 @@ int main(void)
...
+	/* SYSTICK_IRQ with low priority */
+	nvic_set_priority(NVIC_SYSTICK_IRQ, 14U << 4U);
+	systick_interrupt_enable();
+	/* Start the heartbeat timer */
+	systick_counter_enable();
+}
+
+void sys_tick_handler(void)
+{
+	static int count = 0;
```
dragonmux, line 105: Given the switch-case below, please make this at least unsigned - preferably a fixed width type like `uint8_t` or something like `size_t`. This should give you better codegen and prevent any possible undefined things happening with those `count--;`.

src/platforms/common/blackpill-f4/usbdfu.c
```diff
@@ -75,4 +80,46 @@ int main(void)
...
+		return;
+	}
+	switch (count) {
+	case 0:
+		/* Reload downcounter */
+		count = 10;
+		SET_IDLE_STATE(false);
+		break;
+	case 1:
+		count--;
```
dragonmux, line 117: If you wanted to, you could eliminate this line and the break on this case, putting in `/* fallthrough */` in place of the break, wonder if that would decrease the footprint of this slightly?

src/platforms/common/stm32/dfu_f4.c
```diff
@@ -41,7 +41,29 @@ static uint32_t sector_addr[] = {
 	0,
 };
 
-static uint16_t sector_erase_time[] = {500, 500, 500, 500, 1100, 2600, 2600, 2600, 2600, 2600, 2600, 2600, 2600};
+/* Sector erase times in milliseconds, typ, for x32 parallelism at 2.7-3.6v */
+typedef enum erase_times_f4 {
+	ERASE_TIME_16KB = 250,   /*  500 * 0.5 */
+	ERASE_TIME_64KB = 550,   /* 1100 * 0.5 */
+	ERASE_TIME_128KB = 1000, /* 2000 * 0.5 */
```
dragonmux, line 48: This decreases the timing from 2600 to 2000 - what's the basis for this?

src/platforms/common/stm32/dfu_f4.c
```diff
@@ -87,8 +112,8 @@ uint32_t dfu_poll_timeout(uint8_t cmd, uint32_t addr, uint16_t blocknum)
 			return sector_erase_time[sector_num];
 	}
 
-	/* Programming 256 word with 100 us(max) per word*/
-	return 26U;
+	/* Programming 256 words (32-bit) with 16 us(typ), 100 us(max) per word */
+	return 16 * 1024 / 4 / 1000;
```
dragonmux, line 116: Please sprinkle in some `U` here so this is an unsigned division for an unsigned return and we don't tango with UB.

src/platforms/common/blackpill-f4/usbdfu.c
```diff
@@ -75,4 +80,46 @@ int main(void)
...
+	/* Blink it ourself */
+	SET_IDLE_STATE(idle_state);
+	idle_state = !idle_state;
+}
+
+static void sys_tick_init(void)
+{
+	/* Use SysTick at 10Hz to toggle-blink the blue LED at 5 Hz */
+	systick_set_clocksource(STK_CSR_CLKSOURCE_AHB_DIV8);
+	systick_set_reload(rcc_ahb_frequency / 8 / 10);
```
dragonmux, line 95: `U` on the constants please

src/platforms/common/blackpill-f4/usbdfu.c
```diff
@@ -30,6 +31,9 @@
 
 uintptr_t app_address = 0x08004000U;
 volatile uint32_t magic[2] __attribute__((section(".noinit")));
+static int dfu_activity_counter;
```
dragonmux, line 34: Please use a fixed width unsigned type like `uint8_t` or `size_t` here.

dragonmux (changes requested):
We've got a couple of queries and some stylistic items to address, but we think this is looking pretty decent and is close to being merge-able.

src/platforms/common/blackpill-f4/blackpill-f4.c
ALTracer: Sorry for not copying data out of the commit message so that it's visible to you.
* Drop `RCC_OTGFS` enabling because it happens in `stm32f107_usbd_init()`

https://github.com/libopencm3/libopencm3/blob/1f3abd43763fa39d23e737602b6d0011a45c70b2/lib/usb/usb_f107.c#L56
I am not asking the question why all the other F4/dwc platforms repeat the mantra. Maybe `libopencm3` didn't do this prior to 2016 is one possible reason. We're tracking current libopencm3.

dragonmux: Ahhhh, ok, that makes much more sense then - yeah, letting locm3 do its job and not repeating that in the platform makes total sense.

src/platforms/common/stm32/dfu_f4.c
ALTracer: * DS10314 STM32F411xC/xE Rev 7 pg92, Table 45. Flash memory programming
* DS8626 STM32F405xx/407xx Rev 9 pg108, Table 40. Flash memory programming
* DS9405 STM32F427xx/429xx Rev 10 pg129, Table 48. Flash memory programming

Should I hook the datasheets into the dfu_f4.c code comments?
2600ms is max t_ERASE128KB at PSIZE=x16, we're using x32, so 2000ms is more appropriate.

dragonmux: Yes, please - that's exactly the sort of information we've love to have in the source for this exact reason and question 🙂 - we'll mark this as resolved once comments for each one linking the DS number, revision, page number and table number are added much like in your reply are in the source for future reference.

src/platforms/common/blackpill-f4/usbdfu.c
ALTracer: Item accepted.
For this feature I was looking at your native hardware platform-specific code, namely https://github.com/blackmagic-debug/blackmagic/blob/main/src/platforms/native/usbdfu.c -- is that not good enough or outdated? I am often faced with review items on (mostly) copied code existing in adjacent platforms or drivers.

dragonmux: That's outdated and needs the same update for correctness done - that probably means all of the usbdfu.c's need the same check done on them.

src/platforms/common/blackpill-f4/usbdfu.c
ALTracer: This is a mentally taxing transformation, I just wanted to express intent (of changing the counter to next step) and let the compiler deal with what's necessary and possible. Thinking harder I see how this could work:
```c
	case 1:
		/* Blink like a very slow PWM */
		SET_IDLE_STATE(true);
		/* fallthrough */
	default:
		count--;
		break;
	}
```
Again, the first sector of F4 is 16 KiB, and the bootloader is currently slightly above 8 KiB (or slightly below when resorting to newlib-nano) because of all of the libopencm3 usb code etc. Space is not an issue here. I'd rather see (or at least test) BOS descriptors for DFU added, if they are available (and how much space do they take up).

dragonmux: Yeah, that's the transformation we were thinking of, but that's a fair point about additional mental burden - in that case, don't do it. This was very much a suggestion to talk about it, not something that has to be done.

Good questions about the MS OS Descriptors and BOS descriptor - we'd guess it should cause around 100 to 150 bytes of usage, but that'd be fascinating to see. Guessing you're planning on poking that in a follow-up PR?

src/platforms/common/stm32/dfu_f4.c
ALTracer: Listed datasheet numbers and ST automatic URLs after top block comment. Also listed platforms corresponding to them. That is f4discovery, all blackpill-f4, and stlinkv3. Others seem to not use it: 96b_carbon, hydrabus. These platforms may have a different flash IP: f072_if, f3_if.

src/platforms/common/blackpill-f4/usbdfu.c
ALTracer: No, I am not that well versed in USB descriptors, especially MS extensions. It would be nice to try to add them so that WINUSB gets assigned automatically to new builds of BMPBootloader, and compare that with the bmputil/libwdi workflow/UX. There are 150 lines of descriptor structs and macros in `usb_descriptors.h` for the payload part, including a WINUSB-compatible DFU runtime interface description.

dragonmux: Ah, fair - no worries. Well, we can look at poking it and we can run you through how it works but that'll be in a couple of days at the current rate of things. Agreed that it would be nice!

dragonmux (commented):
This all looks good to us now - if you can do final pass with squashing the fixup commits, we'll get this merged.

ALTracer:
Propagated the "unsigned type" and commentaries to relevant commits, it wasn't as simple as a fixup autosquash.
Please state what commit prefix should be used for files `src/common/stm32/dfu_f4.c`, `src/common/blackpill-f4/{blackpill-f4.c,usbdfu.c}`. I could make them consistent. Otherwise I think this is up-to date against `main`. Please check with clang-tidy or whatnot, I don't have that set up (which would have prevented me adding pre-standards code with lints etc.)

dragonmux:
For `dfu_f4.c`, it'd be `common/stm32/dfu_f4:` and for the Black Pill bits, `common/blackpill-f4:` or with the file name minus `.c` also included before the `:` depending on combined-or-separate commits for each file.

dragonmux (approved):
This all LGTM now, so merging. Thank you for the contribution!

PR #1719, opened 2024-01-08
target/stm32f1: fix split-bank erase

tlyu:
Add a missing `bank_offset` when checking for bank unlock completion.

## Detailed description

Discovered while testing #1709 on a GD32F303CGT6 target. It probably affects all split-bank STM32F1 parts or compatibles, as long as they have split flash banks.

The check for flash bank unlock completion failed for bank 2, because it was checking the register for bank 1. This probably went unnoticed for quite a while, because bank 2 is likely used for application data storage instead of code.

Only tested against GD32F303CGT6, which has split flash banks. Testing could only occur in conjunction with #1709, with which it merged cleanly. I don't have access to any other affected targets.

## Closing issues

N/A

### DIFF
--- a/src/target/stm32f1.c
+++ b/src/target/stm32f1.c
@@ -573,7 +573,7 @@ static bool stm32f1_flash_unlock(target_s *target, uint32_t bank_offset)
 {
 	target_mem_write32(target, FLASH_KEYR + bank_offset, KEY1);
 	target_mem_write32(target, FLASH_KEYR + bank_offset, KEY2);
-	uint32_t ctrl = target_mem_read32(target, FLASH_CR);
+	uint32_t ctrl = target_mem_read32(target, FLASH_CR + bank_offset);
 	if (ctrl & FLASH_CR_LOCK)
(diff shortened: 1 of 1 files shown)


dragonmux (approved):
LGTM, and based on reading TRMs and the various PR threads on it, this looks like the right fix. We'll merge this once ALTracer's PR is merged. Thank you for the contribution!

ALTracer:
Adding supporting evidence for this PR. The key to reproducing the behaviour was to shrink all loadable sections (via linkerscript) to target only bank 2.
An AT32F403ACGU7 target (as found on "WeActStudio.BlackPill", not to be confused with "MiniSTM32F4x1") used to trigger an `Error erasing flash with vFlashErase packet.` GDB message, because bank 1 had not been unlocked by the write into bank 2 key register or by prior operations (like loads). Or these BMDA diagnostic logs:
```
stm32f1_flash_erase: at 08080000
unlock failed, cr: 0x00000080
Erase failed at 8080000
```
So the Arterytek families of stm32f1-quacking flash targets were affected at least as far as #1685 .
Testing uploads on this commit confirms correct unlocking/checking now.

PR #1720, opened 2024-01-09
Feature: Support stlink v2 isol variant

HrMitrev:
## Detailed description
The [stlink-isol-v2](https://estore.st.com/en/st-link-v2-isol-cpn.html) variant differs from other stlink-v2 variants with how the SWDIO pin is multiplexed. This commit adds support for it. I have tested the commit on my STlink ISOL and it seems to work fine.

## Closing issues
N/A

### DIFF
--- a/src/platforms/stlink/platform.c
+++ b/src/platforms/stlink/platform.c
@@ -52,6 +52,19 @@ void platform_init(void)
 #ifdef BLUEPILL
 	led_idle_run = GPIO13;
 	nrst_pin = NRST_PIN_V1;
+#elif defined(STLINK_V2_ISOL)
+	led_idle_run = GPIO9;
+	nrst_pin = NRST_PIN_V2;
+	/* PB12 is SWDIO_IN */
+	gpio_set_mode(GPIOB, GPIO_MODE_INPUT, GPIO_CNF_INPUT_FLOAT, GPIO12);
+	/* PA4 is used to set SWDCLK floating when set to 1 */
+	gpio_set_mode(GPIOA, GPIO_MODE_OUTPUT_2_MHZ, GPIO_CNF_OUTPUT_PUSHPULL, GPIO4);
+	gpio_clear(GPIOA, GPIO4);
+	/* PA1 is used to set SWDIO floating and MUXED to SWDIO_IN when set to 1 */
+	gpio_set_mode(GPIOA, GPIO_MODE_OUTPUT_2_MHZ, GPIO_CNF_OUTPUT_PUSHPULL, GPIO1);
+#elif defined(STLINK_FORCE_CLONE)
+	led_idle_run = GPIO9;
+	nrst_pin = NRST_PIN_CLONE;
 #else
 	switch (rev) {
 	case 0:
@@ -69,7 +82,13 @@ void platform_init(void)
 	}
 #endif
 	/* Setup GPIO ports */
+#ifdef STLINK_V2_ISOL
+	/* In case of ISOL variant, this pin is never set to high impedance */
+	gpio_set_mode(TMS_PORT, GPIO_MODE_OUTPUT_2_MHZ, GPIO_CNF_OUTPUT_PUSHPULL, TMS_PIN);
+#else
+	/* In all other variants, this pin is initialized as high impedance */
 	gpio_set_mode(TMS_PORT, GPIO_MODE_OUTPUT_2_MHZ, GPIO_CNF_INPUT_FLOAT, TMS_PIN);
+#endif
 	gpio_set_mode(TCK_PORT, GPIO_MODE_OUTPUT_2_MHZ, GPIO_CNF_OUTPUT_PUSHPULL, TCK_PIN);
 	gpio_set_mode(TDI_PORT, GPIO_MODE_OUTPUT_2_MHZ, GPIO_CNF_OUTPUT_PUSHPULL, TDI_PIN);
 
--- a/src/platforms/stlink/platform.h
+++ b/src/platforms/stlink/platform.h
@@ -53,6 +53,11 @@ extern bool debug_bmp;
 #define SWDIO_PIN  TMS_PIN
 #define SWCLK_PIN  TCK_PIN
 
+#ifdef STLINK_V2_ISOL
+#define SWDIO_IN_PORT GPIOB
+#define SWDIO_IN_PIN  GPIO12
+#endif
+
 #define NRST_PORT      GPIOB
 #define NRST_PIN_V1    GPIO1
 #define NRST_PIN_V2    GPIO0
@@ -77,7 +82,22 @@ extern bool debug_bmp;
 #define SWD_CR      GPIO_CRH(SWDIO_PORT)
 #define SWD_CR_MULT (1U << ((14U - 8U) << 2U))
 
+#define SWDIODIR_ODR GPIO_ODR(GPIOA)
+
 #define TMS_SET_MODE() gpio_set_mode(TMS_PORT, GPIO_MODE_OUTPUT_2_MHZ, GPIO_CNF_OUTPUT_PUSHPULL, TMS_PIN);
+
+#ifdef STLINK_V2_ISOL
+/* The ISOL variant floats SWDIO with GPIO A1 */
+#define SWDIO_MODE_FLOAT()       \
+	do {                         \
+		SWDIODIR_ODR |= (GPIO1); \
+	} while (0)
+#define SWDIO_MODE_DRIVE()        \
+	do {                          \
+		SWDIODIR_ODR &= ~(GPIO1); \
+	} while (0)
+#else
+/* All other variants just set SWDIO_PIN to floating */
 #define SWDIO_MODE_FLOAT()           \
 	do {                             \
 		uint32_t cr = SWD_CR;        \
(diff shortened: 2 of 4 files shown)


src/platforms/stlink/stlink_common.c
```diff
@@ -45,6 +48,17 @@ uint32_t detect_rev(void)
 	rcc_periph_reset_pulse(RST_USB);
 	rcc_periph_clock_enable(RCC_AFIO);
 	rcc_periph_clock_enable(RCC_CRC);
+
+#if defined(STLINK_FORCE_CLONE)
+	/* PA12 is USB D+ pin */
+	gpio_clear(GPIOA, GPIO12);
+	gpio_set_mode(GPIOA, GPIO_MODE_OUTPUT_2_MHZ, GPIO_CNF_OUTPUT_OPENDRAIN, GPIO12);
```
dragonmux, line 55: This appears to be new pin setup code for clone models - could you explain briefly why this is required and correct?

src/platforms/stlink/platform.h
```diff
@@ -78,6 +83,19 @@ extern bool debug_bmp;
 #define SWD_CR_MULT (1U << ((14U - 8U) << 2U))
 
 #define TMS_SET_MODE() gpio_set_mode(TMS_PORT, GPIO_MODE_OUTPUT_2_MHZ, GPIO_CNF_OUTPUT_PUSHPULL, TMS_PIN);
+
+#ifdef STLINK_V2_ISOL
+/* The ISOL variant floats SWDIO with GPIO A1 */
+#define SWDIO_MODE_FLOAT()      \
+	do {                        \
+		gpio_set(GPIOA, GPIO1); \
```
dragonmux, line 91: Ideally this and the clear below would use the unpacked/inlined version of the code this calls, but this works. Just to clarify though - PA1 is the SWDIO direction setting pin? Want to be sure we've understood the purpose of this code.

dragonmux (commented):
We've got two queries that could do with being answered, and then this looks like it's good to merge.

src/platforms/stlink/stlink_common.c
HrMitrev: It is not new, it was executed before at the end of the function(the if statement based on rev) and talking to @ALTracer it is used to force USB re enumeration after reset from the bootloader. I decided to delete the code which is used for rev detection and moved these 2 lines there

src/platforms/stlink/platform.h
HrMitrev: Good point - I can inline the code

src/platforms/stlink/stlink_common.c
dragonmux: Ohhh, gotcha - ok, thank you for clarifying!

src/platforms/stlink/platform.c
```diff
@@ -51,6 +51,19 @@ void platform_init(void)
...
+#elif defined(STLINK_V2_ISOL)
+	led_idle_run = GPIO9;
+	nrst_pin = NRST_PIN_V2;
+	/* PB12 is SWDIO_IN */
+	gpio_set_mode(GPIOB, GPIO_MODE_INPUT, GPIO_CNF_INPUT_FLOAT, GPIO12);
+	/* PA4 is used to set SWDCLK floating when set to 1 */
+	gpio_set_mode(GPIOA, GPIO_MODE_OUTPUT_2_MHZ, GPIO_CNF_OUTPUT_PUSHPULL, GPIO4);
+	gpio_clear(GPIOA, GPIO4);
+	/* PA1 is used to set SWDIO floating and MUXED to SWDIO_IN when set to 1 */
+	gpio_set_mode(GPIOA, GPIO_MODE_OUTPUT_2_MHZ, GPIO_CNF_OUTPUT_PUSHPULL, GPIO1);
```
ALTracer, line 63: If you wanted to, you could implement feature parity with BMP v2.3 HW rev 6 w.r.t. floating the SWCLK pin, required to support stm32g0 in small pin-count packages. See Issue 945 and PR1192 (intentional not-a-mention), specifically commit https://github.com/blackmagic-debug/blackmagic/pull/1192/commits/a573c268d5cae2622c8bf659e608313c61d9801d for `native` platform.
PB12 is `T_SWDIO_IN` on **most/all** stlinkv2 schematics, and you made it a macro -- please substitute it into here (`SWDIO_IN_PIN`).
Accordingly, PA4 sounds like `TCK_DIR_PIN` and PA1 like `TMS_DIR_PIN`/`SWDIO_DIR_PIN`. This makes for less unused/magic pin numbers, even if you described their purpose in adjacent comments (nice!).

src/platforms/stlink/platform.c
```diff
@@ -68,7 +81,13 @@ void platform_init(void)
 	}
 #endif
 	/* Setup GPIO ports */
+#ifdef STLINK_V2_ISOL
+	/* In case of ISOL variant, this pin is never set to high impedance */
+	gpio_set_mode(TMS_PORT, GPIO_MODE_OUTPUT_2_MHZ, GPIO_CNF_OUTPUT_PUSHPULL, TMS_PIN);
```
ALTracer, line 86: I wonder if that's true -- PB14 being T_JTMS (effectively T_SWDIO_OUT) was toggled to become input across BMF previously. Do you think PB14 forcing high/low SWDIO in no way interferes with receiving data on PB12 across the isolation circuitry? Although I don't have the exact schematics and don't know where the 100R between SWDIO_IN & SWDIO_OUT resides for ISOL PCB. But you seem to have the means to test it with live targets.

src/platforms/stlink/platform.h
ALTracer: * Interesting when you put it that way. Does PA1 high *switch* the direction to input of levelshifter between T_SWDIO_OUT and PB14 coupler channel output, or *disconnect* it? I wonder if the shifter as input and coupler as output will fight (as the target drives its data against a fixed-level BMF TMS). In that case, consider also switching SWDIO_OUT to digital input mode. SWDIO_IN can and should stay.
* PA1 might be `TMS_DIR_PIN` (see the other comment). What about `TMS_SET_MODE()`? jtagtap will call it, is it identical in function to `SWDIO_MODE_DRIVE()`? I expect that after a SWD detach, your new platform may fail to drive JTAG TMS because of stuck direction pin. Please post logs with JTAG scan tests.

src/platforms/stlink/stlink_common.c
ALTracer: Indeed I gathered that this is a last-resort hack to present a long SE0 state on fixed 1.5k pullup boards to force re-enumeration. By the way, how do you expect to do it? Is PA15 USB_RENUM wired and available? Or this platform also has to perform the hack? (of inducing a babble (?) condition for lack of togglable pullup voltage source pin per USB 1.1/2.0 spec) Please confirm that the board works as BMF with ST_BOOTLOADER=1, unless you already wiped your stock bootloader with BMPBootloader/F1 (it may have the toggling because it calls `detect_rev()`)

ALTracer (commented):
There are macro naming questions from me and a proposal to bring feature parity with SWCLK Hi-Z, which leads into considerations about drive contention somewhere on the PCB. I cannot "Review -1/+1" as I am not a maintainer, and I am very unlikely to get my hands on an ISOL, but you could test anything before the PR is merged, and other less knowledgable users started (hypothetical worst case scenario) damaging their boards. 
I assume you already live-tested your PR branch build on a unit?

src/platforms/stlink/platform.h
HrMitrev: About the SWDIO multiplexing - the probe uses Si8420BB RF isolators which are fixed direction. If we ignore the isolators the schematic looks like this: 
image
PB14 and PA1 are always forced to output and PB12 is always forced to input because of the unidirectional RF isolators used to interface between the 2G125 IC and the STM32F01

SWDCLK should be similar, but I think PA4 is used to enable the output propagation or set it to high impedance

HrMitrev:
> Ideally this and the clear below would use the unpacked/inlined version of the code this calls, but this works.

I have unpacked/inlined the mode switch of SWDIO and confirmed to be working on my isol probe. Also I have rebased. Currently I have used this patch for more than 2 weeks every day and my probe seems to going strong. I haven't had the chanse to confirm it is working with JTAG instead of SWD. In the future if I have more time and HW to test JTAG I will revisit. For now I think this can be merged

Cheers!

dragonmux (approved):
LGTM, thank you for tweaking and sorting that. We'll get this merged once the builds complete - thank you for the contribution!

src/platforms/stlink/platform.h
HrMitrev: > About the SWDIO multiplexing - the probe uses Si8420BB RF isolators which are fixed direction. If we ignore the isolators the schematic looks like this: image PB14 and PA1 are always forced to output and PB12 is always forced to input because of the unidirectional RF isolators used to interface between the 2G125 IC and the STM32F01
> 
> SWDCLK should be similar, but I think PA4 is used to enable the output propagation or set it to high impedance

This was hidden after the changes, but may be a good idea to be easily accessible

HrMitrev:
> About the SWDIO multiplexing - the probe uses Si8420BB RF isolators which are fixed direction. If we ignore the isolators the schematic looks like this: image PB14 and PA1 are always forced to output and PB12 is always forced to input because of the unidirectional RF isolators used to interface between the 2G125 IC and the STM32F01
> 
> SWDCLK should be similar, but I think PA4 is used to enable the output propagation or set it to high impedance

This was hidden after the changes, but may be a good idea to be easily accessible

PR #1722, opened 2024-01-11
Feature: Faster BMP -> BMDA communications

dragonmux:
## Detailed description

After having a couple of conversations with ALTracer about the speed and performance of BMDA interacting with the firmware, we realised that the current 1-byte-at-a-time solution for reads, while correct, is extremely far from performant but there's a simple solution to the problem that doesn't sacrifice correctness - slipping a large buffer in between the reads and the reading.

This PR addresses this by adding a 4KiB buffer in between the probe reads and BMDA's platform_buffer_read() on both Linux/macOS and Windows. This significantly reduces the number of syscalls being made and streamlines the data transport through BMDA from the probe. A reasonable speedup on current firmware was observed with no drawbacks.

### DIFF
--- a/src/platforms/hosted/serial_unix.c
+++ b/src/platforms/hosted/serial_unix.c
@@ -32,7 +32,14 @@
 #include "utils.h"
 #include "cortexm.h"
 
-static int fd; /* File descriptor for connection to GDB remote */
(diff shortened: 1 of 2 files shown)


esden (approved):
LGTM

PR #1731, opened 2024-01-14
Fix: CMSIS-DAP packet size handling

dragonmux:
## Detailed description

This PR fixes the way the CMSIS-DAP code handles adaptor packet sizing and the differences between HID and Bulk endpoint (v1 and v2) based firmware.

Tested working against ORBTrace and picoprobe (which is based directly on the ARM reference implementation).

This finally solves the mystery too of how the packet size and buffer interact with one another and the HIDAPI code, documenting this properly.

### DIFF
--- a/src/platforms/hosted/bmp_libusb.c
+++ b/src/platforms/hosted/bmp_libusb.c
@@ -406,6 +406,11 @@ static void check_cmsis_interface_type(libusb_device *const device, bmda_probe_s
 			if (strstr(interface_string, "CMSIS") == NULL)
 				continue;
 
+			/* Scan through the endpoints finding the one with the narrowest max transfer length */
+			info->max_packet_length = UINT16_MAX;
+			for (uint8_t index = 0; index < descriptor->bNumEndpoints; ++index)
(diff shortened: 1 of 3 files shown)


src/platforms/hosted/bmp_libusb.c
```diff
@@ -413,6 +413,7 @@ static void check_cmsis_interface_type(libusb_device *const device, bmda_probe_s
 
 			/* Check if it's a CMSIS-DAP v2 interface */
 			if (descriptor->bInterfaceClass == 0xffU && descriptor->bNumEndpoints == 2U) {
+				break;
```
esden, line 416: I am not sure if I see it right but it seems to just skip the rest of the code block by breaking here.

esden (changes requested):
Just found one little oddness otherwise it looks good.

src/platforms/hosted/bmp_libusb.c
dragonmux: Hah, that's a rebase fail - good spot! We've removed the offending commit as that would have introduced a bug (never allowing CMSIS-DAP v2 adaptors).

esden (approved):
LGTM

PR #1732, opened 2024-01-14
Fix: semihosting BMDA correctness

dragonmux:
## Detailed description

This PR addresses the correctness of the semihosting implementation in BMDA and represents the last in the series.

We do this in two major steps - fixing the mechanism by which I/O with the probe is done for SYS_FLEN, SYS_TIME, and SYS_READC which solves issues with how GDB sees these transfers; and then fixing how and when BMDA inserts its own libc syscalls in to speed things up.

This also provides support for BMDA to do stdio redirection just like the firmware, so the user gets the expected behaviour by default (semihosting output and interaction in GDB), while then being able to punt that onto the controlling TTY for BMDA.

As best as we can tell, with this PR we now have a conforming semihosting implementation that works the same in both firmware and as BMDA and which meets the [specifications set forth by ARM](https://github.com/ARM-software/abi-aa/blob/main/semihosting/semihosting.rst).

### DIFF
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -631,36 +631,40 @@ static void handle_v_packet(char *packet, const size_t plen)
 		gdb_putpacketz("OK");
 
 	} else if (!strncmp(packet, "vRun", 4U)) {
-		/* Parse command line for get_cmdline semihosting call */
-		char cmdline[83];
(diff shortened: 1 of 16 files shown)


esden (approved):
LGTM! Fantastic cleanup and fixup of the whole subsystem!

PR #1734, opened 2024-01-14
target: properly fix weak aliasing of probe NOPs on macOS

amyspark:
## Detailed description
@dragonmux got a report on Discord that the weak aliasing of probe functions caused a linking error on macOS. I was able to reproduce the issue by building BMP once with `-Dprobe=''`, then reconfiguring with `-Dprobe='native'` and rebuilding.

Upon looking into the functions with Binary Ninja, I realised that the true issue was that AppleClang does not support aliases *at all*. Therefore, when saying "attribute weak" on the probe declaration, since the alias never takes effect, the linker would then expect a framework or library to supply the symbol [1].

The fix is to define a weak linkable pivot that will call the alias. If a strong definition appears later, the linker will emit just that definition. Otherwise, it'll emit an overridable pointer to the weak definition in the symbol table.

This also lets us restore the visibility and inlining of the NOP probes.

This partially reverts commit 5e98f30ff4d368c4c705deee32a250b5856fbb55.

[1]: https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/MachOTopics/1-Articles/executing_files.html#//apple_ref/doc/uid/TP40001829-98432-TPXREF120

### DIFF
--- a/src/target/target_probe.c
+++ b/src/target/target_probe.c
@@ -20,19 +20,34 @@
 #include "target_probe.h"
 
 #ifdef __APPLE__
-// https://bugs.llvm.org/show_bug.cgi?id=17775
-#define APPLE_STATIC
-// https://gcc.gnu.org/onlinedocs/cpp/Pragmas.html
-#define DO_PRAGMA_(x) _Pragma(#x)
-#define DO_PRAGMA(x)  DO_PRAGMA_(x)
-// __attribute__((alias)) is not supported in AppleClang.
-#define weak_alias(name, aliasname)     DO_PRAGMA(weak name = aliasname)
-#define CORTEXAR_PROBE_WEAK_NOP(name)   weak_alias(name, cortexar_probe_nop)
-#define CORTEXM_PROBE_WEAK_NOP(name)    weak_alias(name, cortexm_probe_nop)
-#define TARGET_PROBE_WEAK_NOP(name)     weak_alias(name, target_probe_nop)
-#define LPC55_DP_PREPARE_WEAK_NOP(name) weak_alias(name, lpc55_dp_prepare_nop)
+// __attribute__((alias)) is not supported in AppleClang, we need to define a
+// __attribute__((weak)) placeholder body that'll get pivoted by the linker.
+// if a "strong" definition is later available, the definition here will be
+// removed at linking time.
+// See:
+// - https://bugs.llvm.org/show_bug.cgi?id=17775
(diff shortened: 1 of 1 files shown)


tlyu:
Thanks! This seems to work for me, allowing BMDA to link without `renesas_rz.c`. I think while the revert of https://github.com/blackmagic-debug/blackmagic/commit/a547261d1d3c0c08868f9270f42764da1b8494cd is useful for testing the weak function fix, it isn't the right thing for BMDA, because BMDA is meant to support all available targets.

`llvm-objdump` confirms that the NOP stubs are exported as weak functions.

Interestingly, the NOP stubs almost all turn into jumps to the same local function:

```
_OUTLINED_FUNCTION_0:
      dc:       00 00 80 52     mov     w0, #0
      e0:       c0 03 5f d6     ret
```

which I guess is smaller code than inlining it for each stub.

amyspark (commented):
Updated the PR accordingly. Let me know if it looks ok with you @dragonmux.

src/target/target_probe.c
```diff
@@ -20,19 +20,34 @@
...
-#define LPC55_DP_PREPARE_WEAK_NOP(name) weak_alias(name, lpc55_dp_prepare_nop)
+// __attribute__((alias)) is not supported in AppleClang, we need to define a
+// __attribute__((weak)) placeholder body that'll get pivoted by the linker.
+// if a "strong" definition is later available, the definition here will be
+// removed at linking time.
+// See:
+// - https://bugs.llvm.org/show_bug.cgi?id=17775
+// - "coalesced weak reference" in https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/MachOTopics/1-Articles/executing_files.html#//apple_ref/doc/uid/TP40001829-98432-TPXREF120
+#define CORTEXAR_PROBE_WEAK_NOP(name)                                        \
+	__attribute__((weak)) bool name(adiv5_access_port_s *u, target_addr_t v) \
```
dragonmux, line 31: These should really maintain the proper parameter naming - `access_port` and `base_address` here, for example - to maintain code base style and not introduce more clang-tidy lints. Otherwise this looks great as a solution!

dragonmux (changes requested):
There's a naming convention item to sort, but with that taken care of, we're happy to merge this.

dragonmux (approved):
LGTM, and tlyu has confirmed it fixes their bug so we're happy to merge this. Thank you for the contribution!

PR #1743, opened 2024-01-21
Add MSVC compatibility for BMDA

amyspark:
## Detailed description

We've been discussing with @dragonmux, that one of the last outstanding bits for building BMDA on Windows was the MSVC compatibility.

Since there's a Meson build system now, that greatly reduces the barrier to use it, so I'm making this PR to add the necessary changes.

Some features added with this PR:

- HIDAPI is now Mesonified, so I've pulled the library as a wrap. Do notice that their build calls up to CMake, so that'll be another requirement (let me know if you agree and I should document it here)
- there is a typo on one of the uses of `__popcnt` under `_MSC_VER` which I fixed here
- I added Mesa's version of OpenBSD/NetBSD's `getopt` library for Windows (BSD-3-Clause / MIT licensed)
- Fixed UCRT detection under newer MinGWs
- Changed the place of definition of `vasprintf` since gdb_packet.c also requires it

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -90,6 +90,7 @@ endif
 # Project wide flags
 extended_warnings = [
 	'-Warith-conversion',
+	'-w24244', # 'conversion' conversion from 'type1' to 'type2', possible loss of data (integer version)
 	'-Wbad-function-cast',
 	# '-Wcast-align=strict',
 	'-Wcast-function-type',
@@ -98,13 +99,18 @@ extended_warnings = [
 	'-Wdangling-else',
 	'-Wdouble-promotion',
 	'-Wduplicated-branches',
+	'-w24754', # Conversion rules for arithmetic operations in a comparison mean that one branch cannot be executed.
 	'-Wfloat-conversion',
 	# '-Wformat-overflow=2',
 	# '-Wformat-signedness',
 	'-Wformat-truncation',
 	'-Wformat=2',
+	'-w24774', # ‘<function>’ : format string expected in argument <position> is not a string literal
+	'-w24777', #‘<function>’ : format string ‘<format-string>’ requires an argument of type ‘<type>’, but variadic argument <position> has type ‘<type>’ 
 	'-Wimplicit-fallthrough',
 	'-Wmaybe-uninitialized',
+	'-w24701', # Potentially uninitialized local variable 'name' used
+	'-w24703', # Potentially uninitialized local pointer variable 'name' used
 	'-Wmissing-attributes',
 	'-Wmissing-braces',
 	'-Wno-char-subscripts',
@@ -112,20 +118,25 @@ extended_warnings = [
 	# '-Wpacked',
 	'-Wredundant-decls',
 	'-Wreturn-type',
+	'-w24013', # 'function' undefined; assuming extern returning int
 	'-Wsequence-point',
 	'-Wshadow=local',
+	'-w24456', # declaration of 'identifier' hides previous local declaration
+	'-w24457', # declaration of 'identifier' hides function parameter
 	# '-Wsign-conversion',
 	# '-Wstack-protector',
 	'-Wstrict-aliasing',
 	'-Wstrict-overflow=3',
 	'-Wstring-compare',
 	'-Wstringop-overflow',
-	'-Wunknown-pragmas',
+	'-Wunknown-pragmas', # MSVC's C4081 and it's a level 1 warning
 	'-Wunsafe-loop-optimizations',
 	'-Wunsuffixed-float-constant',
 	'-Wunused-const-variable=2',
+	'-w24189', # 'identifier' : local variable is initialized but not referenced
 	'-Wunused-local-typedefs',
 	'-Wunused',
+	'-w24101', # 'identifier' : unreferenced local variable
 	'-Wvla-parameter',
 	'-Wvla',
 ]
--- a/src/include/gdb_packet.h
+++ b/src/include/gdb_packet.h
@@ -34,17 +34,25 @@
 #define GDB_PACKET_NOTIFICATION_START '%'
 #define GDB_PACKET_ESCAPE_XOR         (0x20U)
 
+#if defined(__MINGW32__) || defined(__MINGW64__) || defined(__CYGWIN__)
+#define GDB_FORMAT_ATTR __attribute__((format(__MINGW_PRINTF_FORMAT, 1, 2)))
+#elif defined(__GNUC__) || defined(__clang__)
+#define GDB_FORMAT_ATTR __attribute__((format(printf, 1, 2)))
+#else
+#define GDB_FORMAT_ATTR
+#endif
+
 void gdb_set_noackmode(bool enable);
 size_t gdb_getpacket(char *packet, size_t size);
 void gdb_putpacket(const char *packet, size_t size);
 void gdb_putpacket2(const char *packet1, size_t size1, const char *packet2, size_t size2);
 #define gdb_putpacketz(packet) gdb_putpacket((packet), strlen(packet))
-void gdb_putpacket_f(const char *packet, ...) __attribute__((format(printf, 1, 2)));
+void gdb_putpacket_f(const char *packet, ...) GDB_FORMAT_ATTR;
 void gdb_put_notification(const char *packet, size_t size);
 #define gdb_put_notificationz(packet) gdb_put_notification((packet), strlen(packet))
 
 void gdb_out(const char *buf);
 void gdb_voutf(const char *fmt, va_list);
-void gdb_outf(const char *fmt, ...) __attribute__((format(printf, 1, 2)));
+void gdb_outf(const char *fmt, ...) GDB_FORMAT_ATTR;
 
 #endif /* INCLUDE_GDB_PACKET_H */
--- a/src/include/general.h
+++ b/src/include/general.h
@@ -140,6 +140,21 @@ void debug_serial_send_stdout(const uint8_t *data, size_t len);
 #define strcasecmp  _stricmp
 #define strncasecmp _strnicmp
 
-#endif
+// FIXME: BMDA still uses this function in gdb_packet.c
+// It's defined here as an export from utils.c would pollute the ABI of libbmd
+static inline int vasprintf(char **strp, const char *const fmt, va_list ap)
+{
+	const int actual_size = vsnprintf(NULL, 0, fmt, ap);
+	if (actual_size < 0)
+		return -1;
+
+	*strp = malloc(actual_size + 1);
+	if (!*strp)
+		return -1;
+
+	return vsnprintf(*strp, actual_size + 1, fmt, ap);
+}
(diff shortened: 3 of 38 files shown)


src/platforms/hosted/meson.build
```diff
@@ -103,16 +108,23 @@ if build_machine.system() in ['windows', 'cygwin']
...
+	endif
 else
 	libftdi = dependency('libftdi1', required: false, method: 'pkg-config', native: is_cross_build)
 	bmda_sources += files('serial_unix.c')
 endif
 
 # Pick the appropriate HIDAPI depending on platform
 if build_machine.system() == 'linux'
-	hidapi = dependency('hidapi-hidraw', required: false, method: 'pkg-config', native: is_cross_build)
+	hidapi = dependency('hidapi-hidraw', required: false, method: 'pkg-config', fallback: 'hidapi', native: is_cross_build)
```
dragonmux, line 124: `required: false` here can be dropped given this always has the fallback now. The only reason for putting `required: false` in here at all was so BMDA not finding its deps wouldn't explode - hence why `libusb` having a fallback, also doesn't specify `required: false` as the aim is to get all deps to this state eventually.

src/include/timeofday.h
```diff
@@ -0,0 +1,68 @@
...
+#include <winsock.h>
+
+static inline int gettimeofday(PTIMEVAL current_time, void* ignoreme)
+{
+	(void)ignoreme;
+
+	if (current_time == NULL)
+		return FALSE;
+
+	LARGE_INTEGER pf, pc;
```
dragonmux, line 51: Might be worth expanding these to name them more clearly - `frequency` (`freq` is fine too) and `count`? These should also be initialised to something even though the query functions should then rewrite their values - as belt-and-braces insurance that we can't get use-before-init issues.

src/include/timeofday.h
```diff
@@ -0,0 +1,68 @@
...
+		return FALSE;
+
+	current_time->tv_sec = pc.QuadPart / pf.QuadPart;
+	current_time->tv_usec = (((pc.QuadPart % pf.QuadPart) * 1000000000 + (pf.QuadPart >> 1)) / pf.QuadPart) / 1000;
+	if (current_time->tv_usec >= 1000000) {
+		current_time->tv_sec++;
+		current_time->tv_usec -= 1000000;
+	}
+
+	return TRUE;
```
dragonmux, line 65: Clarity: is this `0` in the Windows headers, and `FALSE` something not-0?

src/include/general.h
```diff
@@ -128,4 +130,43 @@ void debug_serial_send_stdout(const uint8_t *data, size_t len);
...
+#endif
+
+#ifdef _MSC_VER
+#define strcasecmp _stricmp
+static inline int strncasecmp(const char* const a, const char* const b, size_t size)
+{
+	if (a == b)
+		return 0;
+
+	char *c = malloc(size * sizeof(char) + 1);
```
dragonmux, line 146: Is it possible to use `_strnicmp` here to avoid allocating? If not, please check the result of these allocations as this is liable to strncpy_s() to a `NULL` pointer otherwise.

meson.build
```diff
@@ -98,34 +99,44 @@ extended_warnings = [
 	'-Wdangling-else',
 	'-Wdouble-promotion',
 	'-Wduplicated-branches',
+	'-w24754', # Conversion rules for arithmetic operations in a comparison mean that one branch cannot be executed.
 	'-Wfloat-conversion',
 	# '-Wformat-overflow=2',
 	# '-Wformat-signedness',
 	'-Wformat-truncation',
 	'-Wformat=2',
+	'-w24774, # ‘<function>’ : format string expected in argument <position> is not a string literal '
```
dragonmux, line 108: The `'` at the end here wants moving to just before the `,` - this should fix the Meson build failures.

src/include/gdb_packet.h
```diff
@@ -25,6 +25,8 @@
 #include <stdarg.h>
 #include <stdbool.h>
 
+#include "debug.h"
```
dragonmux, line 28: This header is BMDA-only - we should probably put a clone of the printf-like format string logic into the general header to work around this, but customised to take the two parameter numbers explicitly so it can be applied later to things like tc_printf().

src/platforms/hosted/utils.c
```diff
@@ -26,27 +26,11 @@
...
 #include <string.h>
-#include <unistd.h>
-#include <sys/time.h>
 
+#include "timeofday.h"
 #include "timing.h"
 #include "bmp_hosted.h"
 
-#if defined(_WIN32) && !defined(__MINGW32__)
-int vasprintf(char **strp, const char *const fmt, va_list ap)
```
dragonmux, line 36: We're not opposed to this change, but would appreciate understanding the rationale and possibly having a comment in general.h as to why it's there please. (That is, we're interested in why moving the function rather than just making a decl of it available in general.h)

src/target/nxpke04.c
```diff
@@ -191,7 +191,7 @@ bool ke04_probe(target_s *t)
 	case SRSID_PIN_80: {
 		/* We have either a KE04Z64 or 128 */
 		/* Try to read a flash address not available in a Z64 */
-		volatile uint32_t __attribute__((unused)) dummy = target_mem_read32(t, 0x00010000U);
+		volatile uint32_t UNUSED dummy = target_mem_read32(t, 0x00010000U);
```
dragonmux, line 194: This macro needs defining in general.h so it's visible to this code please. It should probably be prefixed `BMD_` as in `BMD_UNUSED` for naming consistency.

src/platforms/hosted/ftdi_swd.c
```diff
@@ -75,6 +74,15 @@ bool ftdi_swd_possible(void)
 	return true;
 }
 
+#if defined(_MSC_VER) && !defined(__clang__)
+static inline uint32_t __builtin_ctz(uint32_t i)
```
dragonmux, line 78: `value` instead of `i`? We'd like not to introduce a new instance of the clang-tidy lint for too short a variable name. Please initialise `result` to 0 for use-before-init safety.

dragonmux (commented):
This is definitely looking like a great start. We do have some queries below and there are a few items to clean up a bit, but otherwise no complaints. Please make sure to format the commits using `clang-format` 16 or 17.

Regarding the CMake query - as long as it gets documented somewhere, given that is an external dependency and if the user installs the dev files for it that fallback build won't be triggered, think that's fine. We'll let Esden also weigh in on that but think he'll probably agree.

src/include/general.h
amyspark: It seems like it, I couldn't find this function previously. Thanks!

src/include/timeofday.h
amyspark: `0` if false, `1` if true -- see [minwindef.h](https://github.com/tpn/winsdk-10/blob/9b69fd26ac0c7d0b83d378dba01080e93349c2ed/Include/10.0.14393.0/shared/minwindef.h#L70-L76).

src/platforms/hosted/utils.c
amyspark: There are two uses of  `vasprintf` that seem to not have been tested on MSVC:

https://github.com/blackmagic-debug/blackmagic/blob/88b132edbe9f63da0c9f9754db98724632ac3b53/src/gdb_packet.c#L331

and 

https://github.com/blackmagic-debug/blackmagic/blob/88b132edbe9f63da0c9f9754db98724632ac3b53/src/gdb_packet.c#L353

Moving it is the only avenue because, as you explained in another thread, this affects the non-hosted part of BMP.

src/include/timeofday.h
dragonmux: Ah, ok - the reason for asking is that per [the man page for `gettimeofday()`](https://linux.die.net/man/2/gettimeofday), this needs to return 0 for success and -1 (non-zero is fine) for failure otherwise the logic in semihosting.c will think this failed when it didn't.

src/platforms/hosted/utils.c
dragonmux: Hmm, for non-BMDA (ie, firmware) we should be hitting [`vasiprintf`](https://github.com/mirror/newlib-cygwin/blob/fe2545e9faaf4bf9586f61a7b83d5cb5af501194/newlib/libc/stdio/vasprintf.c#L43) from newlib for this, which is why we were wondering - hence too why the function was only in BMDA's utilities impl. Hence too the guarding on the decl as this is also found in the Cygwin CRT (part of its POSIX compat layer).

amyspark: I meant BMDA under MSVC, not BMDA under MinGW or Cygwin.

src/include/timeofday.h
amyspark: I'll adjust the return values now.

src/platforms/hosted/utils.c
dragonmux: Ok, that helps with clarity - thank you. Guess that our main question then is why a decl like the following couldn't have been added to general.h to cover this case?:

```c
#if defined(_MSC_VER)
int vasprintf(char **strp, const char *const fmt, va_list ap);
#endif
```

We're interested in what makes this impractical as there feels like a comment or some documentation should be written given that.

amyspark: From a semantic point of view, such a declaration would be making the BMD library export a function that is defined in BMDA, a completely separate binary target.

dragonmux: Ahhh, ok - that makes a lot more sense then. In that case, please document that with the function's definition in general.h, noting that we think `#if defined(_WIN32) && !defined(__MINGW32__)` needs to be maintained so that it appears in all appropriate builds.

We would like to write the function out of the code base in the future but at least that'll prevent anyone from thinking the same way and making the same mistake as we nearly did here.

src/include/gdb_packet.h
```diff
@@ -34,17 +34,25 @@
 #define GDB_PACKET_NOTIFICATION_START '%'
 #define GDB_PACKET_ESCAPE_XOR         (0x20U)
 
+#if defined(__MINGW32__) || defined(__MINGW64__) || defined(__CYGWIN__)
+#define GDB_DEBUG_FORMAT_ATTR __attribute__((format(__MINGW_PRINTF_FORMAT, 1, 2)))
```
dragonmux, line 38: Given where this is used, the `_DEBUG` part can be dropped - the functions below aren't debugging functions but rather are used to tell the user about goings on via their GDB connection.

dragonmux (changes requested):
We've got a couple more files to review, but this is looking about ready to merge. We have one nomenclature note and unless we spot something glaringly wrong with one of the includes changes when we double-check those, this looks about ready for merge. Good work!

dragonmux (approved):
LGTM, we'll merge this once the builds complete. Thank you for the contribution!

PR #1746, opened 2024-01-24
stlink: replace odr register for the isol variant

HrMitrev:
## Detailed description
This is a follow up to [PR1720](https://github.com/blackmagic-debug/blackmagic/pull/1720) where I used the Output Data Register of the GPIO peripheral, causing read modify write. Now we use BSRR and BRR which only require a single write. I benchmarked the speed(when flashing 128kb firmware file) of the probe after this patch and it seems to be very similar, however the code size is reduced by a few bytes and as we know - every byte counts

## Closing issues
N/A


dragonmux (approved):
LGTM, merging - thank you for the contribution!

PR #1742, opened 2024-01-19
Fix: Clang fall-through warnings

dragonmux:
## Detailed description

This PR addresses the Clang warnings about switch-case fall-through caused by it not understanding the magic comment `/* fall through */`. To do this we've introduced a new macro and definition logic to use the compiler attribute for the job.

Tested on GCC 12, ARM GCC 12 and Clang 18. CI build items should be added in the future but as CI needs an overhaul post-Meson-PR we'll do that in a separate PR.

## Your checklist for this pull request

* [x] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md)
* [x] I've read the [guidelines for contributing](https://github.com/blackmagic-debug/blackmagic/blob/main/CONTRIBUTING.md) to this repository
* [x] It builds for hardware native (see [Building the firmware](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-firmware))
* [x] It builds as BMDA (see [Building the BMDA](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-app))
* [x] I've tested it to the best of my ability
* [x] My commit messages provide a useful short description of what the commits do

PR #1748, opened 2024-01-26
Add MSVC and macOS coverage

amyspark:
## Detailed description

This PR implements CI support for MSVC, Apple Xcode, and Homebrewed GCC for macOS.

For the latter two, dependencies are fetched from Homebrew as:

- `libftdi` depends on boost and others to be built from scratch
- GCC cannot build hidapi due to extensive usage of Objective-C constructs

This PR also includes the relevant fixes to make MSVC builds happy, in particular around standards compliance (#1742).

I'm open to any feedback. Let me know what you think!

### DIFF
--- a/src/target/riscv_debug.c
+++ b/src/target/riscv_debug.c
@@ -523,7 +523,7 @@ static uint32_t riscv_hart_discover_isa(riscv_hart_s *const hart)
 	do {
 		DEBUG_INFO("Attempting %u-bit read on misa\n", hart->access_width);
 		/* Try reading the register on the guessed width */
-		uint32_t isa_data[4] = {};
+		uint32_t isa_data[4] = {0U};
 		bool result = riscv_csr_read(hart, RV_ISA, isa_data);
 		if (result) {
 			if (hart->access_width == 128U)
--- a/src/target/semihosting.c
+++ b/src/target/semihosting.c
@@ -870,7 +870,7 @@ int32_t semihosting_request(target_s *const target, const uint32_t syscall, cons
 	target->tc->interrupted = false;
 
 	/* Set up the request block appropriately */
-	semihosting_s request = {r1, {}};
+	semihosting_s request = {r1, {0U}};
 	if (syscall != SEMIHOSTING_SYS_EXIT)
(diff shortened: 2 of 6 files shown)


meson.build
```diff
@@ -152,6 +152,28 @@ if is_cross_build
 	)
 endif
 
+# Ensure that MSVC is switched to standards compliant mode
```
dragonmux, line 155: Perhaps better to put this logic in `src/platforms/hosted/meson.build`? That way you can directly fix up `bmda_args` appropriately and use `cc`; It also puts the logic in with the rest of the Windows build logic.

src/target/riscv_debug.c
```diff
@@ -523,7 +523,7 @@ static uint32_t riscv_hart_discover_isa(riscv_hart_s *const hart)
 	do {
 		DEBUG_INFO("Attempting %u-bit read on misa\n", hart->access_width);
 		/* Try reading the register on the guessed width */
-		uint32_t isa_data[4] = {};
+		uint32_t isa_data[4] = {0, 0, 0, 0};
```
dragonmux, line 526: Ah, our bad! These are missing `U` on the end of each 0 to match the data type signedness, and you can drop all bar one leaving just `{0U}` if you want (C11 says the implied members are all 0's).

src/target/semihosting.c
```diff
@@ -870,7 +870,7 @@ int32_t semihosting_request(target_s *const target, const uint32_t syscall, cons
 	target->tc->interrupted = false;
 
 	/* Set up the request block appropriately */
-	semihosting_s request = {r1, {}};
+	semihosting_s request = {r1, {0, 0, 0, 0}};
```
dragonmux, line 873: Same comment as riscv_debug.c. The alternative patch (that works for us too) is:
```suggestion
	semihosting_s request = {r1};
```

dragonmux (changes requested):
Only got a couple of notes as this all looks pretty great - with them addressed, this looks good to merge.

amyspark (commented):
@dragonmux thanks for the comments! Bits should've been addressed now.

dragonmux (approved):

PR #1755, opened 2024-01-30
Fix `stm32h7_cmd_rev`

ALTracer:
## Detailed description

* The existing problem is incomplete revision printout in stm32h7 target driver.
* This PR starts fixing that by aligning part IDs with stm32h5 driver style.

There may be more changes needed, like using `ap->partno` or all-out reading of `DBGMCU_IDC` over whatever AP (and address alias) is appropriate.
I verified across 4 RMs that the documented revision hex values match to given letters on given chip families, so it's valid to attempt to match across entire array.

### DIFF
--- a/src/target/stm32h7.c
+++ b/src/target/stm32h7.c
@@ -133,9 +133,9 @@
 #define NUM_SECTOR_PER_BANK         8U
 #define FLASH_SECTOR_SIZE           0x20000U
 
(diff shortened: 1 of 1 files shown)


dragonmux (approved):
This LGTM. We've already talked about why the probe routine is retaining use of `target->part_id` rather than looking at `ap->partno`, so that's fine for now. Merging - thank you for the contribution!

PR #1756, opened 2024-01-30
Feature: STM32 DESig UID decoder

ALTracer:
## Detailed description

* This is a new feature expanding the existing stm32{h7,mp15} (potentially h5) UID-printing targe commands.
* The existing problems are plain hex in multiple O-packet and unused pattern/decode info.
* This PR solves these by buffering hex into a stack `char[]` (sprintf) and adding a small parser/decoder based on info from RMs.

TODO includes migrating sprintf usage to utoa (upper or lower) and testing the decoder on other families like STM32F4/F2/F7 which also include a Digital Electronic Signature containing Lot and Wafer info.

### DIFF
--- a/src/target/stm32_common.c
+++ b/src/target/stm32_common.c
@@ -1,0 +1,58 @@
+/*
+ * This file is part of the Black Magic Debug project.
+ *
+ * Copyright (C) 2023-2024 1BitSquared <info@1bitsquared.com>
+ * Written by ALTracer <tolstov_den@mail.ru>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "stm32_common.h"
+#include "buffer_utils.h"
+
+typedef struct stm32_uid_fields {
+	uint16_t wafer_xcoord;
+	uint16_t wafer_ycoord;
+	uint8_t wafer_number;
+	uint8_t lot_number[7];
+} stm32_uid_s;
+
+/*
+ * Print the Unique device ID.
+ * Can be reused for other STM32 devices with uid_base as parameter.
+ */
+bool stm32_uid(target_s *const target, const target_addr_t uid_base)
+{
+	char uid_hex[25] = {0};
+	uint8_t uid_bytes[12] = {0};
+	if (target_mem32_read(target, uid_bytes, uid_base, 12))
+		return false;
+
+	for (size_t i = 0; i < 12U; i += 4U) {
+		const uint32_t value = read_le4(uid_bytes, i);
+		//utoa_upper(value, &uid_hex[i * 2U], 16);
+		snprintf(uid_hex + i * 2U, 9, "%08" PRIX32, value);
+	}
+	tc_printf(target, "0x%s\n", uid_hex);
+
+	stm32_uid_s uid;
+	uid.wafer_xcoord = read_le2(uid_bytes, 0);
+	uid.wafer_ycoord = read_le2(uid_bytes, 2);
+	uid.wafer_number = uid_bytes[4];
+	memcpy(uid.lot_number, &uid_bytes[5], 7);
+
+	tc_printf(target, "Wafer coords X=%u, Y=%u, number %u; Lot number %.7s\n", uid.wafer_xcoord, uid.wafer_ycoord,
+		uid.wafer_number, uid.lot_number);
+	return true;
+}
--- a/src/target/stm32h7.c
+++ b/src/target/stm32h7.c
@@ -619,25 +619,12 @@ static uint32_t stm32h7_part_uid_addr(target_s *const target)
 	return 0x1ff1e800U;
 }
 
-/*
- * Print the Unique device ID.
- * Can be reused for other STM32 devices with uid as parameter.
- */
 static bool stm32h7_uid(target_s *target, int argc, const char **argv)
 {
 	(void)argc;
 	(void)argv;
-
 	const uint32_t uid_addr = stm32h7_part_uid_addr(target);
-
-	tc_printf(target, "0x");
-	for (size_t i = 0; i < 12U; i += 4U) {
-		const uint32_t value = target_mem32_read32(target, uid_addr + i);
-		tc_printf(target, "%02X%02X%02X%02X", (value >> 24U) & 0xffU, (value >> 16U) & 0xffU, (value >> 8U) & 0xffU,
-			value & 0xffU);
-	}
-	tc_printf(target, "\n");
-	return true;
+	return stm32_uid(target, uid_addr);
 }
 
 static bool stm32h7_crc_bank(target_s *target, uint32_t addr)
--- a/src/Makefile
+++ b/src/Makefile
@@ -91,6 +91,7 @@ SRC =              \
 	stm32l0.c      \
 	stm32l4.c      \
 	stm32g0.c      \
+	stm32_common.c \
 	renesas_ra.c   \
 	target.c       \
 	target_flash.c \
--- a/src/target/meson.build
+++ b/src/target/meson.build
@@ -252,6 +252,7 @@ target_stm = declare_dependency(
 		'stm32l0.c',
 		'stm32l4.c',
 		'stm32mp15.c',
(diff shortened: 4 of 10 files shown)


src/target/stm32h7.c
```diff
@@ -457,6 +457,18 @@ static bool stm32h7_mass_erase(target_s *target)
 	return stm32h7_check_bank(target, FPEC1_BASE) && stm32h7_check_bank(target, FPEC2_BASE);
 }
 
+/*
+ * RM0399 §64.1 (pg3510) and RM0433 §61.1 (pg3290)
+ * Unique device ID register
+ * Other ref.manuals omit the decoding details.
+ */
+struct __attribute__((packed)) stm32h7_uid {
```
dragonmux, line 465: This needs the `typedef`'ing treatment like the one in stm32mp15.c

src/target/stm32h7.c
```diff
@@ -473,15 +485,22 @@ static bool stm32h7_uid(target_s *target, int argc, const char **argv)
 	(void)argc;
 	(void)argv;
 
+	struct stm32h7_uid uid = {0};
```
dragonmux, line 488: This likewise wants moving down to the memcpy and the `= {0}` optionally dropped same as the stm32mp15.c code

dragonmux (changes requested):
This is definitely getting there.

To fix the windows builds, the packing attribute needs defining in general.h - `__attribute__((packed))` for GCC and Clang, and you'll have to deal with the [packing pragmas](https://learn.microsoft.com/en-us/cpp/preprocessor/pack?view=msvc-170) for MSVC.

ALTracer:
Revisiting. I dropped the duplicate changes in stm32h7.c and developed functionality in the stm32mp15.c driver just so I can be DRY (and stay in bounds of a familiar part). Backpropagated existing review items.
Rewritten the "decode" part to not rely on packed structs, and (subjectively) simplified the readout process.
Tested live on a MP15 board (DK1 STLink/V2-1 was not cooperating) to match both hex printout and decoded human-readable elements.

ALTracer:
Everything after first 4 commits is me moving the implementation into a common file and hooking (both existing and new) target commands to call it with a relevant UID base address (usually around OTP and F_SIZE). It might increase BMF size, but I believe it is done in an efficient enough manner.
Still, the "decoder" part can be disabled/suspended for stm32f1.c driver because the few chips I tested it on don't decode sensibly -- and the hex printout is still useful (for when you don't want to spin a gdb session just to grab 96 bits from a part-specific location outside of flash).
Passing by stm32l4 driver I replaced a couple `gdb_out()/gdb_outf()` with `tc_printf()` and removed `gdb_packet.h` which don't belong there (were used in mass-erase commands, and the proper way is O-packets).
TODO: needs rebasing on `main` and registering new TU in `meson.build`; optionally connecting with `utoa_upper()` of serialno.c, which is currently both static and platform-specific.

src/target/stm32_common.c
```diff
@@ -0,0 +1,62 @@
...
+{
+	char uid_hex[25] = {0};
+	uint8_t uid_bytes[12] = {0};
+	if (target_mem_read(target, uid_bytes, uid_base, 12))
+		return false;
+
+	for (size_t i = 0; i < 12U; i += 4U) {
+#if 0
+		const uint32_t value = read_le4(uid_bytes, i);
+		utoa_upper(value, &uid_hex[i], 16);
```
dragonmux, line 45: This looks so much nicer than the `snprintf()` monstrosity, please include making `utoa_upper()` a part of the hex utils in this PR and make use of it here!

src/target/stm32_common.c
```diff
@@ -0,0 +1,62 @@
...
+#endif
+		snprintf(uid_hex + i * 2U, 9, "%02" PRIX8 "%02" PRIX8 "%02" PRIX8 "%02" PRIX8, uid_bytes[i + 3U],
+			uid_bytes[i + 2U], uid_bytes[i + 1U], uid_bytes[i + 0U]);
+	}
+	tc_printf(target, "0x%s\n", uid_hex);
+
+	stm32_uid_s uid;
+	//	memcpy(&uid, values, 12U);
```
dragonmux, line 53: Don't forget to remove this having changed to using buffer utils and the per-member value extraction.

dragonmux (commented):
We like the new approach much better, nice work! Only have the two review notes on this which are just re-iterating the conversation we already had on Discord about `utoa_upper()` really. We look forward to getting to merge this greatly!

ALTracer:
Rebased on main, hooked to meson.build, no other changes yet -- let's see whether it builds first

src/target/stm32_common.c
ALTracer: The migration will also require touching bootloader-related build system parts, so I'd rather drop the lines from here and do a separate PR.

src/target/stm32_common.c
ALTracer: Removed. Not zero-initializing the stack-allocated struct, because I only ever use its fields, and not the undefined padding between them.

src/target/stm32_common.c
dragonmux: Given what this is, we encourage you to add hex_utils.c to the bootloader builds (CI will take care of verifying you didn't break anything) and use utoa_upper()..

src/target/stm32_common.c
dragonmux: Yep, that's all fine :D that actually would let you do this if you wanted:
```c
	stm32_uid_s uid = {
		.wafer_xcoord = read_le2(uid_bytes, 0),
		.wafer_ycoord = read_le2(uid_bytes, 2),
		.wafer_number = uid_bytes[4],
	};
	memcpy(uid.lot_number, &uid_bytes[5], 7);
```

dragonmux (dismissed):
We've already noted about it being nice to switch the `snprintf()` out for `utoa_upper()` and simplify, but we understand your hesitation to do that change in this PR, so.. if you could please rebase this on main, we'll get this merged as it looks pretty sane otherwise and all review items have been addressed (with thanks!).

ALTracer:
> rebase this on main

Looks like there has been major changes to `target_mem*_read*()` API (PR1793) and medium changes to stm32h7.c, stm32f4.c (cosmetic `t`->`target` rename from PR1814) -- which give me the merge conflicts locally. While I understand the `s/target_mem_read32/target_mem32_read32/` change as present in merged PR1819, which function should I use for read_bytes for UID readout -- `s/target_mem_read/target_mem32_read/`?
Another problem is I replaced %02PRIX32 with %02PRIX8 in the commit before the move to stm32_common, because of -Wformat, and it works for BMDA; however for BMF it generates a %02hhX, but newlib-nano only (not newlib full, I tested both) omits support for the C99 char specifiers %hh and long long %ll, resulting in `(gdb) monitor uid` giving a constant string with no values: `0x0hX0hX0h0hX0hX0h0hX0hX0h`. Revert to PRIX32 and casts? `utoa` would obsolete this, but I'm not about to involve bootloader regress-testing (and meson learning) just yet.

dragonmux:
> While I understand the s/target_mem_read32/target_mem32_read32/ change as present in merged PR1819, which function should I use for read_bytes for UID readout -- s/target_mem_read/target_mem32_read/?

Yes, the insertion of the 32 is the address width used - these are 32-bit parts, so they're `mem32`. If they were 64-bit, they'd be `mem64`. The read width meaning of the suffix when present has not changed.

> Another problem is I replaced %02PRIX32 with %02PRIX8 in the commit before the move to stm32_common, because of -Wformat, and it works for BMDA; however for BMF it generates a %02hhX, but newlib-nano only (not newlib full, I tested both) omits support for the C99 char specifiers %hh and long long %ll, resulting in (gdb) monitor uid giving a constant string with no values: 0x0hX0hX0h0hX0hX0h0hX0hX0h. Revert to PRIX32 and casts?

Because those are formatting strings for uint8_t's, just `%02X` is fine, which solves for that problem in a reasonable enough way. It's not ideal, but it is how things are done elsewhere in the code base. As noted, once replaced with `utoa_upper()`, the problem with the warning permanently goes away anyway, so that's more than fine.

ALTracer:
> Because those are formatting strings for uint8_t's, just `%02X` is fine, which solves for that problem in a reasonable enough way. It's not ideal, but it is how things are done elsewhere in the code base.

I wrote this in one of commit messages:
> As snprintf is Wformat-checked (tc_printf was not), replace %02X with %02PRIX32.

However, I can't get it to trigger the warning? Maybe it was %08X for uint32_t `value & 0xffU`, then I refactored the code to do %02X for uint8_t `uid_bytes[i]`, now I'm at %08X for uint32_t `value` as-is -- it's already processed by `read_le4()` now that buffer utils exist.
If only `void tc_printf(target_s *target, const char *fmt, ...)` was marked `__attribute__((format(printf, 2, 3)))`...

Rebased to current main.

dragonmux:
> Maybe it was %08X for uint32_t value & 0xffU, then I refactored the code to do %02X for uint8_t uid_bytes[i]

This tracks with what we remember, and because it's now an array of `uint8_t`, you can just `%02X` them without worry.

dragonmux (approved):
LGTM, merging once the builds complete.

PR #1757, opened 2024-01-30
Feature: Multi-AP handling on STM32MP15

ALTracer:
## Detailed description

* This is a completely new feature.
* There are existing problems on multi-AP chips like STM32H7, H5, MP15 regarding miscellaneous Debug features setup, some of which has to go through APB-D AP and not MCU's MEM-AP (AHB-AP).
* This PR solves some of these problems by introducing runtime management and hooks for AXI-AP AP0 and APB-D AP1 of MP15.

This was written as a PoC in December, before the cortexar STC/LDC DTRRX speedups got developed and tested. Using AXI-AP may provide background access to entire DDR3 and SYSRAM without halting either CA7 of the platform (or single CA7 on MP151).
I need in particular the APB-D to be able to setup the CoreSight SWO block in SoC system scope, which is shared between CA7 MPCore and CM4 and hence neither of them can write to its configuration registers page. The rest (ITM, DWT, RCC) are accessible in some sort and can be poked via Orbuculum's gdbtrace.init script.


src/target/stm32mp15.c
```diff
@@ -122,12 +176,61 @@ bool stm32mp15_cm4_probe(target_s *const target)
 }
 
 #ifdef ENABLE_CORTEXAR
+extern void cortexar_detach(target_s *target);
+
+/*
+ * Override memory r/w operations to go via the MEM-AP
```
dragonmux, line 182: They're all MEM-AP's, think you mean `AXI AP (AP0)`?

src/target/stm32mp15.c
```diff
@@ -96,6 +105,51 @@ static bool stm32mp15_ident(target_s *const target, const bool cortexm)
...
+	 */
+	uint32_t acpr = 0;
+	adiv5_mem_read(ap_apbd, &acpr, SWO_ACPR, 4);
+	acpr = 32;
+	adiv5_mem_write(ap_apbd, SWO_ACPR, &acpr, 4);
+
+	return true;
+}
+
+static void stm32mp15_setup_apbd_ap(target_s *const target)
```
dragonmux, line 134: Given how the setup functions are named (and we agree with this naming, though aren't sure where the `d` comes from in `apbd` here?), we suggest being absolutely consistent below - `apb_ap`, `axi_ap`, so the code's a little easier to follow. Either that or using the AP number instead of bus type for the naming - `ap0`, `ap1`? That one's up to you which you think would be more beneficial here.

src/target/stm32mp15.c
```diff
@@ -157,13 +260,20 @@ static bool stm32mp15_attach(target_s *const target)
 	/* Disable C-Sleep, C-Stop, C-Standby for debugging */
 	target_mem_write32(target, DBGMCU_CTRL, DBGMCU_CTRL_DBGSLEEP | DBGMCU_CTRL_DBGSTOP | DBGMCU_CTRL_DBGSTBY);
 
+	/* Reference the APB-D in target storage for 0xe0000000 region manipulations */
+	stm32mp15_setup_apbd_ap(target);
 	return true;
 }
 
 static void stm32mp15_detach(target_s *const target)
```
dragonmux, line 268: Given the Cortex-A attach/detach functions introduced above, we suggest these get refactored in name to include that they're specifically for the Cortex-M core - to help reduce code soupiness and confusion.

src/target/stm32mp15.c
```diff
@@ -62,16 +65,22 @@
...
 } stm32mp15_priv_s;
 
 static bool stm32mp15_uid(target_s *target, int argc, const char **argv);
 static bool stm32mp15_cmd_rev(target_s *target, int argc, const char **argv);
+static bool stm32mp15_cmd_swo(target_s *target, int argc, const char **argv);
 
 const command_s stm32mp15_cmd_list[] = {
 	{"uid", stm32mp15_uid, "Print unique device ID"},
 	{"revision", stm32mp15_cmd_rev, "Returns the Device ID and Revision"},
+	{"mp15_swo", stm32mp15_cmd_swo, "Set up SWO mode <1/2> and divisor <0x42>"},
```
dragonmux, line 83: Think it's fine to name this command just `swo` instead of `mp15_swo`, or `conf_swo` if you still want to prefix it.

dragonmux (changes requested):
This is looking pretty decent and we can mostly work out the rationale behind each change (even if it took us a minute). We have some suggestions to improve that and a request to write out a little more of the intent in comments to aid with that, but with those taken care of we think this will be good to merge.

src/target/stm32mp15.c
dragonmux: Correction to this - now that `mon swo` is taken by the in-probe SWO implementation, either `mp15_swo` or `conf_swo` - think the latter is a little more clear as to what it does, but both are acceptable.

ALTracer: Renamed to `conf_swo` in hopes that h7/h5 get a similar command. OTOH it is redundant if C5ACCE55 is used (possibly on PE-visible alias at 0x50083000). This is for orbuculum gdbtrace.init scripts, not BMF executable code (historically, at least)

src/target/stm32mp15.c
```diff
@@ -122,12 +176,61 @@ bool stm32mp15_cm4_probe(target_s *const target)
 }
 
 #ifdef ENABLE_CORTEXAR
+extern void cortexar_detach(target_s *target);
```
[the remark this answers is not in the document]
ALTracer: Introduced a commit touching cortexar.c/h to allow for this.

src/target/stm32mp15.c
ALTracer: They're not CM4-specific, SoC SWO the peripheral is shared; but I don't see it being useful for CA7 (yet) and I wrote this PR such that it reuses target_storage for AXI-AP, so those are exclusive. Renaming.

dragonmux: The statement that it's core specific comes from it calling `cortexm_` functions directly. It would require extra work to make it generic to both cores (probably in the form of having the core common code in one function and having a `stm32mp15_cm4_detach()` and a `stm32mp15_ca7_detach()` that call that common routine and the core-type-specific detach routines. Thank you for doing the rename.

src/target/stm32mp15.c
```diff
@@ -118,6 +126,51 @@ static bool stm32mp15_ident(target_s *const target, const bool cortexm)
 	return true;
 }
 
+static bool stm32mp15_cmd_swo(target_s *target, int argc, const char **argv)
+{
+	(void)argc;
+	(void)argv;
+	/* TODO: argv parsing for mode and baudrate */
+	adiv5_access_port_s *const ap_apbd = (adiv5_access_port_s *)target->target_storage;
```
dragonmux, line 134: As with the other places this `apbd` naming has come up, we do not know where this `d` is coming from in the naming - please either name by the AP number this AP represents, or just say `ap_apb` - if it's supposed to mean "debug", that's already implied by it being the bus attached to the AP, but those busses aren't some special thing.. they are  normal APB's and AHB's, so including the `d` in the nomenclature just muddies things in our opinion. Readers not familiar with things would try searching for an `APBD` and get no results which is counter-productive, for example.

PR #1759, opened 2024-01-31
riscv: make the debugger stop on ebreak

mean00:
riscv: make the debugger stop on ebreak

## Detailed description

This is a temptative bugfix for the (*)ebreak statement not working with the blackmagic probe
Without the modification, the ebreak is completely ignored.
This is problematic for flashstub and for chips that do not have triggers at all  (e.g. ch32v20x and ch32v00x)
The patch just enables the ebreak stop in the debug register. Not sure it is the righ way to do it but it works for me.

* [x ] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md)
* [ x] I've read the [guidelines for contributing](https://github.com/blackmagic-debug/blackmagic/blob/main/CONTRIBUTING.md) to this repository
* [ out of flash but not related to my patch ] It builds for hardware native (see [Building the firmware](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-firmware))
* [x ] It builds as BMDA (see [Building the BMDA](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-app))
I'm using a custom blackmagic build which has native support for the CH32Vxxx chips. Cannot test that with normal BMP at the moment. It works with my build.

### DIFF
--- a/src/target/riscv_debug.c
+++ b/src/target/riscv_debug.c
@@ -107,9 +107,10 @@
 #define RV_VENDOR_JEP106_CONT_MASK 0x7fffff80U
 #define RV_VENDOR_JEP106_CODE_MASK 0x7fU
 
-#define RV_DCSR_STEP       0x00000004U
-#define RV_DCSR_CAUSE_MASK 0x000001c0U
-#define RV_DCSR_STEPIE     0x00000800U
+#define RV_DCSR_STEP           0x00000004U
+#define RV_DCSR_CAUSE_MASK     0x000001c0U
+#define RV_DCSR_STEPIE         0x00000800U
+#define RV_DCSR_EBREAK_MACHINE 0x00008000U
 
 #define RV_GPRS_COUNT 32U
 
@@ -832,8 +833,10 @@ static void riscv_halt_resume(target_s *target, const bool step)
 		return;
 	if (step)
 		stepping_config |= RV_DCSR_STEP | RV_DCSR_STEPIE;
-	else
+	else {
 		stepping_config &= ~(RV_DCSR_STEP | RV_DCSR_STEPIE);
+		stepping_config |= RV_DCSR_EBREAK_MACHINE;
+	}
 	if (!riscv_csr_write(hart, RV_DCSR | RV_CSR_FORCE_32_BIT, &stepping_config))
 		return;
 	/* Request the hart to resume */


mean00:
As a sidenote, to test it just add that line of code
__asm__("ebreak");
without the patch it does nothing
with the patch the debugger stops

src/target/riscv_debug.c
```diff
@@ -834,6 +835,7 @@ static void riscv_halt_resume(target_s *target, const bool step)
 		stepping_config |= RV_DCSR_STEP | RV_DCSR_STEPIE;
 	else
 		stepping_config &= ~(RV_DCSR_STEP | RV_DCSR_STEPIE);
+	stepping_config |= RV_DCSR_EBREAKM;
```
dragonmux, line 838: This looks almost correct - not sure it's semantically correct to be enabled when we're doing single stepping though so it might need pushing into the else statement here as in:
```c
	else {
		stepping_config &= ~(RV_DCSR_STEP | RV_DCSR_STEPIE);
		stepping_config |= RV_DCSR_EBREAKM;
	}
```

We'll go read the debug spec to double check this though and re-review accordingly.

dragonmux: Ok, having read the specs - this looks good, though we stand by our previous observation around single stepping. Perhaps the only small change we'd make is to rename the macro from `EBREAKM` to `EBREAK_MACHINE` to indicate what the `m` means - or at least `EBREAK_M` to indicate it's intentional. The reason for this is we should probably also definedn `RV_DCSR_EBREAK_SUPERVISOR` and `RV_DCSR_EBREAK_USER` macros for the other two modes this can be done for.

dragonmux (changes requested):

src/target/riscv_debug.c
mean00: Thank you
Normally i've taken the remarks into account & squash the commits.

next are flashstub for riscv & wch chip supports.

dragonmux (approved):

Unsure what you mean about the flashstub bit - but we'll be interested to see what you've come up with, noting that most WCH devices need their custom debug protocol transports which there is a draft PR for from Perigoso. It would definitely be surprising if a Flash stub was actually required to reprogram their devices - but we're prepared to be surprised!

mean00:
Thanks!
Let me detail a bit : 
I've written the WCH protocol continuing what Perigoso has done + MCU support for CH32V2xx and CH32V3xx.
So using just a blackmagic "style" probe  works fine  but is a bit slow (like 1kB /s writing speed).

Using a flashstub doubles the speed. It is  still slow but perfectly usable. I get about 4kB/s average flash writing speed with a RP2040 probe.

For info, i've looked into the WCH link blob  out of curiosity and it is definitively a flashstub. I've found strong similarities with my own.

I'll submit MRs for that, i need to port them from the mix of rust & c++ i'm using to C.

PR #1764, opened 2024-02-08
riscv: add flashstub framework and helper functions (riscv32 only)

mean00:
## Detailed description

Hi
This is a temptative patch to add some flashstub helper functions for the rv32 targets.
It tries to be very similar to the cortexm version. i.e. 
Call _riscv32_run_stub( target, loadaddr, param1, param2, param3, param4)_
and the flashstub must end up with  _riscv_stub_exit(errorcode)_

The main difference in behavior is the code saves and restore PC & MIE to avoid letting things dangling in ram.
Not completely sure this is needed, just to be on the safe side.

I've tested it with a CH32V307 board using the upcoming board & rvswd protocol support with a ~2x speedup.

As a sidenote and for the record, my original version was providing a temporary stack for the stub to avoid strict compiler rules about not using stack. Dont know if it is a valid idea. Nonetheless, this MR is assuming the stack is not used at all by the stub.

Thank you.

* [X ] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md)
* [X ] I've read the [guidelines for contributing](https://github.com/blackmagic-debug/blackmagic/blob/main/CONTRIBUTING.md) to this repository
* [Builds till linking then out of flash  ] It builds for hardware native (see [Building the firmware](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-firmware))
* [X ] It builds as BMDA (see [Building the BMDA](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-app))
Tested with a BMP derivative using CH32V3xx chips. Cant test (yet!) with vanilla bmp


dragonmux:
We'll dive in and review this tomorrow, but as for the stack thing - the stub *can* use stack, it just has to set one up itself - see the RP2040 stub in #1609 which does this to improve code density and size. It's not really viable for the stub runner to do this itself, and with the `lui` instruction in RISC-V, it would be trivial to load a 4KiB-aligned stack pointer in the stub.

mean00:
Hi,
This MR exhibits an issue related to the CSR register management introduced earlier (my bad) because it manipulates some CSR directly.
The CSR registers are offset by RV_CSR_GDB_OFFSET to present them as regular register to gdb 
It means that even internall, something like
   _**t->reg_write(t, RV_CSR_MIE, &zero, 4);**_
should actually be 
  **_t->reg_write(t, RV_CSR_MIE+RV_CSR_GDB_OFFSET, &zero, 4);_**

This looks ugly and a bit error prone. Not sure what is the best way to deal with that.

dragonmux:
Given that going through the target API there is just a convoluted way to access riscv_csr_read/riscv_csr_write - we would suggest that the logic use these functions directly, solving the need to deal with the GDB offset and specify the register width that way (you can use `RV_CSR_FORCE_32_BIT` with `RV_CSR_MIE` to ensure the right width access will always happen, or just assume because it's a RISC-V 32 part that it's the right width)

mean00:
Indeed, thank you for the pointer, that's way better.

src/target/riscv32.c
```diff
@@ -666,3 +666,92 @@ static int riscv32_breakwatch_clear(target_s *const target, breakwatch_s *const
...
+    while (reason == TARGET_HALT_RUNNING)
+    {
+        if (platform_timeout_is_expired(&timeout))
+        {
+            goto the_end;
+        }
+        reason = t->halt_poll(t, NULL);
+    }
+
+    if (reason == TARGET_HALT_ERROR)
```
dragonmux, line 734: This condition and the next look merge-able, and more importantly, look to be able to be simplified such that it removes the gotos (we'd *really* like not to introduce any, they make for horrible codegen and a horrid time understanding code flow.

For example:
```c
if (reason == TARGET_HALT_REQUEST)
   ret = true;
target->halt_request(target);
```

Now, we have questions about `ret` as well and its (re)use here as that logic doesn't make sense right now, but that should also simplify the while loop too so as to just need a `break;` statement from that check for timeout expiry.

Please write comments detailing your intent with this code - it's very difficult to tell from the plain text of this code what you are intending to have happen here and why, vs what this actually does.

src/target/riscv32.c
```diff
@@ -666,3 +666,92 @@ static int riscv32_breakwatch_clear(target_s *const target, breakwatch_s *const
...
+        uint32_t a0;
+        t->reg_read(t, RISCV_REG_A0, &a0, 4);
+        ret = (a0 == 0);
+    }
+    // restore PC & MIE
+    riscv32_target_csr_write(t, RV_CSR_MIE, mie); // put back MIE
+    t->reg_write(t, RISCV_REG_PC, &pc, 4);
+    return ret;
+}
+//----
```
dragonmux, line 757: Please remove this trailing comment.

src/target/riscv_debug.h
```diff
@@ -234,4 +234,15 @@ uint32_t riscv32_pack_data(const void *src, uint8_t access_width);
 void riscv32_mem_read(target_s *target, void *dest, target_addr_t src, size_t len);
 void riscv32_mem_write(target_s *target, target_addr_t dest, const void *src, size_t len);
 
+/*
```
dragonmux, line 237: Same comment here about comment style as in riscv32.c

src/target/riscv32.c
```diff
@@ -666,3 +666,92 @@ static int riscv32_breakwatch_clear(target_s *const target, breakwatch_s *const
...
+#define RISCV_REG_A2 12
+#define RISCV_REG_A3 13
+#define RISCV_REG_PC 32
+#define RISCV_REG_SP 2
+
+/*
+    Small helper function to translate target to hart and simplify parameters
+    We assume it is 32 bits
+*/
+static bool riscv32_target_csr_write(target_s *target, const uint16_t reg, uint32_t val)
```
dragonmux, line 681: Given how this gets used and how small a function this is, but with it not marked inline, we expect this to be contributing a significant amount to the codegen for riscv32_run_stub() - if you grab the hart at the top of that function, you can then call the underlying read and write routines directly, which should result in leaner code generation. Same with the read function.

src/target/riscv32.c
```diff
@@ -666,3 +666,92 @@ static int riscv32_breakwatch_clear(target_s *const target, breakwatch_s *const
 	const uint32_t address = 0;
 	return riscv_config_trigger(hart, breakwatch->reserved[0], RISCV_TRIGGER_MODE_UNUSED, &config, &address) ? 0 : -1;
 }
+
+#define RISCV_REG_A0 10
+#define RISCV_REG_A1 11
+#define RISCV_REG_A2 12
+#define RISCV_REG_A3 13
+#define RISCV_REG_PC 32
+#define RISCV_REG_SP 2
```
dragonmux, line 675: These need suffixing with `U` to make them unsigned. It looks like these should really belong in the riscv_debug.h header so all RISC-V debug code can benefit from them.

src/target/riscv32.c
```diff
@@ -666,3 +666,92 @@ static int riscv32_breakwatch_clear(target_s *const target, breakwatch_s *const
...
+    t->reg_write(t, RISCV_REG_PC, &codeexec, 4);
+
+    target_halt_reason_e reason = TARGET_HALT_RUNNING;
+    t->halt_resume(t, false); // go!
+    platform_timeout_s timeout;
+    platform_timeout_set(&timeout, 10000);
+    while (reason == TARGET_HALT_RUNNING)
+    {
+        if (platform_timeout_is_expired(&timeout))
+        {
```
dragonmux, line 728: The braces here can be (should be) removed. Please also run clang-format as this is not formatted properly for this code base.

dragonmux (commented):
This is looking good, though there are some items to address and talk about before this can be considered ready for merge. Most of the items are about formatting and code style though.

mean00:
I think i've taken all or about all of the comments into account.

I've left the two helper functions as inline to have a simpler code. I guess they might be helpful later on.
The generated codesize should be the same.

Similarily i've let the error handling a bit detailed as to allow further more detailed use cases. It should not cause binary bloat.


( On a side note, clang-tidy 17 does not like much the .clang-tidy due to varying indentation and after fixing it complains really a lot  )

dragonmux (commented):
This code is looking significantly better - good work! We do have a few small items found in re-reviewing, but with those fixed we'll be very happy to get this merged.

PR #1766, opened 2024-02-15
riscv : add basic single precision hw float support

### LINKED ISSUE #1765: riscv: missing FPU register declaration makes gdb flatly reject the file
Hi,
With a  RV32 binary generated using hw fp ( such as available on the ch32v3xxx), the resulting binary is flatly rejected by blackmagic  + gdb (tried several versions including gdb 13.2)

It seems the underlying reason is that gdb is trying to find the relevant registers  in the xml description, does not find them, and then let the descriptor incomplete and full of null pointers. 

As a result it sort of asserts with a message  along  "bfd requires flen 4, but target has flen 0"

I have an incomplete patch, which declares the relevant registers (in single precision) if the single fp extension is there, so that gdb is happy and loads the file.

It is incomplete, accessing the FPU registers ends up with warnings/errors (no crash though)

I dont know if submitting that incomplete patch is better than nothing or if you prefer to wait till someone (may take some times in my case ) provides a fully operational patch.

Thank you

mean00:
## Detailed description

* This adds basic support for hw single precision fpu (rv32)
* Without the patch gdb will reject the binary as the register list is incomplete
* This is not exactly the same as the patch posted in the bug report, it actually connects the registers
* With that patch on a CH32V307 , i can load the binary and print the registers with a value that looks correct (limited testing), seems ok for fs0, f1, fs2 , fs3
* Tested only on rv32
* May lack some CSR relatated to FPU declared as CSR , not clear.

* [up to linking then out of rom ] It builds for hardware native (see [Building the firmware](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-firmware))

## Closing issues
fixes #1765

### DIFF
--- a/src/target/riscv32.c
+++ b/src/target/riscv32.c
@@ -153,6 +153,8 @@ static ssize_t riscv32_reg_read(target_s *target, const uint32_t reg, void *data
 		return riscv32_bool_to_4(riscv_csr_read(hart, RV_DPC, data));
 	if (reg >= RV_CSR_GDB_OFFSET)
 		return riscv32_bool_to_4(riscv_csr_read(hart, reg - RV_CSR_GDB_OFFSET, data));
+	if (reg >= RV_FPU_GDB_OFFSET)
+		return riscv32_bool_to_4(riscv_csr_read(hart, RV_FP_BASE + reg - RV_FPU_GDB_OFFSET, data));
 	return -1;
 }
 
@@ -168,6 +170,8 @@ static ssize_t riscv32_reg_write(target_s *const target, const uint32_t reg, con
 		return riscv32_bool_to_4(riscv_csr_write(hart, RV_DPC, data));
 	if (reg >= RV_CSR_GDB_OFFSET)
 		return riscv32_bool_to_4(riscv_csr_write(hart, reg - RV_CSR_GDB_OFFSET, data));
+	if (reg >= RV_FPU_GDB_OFFSET)
+		return riscv32_bool_to_4(riscv_csr_write(hart, RV_FP_BASE + reg - RV_FPU_GDB_OFFSET, data));
 	return -1;
 }
 
--- a/src/target/riscv_debug.c
+++ b/src/target/riscv_debug.c
@@ -141,24 +141,66 @@ static const char *const riscv_gpr_names[RV_GPRS_COUNT] = {
 	"t3", "t4", "t5", "t6",
 };
 
-typedef struct riscv_csr_descriptor  {
+// clang-format on
+typedef struct riscv_csr_descriptor {
 	const char *name;
 	const uint32_t csr_number; // fits in 16 bits actually (?)
 } riscv_csr_descriptor_s;
 
-static const riscv_csr_descriptor_s riscv_csrs[]={
-	{"mstatus",RV_CSR_STATUS},
-	{"misa",RV_CSR_MISA},
-	{"mie",	RV_CSR_MIE},
-	{"mtvec",	RV_CSR_MTVEC},
-	{"tscratch",RV_CSR_MSCRATCH},
-	{"mepc",	RV_CSR_MEPC},
-	{"mcause",RV_CSR_MCAUSE},
-	{"mtval",	RV_CSR_MTVAL},
-	{"mip",	RV_CSR_MIP},
+static const riscv_csr_descriptor_s riscv_csrs[] = {
+	{"mstatus", RV_CSR_STATUS},
+	{"misa", RV_CSR_MISA},
+	{"mie", RV_CSR_MIE},
+	{"mtvec", RV_CSR_MTVEC},
+	{"tscratch", RV_CSR_MSCRATCH},
+	{"mepc", RV_CSR_MEPC},
+	{"mcause", RV_CSR_MCAUSE},
+	{"mtval", RV_CSR_MTVAL},
+	{"mip", RV_CSR_MIP},
 };
-
-// clang-format on
+// fpu registers straight from gdb https://patchwork.kernel.org/project/qemu-devel/patch/20181228220731.4753-1-jimw@sifive.com/
+// actual order from gdb 13.2
+// csr registers declared in the "normal" registers section, they may also need to be in the csr section
+static const char *riscv_fpu_ctrl_regs[] = {
+	"flags",
+	"rm",
+	"csr",
+};
+// Regular  FPU register suffixes
+static const char *riscv_fpu_regs[] = {
+	"t0",
+	"t1",
+	"t2",
+	"t3",
+	"t4",
+	"t5",
+	"t6",
+	"t7",
+	"s0",
+	"s1",
+	"a0",
+	"a1",
+	"a2",
+	"a3",
+	"a4",
+	"a5",
+	"a6",
+	"a7",
+	"s2",
+	"s3",
+	"s4",
+	"s5",
+	"s6",
+	"s7",
+	"s8",
+	"s9",
+	"s10",
+	"s11",
+	"t8",
+	"t9",
+	"t10",
+	"t11",
+};
 
 /* General-purpose register types */
 static const gdb_reg_type_e riscv_gpr_types[RV_GPRS_COUNT] = {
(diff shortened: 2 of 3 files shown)


src/target/riscv32.c
```diff
@@ -153,6 +153,9 @@ static ssize_t riscv32_reg_read(target_s *target, const uint32_t reg, void *data
 		return riscv32_bool_to_4(riscv_csr_read(hart, RV_DPC, data));
 	if (reg >= RV_CSR_GDB_OFFSET)
 		return riscv32_bool_to_4(riscv_csr_read(hart, reg - RV_CSR_GDB_OFFSET, data));
+	if(reg >= RV_FPU_GDB_FIRST_REGISTER) {
```
dragonmux, line 156: Please drop the additional braces in this and the register write if statement additions - they're not necessary. To be more consistent with the style of the existing code and how things are done in the other architecture implementations, may we suggest `RV_FPU_GDB_OFFSET` be used for this constant?

Please also run clang-format - there's an extra space in the second change here, and this + the other added if statement are missing one between `if` and the `(` for the condition.

src/target/riscv_debug.c
```diff
@@ -157,6 +157,15 @@ static const riscv_csr_descriptor_s riscv_csrs[]={
 	{"mtval",	RV_CSR_MTVAL},
 	{"mip",	RV_CSR_MIP},
 };
+// fpu registers straight from gdb https://patchwork.kernel.org/project/qemu-devel/patch/20181228220731.4753-1-jimw@sifive.com/
+// actual order from gdb 13.2
+static const char *riscv_fpu_regs[] ={"t0","t1","t2", "t3","t4","t5","t6","t7",
+										"s0","s1","a0","a1","a2","a3","a4",
+										"a5","a6","a7","s2","s3","s4","s5",
+										"s6","s7","s8","s9","s10","s11","t8",
+										"t9","t10","t11"};
```
dragonmux, line 166: The final string needs to be followed by a `,` to get clang-format to format this properly.

src/target/riscv_debug.c
```diff
@@ -999,7 +1047,34 @@ static size_t riscv_build_target_description(
 	offset += snprintf(buffer + offset, print_size, "<reg name=\"pc\" bitsize=\"%u\"%s/>", address_width,
 		gdb_reg_type_strings[GDB_TYPE_CODE_PTR]);
 
-	/* XXX: TODO - implement generation of the FPU feature and registers */
+	
+
+	/* Basic single precision support */
+	if (extensions & RV_ISA_EXT_SINGLE_FLOAT) {
```
dragonmux, line 1053: Given this needs to essentially be done again for doubles, please move this into its own function above this one. That way it can be re-used more easily to build out the `RV_ISA_EXT_DOUBLE_FLOAT` handling.

src/target/riscv_debug.h
```diff
@@ -204,6 +204,8 @@ typedef struct riscv_hart {
 // so they cannot collide with GPR registers, so you have to subtract
 // RV_CSR_GDB_OFFSET from the value received from gdb
 #define RV_CSR_GDB_OFFSET 128
+#define RV_FPU_GDB_FIRST_REGISTER 33
```
dragonmux, line 207: The original idea with this section of defines is that it was a block just for the CSRs - perhaps move these two new defines below to line 218 as a new block just for FPU definitions?

src/target/riscv_debug.c
```diff
@@ -999,7 +1047,34 @@ static size_t riscv_build_target_description(
...
+
+		if (max_length != 0)
+			print_size = max_length - (size_t)offset;
+		offset += snprintf(buffer + offset, print_size, "</feature><feature name=\"org.gnu.gdb.riscv.fpu\">");
+
+		for (unsigned int i = 0; i < sizeof(riscv_fpu_regs) / sizeof(riscv_fpu_regs[0]); i++) {
+			if (max_length != 0)
+				print_size = max_length - (size_t)offset;
+			offset += snprintf(buffer + offset, print_size,
+				"<reg name=\"f%s\" bitsize=\"32\" type=\"ieee_single\" regnum=\"%u\" group=\"float\"/>",
```
dragonmux, line 1065: One notable thing from our prior work with GDB here, we can likely drop both the `group=` and `type=` stanzas which reduces the size of the feature blob generated here, which can otherwise get rather large. This reduces the likelihood of an OOM condition occurring on STM32F1-based probes.

src/target/riscv_debug.c
```diff
@@ -999,7 +1047,34 @@ static size_t riscv_build_target_description(
...
+	/* Basic single precision support */
+	if (extensions & RV_ISA_EXT_SINGLE_FLOAT) {
+		const int first_fpu_register = RV_FPU_GDB_FIRST_REGISTER;         // fixme!
+		const int first_fpu_control_register = RV_FPU_GDB_CSR_REGISTER; // fixme!
+
+		if (max_length != 0)
+			print_size = max_length - (size_t)offset;
+		offset += snprintf(buffer + offset, print_size, "</feature><feature name=\"org.gnu.gdb.riscv.fpu\">");
+
+		for (unsigned int i = 0; i < sizeof(riscv_fpu_regs) / sizeof(riscv_fpu_regs[0]); i++) {
```
dragonmux, line 1061: Please use the `ARRAY_LENGTH()` macro rather than computing the array length yourself here. Likewise `size_t` not `unsigned int` - the type of sizeof() is always size_t, not using this type alias may lead to type conversions.

dragonmux (commented):
This is looking good though needs `clang-format` run - please see our review notes below for the stylistic items we've spotted and some code organisation suggestions.

With those taken care of, this should be good to merge.

mean00:
Thank you
I've fixed most if not all as a separate commit. If it's okay i'll squash them

dragonmux:
It's looking much better. Please make sure you run `clang-format` on the changes as there are still inconsistencies that it will take care of, and that leaves just one review item to discuss and address.

mean00:
I have both git pre commit installed & manually ran clang-format before commiting (twice).
 I'm beginning to suspect having clang-fmt 17 is causing the issues. It looks like bmp is setup for clang-fmt16.  Maybe a known gotcha ?

Regarding extracting the function, i would have to forward fpu size (32/64) but also offset, buffer,  .... as parameters
Am i missing something or is it the idea ?

Thank you

dragonmux:
clang-format-17 should work fine (it's clang-format-18 that does extra stuff that's unwanted here), but failing that use clang-format-16 via `pip` (https://github.com/ssciwr/clang-format-wheel)

Re extracting the function: yes, that's exactly the idea - see https://github.com/blackmagic-debug/blackmagic/blob/c4015cdf5efaeaca35075f04f461d72132aa3a37/src/target/cortexar.c#L1682-L1704 and the Cortex-M equivalent which explicitly and intentionally splits out FPU block generation to manage routine complexity and size. Ideally the CSR block would likewise be split out.

mean00:
Thank you for the details
Hopefully this time it is ok.
I've moved the fpu part to its own function (like for CM-AR) and added a fpu size parameter so it should be easier to add support for double precision later (dont have any of those available to test).

(triple checked clang-format (17) was not complaining, not too hopeful)

src/target/riscv_debug.c
```diff
@@ -965,6 +1004,45 @@ static const char *riscv_fpu_ext_string(const uint32_t extensions)
  * Both are then continued with:
  * "		<reg name=\"pc\" bitsize=\"[address_width]\" type=\"code_ptr\"/>"
  * "	</feature>"
+ * The following are only generated for an F core (single precision HW float support):
```
dragonmux, line 1007: We would move this comment block to the front of the FPU block function so the documentation for it is kept together with the function.

src/target/riscv_debug.c
```diff
@@ -918,6 +927,36 @@ static const char *riscv_fpu_ext_string(const uint32_t extensions)
 	return "";
 }
 
+/*
+ * Generate the fpu section of the description
+ * fpu_size =32 single precision float
+ * fpu_size =64 double precision float
+ */
+static size_t riscv_build_target_fpu_description(size_t fpu_size, char *const buffer, size_t max_length)
```
dragonmux, line 935: Given the order of parameters on the base function, the compiler will likely generate a smaller amount of code for the call frame if `fpu_size` is made the last parameter. This is fine otherwise.

src/target/riscv_debug.c
```diff
@@ -918,6 +927,36 @@ static const char *riscv_fpu_ext_string(const uint32_t extensions)
...
+{
+	const size_t first_fpu_register = RV_FPU_GDB_OFFSET;             // fixme!
+	const size_t first_fpu_control_register = RV_FPU_GDB_CSR_OFFSET; // fixme!
```
dragonmux, line 938: It would be useful with this and the line above if you could expand on what needs fixing about them - it's very curious what those "fixme!" comments are about.

dragonmux (commented):
Think we've finally figured out why clang-format was not formatting those FPU regs strings - see our review notes. This is almost there. There's a query we've got on two lines in the FPU block generator and a couple of suggestions and once those are answered and taken care of w/ the one necessary fix made, this should be good to merge.

src/target/riscv_debug.c
mean00: No sure i got it correctly, i think the idea was to move the clang-format on up
+ fixed a mistake on fpu csr param that were swapped
Thank you

dragonmux: Think you'd meant to reply to the first of these 4 review items. Yes, that is the change we were thinking though you've moved it further up than we were anticipating - but think that's ok as that other code needed some formatting love too.

The only remaining issue there is the lack of a blank line between the preceding array and your addition, but as the CSR stuff needs some work, don't worry about that. We can fix that after.

dragonmux (approved):
This LGTM now, we'll get this merged as we have no reasonable way to test it locally.

mean00:
Squashed, thank you

PR #1770, opened 2024-02-20
Fix: SAMD/SAMx5x part idententification

dragonmux:
## Detailed description

In response to a user report in Discord of troubles with part identification on an ATSAMD21E18A, we took a look into how the SAMD support code identifies parts that might be ATSAMD's and saw it manually reading out the DSU component and peripheral identifiers rather than using the information already in the target structure.

This PR replaces this logic by the normal target structure values which match up and are already checked. It does the same for the SAMx5x support as this does identical things to check for those devices.

Tested working on an ATSAMD11D14AM. We have no reason to suspect this will create any kind of regressions.

### DIFF
--- a/src/target/adiv5.c
+++ b/src/target/adiv5.c
@@ -377,7 +377,7 @@ static uint32_t adiv5_ap_read_id(adiv5_access_port_s *ap, uint32_t addr)
 	return res;
 }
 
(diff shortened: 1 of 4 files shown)


esden (approved):
LGTM

PR #1772, opened 2024-02-26
Fix: CMSIS-DAP response buffer sizing

dragonmux:
## Detailed description

This PR fixes a bug accidentally introduced in PR #1731 in the underlying DAP command request-response code.

This code defines a stack buffer for the response but this was now vastly undersized to the max response size possible, resulting in stack smashing and badness. We fix this by upping the buffer to handle the same amount of data as the max packet size should be and add a check to make sure this assumption is not violated.

Tested on ORBTrace which triggers this issue with its 512 byte packets.


esden (approved):
LGTM

PR #1773, opened 2024-02-29
libgpiod based BMDA gpio bitbang backend

### LINKED ISSUE #1063: Direct SWD bit bang programming
Does blackmagic have the capability flash files using SWD pins that are wired directly to an embedded linux processor? As in, there is no USB device. The gpio from the microprocessors SWD pins are connected directly to the main application processor.

TechnoMancer:
## Detailed description

This change adds a new backend/probe type to the "hosted"/BMDA platform which uses libgpiod to connect up the `jtagtap.c` and `swdptap.c` bitbanging routines to gpios.  This is of course linux specific.

* A meson build option is added to enable this backend, this only has an effect on linux and defaults to attempting to find the dependency but disabling the feature if libgpiod is not found.
* A new probe type for gpiod is added in the BMDA code.
* An include file is added which provides mappings from the `gpio_set`, `gpio_clear`, `gpio_get`, `gpio_set_val` routines to the libgpiod backed routines for compiling the jtagtap and swdptap code.
* A new cli option is added when the feature is enabled which allows the user to set a mapping of debug signals to GPIOs if this is specified then it is assumed that the gpio bitbanging backend is to be used.  The format of the mapping is as follows:
    <gpio_mapping>=<signal_mapping>(`,`<signal_mapping>)*
    <signal_mapping>=<debug_signal>`=`<gpio_chip_name>`:`<gpio_offset>
    Where the debug signals are `swdio`, `swclk`, `tms`, `tck`, `tdi`, `tdo` and gpio_chip_name is the name of a gpiochip character device to use and gpio_offset is the numeric offset to the required gpio on that chip.
* The swdptap and jtagtap routines in `/src/platform/common` are built into the BMDA.  This requires conditionally removing the definition of `swd_proc` and `jtag_proc` in the `hosted/platform.c` file as these are usually defined in the relevant tap files.
* The required calls to init and use the relevant routines are added to the BMDA code, this requires the `jtagtap_init` routine to be unconditionally declared by `jtagtap.h`
* The required calls to perform swd or jtag scan are added to the BMDA code.
* This backend sets `target_clk_divider` to `UINT32_MAX` in order to disable delays, the calls to the kernel to change the GPIO status appear to have somewhat significant overhead anyway so the maximum clock rate observed was on the order of kilohertz.
* Similarly no support for getting or setting a debug clock frequency is provided.
* No support for a reset line is included but could be added with a little effort.
* JTAG and SWD lines are assumed disjoint, support for sharing similarly to probe hardware could be implemented later but requires more effort.

The backend has been tested with SWD on a Radxa Rock 4 C+ to a Raspberry Pi Pico.  JTAG has not been tested at time of submitting this PR.

## Closing issues

fixes #1063

### DIFF
--- a/src/platforms/hosted/bmda_gpiod.c
+++ b/src/platforms/hosted/bmda_gpiod.c
@@ -117,3 +117,11 @@ void bmda_gpiod_mode_output(struct gpiod_line *pin)
 	} else
 		DEBUG_ERROR("BUG! attempt to set uninit GPIO to output");
 }
+
+bool bmda_gpiod_init(bmda_cli_options_s *const cl_opts)
+{
+	if (!cl_opts->opt_gpio_map)
+		return false;
+
+	return false;
+}
--- a/src/platforms/hosted/bmda_gpiod.c
+++ b/src/platforms/hosted/bmda_gpiod.c
@@ -237,3 +237,13 @@ bool bmda_gpiod_init(bmda_cli_options_s *const cl_opts)
 
 	return bmda_gpiod_jtag_ok || bmda_gpiod_swd_ok;
 }
+
+bool bmda_gpiod_swd_init(void)
+{
+	if (!bmda_gpiod_swd_ok)
+		return false;
+
+	swdptap_init();
+
+	return true;
+}
--- a/src/platforms/hosted/bmda_gpiod.c
+++ b/src/platforms/hosted/bmda_gpiod.c
@@ -51,9 +51,25 @@ struct gpiod_line *bmda_gpiod_swclk_pin;
 
 uint32_t target_clk_divider = UINT32_MAX;
 
+static void bmda_gpiod_debug_pin(struct gpiod_line *line, const char *op, bool print, bool val)
+{
+#ifdef DEBUG
+	DEBUG_WIRE("GPIO %s %s", gpiod_line_consumer(line), op);
+	if (print)
+		DEBUG_WIRE("=%d", val);
+	DEBUG_WIRE("\n");
+#else
+	(void)line;
+	(void)op;
+	(void)print;
+	(void)val;
+#endif
+}
+
 void bmda_gpiod_set_pin(struct gpiod_line *pin, bool val)
 {
 	if (pin) {
+		bmda_gpiod_debug_pin(pin, "set", true, val);
 		if (gpiod_line_set_value(pin, val ? 1 : 0)) {
 			DEBUG_ERROR("Failed to set pin to value %d errno: %d", val, errno);
 			exit(1);
@@ -70,6 +86,7 @@ bool bmda_gpiod_get_pin(struct gpiod_line *pin)
 			DEBUG_ERROR("Failed to get pin value errno: %d", errno);
 			exit(1);
 		}
+		bmda_gpiod_debug_pin(pin, "read", true, ret);
 		return ret;
 	} else {
 		DEBUG_ERROR("BUG! attempt to read uninit GPIO");
@@ -80,6 +97,7 @@ bool bmda_gpiod_get_pin(struct gpiod_line *pin)
 void bmda_gpiod_mode_input(struct gpiod_line *pin)
 {
 	if (pin) {
+		bmda_gpiod_debug_pin(pin, "input", false, false);
 		if (gpiod_line_set_direction_input(pin)) {
 			DEBUG_ERROR("Failed to set pin to input errno: %d", errno);
 			exit(1);
@@ -91,6 +109,7 @@ void bmda_gpiod_mode_input(struct gpiod_line *pin)
 void bmda_gpiod_mode_output(struct gpiod_line *pin)
 {
 	if (pin) {
+		bmda_gpiod_debug_pin(pin, "output", false, false);
 		if (gpiod_line_set_direction_output(pin, 0)) {
 			DEBUG_ERROR("Failed to set pin to output errno: %d", errno);
 			exit(1);
--- a/src/platforms/hosted/bmda_gpiod.c
+++ b/src/platforms/hosted/bmda_gpiod.c
@@ -33,9 +33,12 @@
 
 /* Implement libgpiod based GPIO backend */
 
+#define _POSIX_C_SOURCE 200809L
+
 #include <gpiod.h>
 #include "general.h"
 #include <errno.h>
+#include <string.h>
 #include <limits.h>
 #include <stdbool.h>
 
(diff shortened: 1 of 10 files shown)


src/platforms/hosted/bmda_gpiod.h
```diff
@@ -0,0 +1,44 @@
...
+
+#ifndef BMDA_GPIOD_H
+#define BMDA_GPIOD_H
+
+#include "cli.h"
+
+bool bmda_gpiod_init(bmda_cli_options_s *const cl_opts);
+
+bool bmda_gpiod_jtag_init();
+bool bmda_gpiod_swd_init();
```
dragonmux, line 42: The parens here need `void` inside them to avoid declaring an any-parameters function. See style 3 of [function declaration](https://en.cppreference.com/w/c/language/function_declaration) for details.

src/platforms/hosted/bmda_gpiod.c
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	if (bmda_gpiod_tck_pin && bmda_gpiod_tdi_pin && bmda_gpiod_tdo_pin && bmda_gpiod_tms_pin) {
+		bmda_gpiod_jtag_ok = true;
+	}
+
+	return bmda_gpiod_jtag_ok || bmda_gpiod_swd_ok;
+}
+
+bool bmda_gpiod_jtag_init()
```
dragonmux, line 246: This and the function below get the same comment about needing `void` in the parens to avoid the potentially bad spicy kind of function definition.

src/platforms/hosted/bmda_gpiod.c
```diff
@@ -0,0 +1,264 @@
...
+out_fail_close:
+	DEBUG_ERROR("Requesting gpio failed errno: %d", errno);
+	gpiod_chip_close(gpiod_line_get_chip(line));
+	return false;
+}
+
+static bool bmda_gpiod_parse_gpiomap(const char *gpio_map)
+{
+	bool ret = true;
+	char *saveptr;
```
dragonmux, line 201: If possible, please initialise this and move it next to the first strtok_r() it gets used in, so we can try and prevent possible future UB read-before-init issues.

src/platforms/hosted/bmda_gpiod.c
```diff
@@ -0,0 +1,264 @@
...
+static bool bmda_gpiod_parse_gpio(const char *name, char *gpio)
+{
+	DEBUG_INFO("GPIO set %s: %s\n", name, gpio);
+	char *offset = strstr(gpio, ":");
+	if (!offset)
+		return false;
+	*offset = '\0';
+	offset++;
+	char *end = offset + strlen(offset);
+	char *valid;
```
dragonmux, line 140: Please initialise this to `NULL` as strtoul() may fail and not touch this on some libc's, resulting in UB read-before-init in the if check below.

src/platforms/hosted/bmda_gpiod.c
```diff
@@ -0,0 +1,264 @@
...
+{
+	if (pin) {
+		bmda_gpiod_debug_pin(pin, "set", true, val);
+		if (gpiod_line_set_value(pin, val ? 1 : 0)) {
+			DEBUG_ERROR("Failed to set pin to value %d errno: %d", val, errno);
+			exit(1);
+		}
+	} else {
+		DEBUG_ERROR("BUG! attempt to write uninit GPIO");
+	}
```
dragonmux, line 86: Please drop the extra braces, they're unnecessary. Same for the instances below for single-statement block bodies.

src/platforms/hosted/bmda_gpiod.c
```diff
@@ -0,0 +1,264 @@
...
+bool bmda_gpiod_get_pin(struct gpiod_line *pin)
+{
+	if (pin) {
+		int ret = gpiod_line_get_value(pin);
+		if (ret < 0) {
+			DEBUG_ERROR("Failed to get pin value errno: %d", errno);
+			exit(1);
+		}
+		bmda_gpiod_debug_pin(pin, "read", true, !!ret);
+		return !!ret;
```
dragonmux, line 98: Using `!!` here is not necessary nor the style of the code base, you may simply let implicit bool conversion do its thing.

dragonmux (changes requested):
This is looking excellent! There are a couple of items to address or at least discuss, and once they're taken care of we'll approve this for merge.

Thank you for the contribution! Hopefully users on RPi's and other similar systems will be glad to be able to use the GPIO facilities built into their SBCs for debug.

src/platforms/hosted/bmda_gpiod.c
TechnoMancer: I have removed the `!!`s

src/platforms/hosted/bmda_gpiod.c
TechnoMancer: I think I have sorted this out.

src/platforms/hosted/bmda_gpiod.c
TechnoMancer: Done

src/platforms/hosted/bmda_gpiod.c
TechnoMancer: Done

src/platforms/hosted/bmda_gpiod.h
TechnoMancer: Done

src/platforms/hosted/bmda_gpiod_platform.h
```diff
@@ -0,0 +1,73 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef BMDA_GPIOD_PLATFORM_H
+#define BMDA_GPIOD_PLATFORM_H
+
+#include <stdbool.h>
+
+struct gpiod_line;
```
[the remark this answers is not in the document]
TechnoMancer: Done

src/platforms/hosted/bmda_gpiod.c
TechnoMancer: I missed 142 before, thanks.

Should be sorted now.

dragonmux (approved):
LGTM, merging once the builds complete. Thank you again for the contribution!

PR #1775, opened 2024-03-02
Fix: Update dependencies

dragonmux:
## Detailed description

In this PR we update the BMDA dependencies libusb and HIDAPI. The fork of libusb we use for Meson support has been moved into the organisation and updated to v1.0.27. HIDAPI has been forked to fully Meson-ify it so we can avoid mesonbuild/meson#12761 which Xobs hit on his M3. This also locks HIDAPI to v0.14.0.

Note that in both cases, these are fallbacks for if the libraries and their build files are not found on a user's system, and so we will prefer the user's versions if possible w/ the build system allowing much older versions of both to be used.

## Your checklist for this pull request

* [x] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md)
* [x] I've read the [guidelines for contributing](https://github.com/blackmagic-debug/blackmagic/blob/main/CONTRIBUTING.md) to this repository
* [x] It builds for hardware native (see [Building the firmware](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-firmware))
* [x] It builds as BMDA (see [Building the BMDA](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-app))
* [x] I've tested it to the best of my ability
* [x] My commit messages provide a useful short description of what the commits do

## Closing issues

Fixes #1774

PR #1776, opened 2024-03-02
Fix dual stack listening

TechnoMancer:
## Detailed description

BMDA is supposed to listen for connections on both IPv6 and IPv4 if possible, this is accomplished by dual stack support when using an IPv6 listening socket.  However there are two problems that prevent this from working as intended.

* When `getaddrinfo` returns results the first result is not always the IPv6 address to bind to if one is returned.  To fix this the results are walked over to select the first IPv6 result if one exists.  The first result is used if no IPv6 address can be found.
* An IPv6 listening socket accepting IPv4 connections relies on the `IPV6_V6ONLY ` socket option being set to 0 (off, dual stack) on the socket before it is bound.  On windows this option defaults to 1 (on, v6 only) and must be explicitly disabled to allow dual stack listening sockets.  On linux this option gets it's default from a sysctl setting (`net.ipv6.bindv6only`) and so explicitly setting it there is also ideal.  The option exists on Mac OS but I do not know it's default state.  Searching suggests that if a system is IPv6 only (no v4 support at all) then disabling this option will fail, however in that case v6 only listening is the only possible option.  So to resolve this when binding to an IPv6 address attempt to disable this socket option and print a warning that IPv6 only listening will occur if setting fails.

### DIFF
--- a/src/platforms/hosted/gdb_if.c
+++ b/src/platforms/hosted/gdb_if.c
@@ -252,6 +252,12 @@ int gdb_if_init(void)
 			!socket_set_int_opt(gdb_if_serv, IPPROTO_TCP, TCP_NODELAY, 1))
 			continue;
 
+		if (addr.ss_family == AF_INET6) {
+			DEBUG_INFO("Setting V6ONLY to off for dual stack listening.\n");
+			if (!socket_set_int_opt(gdb_if_serv, IPPROTO_IPV6, IPV6_V6ONLY, 0))
+				DEBUG_WARN("Listening on IPv6 only.\n");
+		}
+
 		if (bind(gdb_if_serv, (sockaddr_s *)&addr, family_to_size(addr.ss_family)) == -1) {
 			handle_error(gdb_if_serv, "binding socket");
 			continue;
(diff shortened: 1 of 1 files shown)


src/platforms/hosted/gdb_if.c
```diff
@@ -140,9 +140,15 @@ static sockaddr_storage_s sockaddr_prepare(const uint16_t port)
 		return (sockaddr_storage_s){AF_UNSPEC};
 	}
 
-	/* Pick the first result, copy it and free the list structure getaddrinfo() returns */
+	/* Try to find an IPv6 result but fall back to the first result if none can be found */
 	sockaddr_storage_s service;
 	memcpy(&service, results->ai_addr, family_to_size(results->ai_addr->sa_family));
+	for (addrinfo_s *curr = results; curr; curr = curr->ai_next)
```
dragonmux, line 146: Because of the inner multi-statement block, the `for` loop is considered to be a multi-statement block too for the style guide, please include the `{}`'s here even though this is technically correct (reduces confusion).

dragonmux (changes requested):
Just the one review note - everything else looks good. Nice work! Look forward to merging this.

dragonmux (approved):
LGTM, merging once the builds complete. Thank you for the contribution!

We're aware that the MSVC build is broken here and that it's not your fault. That'll get resolved when #1775 is merged.

PR #1781, opened 2024-03-06
Fix: BMP on M3 reenumeration

dragonmux:
## Detailed description

This PR attempts to improve the re-enumeration situation on M3 silicon powered machines as the controller built into the M3 does not properly recognise our setting up the conditions for re-enumeration for re-entry into the bootloader as reported on Discord by @xobs.

We attempt to improve things by ensuring we drive the reset condition described by the USB 2.0 specification version 20230224 §7.1.7.1 Table 7-2 pg145:
image

### DIFF
--- a/src/platforms/native/platform.c
+++ b/src/platforms/native/platform.c
@@ -413,6 +413,8 @@ void platform_request_boot(void)
 {
 	/* Disconnect USB cable */
 	gpio_set_mode(USB_PU_PORT, GPIO_MODE_INPUT, GPIO_CNF_INPUT_FLOAT, USB_PU_PIN);
+	gpio_set_mode(USB_PORT, GPIO_MODE_INPUT, GPIO_CNF_INPUT_PULL_UPDOWN, USB_DP_PIN | USB_DM_PIN);
+	gpio_clear(USB_PORT, USB_DP_PIN | USB_DM_PIN);
 
 	/* Drive boot request pin */
 	gpio_set_mode(GPIOB, GPIO_MODE_OUTPUT_2_MHZ, GPIO_CNF_OUTPUT_PUSHPULL, GPIO12);
--- a/src/platforms/native/platform.c
+++ b/src/platforms/native/platform.c
@@ -39,6 +39,7 @@
 #include <libopencm3/stm32/spi.h>
 #include <libopencm3/stm32/flash.h>
 #include <libopencm3/stm32/timer.h>
+#include <libopencm3/cm3/systick.h>
 
 static void adc_init(void);
 static void setup_vbus_irq(void);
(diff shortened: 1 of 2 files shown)


xobs:
I can confirm that this solves the detach problem!

```
seancross in 🌐 bestways in blackmagic (8320b5f) [?] is 📦 v1.10.0
❯ dfu-util -l
dfu-util 0.11

Copyright 2005-2009 Weston Schmidt, Harald Welte and OpenMoko Inc.
Copyright 2010-2021 Tormod Volden and Stefan Schmidt
This program is Free Software and has ABSOLUTELY NO WARRANTY
Please report bugs to http://sourceforge.net/p/dfu-util/tickets/

Found DFU: [1d50:6017] ver=0100, devnum=2, cfg=1, intf=0, path="0-1", alt=0, name="@Internal Flash   /0x08000000/8*001Ka,120*001Kg", serial="97B6DC14"

seancross in 🌐 bestways in blackmagic (8320b5f) [?] is 📦 v1.10.0
❯ bmputil debug detach
Requesting device detach from DFU mode to runtime mode...

seancross in 🌐 bestways in blackmagic (8320b5f) [?] is 📦 v1.10.0
❯ dfu-util -l
dfu-util 0.11

Copyright 2005-2009 Weston Schmidt, Harald Welte and OpenMoko Inc.
Copyright 2010-2021 Tormod Volden and Stefan Schmidt
This program is Free Software and has ABSOLUTELY NO WARRANTY
Please report bugs to http://sourceforge.net/p/dfu-util/tickets/

Found Runtime: [1d50:6018] ver=0109, devnum=1, cfg=1, intf=4, path="0-1", alt=0, name="Black Magic DFU", serial="97B6DC14"

seancross in 🌐 bestways in blackmagic (8320b5f) [?] is 📦 v1.10.0
❯
```

dragonmux:
Oh fantastic! Thank you for checking that

tlyu:
Oh interesting. The relevant line from Table 7-2 is the "Disconnect (at downstream port)" one, which says SE0 for at least 2.5µs. I guess it's possible (but unlikely?) that the bootloader brings up the D+ pullup before the cable fully discharges (I timed around 45µs on my setup). Or maybe there are still transactions occurring on the USB peripheral that are actively driving the USB lines, which defeats the removal of the D+ pullup drive. I think an earlier iteration of my patch did disable the USB IRQ and/or shut down the USB peripheral. I guess explicitly setting the pads to GPIO input might work too, but it seems cleaner to shut down the peripheral.

dragonmux:
We've been contemplating adding 10ms to help guarantee that reset behaviour target. Unsure how best to go about shutting the peripheral down but figured that, well, it's about to see a core reset occur on transition to the bootloader and so it won't matter that it's talking into the void for a few µs/ms before the bootloader resets it. After all, to be connected through to the pads and see what they're doing, the GPIOs have to be in their altfn state. Similarly even if the peripheral can see, it should register a USB Reset having occurred - the goal of this code.

tlyu:
Yeah, I think you're right about the USB peripheral being effectively disconnected by having its pads no longer be set to AF.

I think you don't need a full 10ms, because the reset is something sent by the hub/host. The trick is making sure that the D+ pullup floats for long enough for the hub to see SE0 for at least 2.5µs, which means no traffic sent by the device or the hub/host. But 10ms is also fairly short in terms of human perception, so might as well?

The hub/host will reset the bus after seeing the re-attach by the bootloader, anyway, which should reset the remaining peripheral state.

dragonmux:
We'll do that then. This PR is already only ensuring D+ and D- are pulled low via the weak internal pulls, so if the host really is trying to drive the lines for some reason there should be no drive conflicts, and you're correct that once the host sees the pullup again, it should USB Reset all over anyway.

esden (approved):
LGTM

PR #1782, opened 2024-03-08
Some fixes to single register read/write

TechnoMancer:
## Detailed description

When testing the BMDA gpiod support with JTAG I came across a segfault when running `info registers` on a riscv32 target (GD32VF103).  Investigating this led to discovering a series of issues around single register reads. This was during dumping of the `tscratch` register.

To improve matters here a few things are done.  Firstly the target_reg_read and target_reg_write functions are altered to return `size_t` instead of `ssize_t` and to return 0 to indicate failure rather than -1.  The use in `gdb_main_loop` was incorrectly storing the result in a `size_t` variable which resulted in failure attempting to hexify an unreasonably large number of bytes because the failure return of -1 was not caught by the < 0 check as `size_t` is unsigned.

After this the reading of `tscratch` would still fail, at first I assumed this register was just missing on this target but some comparison of the riscv32 target code to the cortexm target code while making the above change led to realising that the 'p' packet implementation will always call target_reg_read with a size of 8, however the riscv32 reg_read function was assuming that only a size of 4 was acceptable and so would always error in this case.  This is fixed by changing riscv32 to only error out early if the passed buffer size is less than 4.

Additionally when looking at the implementation of the 'P' packet type (single register write) it was noticed that the code uses a 4 byte value buffer for reading the given hex value into, however this code will just use the length of the given hex value to determine how many bytes to try and dehexify, this would overrun the stack allocated value buffer if more than 4 bytes worth of hex value are supplied.  To correct this a check on the length of supplied hex data is added to error if it is larger than the value buffer.

These fixes result in `info registers` working on the GD32VF103 without segfaults.

### DIFF
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -234,7 +234,7 @@ int gdb_main_loop(target_controller_s *tc, char *pbuf, size_t pbuf_size, size_t 
 			sscanf(pbuf, "p%" SCNx32, &reg);
 			uint8_t val[8];
 			size_t s = target_reg_read(cur_target, reg, val, sizeof(val));
-			if (s > 0)
+			if (s != 0)
 				gdb_putpacket(hexify(pbuf, val, s), s * 2U);
 			else
 				gdb_putpacketz("EFF");
@@ -267,7 +267,7 @@ int gdb_main_loop(target_controller_s *tc, char *pbuf, size_t pbuf_size, size_t 
 			uint8_t value[4] = {0};
 			unhexify(value, packet, value_length);
 			/* Finally, write the converted value to the target */
-			if (target_reg_write(cur_target, reg, value, sizeof(value)) > 0)
+			if (target_reg_write(cur_target, reg, value, sizeof(value)) != 0)
 				gdb_putpacketz("OK");
 			else
 				gdb_putpacketz("EFF");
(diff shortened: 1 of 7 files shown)


src/gdb_main.c
```diff
@@ -264,10 +264,15 @@ int gdb_main_loop(target_controller_s *tc, char *pbuf, size_t pbuf_size, size_t
 			/* Skip past the '=' and convert the value */
 			++packet;
 			const size_t value_length = strlen(packet) / 2U;
+			/* If the value is bigger than 4 bytes report error */
+			if (value_length > 4) {
```
dragonmux, line 268: This and the above should ideally use `U` suffixed constants so the comparison is always done unsigned.

Looks good otherwise with the note that we should make the value compared here depend on the target so it works with 64-bit targets too, or upgrade the buffer to 8 bytes always, but this is for later work and not a problem for this PR (unless you want to tackle it!).

dragonmux (changes requested):
Only got the one review note, the rest of this looks great and we'll be happy to merge this once the single note is addressed 🙂

src/gdb_main.c
TechnoMancer: I will fix the constant. I think the target dependence stuff should be dealt with separately.

dragonmux: Aye, we mostly wrote the second half of that so it's noted somewhere 😄

dragonmux (approved):
LGTM. Ideally all comparison constants match the signedness of the thing they're being compared against to avoid any conversions (eg, `s != 0U`) for note in the future. Happy to merge this, thank you for the contribution!

PR #1785, opened 2024-03-15
target/nrf91: add mass_erase and recovery probe

### LINKED ISSUE #1778: extending nRF91 support
Hi!
I've been looking at nRF91 support in BMP. It's nice that we have basic support already, but two things are missing:
* APPROTECT recovery
* modem firmware update

I could try to help implementing this. This recovery mechanism seems to relatively straightforward to integrate into BMP, but what about the modem firmware? We deliver it as a zip file and there is a long-ish sequence to get the job done (see [here](https://github.com/pyocd/pyOCD/blob/5ab5c327987c4c3ba5a37928cac25263d4db7586/pyocd/target/family/target_nRF91.py#L629)). How would you approach this?

maxd-nordic:
## Detailed description

* add nrf91 rescue probe
* add nrf91 mass erase
* add detailed nrf91 identification
* add nrf91 UICR writing
* TODO: add UICR APPROTECT initialization
* TODO: add UICR overwrite warning (recommend mass erase if value cannot be written)

## Closing issues

fixes #1778


maxd-nordic:
@dragonmux I added most of the missing parts for nrf91. Maybe you could give me pointers on how to handle the TODOs? The UICR APPROTECT part basically needs to do a flash write after each mass-erase to tell the "hardenend APPROTECT" system not to lock down the device after the next reset. This is also relevant for nRF53 support.

dragonmux:
The post-Flash write and mass-erase UICR fix-up sounds like something that should be a target [`exit_flash_mode`](https://github.com/blackmagic-debug/blackmagic/blob/fa79ab6b8bba4f9807354e48fca477fc565854e8/src/target/target_internal.h#L141) function and then call that at the end of the nRF91 mass erase implementation too - that should take care of the problem.

src/target/nrf91.c
```diff
@@ -81,25 +134,180 @@ static void nrf91_add_flash(target_s *target, uint32_t addr, size_t length, size
...
+	adiv5_access_port_s ctrl_ap = {
+		.dp = dp,
+		.apsel = 0x4U,
+	};
+	ahb_ap.idr = adiv5_ap_read(&ahb_ap, ADIV5_AP_IDR);
+	ahb_ap.csw = adiv5_ap_read(&ahb_ap, ADIV5_AP_CSW);
+	ctrl_ap.idr = adiv5_ap_read(&ctrl_ap, ADIV5_AP_IDR);
+
+	if (ahb_ap.idr != AHB_AP_IDR_EXPECTED) {
+		DEBUG_ERROR("nRF91: AHB-AP IDR is 0x%08x, expected 0x84770001\n", ahb_ap.idr);
```
dragonmux, line 294: Please use `PRIx32` for the format specifier, not `x`, as idr is a `uint32_t` and so this doesn't work on Xtensa or several desktop platforms.

eg: `DEBUG_ERROR("nRF91: AHB-AP IDR is 0x%08" PRIx32 ", expected 0x84770001\n", ahb_ap.idr);`

It would also be better for Flash space if the expected value were formatted in the same way.

src/target/target_probe.h
```diff
@@ -98,5 +98,6 @@ bool stm32mp15_cm4_probe(target_s *target);
 bool zynq7_probe(target_s *target);
 
 void lpc55_dp_prepare(adiv5_debug_port_s *dp);
+int nrf91_dp_prepare(adiv5_debug_port_s *const dp);
```
dragonmux, line 101: `const` here will be ignored by the compiler. This should be a `clang-tidy` lint. Having it in the function definitions is correct though.

src/target/nrf91.c
```diff
@@ -81,25 +134,180 @@ static void nrf91_add_flash(target_s *target, uint32_t addr, size_t length, size
...
-		target->target_options |= TOPT_INHIBIT_NRST;
-		target_add_ram(target, 0x20000000, 256U * 1024U);
-		nrf91_add_flash(target, 0, 4096U * 256U, 4096U);
-		break;
-	default:
+	uint32_t buf[3] = {0};
+
+	ap->dp->mem_read(ap, buf, 0x00FF0140, 3*sizeof(uint32_t));
+
+	gdb_outf("nRF%04x %4s%4s detected!\n", buf[0], (const char *)&buf[2], (const char *)&buf[1]);
```
dragonmux, line 177: This would be fine as a DEBUG_* statement, but should be avoided in the scan output as it's not guaranteed where exactly this will end up in the information display, while also duplicating the information to the user from the target->driver strings.

Same comment as below on uint32_t needing PRIx32.

src/target/nrf91.c
```diff
@@ -81,25 +134,180 @@ static void nrf91_add_flash(target_s *target, uint32_t addr, size_t length, size
...
+
+	return nrf91_ctrl_ap_mass_erase(&ctrl_ap);
 }
 
 bool nrf91_probe(target_s *target)
 {
 	adiv5_access_port_s *ap = cortex_ap(target);
 
-	if (ap->dp->version < 2U)
+	if (ap->dp->version < 2U || ap->dp->target_partno != 0x90U)
```
dragonmux, line 170: Please avoid a magic number here - `0x90U` should ideally be `#define`'d above as, eg, `ID_NRF91` so it's clearer what it is and where it comes from. Ideally a reference to the TRM section and page this information can be found from would be great, but we know that's not always possible.

src/target/nrf91.c
```diff
@@ -15,6 +16,49 @@
...
+#define CTRL_AP_RESET  ADIV5_AP_REG(0x000)
+#define CTRL_AP_ERASEALL  ADIV5_AP_REG(0x004)
+#define CTRL_IDR_EXPECTED  0x12880000
+#define AHB_AP_IDR_EXPECTED  0x84770001
+#define CTRL_AP_ERASEALLSTATUS  ADIV5_AP_REG(0x008)
+
+#define UICR_APPROTECT 0x00FF8000U
+#define UICR_SECUREAPPROTECT 0x00FF802CU
+#define UICR_APPROTECT_UNPROTECT_VAL 0x50FA50FAU
+#define UICR_ERASED_VAL 0xFFFFFFFFU
```
dragonmux, line 28: Please prefix these with `NRF91_` to maintain the macro prefixing for the file.

src/target/nrf91.c
```diff
@@ -81,25 +134,180 @@ static void nrf91_add_flash(target_s *target, uint32_t addr, size_t length, size
...
+//		target_mem_write32(target, NRF91_NVMC_CONFIG, NRF91_NVMC_CONFIG_REN);
+//	}
+
+	return true;
+}
+
+
+
+
+static bool nrf91_rescue_do_recover(target_s *t)
```
dragonmux, line 230: Please avoid single letter variable naming for this, please use `target` instead. Likewise for `nrf91_rescue_probe()` on return from `target_new()`.

This is the style we've been gradually moving the whole code base to.

src/target/nrf91.c
```diff
@@ -81,25 +134,180 @@ static void nrf91_add_flash(target_s *target, uint32_t addr, size_t length, size
...
+			platform_nrst_set_val(false);
+
+			/* repetition not needed and debug port inactive at this point */
+			return false;
+		}
+
+		//check if CSW DEVICEEN is set
+		struct adiv5_access_port ahb_ap = *ap;
+		ahb_ap.apsel = 0x0U;
+		uint32_t csw = ap->dp->ap_read(&ahb_ap, ADIV5_AP_CSW);
```
dragonmux, line 254: This can be made `const`, so lets do so.

dragonmux (commented):
Done an initial review, there are a lot of interesting details here that we're very grateful to get to know about, like the TARGETID revision number being used for once in a part and having bearing on the part number.

Please don't forget to run `clang-format` on the contribution and fix the few hex constant capitalisation lints - the `clang-format` lint pass is currently broken but we are still enforcing it.

src/target/nrf91.c
```diff
@@ -49,6 +93,14 @@ static bool nrf91_flash_erase(target_flash_s *flash, target_addr_t addr, size_t
 	return nrf91_wait_ready(target, NULL);
 }
 
+static bool nrf91_uicr_flash_erase(target_flash_s *flash, target_addr_t addr, size_t len)
+{
+	/* can't erase UICR without a mass-erase */
+	/* TODO: this will silently fail if you actually need to erase */
```
dragonmux, line 99: Easiest way to fix this would be to `gdb_outf()` a message like "Skipping UICR erase, operation requires mass erase". If there's a good way to blank check the UICR that can be used to have this routine `return true` with no message, and then `return false` if the UICR is not blank.

maxd-nordic:
> The post-Flash write and mass-erase UICR fix-up sounds like something that should be a target [`exit_flash_mode`](https://github.com/blackmagic-debug/blackmagic/blob/fa79ab6b8bba4f9807354e48fca477fc565854e8/src/target/target_internal.h#L141) function and then call that at the end of the nRF91 mass erase implementation too - that should take care of the problem.

Thanks! Got it to work now. Just putting it in exit_flash_mode is enough IMO because a blank app will also cause the protection to engage.

dragonmux:
The reason we said "and at the end of the mass erase function" is that it's up to the target support too override `target->mass_erase` with its own implementation (it's NULL by default) and the implementation will only do what you put in the function - so if you need to call `target->exit_flash_mode` then you'll need to include that in the implementation specifically. We'd be concerned for newly mass erased parts winding up protected without it.

maxd-nordic:
> The reason we said "and at the end of the mass erase function" is that it's up to the target support too override `target->mass_erase` with its own implementation (it's NULL by default) and the implementation will only do what you put in the function - so if you need to call `target->exit_flash_mode` then you'll need to include that in the implementation specifically. We'd be concerned for newly mass erased parts winding up protected without it.

`nrfjprog` flashes a placeholder app onto the target when doing recovery. Should I include that here as well? (it would use up quite a bit of space) Just writing UICR is not enough to keep protection off. The way it's implemented now, you can do a mass-erase and UICR is written alongside your app.

maxd-nordic:
I got to check the flash write function again - we can skip the readynext wait if the page is erased beforehand.

maxd-nordic:
@dragonmux I managed to make the app image much smaller and include it here - the target is now unprotected after mass-erase. :)

maxd-nordic:
It might actually be better to implement target commands and show the CTRL-AP in any case. There is also a rarely-used feature that could lock mass-erase with a password.

src/target/nrf91.c
```diff
@@ -81,25 +158,208 @@ static void nrf91_add_flash(target_s *target, uint32_t addr, size_t length, size
...
 {
 	adiv5_access_port_s *ap = cortex_ap(target);
 
-	if (ap->dp->version < 2U)
+	if (ap->dp->version < 2U || ap->dp->target_partno != NRF91_PARTNO)
 		return false;
 
-	switch (ap->dp->target_partno) {
-	case 0x90:
+#ifndef ENABLE_DEBUG
```
dragonmux, line 251: This wants to be `#ifndef DEBUG_INFO_IS_NOOP` to correctly guard these memory read-backs.

src/target/nrf91.c
```diff
@@ -3,18 +3,75 @@
...
 #define NRF91_NVMC_CONFIG_EEN  0x2U // Erase enable
 #define NRF91_NVMC_CONFIG_PEEN 0x3U // Partial erase enable
 
+/* https://infocenter.nordicsemi.com/topic/ps_nrf9160/dif.html */
+#define NRF91_PARTNO 0x90U
+
+#define NRF91_CTRL_AP_RESET          ADIV5_AP_REG(0x000)
+#define NRF91_CTRL_AP_ERASEALL       ADIV5_AP_REG(0x004)
+#define NRF91_CTRL_IDR_EXPECTED      0x12880000
+#define NRF91_AHB_AP_IDR_EXPECTED    0x84770001
```
dragonmux, line 28: These are missing their `U` suffixes to ensure they're defined as unsigned values. Because the AP IDR value (as an example) sets the sign bit on 32-bit platforms, if this value is then widened for any reason, then the upper bytes of the new value will be all set high, and if it is narrowed, this invokes UB (due to changing the value of the sign bit). Neither of these problems occur with unsigned constants.

src/target/nrf91.c
```diff
@@ -49,6 +106,25 @@ static bool nrf91_flash_erase(target_flash_s *flash, target_addr_t addr, size_t
...
+	for (size_t offset = 0; offset < len; offset += 4) {
+		if (target_mem_read32(target, addr + offset) != NRF91_UICR_ERASED_VAL) {
+			erase_needed = true;
+			break;
+		}
+	}
+
+	if (erase_needed) {
+		gdb_out("Skipping UICR erase, mass erase might be needed\n");
+	}
```
dragonmux, line 124: The braces are superfluous, please drop them.

dragonmux (commented):
Apologies this PR got neglected and the code review we'd started some time back in March didn't get posted. If you could please rebase this PR on `main` and fix the lint pass issues along with this new review, that would be greatly appreciated as we would like to be able to land this before v2.0

src/target/nrf91.c
dragonmux: Please note that the choice of suggesting `ID_NRF91` is for consistency with all the other target support implementations - while this is at least now a define that describes what the magic number is, it would be best to have it use the same style as the other target implementations.

PR #1788, opened 2024-03-22
Add CH579 target support

ArcaneNibble:
## Detailed description

This (WIP) pull request adds target support for the [CH579](https://wch-ic.com/products/CH579.html) microcontrollers, which feature a Cortex-M0, BLE, and 10BASE-T Ethernet as "flagship" features. This is one of the WCH / Nanjing Qinheng Microelectronics chips, but its flash interface seems to be completely different from the CH32F1 series which has existing support in the codebase.

This PR adds a basic memory map and flash programming, including support for programming the part of the "info flash" that you're "supposed to" be able to reprogram. More specifically:
* Code flash from [0x00000-0x3E7FF] as well as data flash from [0x3E800-0x3EFFF] can be programmed and are treated as a contiguous region without distinction. 
* The first page of info flash, [0x40000-0x401FF] can be programmed only via target-specific commands (i.e. is not treated as part of the normal flash). The only officially-documented contents of this flash page is the word at 0x40010 which is documented in Table 2-3 of the datasheet and which contains reset enable, SWD enable, code readout protection, and ISP bootloader enable flags. **NOTE**: code readout protection appears to imply SWD disable, even though the datasheet doesn't explicitly say this. Don't be me and brick your chip. This appears to be loaded either by hardware or by hidden boot code (i.e. not the ISP bootloader which will be mentioned extensively later). Addresses 0x40008 and 0x4000c contain undocumented configuration flags for the ISP bootloader. bit0 of 0x4000c seems to switch the ISP bootloader's "force enter" pin from PB22 to PB11 (neither of which are mentioned in the datasheet, but the "CH579M-EVT-R1-1v4.FP" devkit is wired up for the PB22 option). 0x40008 contains an address that the ISP bootloader will jump to in some situation which I have not fully reverse engineered yet. The remainder of the page appears to be writable but does not appear to be used. **The command for flashing the info page is not documented in the datasheet but instead has been reverse engineered from the ISP bootloader.**
* The ISP bootloader from [0x3F000-0x3FFFF] cannot be programmed. It does not sound like you are actually "supposed to" be able to program this, but you actually can as long as the bootloader is still "active".
* The second page of info flash, [0x40200-0x403FF] cannot be programmed. This page officially contains a MAC address and temperature sensor calibration (read from a memory mirror at 0x40600), and it possibly contains other calibration which has not been reverse engineered yet. It does not respond to either the standard page program command or the undocumented info flash program command, so there is probably yet another hidden command which I have not found.

The main improvements that need to be made to this PR (other than cleanup) is to check how many other WCH parts have the same flash IP that this code can then support.

### Bootloader, and other quirks

The ISP bootloader implements the protocol described [here](https://github.com/frank-zago/isp55e0/blob/master/protocol.txt). The bootloader will be active only if bit 6 of 0x40010 is set **and** if no relevant parts of the chip have been soft reset. When it is active, the bootloader will overlay itself over [0x00000-0x3EFFF] repeatedly. When this is the case, the bootloader can be overwritten by directing flash erase/write commands to its address range (which the bootloader code will reject in software, but which the hardware itself accepts).

If "some part" of the chip is ever reset (I haven't fully reverse engineered which part, but I know that `target_reset` in BMP is enough to cause it), the bootloader overlay will disable itself (regardless of the value at 0x40010). This means that, in this PR as it currently is, if the bootloader is enabled, the probe can break into bootloader code. The probe can also write to the bootloader region by manually poking registers. However, any use of a flash function will make the bootloader disappear until a hard power cycle (as the code path ends up calling `target_reset`).

I haven't tested this, but, from static analysis, I believe the bootloader will disable itself and reset into user code as long as the first word of flash isn't 0 nor 0xffffffff (or if the "force enter" pin is triggered). Disabling the bootloader in 0x40010 does not look like it should be necessary.

Note that only the CH579M has SWD. The CH579F does not have SWD.


src/target/ch579.c
```diff
@@ -0,0 +1,139 @@
...
+static bool ch579_flash_erase(target_flash_s *f, target_addr_t addr, size_t len)
+{
+    (void)len;
+    DEBUG_INFO("ch579 flash erase %08x\n", (int)addr);
+    target_mem_write32(f->t, 0x40001804, addr);
+    target_mem_write8(f->t, 0x40001808, 0xa6);
+
+    if (!ch579_wait_flash(f->t, NULL))
+        return false;
+    return true;
```
dragonmux, line 30: The result of the call is already a boolean, `clang-tidy` has a lint about this exact pattern - you can just `return ch578_wait_flash(flash->t, NULL);` here, simplifying this.

src/target/ch579.c
```diff
@@ -0,0 +1,139 @@
...
+}
+static bool ch579_flash_write(target_flash_s *f, target_addr_t dest, const void *src, size_t len)
```
dragonmux, line 32: Missing new line before this one - running `clang-format` should fix that.

src/target/ch579.c
```diff
@@ -0,0 +1,139 @@
...
+}
+
+static bool ch579_cmd_erase_info_dangerous(target_s *target, int argc, const char **argv)
+{
+    (void)argc;
+    (void)argv;
+    target_mem_write8(target, 0x40001809, 0b10001100);
+    target_mem_write32(target, 0x40001804, 0x40000);
+    target_mem_write8(target, 0x40001808, 0xa5);
+    bool okay = ch579_wait_flash(target, NULL);
```
dragonmux, line 65: This can be made `const` so should be.

src/target/ch579.c
```diff
@@ -0,0 +1,139 @@
...
+    target_mem_write8(target, 0x40001809, 0b10000000);
+    return okay;
+}
+
+const command_s ch579_cmd_list[] = {
+	{"void_warranty_erase_infoflash", ch579_cmd_erase_info_dangerous, "Erase info flash sector"},
+	{"void_warranty_write_infoflash", ch579_cmd_write_info_dangerous, "Write to info flash: [address] [value]"},
+    {"disable_bootloader", ch579_cmd_disable_bootloader, "Disables ISP bootloader"},
+	{NULL, NULL, NULL},
+};
```
dragonmux, line 106: This list should be defined at the top of the file, like the [one in the STM32H5 support](https://github.com/blackmagic-debug/blackmagic/blob/d9a4030b94fe9cf95dc51c8b28d802b41a8ab89f/src/target/stm32h5.c#L121)

src/target/ch579.c
```diff
@@ -0,0 +1,139 @@
...
+
+
+bool ch579_probe(target_s *target)
```
dragonmux, line 109: We know that you've put this all the way down here to save on function prototypes, but please don't.

We keep referencing the STM32H5 support as it provides a clear guide on the style we expect from target support - the prototypes exist so the code gets ordered more logically: probe handling, Flash support, monitor commands. It also provides consistency between the different target support implementations.

Likewise, it demonstrates the define prefixing and naming scheme so we can have named addresses and bit values rather than undocumented magic numbers everywhere for the next person who has to touch this code, which could be yourself in 6mo after you've forgotten what it all means.

src/target/ch579.c
```diff
@@ -0,0 +1,139 @@
...
+{
+    DEBUG_INFO("ch579 probe\n");
```
dragonmux, line 111: Please drop this. `PROBE()` already outputs a line indicating entry to this function.

dragonmux (changes requested):
We've done an initial review - overall it's looking promising, but needs some stylistic fixes and documentation.

ArcaneNibble:
Hi, my friend @jediminer543 very kindly worked with me to tidy this PR up. I am now marking this as no longer draft.

ArcaneNibble:
I have no idea why mingw is failing (I understand that Windows doesn't support weak symbols, but I copy-pasted exactly an existing example)

dragonmux:
That would be because while you added the new target to the *Meson* build system you didn't also do so to the Makefile build system - so the .c file for the target is never built under the old build system leading to that link failure.

ArcaneNibble:
I don't know why a variant that isn't enabled by default is causing code size overflow (in size-diff). As far as I can tell, head is already broken.

ArcaneNibble:
Rebased to current main branch, unsure if the code size overflow will go away or not

ArcaneNibble:
*sigh* latest CI/CD error seems to be flakiness in the infrastructure?

dragonmux:
Yes, don't worry about that macOS GCC 13 failure, it doesn't count against you.

ArcaneNibble:
Is there anything else that needs to be done to this before it is ready for merge?

dragonmux:
We'll try to review it in the morning and let you know - the only major thing that immediately pops out is that things like `clang-format` application are currently their own commits - we do not squash merge, so these fixups need to be slipstreamed into the preceding commits so the commits can be cleanly unwound for `git bisect` and `git revert` purposes if at all possible.

Thank you for getting back to this and for your patience - we'll do our best not to keep you waiting on a review beyond what's necessary.

ArcaneNibble:
I've rewritten the history to have fewer, logically separated commits

dragonmux (approved):
There are still some magic numbers and stylistic issues, but nothing massive - this is looking very significantly better. We can run some clean-up later, let's get this merged so people more generally can start testing it. Thank you for the contribution! LGTM, merging.

PR #1790, opened 2024-03-24
[WIP] Feature: correct flash mappings of STM32H7Bx

ALTracer:
## Detailed description

* This is a small new feature (target flavour support).
* The existing problem is that STM32H7Bx chips are detected by BMD with wrong (of H72x) flash mapping.
* This PR solves this problem by adding a switch-case entry for them which adds a correct flash map.

This is somewhat WIP; I would like to see CI passing. I only added the small patch/hunk from early February. If you allow some time before merging (up to next weekend), then I see an easy way to also add H7A3/H7B3 flash, which has not 16 but 128 sectors of 8 KiB. Note that I added an explicit H72x case before default case. I cannot test on any H7Bx parts.

The recently announced STM32H7Rx/H7Sx which hit General Availibility also seem to have the newer Flash IP with 64 KiB of 8-KiB pages, and a part ID I could fish out from the public reference manuals. Or this could be its own very miniature PR.
* [x] STM32H7B0
* [ ] STM32H7B3/H7A3
* [ ] STM32H7R7/H7S7

## Closing issues

I have not found GH issues, but some users on Discord reported problems with H7Bx recently.

### DIFF
--- a/src/target/stm32h7.c
+++ b/src/target/stm32h7.c
@@ -273,6 +273,12 @@ bool stm32h7_probe(target_s *target)
 		}
 		break;
 	}
+	case ID_STM32H7Bx: {
+		/* STM32H7B0nB: 128 KiB in 16 sectors of 8 KiB */
+		stm32h7_add_flash(target, STM32H7_FLASH_BANK1_BASE, 0x20000U, 0x2000U);
+		break;
+	}
+	case ID_STM32H72x:
 	default:
 		stm32h7_add_flash(target, STM32H7_FLASH_BANK1_BASE, STM32H7_FLASH_BANK_SIZE, FLASH_SECTOR_SIZE);
 		stm32h7_add_flash(target, STM32H7_FLASH_BANK2_BASE, STM32H7_FLASH_BANK_SIZE, FLASH_SECTOR_SIZE);


src/target/stm32h7.c
```diff
@@ -273,6 +273,12 @@ bool stm32h7_probe(target_s *target)
 		}
 		break;
 	}
+	case ID_STM32H7Bx: {
```
dragonmux, line 276: Unless you're intending to introduce new variables into the scope this creates, please drop the extra braces as they're not required. The code base only uses them for scope creation in case statements.

dragonmux (commented):
Only the one review item, this is looking great though.

src/target/stm32h7.c
ALTracer: There is a need for an extra `flash_size` scoped variable to support *both* H7A3/H7B3 and H7B0, so I'm keeping it until next commit (for a cleaner diff).

ALTracer:
There are a few constants you may want to have hoisted up into preprocessor defines: flash bank size, erase sector size, product total flash size, and F_SIZE address; most of which do not match H72x/H74x. I can wait on your suggestions of properly updating all that.
```diff
-#define STM32H7_FLASH_SIZE 0x1ff1e800U
+#define STM32H7_FLASH_SIZE 0x1ff1e880U
+#define STM32H7B_FLASH_SIZE 0x08fff80cU
```
I think it's a bug to read F_SIZE from UID base, where wafer X/Y coordinates (16-bit, small number) are. This also might mean that everything is matched by the fallback branch and gets "dual-banked".

Finally, texane/stlink and OpenOCD use H7Ax as main name suffix, why does BMD use H7Bx? Is this difference important?

ALTracer:
I pondered at the code some more and found FLASH_CR bit layout mismatches between H74x/H72x (266-bit) and H7Bx (137-bit) Flash IP. The code looks correct for H74x but will fail to erase sectors or banks of H7Bx. Mass erase, which is implemented as simultaneous bank erase in this driver, will touch bit 7, but should touch bit 5, and SNB/SSN is shifted. The sector select logic can be updated to use `target_flash->blocksize`, but it's not helpful on its own right now.
The driver of H503/H563 actually looks closer to what's needed, sans mass-erase & bank select bits.
Maybe we should NULL out `target_flash->flash_erase` if a H7Bx is detected in the meantime? `flash_write` will try to flip parallelism bits but end up in FW (forced write) and START [erase] (not asking for sector or bank erase). Keeping the regions' mapping is still useful for nudging GDB to auto-use hw breakpoints.

ALTracer:
I implemented flash driver (erase/write) changes required for H7Bx and tested `(gdb) load` ops on Nucleo-H743ZI2 (for regression check) using either `blackpill-f411ce` 8MHz SWD directly or via BMDA, and it passed `compare-sections` for 1.5 MiB of used binary (RTOS). Achieved flash write speed is 75 KiB/s as reported by gdb, read 292 KiB/s as reported by bmd_crc32; BMDA is 3-4 times slower.

Please test on H7B0/H7B3/H7A3 that it indeed initiates sector erases and programs many consecutive sectors in both banks (or merge opportunistically). I can't provide that in at least the next two weeks. 
*Should I update the copyright notice by now? (#1710 and others)*

ALTracer:
After a somewhat eventful testing on a live board, that is [WeActStudio.MiniSTM32H7B0](https://github.com/WeActStudio/WeActStudio.MiniSTM32H7B0), I noticed a few errors and mitigated some of them with more changes.
1. F_SIZE word at `0x08fff80c` contains `0x00010080` which then after left-shift by 10 (scaling by 1024) becomes giant and fails to match any of known bank sizes. Per ref.man we should only be interested in lower 16 bits, and on a little-endian MCU I decided that a `target_mem_read16()` will suffice. If adapter backends misbehave, this can be reverted to `target_mem_read32()` plus masking. After that 128 KiB came online in the memory map. CubeProgrammer and dfu-util think there's 2 megabytes, but I didn't bother writing into them, only reading (and they are 0xff).
2. Driver wants to reset ACR to 0, this destabilizes H7B0 flash. Chip starts up off HSI64 in VOS3 with FLASH->ACR = 0x13 for 3 wait-states, so I reused the value and checked it doesn't impact programming performance with USB FS backends. Per H74x ref.man. they even have an ACR of 0x37 for fairly conservative 7 wait-states despite spinning up also off HSI64 in VOS3, i.e. something like 0x11 would be enough. But that will require testing on more boards.
3. Some of initial loads triggered a Double ECC error which is not sticky but driver does not acknowledge the bit by clearing it. An `erase_mass` also works to remove it (but not `erase_range`). Interesting register operations are visible at DEBUG_INFO and DEBUG_TARGET loglevels.

This is now certainly ready for review and tests on H7A3/H7B3 DK or Nucleo-144 boards.

src/target/stm32h7.c
```diff
@@ -343,25 +371,55 @@ static bool stm32h7_flash_unlock(target_s *const target, const uint32_t addr)
 	return !(target_mem_read32(target, regbase + FLASH_CR) & FLASH_CR_LOCK);
 }
 
+/* Helper for offsetting FLASH_CR bits correctly */
+static uint32_t stm32h7_flash_cr(uint32_t sector_size, const uint32_t ctrl, int snb)
+{
+	uint32_t command = ctrl;
+	/* H74x, H72x IP: 128 KiB and has PSIZE */
+	if (sector_size == FLASH_SECTOR_SIZE) {
+		command = ctrl | (snb * FLASH_CR_SNB_1);
```
dragonmux, line 380: Given `command` is already a copy of `ctrl` at this point, do you see lower Flash usage or any change if you use `|=` instead of `= ctrl |`?

src/target/stm32h7.c
```diff
@@ -343,25 +371,55 @@ static bool stm32h7_flash_unlock(target_s *const target, const uint32_t addr)
...
+		return command;
+	}
+
+	/* H7Bx IP: 8 KiB and no PSIZE */
+	/* Save and right-shift FW, START bits */
+	const uint32_t temp_fw_start = command & (FLASH_CR_FW | FLASH_CR_START);
+	/* Parallelism is ignored */
+	command &= ~(FLASH_CR_PSIZE64 | FLASH_CR_FW | FLASH_CR_START);
+	/* Restore FW, START to H7Bx-correct bits */
+	command |= (temp_fw_start >> 2U);
```
dragonmux, line 391: The parens here are unnecessary, please drop them.

dragonmux (changes requested):
There are a few items to address, but nothing horrid or major. This PR's looking very close to mergable and we look forward to getting to do so.

src/target/stm32h7.c
```diff
@@ -404,7 +404,10 @@ static bool stm32h7_flash_erase(target_flash_s *const target_flash, target_addr_
 	if (!stm32h7_flash_unlock(target, addr))
 		return false;
 	/* We come out of reset with HSI 64 MHz. Adapt FLASH_ACR.*/
-	target_mem_write32(target, flash->regbase + FLASH_ACR, 0);
+	uint32_t acr = 0;
+	/* H7Bx starts up in VOS3 from HSI 64 MHz with default ACR of 0x13 */
+	if (sector_size == 0x2000U) acr = 0x13;
```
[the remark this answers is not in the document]
ALTracer: Can do. By the way, do you think H74x will benefit from increasing their ACR to 0x11 or all the way to 0x37 wrt stability/correctness? Or I could just always write 0x13 for all families if only it worked (less code complexity, less flash space)

src/target/stm32h7.c
```diff
@@ -343,25 +371,55 @@ static bool stm32h7_flash_unlock(target_s *const target, const uint32_t addr)
 	return !(target_mem_read32(target, regbase + FLASH_CR) & FLASH_CR_LOCK);
 }
 
+/* Helper for offsetting FLASH_CR bits correctly */
+static uint32_t stm32h7_flash_cr(uint32_t sector_size, const uint32_t ctrl, int snb)
```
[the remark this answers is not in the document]
ALTracer: Ref.mans state either SNB or SSN for Sector Selection Number bitfield FLASH_CR. Inside BMD driver it's a small unsigned integer 0..7 (0..127) depending on exact part. Previously it was multiplied by some offset bit to land in place, and I didn't feel like introducing more macros; but if that's more readable, then please suggest an alternative to `FLASH_CR_SNB_1` counterpart for H7Bx. I can make it `uint32_t snb` or `unsigned int` or whatever, it can be `const`, the caller doesn't see this, the callee doesn't modify it.

src/target/stm32h7.c
ALTracer: Dropped a few parens. I don't get any highlighting for extraneous parens, so I may add more than what's enough per C rules (and `make clang-tidy` sucks with all the noise from headers)

src/target/stm32h7.c
dragonmux: We'd prefer the `|=` variant then, it's clearer what the intended meaning is and that `ctrl`'s value is already in `command` intentionally. Compiler looks to have done the pivot for us anyway, which is nice, but we'd like to not rely on that.

src/target/stm32h7.c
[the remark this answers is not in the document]
dragonmux: Without looking in the TRM, we don't recall what 0x37 and 0x11 for that control value mean - but if you think it will prevent the micro getting destabilised during Flash programming then we see no harm. We do wonder why the code was modifying a control value like that rather than respecting w/e the user has set already.

src/target/stm32h7.c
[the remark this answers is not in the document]
dragonmux: Please go with `const uint8_t sector_number` then - this should avoid all the unpleasantness and gives a more meaningful name. The shift below is not the problem, the cryptic naming is, along with the bad type choice.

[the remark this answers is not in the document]
dragonmux: Having been poking this a bit ourself for the H723/H733, we additionally suggest replacing `FLASH_CR_SNB_1` with `FLASH_CR_SNB_SHIFT` which gets defined as 8U, and turning the multiplication into a right shift. It makes the whole lot make a bit more sense we think.

src/target/stm32h7.c
```diff
@@ -343,25 +371,56 @@ static bool stm32h7_flash_unlock(target_s *const target, const uint32_t addr)
...
 
-	for (size_t begin_sector = addr / FLASH_SECTOR_SIZE; begin_sector <= end_sector; ++begin_sector) {
+	for (size_t begin_sector = addr / sector_size; begin_sector <= end_sector; ++begin_sector) {
 		/* Erase the current Flash sector */
-		const uint32_t ctrl = (psize * FLASH_CR_PSIZE16) | FLASH_CR_SER | (begin_sector * FLASH_CR_SNB_1);
+		const uint32_t ctrl = stm32h7_flash_cr(sector_size, (psize * FLASH_CR_PSIZE16) | FLASH_CR_SER, begin_sector);
 		target_mem_write32(target, reg_base + FLASH_CR, ctrl);
-		target_mem_write32(target, reg_base + FLASH_CR, ctrl | FLASH_CR_START);
+		const uint32_t ctrl_start = stm32h7_flash_cr(sector_size, ctrl | FLASH_CR_START, begin_sector);
+		target_mem_write32(target, reg_base + FLASH_CR, ctrl_start);
```
dragonmux, line 423: This results in the same `ctrl` value as before with just ` | FLASH_CR_START`, not sure this change is necessary.

ALTracer: This existed in the driver. See RM0433, 4.3.10 Flash Erase Operations. START1 bit may be set *after* SER1 & SNB1 bitfields are set. There is a note that BER1+START1 can be combined for Bank erase, but I'm unsure about Sector erase. This is in all 4 refmans.

dragonmux: Yes, we're not saying eliminate the register write - we're saying eliminate the recomputation of `ctrl`. you should still be able to just `target_mem_write32(target, reg_base + FLASH_CR, ctrl | FLASH_CR_START);` and that cheapens the codegen here.

ALTracer: As much as I'd like to aim at smaller codegen, START1 is BIT(7) for H74x/H72x, but BIT(5) for H7Bx (and H7RS). Hence the extra call to the shifter to deal with it.

dragonmux: Ohhh, we see.. ok, could you please add a comment on the front of this call in each of the places it's made to explain this? We completely missed that detail when reading through this today and that seems like a very important thing not to forget.

It is valid to eliminate the `ctrl_start` variable at least by chaining the `stm32h7_flash_cr()` and `target_mem_write32()` calls though so that might still be a reasonable idea to try and minimise the chances of extra pessimistic codegen here?

Otherwise feel free to disregard this review thread beyond asking for extra comments to guard future readers from making the same mistake.

ALTracer: I trust the compiler to optimize away a stack scope const variable (i.e. keep it in gp register only) at -Os. It also helps me with readability (otherwise I'd have to stare at a very long line of code). I added the comment about START1 position.
In the other case (page program) the second invocation of helper is actually unnecessary -- dropped it. First invocation would unflip PSIZE bits, and _PG position is identical.

src/target/stm32h7.c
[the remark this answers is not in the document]
ALTracer: Renamed argument and introduced macros.

src/target/stm32h7.c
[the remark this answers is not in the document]
ALTracer: Let's just say H7Bx can't practically operate with an ACR of 0 (under a running firmware or after reset). Whether touching it is required at all -- is a better question, and I leave it to you and other part users to test.

dragonmux (approved):
This all looks good now, thank you for your hard work and effort. We'll get this merged and rebase the other H7 fix branch over top so H72x/H73x flashing work.

PR #1802, opened 2024-04-19
Fix STLINK/V2 operation with multi-AP chips and block reads

ALTracer:
## Detailed description

* This is not a feature.
* The existing problems are 1) BMDA over STLINK/V2 (incl. onboard) adapter with original firmware failing to scan/attach to STM32MP15x; 2) RTT not working over these adapters (with errors "Too large!").
* The PR solves that by carefully using what's already there (apsel) and raising the block read limit to at least 4096.

On DK1, its onboard V2-1 is the only way to access and debug the SoC, but BMDA's support for that adapter used to assume AP 0 is the only one usable, which flies for simpler MCUs but not for this MPU -- it has AXI-AP at 0 with no debug components. A bug in the backend then resets `stlink.apsel` to 0 and next accesses targeted at Cortex-A (halt and) detection via APB-D AP 1 land at AP 0 and lock it up, rendering further debug impossible until next power-on reset.
On many Discovery Kits and (cheaper/older) Nucleo-64 boards their V2-1 is a convenient way to collect RTT logs even when the target MCU does not implement TraceSWO (CM0+) or the wire is not connected by default (32F429i-DISC1). However, it issues block reads longer than 64 bytes, which triggers an unexplained failsafe in the backend. I first encountered this in July 2023 when I was benchmarking BMF-internal RTT.
The two patches allow using onboard stlinks with BMDA again out-of-the-box, without resorting to reflashing them to any alternative firmware.
Caveat SWD multidrop, I don't have a working solution for LINERESET+TARGETSEL+READ DPIDR sequence, where the second operation currently throws an uncaught C sjlj "exception" from stlink not seeing an "expected" DP Ack. Hence SWD with F7, H7, L5, MP15 is still broken (JTAG is not, unless some chip errata render it incompatible).

## Closing issues

Fixes support of MP15. Should fix support of H7 and others. Fixes support of RTT.

### DIFF
--- a/src/platforms/hosted/stlinkv2.c
+++ b/src/platforms/hosted/stlinkv2.c
@@ -732,11 +732,14 @@ static void stlink_mem_write(
 	if (!stlink_ensure_ap(ap->apsel))
 		raise_exception(EXCEPTION_ERROR, "ST-Link AP selection error");
 
+	DEBUG_PROBE("%s: @0x%08" PRIx32 "+%zu\n", __func__, dest, len);
+
 	const uint8_t *const data = (const uint8_t *)src;
-	/* Chunk the write up into stlink.block_size blocks */
-	for (size_t offset = 0; offset < len; offset += stlink.block_size) {
+	const uint16_t block_size = (align == ALIGN_8BIT) ? stlink.block_size : STLINK_READMEM_32BIT_MAX_SIZE;
+	/* Chunk the write up into firmware-digestible blocks */
+	for (size_t offset = 0; offset < len; offset += block_size) {
 		/* Figure out how many bytes are in the block and at what start address */
-		const size_t amount = MIN(len - offset, stlink.block_size);
+		const size_t amount = MIN(len - offset, block_size);
 		const uint32_t addr = dest + offset;
 		/* Now generate an appropriate access packet */
 		stlink_mem_command_s command;
--- a/src/platforms/hosted/stlinkv2.c
+++ b/src/platforms/hosted/stlinkv2.c
@@ -676,20 +676,25 @@ static void stlink_mem_read(adiv5_access_port_s *ap, void *dest, uint32_t src, s
 {
 	if (len == 0)
 		return;
-	if (len > stlink.block_size) {
-		DEBUG_WARN("Too large!\n");
-		return;
-	}
 	if (!stlink_ensure_ap(ap->apsel))
 		raise_exception(EXCEPTION_ERROR, "ST-Link AP selection error");
 
 	uint8_t type;
-	if ((src & 1U) || (len & 1U))
+	uint16_t block_size;
+	if ((src & 1U) || (len & 1U)) {
 		type = STLINK_DEBUG_READMEM_8BIT;
-	else if ((src & 3U) || (len & 3U))
+		block_size = stlink.block_size;
+	} else if ((src & 3U) || (len & 3U)) {
 		type = STLINK_DEBUG_APIV2_READMEM_16BIT;
-	else
+		block_size = STLINK_READMEM_32BIT_MAX_SIZE;
+	} else {
 		type = STLINK_DEBUG_READMEM_32BIT;
+		block_size = STLINK_READMEM_32BIT_MAX_SIZE;
+	}
+	if (len > block_size) {
+		DEBUG_ERROR("%s(AP %u @0x%08" PRIx32 "+%zu): Too large! Must be <%u\n", __func__, ap->apsel, src, len, block_size);
+		return;
+	}
 
 	/* Build the command packet and perform the access */
 	stlink_mem_command_s command = stlink_memory_access(type, src, len, ap->apsel);
--- a/src/platforms/hosted/stlinkv2.c
+++ b/src/platforms/hosted/stlinkv2.c
@@ -608,10 +608,15 @@ static int stlink_write_dp_register(const uint16_t apsel, const uint16_t address
 
 uint32_t stlink_raw_access(adiv5_debug_port_s *dp, uint8_t rnw, uint16_t addr, uint32_t request_value)
 {
-	DEBUG_PROBE("%s: Attempting access to addr %04x\n", __func__, addr);
 	uint32_t result_value = 0;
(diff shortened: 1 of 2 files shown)


src/platforms/hosted/stlinkv2.c
```diff
@@ -676,20 +678,25 @@ static void stlink_mem_read(adiv5_access_port_s *ap, void *dest, uint32_t src, s
...
-	if ((src & 1U) || (len & 1U))
+	uint16_t block_size;
+	if ((src & 1U) || (len & 1U)) {
 		type = STLINK_DEBUG_READMEM_8BIT;
-	else if ((src & 3U) || (len & 3U))
+		block_size = stlink.block_size;
+	} else if ((src & 3U) || (len & 3U)) {
 		type = STLINK_DEBUG_APIV2_READMEM_16BIT;
-	else
+		block_size = 4096;
```
dragonmux, line 691: This and the one below are missing their `U`'s. It might be helpful to document why this is OK and where this block size comes from - no objections otherwise.

dragonmux (commented):
There's a couple of items, the major one is the `DEBUG_ERROR` one, but with them taken care of we'll be happy to merge this.

src/platforms/hosted/stlinkv2.c
```diff
@@ -676,20 +678,25 @@ static void stlink_mem_read(adiv5_access_port_s *ap, void *dest, uint32_t src, s
...
+	} else {
 		type = STLINK_DEBUG_READMEM_32BIT;
+		block_size = 4096;
+	}
+	if (len > block_size) {
+		DEBUG_ERROR("%s(*ap, %p, 0x%08X, %lu): Too large! Must be <%u.\n", __func__, dest, src, len, block_size);
```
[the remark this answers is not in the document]
ALTracer: Ah, this was written before -Wformat and before Windows builds were available and on CI. Agreed on `@0x2001ae80+768` format, `dest` on `hosted` can be anywhere, but I still would like to indicate AP number in this message (just in case). Updated. BMDA is not flash-constrained, and this line is not in hot/happy code path.
Can't easily test with `make CC=x86_64-w64-mingw32-gcc -C src/ PROBE_HOST=hosted HOSTED_BMP_ONLY=1 -j4` because it doesn't build with libusb or stlinkv2.o though. And the adapter doesn't support 64-bit [ADIv5, MEM-AP, ...] I think.

src/platforms/hosted/stlinkv2.c
ALTracer: Can add unsigned qualifiers. For size, different impls actually lean towards 6144 bytes per read, which sounds realistic for the 20/64 f103c8, but earlier firmware might only do 4096 (?). I don't have a good testbench for this, and AHB-AP won't autoincrement past 4096 byte boundaries anyways (for CM3/CM4), so this is a sensible compromise between not having any access at all and surfing the dangerous edge. *For JLink fw which has 3 buffers of 2 KiB for TMS, TDO, TDI this also lines up somewhat.* Anecdotally, writes are streamed for up to (65536-4) bytes, but don't @ me on this -- see probe-rs PR377, `src/probe/stlink/mod.rs` for example.

src/platforms/hosted/stlinkv2.c
```diff
@@ -608,10 +608,12 @@ static int stlink_write_dp_register(const uint16_t apsel, const uint16_t address
 
 uint32_t stlink_raw_access(adiv5_debug_port_s *dp, uint8_t rnw, uint16_t addr, uint32_t request_value)
 {
-	DEBUG_PROBE("%s: Attempting access to addr %04x\n", __func__, addr);
 	uint32_t result_value = 0;
-	const int result = rnw ? stlink_read_dp_register(dp, addr < 0x100U ? STLINK_DEBUG_PORT : 0U, addr, &result_value) :
-							 stlink_write_dp_register(addr < 0x100U ? STLINK_DEBUG_PORT : 0U, addr, request_value);
+	/* Accesses to registers lower than 0x100 are called with DP port value; higher with last/current apsel which may be nonzero! */
```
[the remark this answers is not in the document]
ALTracer: Comment updated, especially as the `< 0x100` check was changed to `& ADIv5_APnDP` bit masking.

src/platforms/hosted/stlinkv2.c
dragonmux: Sure, having that as a comment near the top of the file or in the ST-Link v2 protocol header and using a define for it is probably favourite then. That puts the reasoning somewhere easily found in the future that helps describe why the magic number and not, eg, the probe-rs choice, etc.. The value picked is fine, we just want to capture the reasoning somehow.

dragonmux:
Once the second half of that second review note has been taken care of, we'll be happy to merge this.

src/platforms/hosted/stlinkv2.c
ALTracer: Extracted macro and moved to protocol header. Failing to find any public docs I bruteforced the read length for a bit, ending up on 6144.

ALTracer:
Update: I took the liberty of bumping write block size, too. This increases vFlashWrite speed (by 15% against STM32G071) but might start triggering bugs in drivers not polling FPEC status registers often enough etc. due to higher data rate. I'd like to eliminate the `stlink.block_size` variable as it means either 64 ot 512 depending on `stlink.ver_hw == 30`, but maybe next time.
There are a couple extra parens in conditional assignments (aka ternary + capture), which I added after cppcheck complained with [clarifyCalculation], so please look at the style of PR again. BMDA does not support V2J31 and older because of the pre-requisite of DAP (dapdirect) commands, so I didn't bother testing that.
The read_mem32 will still trigger errors when invoked with length > 6144, which happens after bumping either `GDB_MAX_PACKET_SIZE, GDB_PACKET_BUFFER_SIZE` from 1024 to 2*6144+4=12296 (tested), or the stack buffer in `generic_crc32()` from its 4096 to 6148 (not tested). When that happens, you may want to introduce read chunking like the one you wrote in write_mem32 impl. 
For RTT it is good enough -- it wants to find (128 byte wise) and poll the buffer control block (144 bytes) then the buffer contents (1024 byte default), which is the second reason for this PR.

dragonmux (approved):

PR #1803, opened 2024-04-19
Feature: Arterytek option bytes and more part IDs

ALTracer:
## Detailed description

* This is a new feature.
* Currently there's no easy way to extend SRAM of AT32F403A via `monitor option` commands because EOPB0 is out of reach for the normal 16-byte driver.
* The PR solves it by conditionally bumping the option byte ranges.

Past first commit see my work of adding yet more part ID lists for other families in AT32F4 lineup. This is probably getting unmanageable, so I started incorporating F_SIZE and "Product ID" readouts from UID (in their DESig analog). If this is fine, e.g. does not result in problems for RDP1-enabled targets, then all such lists can be dropped or gated under ENABLE_DEBUG to not hog precious BMF flash -- it only needs to know flash parameters (banks, sector size, total size), not exact package & pinout. I may move this to a separate PR (and a split translation unit) as no users have probably asked for F421, F413 or F425. F423 and F402/F405 are plain unimplemented (F405 with USB HS PHY would be nice to have though). I have none of these chips to test on.

* [x] It builds for hardware native (see [Building the firmware](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-firmware)) ...and overflows flash

## Closing issues

Allows operating Arterytek User System Data, including Flash Access Protection enable/disable ops.

### DIFF
--- a/src/target/stm32f1.c
+++ b/src/target/stm32f1.c
@@ -368,6 +368,54 @@ static void gd32vf1_detach(target_s *const target)
 #endif
 #endif
 
+static bool at32f403_detect(target_s *target, const uint16_t part_id)
+{
+	switch (part_id) {
+	case 0x0240U: // AT32F403ZCT6 / LQFP144
+	case 0x0241U: // AT32F403VCT6 / LQFP100
+	case 0x0242U: // AT32F403RCT6 / LQFP64
+	case 0x0243U: // AT32F403CCT6 / LQFP48
+	case 0x024eU: // AT32F403CCU6 / QFN48
+		// Flash (C): 256 KiB / 2 KiB per block
+		stm32f1_add_flash(target, STM32F1_FLASH_BANK1_BASE, 256U * 1024U, 2U * 1024U);
+		break;
+	case 0x02c8U: // AT32F403ZET6 / LQFP144
+	case 0x02c9U: // AT32F403VET6 / LQFP100
+	case 0x02caU: // AT32F403RET6 / LQFP64
+	case 0x02cbU: // AT32F403CET6 / LQFP48
+	case 0x02cdU: // AT32F403CEU6 / QFN48
+		// Flash (E): 512 KiB / 2 KiB per block
+		stm32f1_add_flash(target, STM32F1_FLASH_BANK1_BASE, 512U * 1024U, 2U * 1024U);
+		break;
+	case 0x0344U: // AT32F403ZGT6 / LQFP144
+	case 0x0345U: // AT32F403VGT6 / LQFP100
+	case 0x0346U: // AT32F403RGT6 / LQFP64
+	case 0x0347U: // AT32F403CGT6 / LQFP48
+	case 0x034cU: // AT32F403CGU6 / QFN48
+		// Flash (G): 1024 KiB / 2 KiB per block, dual-bank
+		stm32f1_add_flash(target, STM32F1_FLASH_BANK1_BASE, 512U * 1024U, 2U * 1024U);
+		stm32f1_add_flash(target, STM32F1_FLASH_BANK2_BASE, 512U * 1024U, 2U * 1024U);
+		break;
+	// Unknown/undocumented
+	default:
+		return false;
+	}
+
+	// All parts have 96 KiB SRAM
+	target_add_ram32(target, STM32F1_SRAM_BASE, 96U * 1024U);
+	target->driver = "AT32F403";
+	target->part_id = part_id;
+	target->target_options |= STM32F1_TOPT_32BIT_WRITES;
+	target->mass_erase = stm32f1_mass_erase;
+
+	// AT32F403 has 48 bytes of User System Data
+	target_add_commands(target, stm32f1_cmd_list, target->driver);
+	// TODO: SPIM 0x08400000 bank support
+
+	/* Now we have a stable debug environment, make sure the WDTs + WFI and WFE instructions can't cause problems */
+	return stm32f1_configure_dbgmcu(target, STM32F1_DBGMCU_CONFIG);
+}
+
 static bool at32f40_is_dual_bank(const uint16_t part_id)
 {
 	switch (part_id) {
@@ -493,8 +541,10 @@ bool at32f40x_probe(target_s *target)
 
 	DEBUG_TARGET("%s: idcode = %08" PRIx32 ", project_id = %02x\n", __func__, idcode, project_id);
 
-	/* 0x08: F407 (has EMAC), 0x07: F403A (only CAN+USB). */
+	/* 0x08: F407 (has EMAC), 0x07: F403A (only CAN+USB). 0x02 is the older F403 (200MHz). */
 	if (series == AT32F40_SERIES) {
+		if (project_id == 2U)
+			return at32f403_detect(target, part_id);
 		if (project_id == 7U || project_id == 8U)
 			return at32f403a_407_detect(target, part_id);
 		if (project_id == 0xffU)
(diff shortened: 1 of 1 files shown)


src/target/stm32f1.c
```diff
@@ -820,16 +955,22 @@ static bool stm32f1_option_write_erased(
 	return status == SR_PROG_ERROR;
 }
 
+#define STM32F1_OB_COUNT 8U
+#define AT32F4_OB_COUNT 24U
```
dragonmux, line 959: All `#define`'s should go at the top of the file after the `#include` block please.

dragonmux (commented):
Please can you rebase this on `main` and run `clang-format`. We'll do a proper review at that point rather than just a quick cursory as we're anticipating it's pretty close to already being mergable.

Apologies this one slipped through the cracks and we failed to review it at all till now.

ALTracer:
Rebased and clang-formatted.
I only wanted 48 option bytes for AT32F403A (especially RDP1 and EOPB0). Even that impl is bad -- it would affect GD32F1 etc. because of TOPT_32BIT criteria. The stm32f1-generic flash driver does not record option byte region size. If you have any better ideas -- I'll try to edit them in for a minor but effective change (no large driver-scope refactoring).
I would rather migrate/move the rest of chips/targets into at32f43x.c sometime in the future, but that should be made more generic as well to handle that. And it'd shrink the stm32f1.o (always compiled).

src/target/stm32f1.c
```diff
@@ -418,11 +428,15 @@ static bool at32f40_detect(target_s *target, const uint16_t part_id)
 	target->target_options |= STM32F1_TOPT_32BIT_WRITES;
 	target->mass_erase = stm32f1_mass_erase;
 
+	// AT32F403A/F407 have 48 bytes of User System Data
```
dragonmux, line 431: This is a useful tidbit to know but we don't see how it relates to the line after it or how it impacts things. Please flesh this out as a comment perhaps around line 424 as part facts and how they reflect onto the implementation.

src/target/stm32f1.c
```diff
@@ -457,11 +471,112 @@ static bool at32f41_detect(target_s *target, const uint16_t part_id)
...
+	target->part_id = part_id;
+	target->target_options |= STM32F1_TOPT_32BIT_WRITES;
+	target->mass_erase = stm32f1_mass_erase;
+	// TODO: AT32F421 have 512 bytes of User System Data
+	return stm32f1_configure_dbgmcu(target, STM32F1_DBGMCU_CONFIG);
+}
+
+static bool at32f425_detect(target_s *target, const uint16_t part_id)
+{
+	const uint16_t flash_size = target_mem32_read16(target, AT32F4x_FLASHSIZE);
```
dragonmux, line 568: This doesn't appear to check any ID codes before going "yeah, sure, it's a AT32F425"?

src/target/stm32f1.c
```diff
@@ -473,11 +588,33 @@ bool at32f40x_probe(target_s *target)
...
+			return at32f415_detect(target, part_id);
+	}
+	// Value line
+	if ((series == AT32F421_SERIES || series == AT32F421_SERIES_16KB) && (project_id == 9U))
+		return at32f421_detect(target, part_id);
+
+	if (series == AT32F425_SERIES && project_id == 0x0fU)
+		return at32f425_detect(target, part_id);
+	if ((series == AT32F423_SERIES_256KB || series == AT32F423_SERIES_128KB || series == AT32F423_SERIES_64KB) &&
+		(project_id == 0x12U))
```
dragonmux, line 613: The extra parens here are superfluous and can be dropped - same on line 607 for the clause after the `&&`. They suppress the GCC warning about accidental assignment that we want here.

dragonmux (commented):
There are a couple of remaining review notes to deal with, and once they've been sorted we're happy for this to merge.

src/target/stm32f1.c
ALTracer: Correct, because the checking was done by its caller against `0x5009_2???` series. This function is implemented simpler because my theory is F_SIZE might be enough for BMD purposes.

src/target/stm32f1.c
ALTracer: Removed two pairs of parens for project_id.

ALTracer:
While I don't think it ends good for GD32 in current form, I'd like to ask to `if(!strcmp(target->driver, "AT32"))` instead of TOPT so both would work as expected.
Rebased away the addition of AT32F405/F402 (HS USB) because their FPEC is at 0x40023C00 like F435/F437, not 0x40022000 as this driver is hardcoded to work on. Same for AT32F423 (three series, FPEC at incompatible offset, unsupported here).

src/target/stm32f1.c
dragonmux: Ah, okie - in that case perhaps prefix this with `NB: ` or so to indicate that you're delivering some extra information?

src/target/stm32f1.c
ALTracer: Moved up.

dragonmux:
Once the single outstanding review note has been sorted, we're happy to get this merged. This is looking really good.

src/target/stm32f1.c
ALTracer: Expanded oneline comment to carry more meaning.

ALTracer:
Replaced the option byte size dispatch check from TOPT 32-bit writes (shared with Gigadevices) to direct strncmp. This triggers correctly, 48 bytes (24 halfwords) for AT32F403A/407, but still 16 bytes (8 halfwords) for GD32F103 etc.
Did not register cmd_option for older F403 or newer F413, as I cannot test on them. Did not register it for F415 either, because it has 1024 not 48 bytes.
Added a few more families very likely compatible with STM32F1 flash driver. They all have UID/F_SIZE unreadable under RDP1, so need part ID tables. I started to notice a pattern "part ID <=> flash density", but the optimization would be handled in a follow-up PR, if ever.
Please review again.

src/target/stm32f1.c
```diff
@@ -459,11 +519,143 @@ static bool at32f41_detect(target_s *target, const uint16_t part_id)
...
+	target->part_id = part_id;
+	target->target_options |= STM32F1_TOPT_32BIT_WRITES;
+	target->mass_erase = stm32f1_mass_erase;
+	// TODO: AT32F421 have 512 bytes of User System Data
+	return stm32f1_configure_dbgmcu(target, STM32F1_DBGMCU_CONFIG);
+}
+
+static bool at32f425_detect(target_s *target, const uint16_t part_id)
+{
+#if 1
```
dragonmux, line 618: This preprocessor logic makes the `#else` part moot - please don't include dead code if at all possible. If what's needed can be achieved with just this 16-bit register read, then let's do that instead of having a large part ID switch-case.

dragonmux (changes requested):
Having properly re-reviewed this, there's just one thing we found. Please address it and rebase on `main` then we'll approve this for merge.

src/target/stm32f1.c
ALTracer: No, there is a major problem of us not knowing flash size when RDP1 is active. Part IDs allow that. I would rather delete the easy two-liner and postpone it to the refactor, relying on part ID tables like all other routines (this table is thankfully small). We previously argued that read-locked flash should not be added anyways, just the option byte erase handler; but it's not implemented for 512 byte region.

ALTracer: Inverted macro to gate the easy path and added TODO comments.

dragonmux (approved):
LGTM, all items addressed. Merging. Thank you for the contribution!

PR #1804, opened 2024-04-21
Fix remaining length calculation for 'X' 'M' and 'vFlashWrite'

TechnoMancer:
## Detailed description

When removing use of sscanf from gdb_main the calculation of the length of the remaining data portion of the 'X', 'M', and 'vFlashWrite' packet types was done wrongly. The difference between the start of data pointer and start of buffer pointer was reversed resulting in adding this to the total length rather than subtracting it.  For the 'X' and 'M' packets this causes them to accept lengths longer than the supplied data but correct packets would still function as expected. For the 'vFlashWrite' packet the length passed to the flash write function was about 8 larger than it should have been which could cause problems.

The calculation of the data length portion of the packets has been corrected so all three now work as expected.

Tested vFlashWrite with '-v 2' on BMDA to confirm the correct lengths.
Tested the 'M' and 'X' packets on both BMDA and native firmware by attempting packets with overly long length field for how much data is present and checking that they fail now while the correct length produces no errors.

### DIFF
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -168,7 +168,7 @@ int32_t gdb_main_loop(target_controller_s *tc, char *pbuf, size_t pbuf_size, siz
 		uint32_t len = 0;
 		ERROR_IF_NO_TARGET();
 		if (read_hex32(pbuf + 1, &rest, &addr, ',') && read_hex32(rest, &rest, &len, ':')) {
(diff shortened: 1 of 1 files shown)


dragonmux (approved):
LGTM, thanks for finding and fixing this with ALTracer's query about some funky Flash behaviour! Merging.

PR #1806, opened 2024-04-26
Fix flash for STM32U5xx

rob-the-dude:
The flash size register for the STM32U5xx was in a different place, but all the other registers were compatible with the L5.  So I modified the way the flash size was loaded (moving the register definition to its own location in the device structure), and then revised the code that calls it (turns out that the only read16 call was for flash size, so I changed the name).

That was enough to get it working for U5xx, but it turns out that the block size was also wrong.  I've corrected the block size for U5xx, but I have not changed anything else.  However, I made a comment for where I would suggest we start tackling that problem.

Much thanks to @ALTracer and @dragonmux for helping me get this far so quickly.


dragonmux:
Before we get too deep into reviewing this, could we ask you to confirm with the checkboxes in your PR description which of the checklist steps you've completed, and please reword the commit to conform with the contribution guidelines - `stm32l4: Fix flash for STM32U5xx` - so we can see at a glance which file and so code area got modified.

src/target/stm32l4.c
```diff
@@ -187,6 +191,9 @@ typedef enum stm32l4_family {
 	STM32L4_FAMILY_WLxx,
 } stm32l4_family_e;
 
+/* recommend augmenting this structure with flash block size,
+	which would get rid of the hard-coded magic numbers in
+	stm32l4_attach, but I have not implemented that yet */
```
dragonmux, line 196: As this is a multi-line comment, please use the multi-line comment style for the code base:

```suggestion
/*
 * XXX: Recommend augmenting this structure with flash block size,
 * which would get rid of the hard-coded magic numbers in
 * stm32l4_attach, but that has not been implemented yet
 */
```

src/target/stm32l4.c
```diff
@@ -149,6 +150,9 @@ const command_s stm32l4_cmd_list[] = {
 #define DUAL_BANK     0x80U
 #define RAM_COUNT_MSK 0x07U
 
+/* TODO: add block size constants for other MCUs */
+#define STM32U5_FLASH_BLOCK_SIZE	0x2000
```
dragonmux, line 154: Please include a `U` suffix on this constant - this makes it unsigned and prevents silent bugs down the road.

dragonmux (changes requested):
Only have the two review notes, this otherwise looks great! Please make sure to run `clang-format` as there look to be some overly long lines and with the two review items addressed, we'll be happy to merge this.

rob-the-dude:
I tried to fix my commit, but something got screwed up.  I'm going to close this PR and open a new one with the recommended changes.

PR #1807, opened 2024-04-26
stm32l4: Fix flash for STM32U5xx

rob-the-dude:
The flash size register for the STM32U5xx was in a different place, but all the other registers were compatible with the L5. So I modified the way the flash size was loaded (moving the register definition to its own location in the device structure), and then revised the code that calls it (turns out that the only read16 call was for flash size, so I changed the name).

That was enough to get it working for U5xx, but it turns out that the block size was also wrong. I've corrected the block size for U5xx, but I have not changed anything else. However, I made a comment for where I would suggest we start tackling that problem.

Much thanks to @ALTracer and @dragonmux for helping me get this far so quickly.

### DIFF
--- a/src/target/stm32l4.c
+++ b/src/target/stm32l4.c
@@ -140,6 +140,7 @@ const command_s stm32l4_cmd_list[] = {
 
 #define STM32L4_FLASH_SIZE_REG 0x1fff75e0U
 #define STM32L5_FLASH_SIZE_REG 0x0bfa05e0U
+#define STM32U5_FLASH_SIZE_REG 0x0bfa07a0U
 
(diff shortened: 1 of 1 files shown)


rob-the-dude:
Hi, I ran into some git/github issues trying to fix my previous PR.  Hopefully this one gets things correct.  Looking over it now, I may have still messed up a line being too long.  Let me know if you'd like me to fix it.  I need to get more skilled at adding more commits to a PR.

dragonmux (approved):
Looks good enough to us 🙂 we can run clang-format the next time something touches this file to fix the line lengths stuff.. need to fix the hook config still to make the lint pass work properly again, and that's on us. Merging once CI completes.

PR #1812, opened 2024-04-29
Fix: Timeout mishandling issue

dragonmux:
## Detailed description

This PR addresses an issue in the platform timeouts subsystem where the firmware will execute and handle timeouts very wrongly for a few seconds after ~49.7 days of uptime.

The gist of the bug is that as the time_ms counter approaches UINT32_MAX and platform_timeout_set() starts generating expiration times close to 0, the comparison for whether the timeout has expired will incorrectly determine that it has at the start of the timeout validity interval. This is incredibly wrong and will cause all manner of broken behaviour in anything that depends on timeouts to function properly.

We fix this by detecting this edge condition where the uptime counter has the MSb set and the expiration time does not and force the comparison to false. This is wrong if a timeout is requested that is exceedingly large (~2<sup>31</sup>ms) but as the firmware never does that and that would be a ~25.8 day timeout, this edge-case is irrelevant.

### DIFF
--- a/src/platforms/common/stm32/timing_stm32.c
+++ b/src/platforms/common/stm32/timing_stm32.c
@@ -62,8 +62,8 @@ void platform_timing_init(void)
 {
 	/* Setup heartbeat timer */
 	systick_set_clocksource(STK_CSR_CLKSOURCE_AHB_DIV8);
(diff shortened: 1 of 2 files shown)


esden (approved):
LGTM

PR #1815, opened 2024-05-05
Main

gitgame081:


dragonmux:
We're not sure what you're trying to accomplish with this PR, but deleting the plugdev rules file is definitely not correct. It is required for Debian and derivatives.

We're aware there is a bug with the current rules, which we have a fix for that needs a PR - two of the rules use `ATTR` when they should use `ATTRS` - but that's not this.

dragonmux:
Given the lack of movement to explain this PR, and as it deletes a udev rules file that is very much still necessary as it stands, we're going to close this as changing something that already works as intended. If you intended to update the .rules file with new contents and miss-pushed, then you are welcome to open a new PR with the corrected commit.

PR #1817, opened 2024-05-10
Initial Puya PY32 target support

### LINKED ISSUE #1395: new target - Puya Py32x
I've been playing with a new mcu, the Puya py32f030, a cortex M0+ device, after reading about it on jaycarlson's site ( https://jaycarlson.net/2023/02/04/the-cheapest-flash-microcontroller-you-can-buy-is-actually-an-arm-cortex-m0/ ). I am able to connect to it using daplink, and programs like pyocd work with it when provided the CMSIS pack file. BMP (hosted, w/daplink) reports:

$ src/blackmagic -E
Black Magic Debug App 308aec7
 for Black Magic Probe, ST-Link v2 and v3, CMSIS-DAP, JLink and libftdi/MPSSE
Using c251:f001 LU_2022_8888 jixin.pro
 CMSIS-DAP_LU
CMSIS_DAP < v1.2 can not handle multi-drop, disabling
scan_multidrop: false
dap_read_single failed (fault = 4)
dap_read_single failed (fault = 4)
Please report unknown device with Designer 0x43b Part ID 0x4c0

Documentation (including .pack file) is available from the vendor ( https://www.puyasemi.com/uploadfiles/2022/11/PY-MCU%E8%B5%84%E6%96%99-20221117.rar ). There's also a really handy gcc/make/pyocd build environment someone set up here ( https://github.com/IOsetting/py32f0-template )

ArcaneNibble:
## Detailed description

This pull request adds initial support for [Puya PY32](https://www.puyasemi.com/en/py32_series.html) microcontrollers. These microcontrollers are notable for their low cost. The flash memory interface is very strongly inspired by the STM32 series, but it is not a drop-in compatible clone. Testing of this PR was performed on the PY32F002AW15U.

This PR adds RAM size detection and Flash programming, which unfortunately relies on an undocumented value stored in the factory calibration / reserved section of flash. This undocumented value is used by the built-in serial bootloader.

Note that the PY32F002AW15U which I have been testing on actually contains 32K of Flash and 4K of RAM instead of the listed 20K/3K, and is generally widely suspected to be the same die as the PY32F030xx6.

## Closing issues

Closes #1395.


src/target/puya.c
```diff
@@ -0,0 +1,171 @@
...
+
+	return false;
+}
+
+static bool puya_flash_prepare(target_flash_s *flash)
+{
+	target_mem32_write32(flash->t, FLASH_KEYR, FLASH_KEYR_KEY1);
+	target_mem32_write32(flash->t, FLASH_KEYR, FLASH_KEYR_KEY2);
+
+	int hsi_fs = (target_mem32_read32(flash->t, RCC_ICSCR) >> 13) & 7;
```
dragonmux, line 105: The masking operation and shifting operation should be unsigned (suffix the constants with `U`), making the result a `uint8_t` - signed bitwise operations are a very prickly part of C that are best avoided.

src/target/puya.c
```diff
@@ -0,0 +1,171 @@
+#include "general.h"
+#include "target.h"
+#include "target_internal.h"
+#include "buffer_utils.h"
+#include "cortexm.h"
+#include "adiv5.h"
+
+/* Flash */
+#define FLASH_START     0x08000000U
```
dragonmux, line 9: These definitions need to get the same suffixing treatment (`PUYA_`) as the ones from your other PR did

src/target/puya.c
```diff
@@ -0,0 +1,171 @@
+#include "general.h"
```
dragonmux, line 1: This file must have a license header included and it must go before the first `#include` please. We cannot take the contribution otherwise. Either BSD-3-Clause or GPLv3+ is fine (our preference is BSD-3-Clause, but this is up to you).

src/target/puya.c
```diff
@@ -0,0 +1,171 @@
...
+bool puya_probe(target_s *target)
+{
+	uint32_t ram_size = 0U;
+	size_t flash_size = 0U;
+
+	const uint32_t dbg_idcode = target_mem32_read32(target, DBG_IDCODE);
+	if ((dbg_idcode & 0xfff) == 0) {
+		const uint32_t flash_ram_sz = target_mem32_read32(target, FLASH_RAM_SZ);
+		flash_size = ((flash_ram_sz & 7) + 1) * 0x2000;
+		ram_size = (((flash_ram_sz >> 4) & 3) + 1) * 0x800;
```
dragonmux, line 71: All masking and shifting constants must be suffixed with `U` to make them unsigned. Rather than multiplying by a magic number, please use a left shift or at least turn the magic numbers into proper constants with a comment explaining them. (The shift is preferred to ensure the compiler won't suddenly generate an expensive software multiply shim call)

src/target/puya.c
```diff
@@ -0,0 +1,171 @@
...
+	const uint32_t eppara0 = target_mem32_read32(flash->t, FLASH_TIMING_CAL_BASE + hsi_fs * 20 + 0);
+	const uint32_t eppara1 = target_mem32_read32(flash->t, FLASH_TIMING_CAL_BASE + hsi_fs * 20 + 4);
+	const uint32_t eppara2 = target_mem32_read32(flash->t, FLASH_TIMING_CAL_BASE + hsi_fs * 20 + 8);
+	const uint32_t eppara3 = target_mem32_read32(flash->t, FLASH_TIMING_CAL_BASE + hsi_fs * 20 + 12);
+	const uint32_t eppara4 = target_mem32_read32(flash->t, FLASH_TIMING_CAL_BASE + hsi_fs * 20 + 16);
+	DEBUG_INFO("PY32 flash timing cal 0: %08" PRIx32 "\n", eppara0);
+	DEBUG_INFO("PY32 flash timing cal 1: %08" PRIx32 "\n", eppara1);
+	DEBUG_INFO("PY32 flash timing cal 2: %08" PRIx32 "\n", eppara2);
+	DEBUG_INFO("PY32 flash timing cal 3: %08" PRIx32 "\n", eppara3);
+	DEBUG_INFO("PY32 flash timing cal 4: %08" PRIx32 "\n", eppara4);
```
dragonmux, line 119: What use are these to a typical user of this? Why the INFO level? These should, if kept, go in the TARGET debugging log level. Same for the one above about the HSI.

Similarly, please explain what each of these 5 registers mean/do with your choice of variable naming.

src/target/puya.c
```diff
@@ -0,0 +1,171 @@
...
+		if (timeout)
+			target_print_progress(timeout);
+	}
+	return true;
+}
+
+static bool puya_flash_erase(target_flash_s *flash, target_addr_t addr, size_t len)
+{
+	(void)len;
+	DEBUG_INFO("PY32 flash erase @ %08" PRIx32 "\n", addr);
```
dragonmux, line 154: Please drop this and the DEBUG_INFO in write - suitable debugging is already in the target Flash layer, so this duplicates that.

src/target/puya.c
```diff
@@ -0,0 +1,171 @@
...
+
+static bool puya_wait_flash(target_s *const target, platform_timeout_s *const timeout)
+{
+	while (target_mem32_read32(target, FLASH_SR) & FLASH_SR_BSY) {
+		if (target_check_error(target))
+			return false;
+		if (timeout)
+			target_print_progress(timeout);
+	}
+	return true;
```
dragonmux, line 148: Does this not want to check the error bits in the status register on completion to check if the part's reported a fault?

src/target/puya.c
```diff
@@ -0,0 +1,171 @@
...
+	target_mem32_write32(flash->t, addr, 0);
+	return puya_wait_flash(flash->t, NULL);
+}
+
+static bool puya_flash_write(target_flash_s *flash, target_addr_t dest, const void *src, size_t len)
+{
+	(void)len;
+	DEBUG_INFO("PY32 flash write @ %08" PRIx32 "\n", dest);
+	target_mem32_write32(flash->t, FLASH_CR, FLASH_CR_PG);
+	for (uint32_t i = 0; i < FLASH_PAGE_SIZE; i += 4) {
```
dragonmux, line 165: This is not guaranteed to get `FLASH_PAGE_SIZE` bytes passed to it - `len` is allowed to be `0 < len <= FLASH_PAGE_SIZE`. It is an error to access `src` past `len` bytes.

dragonmux (changes requested):
This is looking good, though needs some signed-unsigned confusion and style fixes applied - with those and the bounds note in the Flash write routine sorted, this looks like it'll be ready for merge

src/target/puya.c
ArcaneNibble: Changed to `DEBUG_TARGET`. I have been defaulting to `DEBUG_INFO` because I could never actually manage to make `DEBUG_TARGET` work (despite `-Ddebug_output=true`)

As for the naming, this is the exact naming used in the reference manual. I'm not sure how they could be explained any better? These parameters are just copied from the calibration sector to hardware registers, and other than controlling "timing" their use isn't explained in the manual.

src/target/puya.c
ArcaneNibble: I fixed this to support an arbitrary len (including unaligned non-multiple-of-4), although this is contrary to the behavior I actually observe and also goes against the assumption made in https://github.com/blackmagic-debug/blackmagic/blob/main/src/target/msp432e4.c#L287-L292 (which is where I copied this from).

src/target/puya.c
dragonmux: `DEBUG_TARGET` is a log level that is intentionally never included in the firmware, but which is included in BMDA - you can access the various log levels with the `-v` option which then allows you to get GDB and higher log levels (the firmware only ever includes ERROR, WARN and INFO when built with debug output enabled, the main reason for this is to manage the binary size which is already very large when we do this).

src/target/puya.c
ArcaneNibble: Indeed, the code in target_flash.c only ever calls write with writesize bytes: https://github.com/blackmagic-debug/blackmagic/blob/main/src/target/target_flash.c#L206

src/target/puya.c
ALTracer: `-Ddebug_output=true`? You're doing on-probe diag logging? `DEBUG_TARGET()` is not currently compiled into in-tree firmware builds, and `DEBUG_INFO()` is. You could mangle `src/include/general.h` to override this, but I'm thinking of escalating this to include said level. It's only 3116 bytes more on latest `main` for `blackpill-f411ce` with every driver enabled: 162808-159692 `build-f411ce/blackmagic_blackpill_f411ce_firmware.bin`, and for f103 platforms even less: 120452-118256=2196 `build-swlink/blackmagic_swlink_firmware.bin` (configured as `meson setup --cross-file cross-file/swlink.ini -Drtt_support=true -Ddebug_output=true -Dtargets=cortexm,hc32,lpc,nxp,rp,stm build-swlink/`)

src/target/puya.c
dragonmux: Argh, ok, you are correct there - our memory was faulty. We'll change the request then to: rather than using the FLASH_PAGE_SIZE hard-coded constant here, please either use `flash->writesize`, or `len` to guard against both other people reading it and making the same mistake, and also in case the write size is changed at some later date so it can't get out of sync.

Don't worry about the other half of our original review note as the target Flash layer is preventing that being a problem.

ArcaneNibble:
Everything should now be resolved other than possibly renaming the epparaN variables

src/target/puya.c
dragonmux: We were thinking of something a bit more like stm32h5_flash_wait_complete(), but the given solution looks functional at least, so we'll take it. The only comment is that it does the same explicit returns thing that clang-tidy correctly explicitly warns about - we suggest making the return logic:
```c
	if (status & PUYA_FLASH_SR_WRPERR)
		DEBUG_ERROR("puya flash erase error: sr 0x%" PRIx32 "\n", status);
	return !(status & PUYA_FLASH_SR_WRPERR);
```

Due to the way the check is then architected, this is not more expensive.

src/target/puya.c
```diff
@@ -0,0 +1,234 @@
...
+}
+
+static bool puya_flash_erase(target_flash_s *flash, target_addr_t addr, size_t len)
+{
+	(void)len;
+	target_mem32_write32(flash->t, PUYA_FLASH_CR, PUYA_FLASH_CR_PER);
+	target_mem32_write32(flash->t, addr, 0);
+	if (!puya_wait_flash(flash->t, NULL)) {
+		return false;
+	}
```
dragonmux, line 218: Please remove the excess braces here and the check below - they're not necessary and should be dropped.

dragonmux (commented):
Only got two review notes this round - with those fixed, this looks like it should be ready to merge.

dragonmux (approved):

We're sure the Puya PY32 users who've been waiting for support will be very happy 😄

dragonmux:
We're going to have to gate the PY32 support on native under the Makefiles because of space constraints, but that's something we'll take care of in a follow-up commit - we know exactly what has to change for that and we'll make the necessary adjustment. Just a couple of lines in the Makefiles, so no big deal.

PR #1820, opened 2024-05-13
Fix: Replace `sprintf()` calls remaining in firmware with `snprintf()`

ALTracer:
## Detailed description

* Not a feature.
* The existing minor problem is inconsistency in usage of sprintf/snprintf in firmware (target drivers).
* This PR solves it by replacing (almost as-is) the three remaining `sprintf` calls as highlighted by `puncover` and `grep` / `the-silver-searcher` with `snprintf` which is used much more often.

Bonus points for bounds checking, as all touched cases deal with copying a runtime-constructed pattern into a target private storage struct, which has a fixed capacity, and it's cheap to take its size [and stuff into second argument] -- not invented by me, I was just looking at other callers of snprintf. There could be stricter bounds enforced, but I won't be researching that.
As measured on gcc-12 and `native` this drops a 64-byte `siprintf` shim function from current `main`, while retaining 104-byte `sniprintf` with 14 callers and `_svfiprintf_r` the impl proper. It may be much harder to get rid of that set of efficient newlib-nano library functions, because of e.g. existing XML builders.

I can't test against any of three touched targets.

## Closing issues

### DIFF
--- a/src/target/efm32.c
+++ b/src/target/efm32.c
@@ -960,7 +960,8 @@ bool efm32_aap_probe(adiv5_access_port_s *ap)
 	DEBUG_INFO("EFM32: AAP STATUS=%08" PRIx32 "\n", adiv5_ap_read(ap, AAP_STATUS));
 
 	efm32_aap_priv_s *priv_storage = calloc(1, sizeof(*priv_storage));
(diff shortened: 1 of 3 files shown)


dragonmux (approved):
LGTM, good call to make all these bounded just in case and as that reduces Flash usage a bit. Merging - thank you for the contribution!

PR #1821, opened 2024-05-13
Please report unknown device with Designer 0x8055 Part ID 0x29b - Add support for another STM32F01 clone

nboehm99:
## Detailed description

Tried to use blackmagic to program/debug one of the infamous "bluepill" boards with a counterfeit STM32F103C8 MCU. BMDA reports `Please report unknown device with Designer 0x8055 Part ID 0x29b` and is not able to flash the board.

I had a deeper look, added a `DEBUG_INFO`to print the PIDR before processing and figured that the PIDR is apparently 0x000000050005529b, which would indicate a manufacturer code of 0x555 and part id 0x29b, which would actually already be recognized if the `PIDR_JEP106_USED` bit would be set. Which is sadly missing. Therefore in `adiv5_component_probe()` the if-clause where the designer code would be re-written to STM is not entered, instead the code is handled as "legacy ascii code" and extended to 0x8055.

To allow recognition of this chip as an STM32F103C8, this pull request adds a code snippet in `cortexm_probe()` to re-map the 0x8055 designer code to `JEP106_MANUFACTURER_STM`.

At time of creation, this PR also contains the additional DEBUG_INFO I added. If you think this is too noisy to be kept I can remove that commit from the PR.

Notes:
* Native build testen on blackpill STM401CC, that's all I have 
* I've read the contributing guidelines, but was unsure if that few lines of code would justify adding my name to the list of contributors in the file header, as in the git history there are far more significant contributions by people not mentioned there.

## Closing issues

None.

### DIFF
--- a/src/target/adiv5.c
+++ b/src/target/adiv5.c
@@ -581,17 +581,19 @@ static void adiv5_component_probe(
 		designer_code = (pidr & PIDR_JEP106_CONT_MASK) >> (PIDR_JEP106_CONT_OFFSET - 8U) |
 			(pidr & PIDR_JEP106_CODE_MASK) >> PIDR_JEP106_CODE_OFFSET;
 
-		if (designer_code == JEP106_MANUFACTURER_ERRATA_STM32WX || designer_code == JEP106_MANUFACTURER_ERRATA_CS) {
-			/**
-			 * see 'JEP-106 code list' for context, here we are aliasing codes that are non compliant with the
-			 * JEP-106 standard to their expected codes, this is later used to determine the correct probe function.
-			 */
-			DEBUG_WARN("Patching Designer code 0x%03" PRIx16 " -> 0x%03u\n", designer_code, JEP106_MANUFACTURER_STM);
-			designer_code = JEP106_MANUFACTURER_STM;
-		}
 	} else {
 		/* legacy ascii code */
 		designer_code = (pidr & PIDR_JEP106_CODE_MASK) >> PIDR_JEP106_CODE_OFFSET | ASCII_CODE_FLAG;
+	}
+
+	if (designer_code == JEP106_MANUFACTURER_ERRATA_STM32WX || designer_code == JEP106_MANUFACTURER_ERRATA_CS ||
+		designer_code == JEP106_MANUFACTURER_ERRATA_CS_ASCII) {
+		/**
+         * see 'JEP-106 code list' for context, here we are aliasing codes that are non compliant with the
+         * JEP-106 standard to their expected codes, this is later used to determine the correct probe function.
+         */
+		DEBUG_WARN("Patching Designer code %03x -> %03x\n", designer_code, JEP106_MANUFACTURER_STM);
+		designer_code = JEP106_MANUFACTURER_STM;
 	}
 
 	/* Extract part number from the part id register. */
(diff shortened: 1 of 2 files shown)


dragonmux:
> At time of creation, this PR also contains the additional DEBUG_INFO I added. If you think this is too noisy to be kept I can remove that commit from the PR.

The unprocessed PIDR value is actually already displayed at the INFO log level just a little further down in the component probe routine - as part of displaying the ROM tables for the part (at least for valid entries that aren't the start of ROM table entry); What we'd suggest is adding the PIDR display to the end of the MFR and part number print on the `ROM: ` output lines as this is really where it's missing from. That'll stop it getting sprayed for every ROM table entry.

src/target/cortexm.c
```diff
@@ -470,6 +470,10 @@ bool cortexm_probe(adiv5_access_port_s *ap)
 		ap->dp->designer_code == JEP106_MANUFACTURER_ARM_CHINA)
 		target->designer_code = JEP106_MANUFACTURER_ARM_CHINA;
 
+	/* Some STM32F103 counterfeits have a broken manufacturer code. Map to STM */
+	if (target->designer_code == JEP106_MANUFACTURER_ERRATA_CS_ASCII)
```
dragonmux, line 474: Why not patch this in adiv5.c the same way as the other `JEP106_MANUFACTURER_ERRATA_CS` is (having it generate the same the debug warning)? Ideally, the MM32 patch that's done just above this would be in adiv5.c too, as this isn't really specific to Cortex-M parts but more generically ADIv5 identity handling..

dragonmux (commented):
Other than the note above, we've only got the one review note, with that addressed and the PIDR display tidied up, we'll be happy to accept this PR for merge.

src/target/cortexm.c
nboehm99: Moved to adiv5.c as suggested. I also moved the designer code re-write and the associated warning out of the `if (pidr & PIDR_JEP106_USED)` clause to avoid duplicating the warning in both _if_ and _else_-clause. Fixed a formatting mistake in the warning on the way (previously it printed a `0x` prefix followed by the manufacturer id in decimal, now it's correctly printed in hexadecimal)

I also removed my extra DEBUG_INFO and added the raw PIDR to the "ROM: " debug output as suggested above.

I decided not to move the MM32 patch from cortexm.c because of several reasons: First, after refactoring my patch would not touch cortexm.c at all, second I wasn't sure where to exactly put it in adiv5.c and most importantly I don't have that hardware available to verify that it still works and didn't want to put anything at risk here.

dragonmux: Ah, good catch on that formatting bug!

src/target/adiv5.c
```diff
@@ -581,19 +581,22 @@ static void adiv5_component_probe(
...
 	}
 
+	if (designer_code == JEP106_MANUFACTURER_ERRATA_STM32WX || designer_code == JEP106_MANUFACTURER_ERRATA_CS ||
+		designer_code == JEP106_MANUFACTURER_ERRATA_CS_ASCII) {
+		/**
+         * see 'JEP-106 code list' for context, here we are aliasing codes that are non compliant with the
+         * JEP-106 standard to their expected codes, this is later used to determine the correct probe function.
+         */
+		DEBUG_WARN(
+			"Patching Designer code 0x%03" PRIx16 " -> 0x%03" PRIx16 "\n", designer_code, JEP106_MANUFACTURER_STM);
```
dragonmux, line 596: Keeping in mind that newlib has some fairly restricted handling of formatting specifiers, it might be a good idea here to just use "%03x" for the format strings as that's perfectly fine with uint16_t's going in..

It does feel weird to us including the `0x` prefix here and below, but we do recognise it creates a certain consistency with the "Please report unknown device" strings, even if we feel as these are always in hexadecimal that saving 4 bytes per string by dropping the prefix is a good space saving which doesn't hamper anything.

dragonmux (commented):
One note about the format strings for the designer/manufacturer and part number codes, and then we think this'll be good to merge.

src/target/adiv5.c
nboehm99: Removed the "0x" prefix for designer and part id and used "%03x" as format string (again) as suggested.

dragonmux (approved):

PR #1827, opened 2024-05-18
Reorder USB endpoints for TraceSWO capture on STM32F4 platforms

### LINKED ISSUE #1671: USB endpoints (including trace) broken on Black Pill (STM32F4xx)
While waiting for my official BMP 2.3 order to arrive, I've been experimenting with "building my own" using a WeAct STM32F411 board.

Basic functionality (debug and flash) is working.

SWO tracing does not work at all, in either the current 1.10 pre-release, or in the 1.9.2 build I did.  The symptom of the failure is that the host attempt to read from the trace endpoint fails immediately, with a "broken pipe" error (transfer stalled, apparently).

The identical version of 1.9.2, built for "stlink" and flashed into a "Baite" ST-Link v2 clone, does not exhibit this error.  SWO doesn't capture anything on the host (due to the fact that the trace line isn't brought out to the connector) but the host-to-USB connection doesn't flail and die.

I've been digging into the problem for a day or so, and I've come to the sad conclusion that trace-capture may never be able to work on the Black Pill.  In fact, I'm sorta surprised that BMP runs as well on this board/chip as it does.  The reason:  the USB OTG controller used in the STM32F411 (and apparently the F107 as well) has hardware support for one control endpoint (#0),  three IN endpoints (0x81-0x83), and three OUT endpoints(1-3)... period.  No others need apply.  You can have zero, one, two or three.  Counteth thee not to four.  Five is right out.

BMP uses four IN endpoints (for the two CDC-ACM), so it's already out of range.

I looked through the logic in libopencm3/.../usb_dwc_common.c to see how it accesses the hardware registers, and was somewhat troubled to see that it doesn't bother to range-check the endpoint index at all.  If you pass it endpoint numbers in the valid ranges (0, 1-3, 0x81-0x83) it hits the right hardware registers.  If you pass larger numbers (like 0x85 for trace) it appears to just go ahead and read/write quite happily to unimplemented register addresses.  This doesn't seem to cause hardware to throw exceptions, but I rather suspect that the results are quite undefined and may result in purple monkey bus dishwasher.

I've looked in detail at the hardware-register definitions, in the hope that one _might_ be allowed three active IN endpoints at a time, and be able to multiplex those three slots among a larger number of logical endpoints.  I haven't found any way to do so - the endpoint index/number is implied by the register offset, and isn't set explicitly in any register I've found.

I looked at a Wireshark/usbmon trace of the traffic between host and Black Pill when I try to start reading the trace data.  The host fires off a bunch of bulk-read URBs in rapid sequence.  A fraction of a second later the Black Pill starts sending responses - a combination of "broken pipe" and "no such address" is how Wireshark interprets them.  I'd guess that the F411 USB-OTG core is taking exception to being asked to send data from endpoint numbers it has no registers/FIFO for.

In theory I suspect that this same problem might affect the VCP CDC-ACM, since it uses endpoint 0x84 for control.  Maybe it's never sending anything of significance on that endpoint?

So, that's the unfortunate news.  Unless I'm gravely wrong (and I certainly may be) it won't be possible to put a full-function BMP firmware into any of the STM32 devices that have this particular USB-OTG core.  Some functionality will have to be sacrificed - either trace won't work, or the VCP "UART" will have to be configured out, and the trace endpoint moved down to 0x83.

ALTracer:
## Detailed description

* This is not exactly a new feature, but more of a support PR for something of bigger importance.
* The existing problem is Endpoint 5 IN used for all platforms, including STM32F4-based, for Trace Capture binary data. This works on STM32F1 FS Device with 8 endpoints, but doesn't work on DWC2 with 4 bi-dir endpoints including EP0 (0-3).
* This PR solves the problem by reordering important IN endpoints to stay under EP3 inclusive, but shifting CDC-ACM Notification Interrupt endpoints away to EP4 and EP5.

This is the first changeset in a series to enable TraceSWO (Async) Capture support on blackpill-f4 family of BMP-compatibles. I have tested the resulting combination on F411CE and observed ~480 KiB/s (of 6Mbaud) traffic in Orbuculum (despite `/dev/ttyBmpTarg` doing the same faster at 600 KiB/s) from STM32MP157C and GD32E508 targets with TDO pinned out, TPIU/SWO set up for UART at 6 Mbaud or other integer divided rates from 133.25 and 180 MHz respectively. This "adapter" itself can push about 1000 KiB/s of fixed payload traffic over a single VCP with a test firmware not using libopencm3. `f4discovery` would benefit this, too.

Next changes are #1529 and possible USART OVER8 in libopencm3 fork to allow in principle 12 Mbaud (96/8). USB FS may not sustain it, but 6.4, 8, 9.6 and 10 mega baudrates are nice to have, too. It may be beneficial to have independent TraceSWO Async *and* second CDC-ACM if the latter relays important info in real time, like physical TTL UART logs from DUT, or occasional RTT messages, or even semihosting redirected stdout (or BMP-originating diagnostic logging), all of which are potentially ASCII-printable. `traceswodecode.c` also prints there, when ITM_printf is sufficient, without binary traffic (like DWT events or TSDL instrumentation) -- but the problem of 5-to-1 mux of traffic is best solved elsewhere. OTOH, Async capture can be achieved by feeding TDO into AUX UART Rx and therefore sacrificing it -- Orbuculum supports serial ports (BMTrace might not). Manchester capture is not feedable into Aux uart Rx but requires two timer channels on a single input pin -- if ported from F1, it **requires** these changes.

I'll need to rebase the branch from November 2023 and drop backported changes, but feel free to review as I've submitted this.

Caveat 1: I expect problems from Windows 7 or MacOS or FreeBSD (or even Android?) during usage of both CDC-ACM ports of 4-EP F4 now, like DTR/RTS line state changes not getting back or ports hanging altogether (previously only second port was affected, first port worked properly), but I've been using Linux Mint (Ubuntu 22.04 LTS) with no problems in dmesg. F1 shouldn't break like this. Testing on other hosts is welcome. I only aim to empower Linux desktop-using developers, I don't have the resources to fight proprietary OS problems as well. No warranty or fittness for a particular purpose, as-is.

Caveat 2: Consumers of BMP Trace Capture need to be adjusted if they hardcode 0x85 (EP 5 IN), becomes 0x83 (EP 3 IN), namely `scripts/swolisten.c`, Orbuculum and Compuphase's BMTrace, really they all should walk USB descriptors using libusb. Interface is still 5, it's a logical number not physical pipe in USBOTG.

Caveat 3: Not sure whether `driver/blackmagic.inf` should need changes and signing, MS BOS already resolves some of that. Worst case users will need to uninstall drivers or clean registry entries if OS caches that, or you could bump .bcdDevice from 0x0109 or something.

* [x] It builds as BMDA (see [Building the BMDA](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-app)) *and does not affect it*

## Closing issues

Fixes #1671.


src/platforms/common/usb_serial.c
```diff
@@ -103,7 +103,11 @@ static usbd_request_return_codes_e gdb_serial_control_request(usbd_device *dev,
 	switch (req->bRequest) {
 	case USB_CDC_REQ_SET_CONTROL_LINE_STATE:
 		/* Send a notification back on the notification endpoint */
-		usb_serial_set_state(dev, req->wIndex, CDCACM_GDB_ENDPOINT + 1U);
+#if defined(STM32F4) && CDCACM_GDB_NOTIF_ENDPOINT < 4
```
dragonmux, line 106: This condition needs amending to only exclude STM32F4 and `CDCACM_GDB_NOTIF_ENDPOINT >= 4` - otherwise this changes the behaviour on the F1 and other parts.

dragonmux (changes requested):
Done a first review and didn't spot anything too major - just a couple of items to address. First is nomenclature related, and the second is a functional nit in the request handling to make sure we don't break things on other platforms.

src/platforms/common/usb_descriptors.h
```diff
@@ -174,10 +177,15 @@ static const usb_iface_assoc_descriptor_s gdb_assoc = {
 
 /* Physical/debug UART interface */
 
+/*
+ * This notification endpoint, when available, allows BMP
+ * to respong with DCD+DSR high, and optionally other SerialState flags:
```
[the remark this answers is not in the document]
ALTracer: Comment block reworded. I only wanted to document what users of stm32f4 would be losing without serial state notifications.

src/platforms/common/usb_serial.c
ALTracer: Good catch. I'll try rewriting the macro condition and moving one set of them into `usb_serial_set_state()`. The other set is dirty because `usb_ep_setup()` is part of libopencm3 API, and I didn't feel like introducing a wrapper.

dragonmux (dismissed):
LGTM, and we're happy this shouldn't change any behaviour it shouldn't now. We'll get it tested on native here and once done, we'll merge this. Thank you for the contribution!

ALTracer:
Amended last commit to keep O-packet contents compatible with any external RSP parsers. `USB EP 0x 3` is too different from `USB EP 5`. I didn't go full `USB EP 0x83` or revert all the way to `serialno:0x05:0x83` for Interface 5 and Endpoint 3.

dragonmux (approved):

dragonmux:
Sorry for it taking so long to get this all tested, etc - we're happy to say though that with this rebased on `main`, it all seems to work perfectly, so merging. Many thanks again!

PR #1828, opened 2024-05-19
Fix: -Wformat for tc_printf

HrMitrev:

## Closing issues

https://github.com/blackmagic-debug/blackmagic/pull/1676

### DIFF
--- a/src/target/ch579.c
+++ b/src/target/ch579.c
@@ -136,7 +136,7 @@ bool ch579_probe(target_s *target)
 {
 	uint8_t chip_id = target_mem32_read8(target, CH579_R8_CHIP_ID);
 	if (chip_id != 0x79) {
-		DEBUG_ERROR("Not CH579! 0x%02" PRIx8 "\n", chip_id);
+		DEBUG_ERROR("Not CH579! 0x%02x\n", chip_id);
 		return false;
 	}
 
--- a/src/target/efm32.c
+++ b/src/target/efm32.c
@@ -581,7 +581,7 @@ bool efm32_probe(target_s *t)
 	/* Setup Target */
 	t->target_options |= TOPT_INHIBIT_NRST;
 	t->driver = priv_storage->efm32_variant_string;
-	tc_printf(t, "flash size %u page size %u\n", flash_size, flash_page_size);
+	tc_printf(t, "flash size %" PRIu32 " page size %" PRIu32 "\n", flash_size, flash_page_size);
 
 	target_add_ram32(t, SRAM_BASE, ram_size);
 	efm32_add_flash(t, 0x00000000, flash_size, flash_page_size);
@@ -744,11 +744,11 @@ static bool efm32_cmd_serial(target_s *t, int argc, const char **argv)
 		break;
 
 	default:
-		tc_printf(t, "Bad DI version %hhu! This driver doesn't know about this DI version\n", di_version);
+		tc_printf(t, "Bad DI version %u! This driver doesn't know about this DI version\n", di_version);
 		return false;
 	}
 
-	tc_printf(t, "Unique Number: 0x%016llx\n", unique);
+	tc_printf(t, "Unique Number: 0x%08" PRIx32 "%08" PRIx32 "\n", (uint32_t)(unique >> 32U), (uint32_t)unique);
 
 	return true;
 }
@@ -768,15 +768,15 @@ static bool efm32_cmd_efm_info(target_s *t, int argc, const char **argv)
 
 	switch (di_version) {
 	case 1:
-		tc_printf(t, "DI version 1 (silabs remix?) base 0x%08" PRIx32 "\n\n", EFM32_V1_DI);
+		tc_printf(t, "DI version 1 (silabs remix?) base 0x%08" PRIx16 "\n\n", EFM32_V1_DI);
 		break;
 
 	case 2:
-		tc_printf(t, "DI version 2 (energy micro remix?) base 0x%08" PRIx32 "\n\n", EFM32_V2_DI);
+		tc_printf(t, "DI version 2 (energy micro remix?) base 0x%08" PRIx16 "\n\n", EFM32_V2_DI);
 		break;
 
 	default:
-		tc_printf(t, "Bad DI version %hhu! This driver doesn't know about this DI version\n", di_version);
+		tc_printf(t, "Bad DI version %u! This driver doesn't know about this DI version\n", di_version);
 		return false;
 	}
 
@@ -791,8 +791,8 @@ static bool efm32_cmd_efm_info(target_s *t, int argc, const char **argv)
 
 	tc_printf(t, "%s %hu F%hu = %s %hukiB flash, %hukiB ram\n", device->name, part_number, flash_kib,
 		device->description, flash_kib, ram_kib);
-	tc_printf(t, "Device says flash page size is %u bytes, we're using %u bytes\n", flash_page_size_reported,
-		flash_page_size);
+	tc_printf(t, "Device says flash page size is %" PRIu32 " bytes, we're using %" PRIu32 " bytes\n",
+		flash_page_size_reported, flash_page_size);
 	if (flash_page_size_reported < flash_page_size) {
 		tc_printf(t, "This is bad, flash writes may be corrupted\n");
 	}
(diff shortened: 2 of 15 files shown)


src/target/stm32l0.c
```diff
@@ -591,23 +591,24 @@ static bool stm32lx_cmd_option(target_s *const target, const int argc, const cha
...
+			options, (uint32_t)read_protection, (options & STM32L1_NVM_OPTR_SPRMOD) ? 1 : 0,
 			(options >> STM32L1_NVM_OPTR_BOR_LEV_S) & STM32L1_NVM_OPTR_BOR_LEV_M,
 			(options & STM32Lx_NVM_OPTR_WDG_SW) ? 1 : 0, (options & STM32L1_NVM_OPTR_nRST_STOP) ? 1 : 0,
 			(options & STM32L1_NVM_OPTR_nRST_STDBY) ? 1 : 0, (options & STM32L1_NVM_OPTR_nBFB2) ? 1 : 0);
 	} else {
-		tc_printf(target, "OPTR: 0x%08" PRIx32 ", RDPROT %u, WPRMOD %u, WDG_SW %u, BOOT1 %u\n", options,
-			read_protection, (options & STM32L0_NVM_OPTR_WPRMOD) ? 1 : 0, (options & STM32Lx_NVM_OPTR_WDG_SW) ? 1 : 0,
-			(options & STM32L0_NVM_OPTR_BOOT1) ? 1 : 0);
+		tc_printf(target, "OPTR: 0x%08" PRIx32 ", RDPROT %" PRIu32 ", WPRMOD %" PRIu16 ", WDG_SW %u, BOOT1 %u\n",
+			options, (uint32_t)read_protection, (options & STM32L0_NVM_OPTR_WPRMOD) ? 1 : 0,
```
HrMitrev, line 610: Here I opted to cast read_protection because it was giving me 2 different integer sizes depending on the platform I think? size_t is define as uint32_t/uint16_t depending on the platform? Maybe it will be better to redefine it as u32 everywhere?

src/target/efm32.c
```diff
@@ -41,6 +41,7 @@
 #include "target_internal.h"
 #include "cortexm.h"
 #include "adiv5.h"
+#include <inttypes.h>
```
HrMitrev, line 44: Even though I included this, I still get compiler error. Any Idea why?

HrMitrev: Thanks @ALTracer for pointing out that PRIx64 is not supported in all embedded platforms. I opted to print unique as 2 u32 ints

src/target/efm32.c
```diff
@@ -744,11 +744,11 @@ static bool efm32_cmd_serial(target_s *t, int argc, const char **argv)
 		break;
 
 	default:
-		tc_printf(t, "Bad DI version %hhu! This driver doesn't know about this DI version\n", di_version);
+		tc_printf(t, "Bad DI version %" PRIu8 "! This driver doesn't know about this DI version\n", di_version);
 		return false;
 	}
 
-	tc_printf(t, "Unique Number: 0x%016llx\n", unique);
+	tc_printf(t, "Unique Number: 0x%08" PRIx32 "%08" PRIx32 "\n", (uint32_t)(unique << 32), (uint32_t)(unique));
```
HrMitrev, line 751: Here @ALTracer reminded me that I cannot rely on PRIx64 for embedded platforms

src/target/ch579.c
```diff
@@ -136,7 +136,7 @@ bool ch579_probe(target_s *target)
 {
 	uint8_t chip_id = target_mem32_read8(target, CH579_R8_CHIP_ID);
 	if (chip_id != 0x79) {
-		DEBUG_ERROR("Not CH579! 0x%02" PRIx8 "\n", chip_id);
+		DEBUG_ERROR("Not CH579! 0x%02x\n", chip_id);
```
HrMitrev, line 139: After conversing with @dragonmux , We concluded that it is not necessary to use PRI*8 specifiers and better to stick to %d/%u/%x

src/target/target_internal.h
```diff
@@ -208,7 +208,15 @@ bool target_mem64_write16(target_s *target, target_addr64_t addr, uint16_t value
 bool target_mem64_write8(target_s *target, target_addr64_t addr, uint8_t value);
 bool target_check_error(target_s *target);
 
+#if defined(__MINGW32__) || defined(__MINGW64__) || defined(__CYGWIN__)
+#define TC_FORMAT_ATTR __attribute__((format(__MINGW_PRINTF_FORMAT, 2, 3)))
+#elif defined(__GNUC__) || defined(__clang__)
+#define TC_FORMAT_ATTR __attribute__((format(printf, 2, 3)))
```
HrMitrev, line 214: Thanks to @ALTracer for suggesting this macro.

ALTracer: If you insist on mentioning contributors, then also shout out to @amyspark -- this is the implementation in PR1743 https://github.com/blackmagic-debug/blackmagic/commit/ce69efd0#diff-4caf95c6659dfe9a43bcd59e329b01aa444e4ae17669bddd063bf8d5b14bd319

src/target/efm32.c
ALTracer: Technically `PRI*8` and `PRI*64` the C99 IO format specifiers are supported by Newlib, but BMF is restricted to newlib-nano on f103 etc. platforms, and that omits such support. I've found it out by rebuilding and linking BMF for blackpills with either newlib or nano and running it live for checking monitor output. Alternatively, examine `strings` resources in resulting .bin once you know which specifiers will be ignored by *printf at run time.

src/target/semihosting.c
```diff
@@ -723,7 +723,7 @@ int32_t semihosting_exit(target_s *const target, const semihosting_exit_reason_e
 	if (reason == EXIT_REASON_APPLICATION_EXIT)
 		tc_printf(target, "exit(%" PRIu32 ")\n", status_code);
 	else
-		tc_printf(target, "Exception trapped: %" PRIx32 " (%" PRIu32 ")\n", reason, status_code);
+		tc_printf(target, "Exception trapped: %" PRIx16 " (%" PRIu32 ")\n", reason, status_code);
```
dragonmux, line 726: Here, due too how enums work in C, the change should probably be to `%x` as enumerations have an underling type of `int` (but are treated by the compiler as unsigned for things like formatting). We can't assume that `PRIx16` is set up to be wide enough for the numerical representation of the enumeration otherwise.

src/target/stm32l0.c
dragonmux: [`size_t` is defined by the compiler](https://en.cppreference.com/w/c/types/size_t) and trying to redefine it would invoke UB. The correct formatting string for it is `%zu`, however newlib-nano does not support this so casting to uint32_t is fine.

dragonmux (changes requested):
We've got just the one review item - nice work! With that addressed, we think this will be ready for merge.

Our only other request is a tweak to the commit message - because it touches so much of the code base, please use the suffix `misc: `, as in `misc: Fixes for tc_printf invocations to enable -Wformat`

dragonmux:
Just a note that, once the remaining review notes have been taken care of, we're happy for this PR to be merged if it could be rebased on `main` (`git rebase main` should do the trick)

src/target/target_internal.h
HrMitrev: BTW, wouldn't it be better to move this macro and the one from @amyspark (located in gdb_packet.h) to some place where we can merge them in one and include it everywhere?

ALTracer: Not really, the argument positions are different.

HrMitrev: But we'll have 2 `#define X_FORMAT_ATTR`  and we'll skip the duplication of `#if` / `#elif`. @dragonmux what do you think?

dragonmux: They are in presently unrelated headers and while perhaps there's an argument to be made for a new header, this creates new inter-dependencies vs this approach and gives the headache that ALTracer mentions about having to say the printf parameter numbers everywhere these are used. Having them as separate macro defines in separate independent headers makes more sense in this context to us for now.

dragonmux (approved):
We don't see any remaining review comments that need addressing, and it all LGTM, so merging. Thank you for the contribution!

dragonmux (approved):

PR #1832, opened 2024-05-26
Fix UART TX and NRST pin conflict  on stlink clones with SWIM_AS_UART

positron96:
## Detailed description

Adding UART to stlink probe introduced in #1087 has created a conflict between UART TX and NRST pins on stlink clones
(both are PB6). More discussion here: https://github.com/blackmagic-debug/blackmagic/pull/1087#issuecomment-2053975402

This PR fixes this by moving NRST pin to PB0 on stlink clones (same one as on original stlink). 

Note that PB0 is not present on the header itself, so NRST functionality is lost (However this seems to be the intention of the original PR).

Also this PR renames build option from SWIM_AS_UART to SWIM_NRST_AS_UART for better clarity that it occupies both SWIM and NRST header pins. 
This is also changed for meson configs.

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -67,10 +67,10 @@ option(
 )
 option('bmd_bootloader', type: 'boolean', value: true, description: 'Use the BMD bootloader (not always applicable)')
 option(
-	'stlink_swim_as_uart',
+	'stlink_swim_nrst_as_uart',
 	type: 'boolean',
 	value: false,
-	description: 'Repurpose the SWIM pins as the UART (only applicable to stlink)'
+	description: 'Repurpose the SWIM and NRST pins as the UART RX and TX (only applicable to stlink)'
 )
 option('serialno', type: 'string', value: '1', description: 'Serial number to report (only applicable to some probes)')
 option(
--- a/src/platforms/stlink/Makefile.inc
+++ b/src/platforms/stlink/Makefile.inc
@@ -26,7 +26,12 @@ LDFLAGS += --specs=nosys.specs
 endif
 
 ifeq ($(SWIM_AS_UART), 1)
-CFLAGS += -DSWIM_AS_UART=1
+SWIM_NRST_AS_UART=1
+$(warning Deprecation: SWIM_AS_UART has been renamed to SWIM_NRST_AS_UART; it might be ignored in the future)
+endif
+
+ifeq ($(SWIM_NRST_AS_UART), 1)
+CFLAGS += -DSWIM_NRST_AS_UART=1
 else
 SRC += traceswoasync.c
 endif
--- a/src/platforms/stlink/platform.h
+++ b/src/platforms/stlink/platform.h
@@ -58,10 +58,14 @@ extern bool debug_bmp;
 #define SWDIO_IN_PIN  GPIO12
 #endif
 
-#define NRST_PORT      GPIOB
-#define NRST_PIN_V1    GPIO1
-#define NRST_PIN_V2    GPIO0
+#define NRST_PORT   GPIOB
+#define NRST_PIN_V1 GPIO1
+#define NRST_PIN_V2 GPIO0
+#ifdef SWIM_NRST_AS_UART
+#define NRST_PIN_CLONE GPIO0
+#else
 #define NRST_PIN_CLONE GPIO6
+#endif
 
 #ifdef BLUEPILL
 #define LED_PORT GPIOC
@@ -72,7 +76,7 @@ extern bool debug_bmp;
 #define LED_PORT_UART GPIOA
 #define LED_UART      GPIO9
 
-#ifndef SWIM_AS_UART
+#ifndef SWIM_NRST_AS_UART
 #define PLATFORM_HAS_TRACESWO 1
 #endif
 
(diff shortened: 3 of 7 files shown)


src/platforms/stlink/platform.h
```diff
@@ -61,7 +61,12 @@ extern bool debug_bmp;
 #define NRST_PORT      GPIOB
 #define NRST_PIN_V1    GPIO1
 #define NRST_PIN_V2    GPIO0
+#ifdef SWIM_NRST_AS_UART
+#pragma message "NRST pin is now PB0"
```
dragonmux, line 65: Think this would be better noted in the documentation - people using a pre-made build of the firmware in this configuration won't see this, and this will hammer the compiler output with this message for every compilation unit that platform.h is included into.

dragonmux (changes requested):
Only got the one review note - with that addressed and the clang-format lints from the lint pass fixed, this looks good to merge.

src/platforms/stlink/platform.h
positron96: where would be the best place in documentation to mention this?

dragonmux: We'd think in the platform README.md around where you had to change the wording for what the Makefile flag does - mentioning around there that the nRST signal is remapped to PB0 would be sufficient.

dragonmux:
Looks good. As we will rebase merge the PR, please would you rebase the two fix commits into the base commit (or at least rebase 44be2e2 in, and then fix the commit message for 5539724 to meet the contribution guidelines). Once that's done we will approve the PR for merge and get it merged.

meson_options.txt
```diff
@@ -67,10 +67,10 @@ option(
 )
 option('bmd_bootloader', type: 'boolean', value: true, description: 'Use the BMD bootloader (not always applicable)')
 option(
-	'stlink_swim_as_uart',
+	'stlink_swim_nrst_as_uart',
 	type: 'boolean',
 	value: false,
-	description: 'Repurpose the SWIM pins as the UART (only applicable to stlink)'
+	description: 'Repurpose the SWIM and NRST pins as the UART RX and TX(only applicable to stlink)'
```
amyspark, line 73:
```suggestion
	description: 'Repurpose the SWIM and NRST pins as the UART RX and TX (only applicable to stlink)'
```

amyspark (changes requested):
Hi! Just noticed the one nit in here, perhaps @dragonmux will want to see it sorted out with the rest?

meson_options.txt
dragonmux: Ah, great catch, yes - please.

src/platforms/stlink/Makefile.inc
```diff
@@ -25,8 +25,8 @@ else
 LDFLAGS += --specs=nosys.specs
 endif
 
-ifeq ($(SWIM_AS_UART), 1)
-CFLAGS += -DSWIM_AS_UART=1
+ifeq ($(SWIM_NRST_AS_UART), 1)
+CFLAGS += -DSWIM_NRST_AS_UART=1
```
ALTracer, line 29: > renames build option

How would be newcoming users following old build guides treated here? `make PROBE_HOST=stlink SWIM_AS_UART=1` will silently build the default configuration now, I think. I have a low-priority suggestion to either keep accepting the old/legacy makeflag (possibly with a warning/deprecation message) or outright fail the build. The preprocessor macro value passed to compiler frontends can stay. Meson is probably not affected.

dragonmux: Checking if SWIM_AS_UART is set in any capacity and turning that into an error w/ mention of the new option name would be a good idea, but is something we can do as a follow-up if the contributor doesn't wish to do this themself. It's certainly a good point though that we hadn't considered.

positron96: I can as well add a check for old flag setting new compiler switch, like this:
```
ifeq ($(SWIM_AS_UART), 1)
CFLAGS += -DSWIM_NRST_AS_UART=1
```

and some deprecation warning.

positron96:
> As we will rebase merge the PR, please would you rebase the two fix commits into the base commit (or at least rebase [44be2e2](https://github.com/blackmagic-debug/blackmagic/commit/44be2e24980a159746ab1379ae938442aecd9ec9) in, and then fix the commit message for [5539724](https://github.com/blackmagic-debug/blackmagic/commit/5539724ec0fe9e6ccfb8e24075b6bb6069e8ff99) to meet the contribution guidelines). Once that's done we will approve the PR for merge and get it merged.

do you mean squashing them all in one commit?

dragonmux:
Sort of - we mean using rebase's fixup command rather than squash, so `git rebase -i main`, then switching `pick` to `f`/`fixup` in the interactive editor on the aforementioned commits before letting it loose. The net effect is the same, the difference is in the commit message handling.

src/platforms/stlink/Makefile.inc
dragonmux: That works. We can turn that into a hard error down the road 👍🏼 .

dragonmux (approved):
LGTM, all the review items from everyone have been addressed. We'll merge this as soon as the builds complete. Thank you for the contribution!

PR #1833, opened 2024-05-27
Restore TraceSWO capture in Manchester mode

ALTracer:
## Detailed description

* This is an existing/missing feature.
* There is an unconfirmed problem of Manchester capture not working on `native`, and a "feature request" to bring it back to swlink (on which it originally existed), stlink (this is new) and blackpill-f4 (neither mode worked due to various reasons).
* This PR solves that by restoring and copying some macros and introducing compile-time selection of traceswo protocol.

Changes from November 2023, so meson buildsystem parts are missing. I wonder how seamlessly it rebases this far forward.
Default capture modes are Manchester for `native` (and I'm not changing that), but Async/UART+DMA for `stlink` and `swlink`, which I tested to be somewhat working. Building with `TRACESWO_PROTOCOL=1` reverts the two latter platforms to TIM-based Manchester. The PR needs testing and minor cleanup.

Known benefits of Manchester mode capture are automatic data rate (up to a certain BMF-specific limit) useful for power-saving systems and otherwise dynamically changing clocks systems; smaller RAM buffer (as of now, because of slow data rate); less external wiring on `stlink BLUEPILL=1` (uses same TDO pin PA6, not PA10) and `blackpill-f4` (can use PB6, not just PB7). Drawbacks are, well, slower data rate, and 50% link efficiency (UART is 80% efficient and leverages DMA). I believe moving from TIM+IRQ to TIM+DMA and large buffers might boost this mode on par or faster than Uart, and still save users from matching baudrates (and holding Traceclk frequencies) without resorting to FPGA-based tools. No other MCU-based adapter I know of implements this.

## Closing issues

I would raise an issue of Manchester not working on blackpill-f4 (wrong pin and channel), but a) there may very well be existing one; b) uart/async is faster and I don't debug dynamically clocked chips.

### DIFF
--- a/src/platforms/common/blackpill-f4/blackpill-f4.h
+++ b/src/platforms/common/blackpill-f4/blackpill-f4.h
@@ -299,12 +299,12 @@ extern bool debug_bmp;
 #define SWO_UART_PIN_AF GPIO_AF7
 
 /* Bind to the same DMA Rx channel */
-#define SWO_DMA_BUS    USBUSART1_DMA_BUS
-#define SWO_DMA_CLK    USBUSART1_DMA_CLK
-#define SWO_DMA_CHAN   USBUSART1_DMA_RX_CHAN
-#define SWO_DMA_IRQ    USBUSART1_DMA_RX_IRQ
-#define SWO_DMA_ISR(x) USBUSART1_DMA_RX_ISRx(x)
-#define SWO_DMA_TRG    DMA_SxCR_CHSEL_4
+#define SWO_DMA_BUS     USBUSART1_DMA_BUS
+#define SWO_DMA_CLK     USBUSART1_DMA_CLK
+#define SWO_DMA_CHAN    USBUSART1_DMA_RX_CHAN
+#define SWO_DMA_IRQ     USBUSART1_DMA_RX_IRQ
+#define SWO_DMA_ISR(x)  USBUSART1_DMA_RX_ISRx(x)
+#define SWO_DMA_TRG     DMA_SxCR_CHSEL_4
 
 #define SET_RUN_STATE(state)      \
 	{                             \
--- a/src/platforms/common/blackpill-f4/blackpill-f4.h
+++ b/src/platforms/common/blackpill-f4/blackpill-f4.h
@@ -32,10 +32,6 @@
 #include "timing.h"
 #include "timing_stm32.h"
 
-#define PLATFORM_HAS_TRACESWO
-#define NUM_TRACE_PACKETS 256U /* 16K buffer */
-#define TRACESWO_PROTOCOL 2U   /* 1 = RZ/Manchester, 2 = NRZ/async/uart */
-
 #if ENABLE_DEBUG == 1
 #define PLATFORM_HAS_DEBUG
 extern bool debug_bmp;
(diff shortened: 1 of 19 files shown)


ALTracer:
I tried to test it by throwing some Manchester traffic at a blackpill-f411ce, but did not get any reaction from it. 

Because F1 GPIO is simpler, and F4 GPIO has an AF matrix/layers, I think it's mandatory to patch the TDO pin to Timer Input function -- the top commit does that in *a way* (there could be a better place to put the calls). Even after that it did not decode nor pass raw traffic (but didn't lock up either). I may misunderstand all the timer API arguments, like what's the Slave mode trigger -- Timer Input, or some internal Timer Channel, or Rising/Falling triggered channel? Depending on RM0008 and RM0383 (which inherits the PWM Input Capture paragraphs almost verbatim) it may be as simple as Always `TIM_SMCR_TS_TI1FP1` for Rising edge (for pulse width decoder logic to work).

Further analysis with upstream TraceSWO capture required (and it's nice to use a platform with unconflicting `jtagtap.c` and SWJ-DP pins!)

src/platforms/stlink/platform.h
```diff
@@ -180,11 +192,13 @@ extern bool debug_bmp;
...
-#define SWO_DMA_BUS    DMA1
-#define SWO_DMA_CLK    RCC_DMA1
-#define SWO_DMA_CHAN   DMA_CHANNEL5
-#define SWO_DMA_IRQ    NVIC_DMA1_CHANNEL5_IRQ
-#define SWO_DMA_ISR(x) dma1_channel5_isr(x)
+#define SWO_DMA_BUS     DMA1
+#define SWO_DMA_CLK     RCC_DMA1
+#define SWO_DMA_CHAN    DMA_CHANNEL5
+#define SWO_DMA_IRQ     NVIC_DMA1_CHANNEL5_IRQ
+#define SWO_DMA_ISR(x)  dma1_channel5_isr(x)
```
dragonmux, line 199: These want running through clang-format to fix the formatting.

src/platforms/swlink/platform.h
```diff
@@ -140,11 +147,13 @@ extern bool debug_bmp;
...
-#define SWO_DMA_BUS    DMA1
-#define SWO_DMA_CLK    RCC_DMA1
-#define SWO_DMA_CHAN   DMA_CHANNEL6
-#define SWO_DMA_IRQ    NVIC_DMA1_CHANNEL6_IRQ
-#define SWO_DMA_ISR(x) dma1_channel6_isr(x)
+#define SWO_DMA_BUS     DMA1
+#define SWO_DMA_CLK     RCC_DMA1
+#define SWO_DMA_CHAN    DMA_CHANNEL6
+#define SWO_DMA_IRQ     NVIC_DMA1_CHANNEL6_IRQ
+#define SWO_DMA_ISR(x)  dma1_channel6_isr(x)
```
dragonmux, line 154: Same comment about clang-format

dragonmux (changes requested):
This looks like a really good start - with the review items highlighted fixed, we'll be happy to merge this with the intent of using it as a basis for fixing the native SWO trace support as this nicely advances that even if, as you note, not everything will be fixed by it.

src/platforms/stlink/platform.h
ALTracer: Notice first commit doing exactly that. Because with contents of second commit of this patchset, pre-commit clang-format hook requires modifying entire block. Should it have 1 or 2 spaces between longest macro and values?

src/platforms/common/blackpill-f4/blackpill-f4.h
```diff
@@ -273,10 +274,15 @@ extern bool debug_bmp;
 #define IRQ_PRI_USBUSART_DMA (2U << 4U)
 #define IRQ_PRI_TRACE        (0U << 4U)
 
-#define TRACE_TIM          TIM3
-#define TRACE_TIM_CLK_EN() rcc_periph_clock_enable(RCC_TIM3)
-#define TRACE_IRQ          NVIC_TIM3_IRQ
-#define TRACE_ISR(x)       tim3_isr(x)
+#if (TRACESWO_PROTOCOL == 1)
```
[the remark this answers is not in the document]
ALTracer: Dropped parens for `#if`, `#elif` here, and in stlink/swlink.

src/platforms/stlink/platform.h
dragonmux: Should be a single space.

ALTracer: Dropped commit doing that.

dragonmux: Huhh, so.. that made the lint pass angry - apparently, unlike anywhere else in the code where it uses one space, it wants two here. If you could please add it back in, we'll resolve this and the comment below as it actually really wants two spaces.

We can look at seeing why clang-format is doing that later.

ALTracer: Restored.

src/platforms/swlink/platform.h
ALTracer: Applied same here and in blackpill-f4.h

dragonmux (approved):
LGTM, merging when the builds complete. Thank you for the contribution!

ALTracer:
Updated the rest of platforms to provide working TRACE_TIM definitions, and dropped TraceSWO for 96b_carbon as I don't see how it should work.
Figured out the `-DTRACESWO_PROTOCOL=2` meson change to match Makefiles. Note that the dual-capable platforms do not `#ifndef-define-endif` it.
Rebasing to main. CI should work now.

PR #1846, opened 2024-06-19
Fix: increase semihosting F-packet detection robustness

### LINKED ISSUE #1845: Semihost: regression from v1.9.3 to aa67538
One of my semihost on-target tests is failing when updating the BlackMagicProbe firmware from v1.9.3 to the latest aa6753833764b41697900536395bb343f611377d in main.

The issue arises when I try to write a negative value through semihost:

```cpp
auto val = semihost_persistence::add("./logs/test_add", -3);
zassert_equal(-3, val.value());
```

(`semihost_persistence::add` will rely on both read as well as write calls)

The debugger will hang in the following manner:

image

Positive values work in both versions.

This is using SAMC21.

ALTracer:
## Detailed description

* This is a fix, not a new feature.
* The existing problem is BMD core code detecting GDB F-reply packets too loosely (contributed in #1686).
* This PR solves it by moving the F-reply check up.

Thanks to @TechnoMancer for the idea of capturing `gdb_getpacket()` results before `gdb_main_loop()` potentially clobbers the single in-BMD `gdb_packet_buffer[]` with `gdb_putpacket()` via m-reply. Such as write syscalls with payload beginning with negative numbers, bytes 0xF0..0xFF.

Tested on `blackpill-f411ce` with extra diagnostic logging enabled, against `gd32f303cc` bluepill-plus running Koen's SemihostingTest.ino I modified to do extra writes of negative int32_t numbers (like -3). Logs from GDB and BMP on request.

v1.10 branch is also affected, but some reorganisation has happened on current `main`, so that is not deemed worth backporting (or using for semihosting).

## Closing issues

Fixes #1845.

### DIFF
--- a/src/target/semihosting.c
+++ b/src/target/semihosting.c
@@ -183,9 +183,10 @@ static int32_t semihosting_get_gdb_response(target_controller_s *const tc)
 		/* If this was an escape packet (or gdb_if reports link closed), fail the call */
 		if (size == 1U && packet_buffer[0] == '\x04')
 			return -1;
+		/* If this was an F-reply packet, we're done. Check before clobbering buffer with m-reply. */
+		const bool done = packet_buffer[0] == 'F';
 		const int32_t result = gdb_main_loop(tc, packet_buffer, GDB_PACKET_BUFFER_SIZE, size, true);
-		/* If this was an F-packet, we're done */
-		if (packet_buffer[0] == 'F')
+		if (done)
 			return result;
 	}
 }


dragonmux (approved):
LGTM, we'll get this merged one the user in the linked issue has confirmed that it fixes the behaviour they're seeing.

maksimdrachov:
Yes, this has fixed my issue!

Compiles for `PROBE_HOST=native` (Using GCC 9.3.1)

PR #1848, opened 2024-06-19
Fix: Ensure F-packet replies from gdb are reliably detected

### LINKED ISSUE #1845: Semihost: regression from v1.9.3 to aa67538

TechnoMancer:
## Detailed description

* This fixes an issue.
* The origin of the issue was the change contributed in #1686 which added the loop in semihosting_get_gdb_response.
* The check for an F-packet reply was too late as gdb_main_loop may have already clobbered the packet by the time this is checked.

Collaborative inspection of debug logs from #1845 and provided by others pointed to inconsistent behaviour dealing with remote file i/o completion, which was narrowed down to the F-packet reply check being done after the packet had been processed and the packet buffer possibly clobbered.

This is corrected by checking if the received packet is an F-packet first then processing it and using the result of the check to exit the loop after the processing.

## Closing issues

Fixes #1845.

### DIFF
--- a/src/target/semihosting.c
+++ b/src/target/semihosting.c
@@ -183,9 +183,13 @@ static int32_t semihosting_get_gdb_response(target_controller_s *const tc)
 		/* If this was an escape packet (or gdb_if reports link closed), fail the call */
 		if (size == 1U && packet_buffer[0] == '\x04')
 			return -1;
+		/* 
+		 * If this was an F-packet, we are done waiting.
+		 * Check before gdb_main_loop as it may clobber the packet buffer.
+		 */
+		const bool done = packet_buffer[0] == 'F';
 		const int32_t result = gdb_main_loop(tc, packet_buffer, GDB_PACKET_BUFFER_SIZE, size, true);
-		/* If this was an F-packet, we're done */
-		if (packet_buffer[0] == 'F')
+		if (done)
 			return result;
 	}
 }


src/target/semihosting.c
```diff
@@ -183,9 +183,12 @@ static int32_t semihosting_get_gdb_response(target_controller_s *const tc)
 		/* If this was an escape packet (or gdb_if reports link closed), fail the call */
 		if (size == 1U && packet_buffer[0] == '\x04')
 			return -1;
+		/* If this was an F-packet, we are done waiting.
```
dragonmux, line 186: As this is going to be a multi-line comment, please use the style
```c
/*
 * Multi-line
 * comment
 */
```

dragonmux (changes requested):
There's one small stylistic change, but with that taken care of, we'll be happy to merge this.

Many thanks to you and ALTracer for this fix even if he had to pull his PR for geopolitical reasons.

dragonmux (approved):

PR #1850, opened 2024-06-22
TI MSPM0 Experimental Support

hardesk:
## Detailed description

The PR adds support for a new probe target, that is Texas Instruments MSPM0 microcontrollers. It also modifies MEM-AP discovery, which is a change needing review-attention.

I imagine the purpose of this PR is mostly to gather feedback on the code as opposed to actually merge it.

According to puncover tool, the code when compiling as release adds 808 bytes (without bcr/bcl register dump functions as those are included only on pc-hosted or debug output builds).


hardesk:
Fixed code formatting and compilation issues, but the problem now is that the firmware doesn't fit into 128k and thus linux and size-diff builds fail. Not sure how to go about it, please advice.

src/target/adiv5.c
```diff
@@ -768,8 +768,7 @@ adiv5_access_port_s *adiv5_new_ap(adiv5_debug_port_s *dp, uint8_t apsel)
 			ap.base |= (uint64_t)adiv5_ap_read(&ap, ADIV5_AP_BASE_HIGH) << 32U;
 		}
 		/* Check the Debug Base Address register for not-present. See ADIv5 Specification C2.6.1 */
-		if (base_flags == (ADIV5_AP_BASE_FORMAT_ADIV5 | ADIV5_AP_BASE_PRESENT_NO_ENTRY) ||
-			(!(ap.flags & ADIV5_AP_FLAGS_64BIT) && (uint32_t)ap.base == ADIV5_AP_BASE_NOT_PRESENT)) {
+		if (!(ap.flags & ADIV5_AP_FLAGS_64BIT) && (uint32_t)ap.base == ADIV5_AP_BASE_NOT_PRESENT) {
```
dragonmux, line 771: The specification for ADIv5 makes it very clear the check that's being removed here is required, as there are no validity requirements around `BASE.BASEADDR[31:12]` - any value in those bits is valid - and only bits [11:0] have validity requirements placed on them.

What is the cause of making this change? We don't see a noted reasoning for it in the PR or anywhere obvious in the commit messages. It might be helpful to attach a ROM table dump (`blackmagic -tv 5`) from the part so we can understand what's going on in the BASE field for it to help us square what's going on here against other parts and the specification.

src/target/mspm0.c
```diff
@@ -0,0 +1,349 @@
...
+static bool mspm0_dump_bcr_config(target_s *const target, const int argc, const char **const argv);
+
+static command_s mspm0_cmds_list[] = {{"dump_factory", mspm0_dump_factory_config, "Display FACTORY registers"},
+	{"dump_bcr", mspm0_dump_bcr_config, "Display NONMAIN (BCR/BSL) registers"},
+	/* {"write_nm", mspm0_write_non_main, "Write data to NONMAIN (configuration) flash. write_nm <address> <value>"}, */
+	{NULL, NULL, NULL}};
+
+typedef struct conf_register {
+	uint16_t offset : 12;
+	uint16_t words : 4;
```
dragonmux, line 104: Please do not use bitfields. They are not well defined by the C specification and so their use is badly liable to causing undefined behaviour and unwanted side-effects, up to and including this structure winding up padded to `alignof(int)` instead of `alignof(uint16_t)`, negating the entire purpose of this.

This is why the code base prefers masking and shifting and an encoding macro instead. Much easier to control and get the behaviour desired.

src/target/mspm0.c
```diff
@@ -0,0 +1,349 @@
...
+static bool mspm0_mass_erase(target_s *target);
+
+#if MSPM0_CONFIG_FLASH_DUMP_SUPPORT
+static bool mspm0_dump_factory_config(target_s *const target, const int argc, const char **const argv);
+static bool mspm0_dump_bcr_config(target_s *const target, const int argc, const char **const argv);
+
+static command_s mspm0_cmds_list[] = {{"dump_factory", mspm0_dump_factory_config, "Display FACTORY registers"},
+	{"dump_bcr", mspm0_dump_bcr_config, "Display NONMAIN (BCR/BSL) registers"},
+	/* {"write_nm", mspm0_write_non_main, "Write data to NONMAIN (configuration) flash. write_nm <address> <value>"}, */
+	{NULL, NULL, NULL}};
```
dragonmux, line 100: Please include a trailing `,` after the closing `}` of the sentinel entry. That will fix the formatting of this list. (Ditto for the register lists below.)

src/target/mspm0.c
```diff
@@ -0,0 +1,349 @@
...
+	{0x50, 1, "APPCRCSTART"}, {0x54, 1, "APPCRCLENGTH"}, {0x58, 1, "APPCRC"}, {0x5c, 1, "BOOTCRC"},
+	{0x100, 1, "BSLCONFIGID"}, {0x104, 1, "BSLPINCFG0"}, {0x108, 1, "BSLPINCFG1"}, {0x10c, 1, "BSLCONFIG0"},
+	{0x110, 8, "BSLPW"}, {0x130, 1, "BSLPLUGINCFG"}, {0x134, 4, "BSLPLUGINHOOK"}, {0x144, 1, "PATCHHOOKID"},
+	{0x148, 1, "SBLADDRESS"}, {0x14c, 1, "BSLAPPVER"}, {0x150, 1, "BSLCONFIG1"}, {0x154, 1, "BSLCRC"}, {0, 0, NULL}};
+
+static void mspm0_dump_regs(target_s *const target, const conf_register_s *regs, uint32_t base)
+{
+	for (conf_register_s const *r = regs; r->id; ++r) {
+		tc_printf(target, "%15s: ", r->id);
+		for (int q = 0; q < r->words; ++q) {
```
dragonmux, line 124: Do not use `int` for this loop. This performs a signed-unsigned conversion and opens the code up to bugs we do not want to tango with here. Either use `uint16_t` to match the intended underlying type of `words`, or `size_t`.

src/target/mspm0.c
```diff
@@ -0,0 +1,349 @@
...
+	target_flash->blocksize = MSPM0_FLASH_SECTOR_SZ;
+	target_flash->writesize = 8U;
+	target_flash->erase = mspm0_flash_erase;
+	target_flash->write = mspm0_flash_write;
+	target_flash->erased = 0xff;
+	target_add_flash(target, target_flash);
+}
+
+#define BIT_LMASK(T, bits) ((T)-1) >> (sizeof(T) * CHAR_BIT - (bits))
+#define BITS(h, l, x)      (uint32_t)((x >> l) & BIT_LMASK(uint32_t, h - l + 1))
```
dragonmux, line 169: This tries too be clever, but it falls foul of two issues which are going to cause eminent pain to maintain here: CHAR_BIT is platform-dependant, so all the shifts this tries to calculate are also platform-dependant; This performs more signed-unsigned conversions due to unsuffixed constants (1 is signed, 1U is unsigned), and may well invoke undefined behaviour (though we can't be certain) with the shifts being defined. This is liable to entirely miscompiling when building BMDA on spicier architectures.

Please instead do as the rest of the code base does and define `MSPM0_` prefixed mask and shift macros and apply those below. This then defines what the bit ranges mean with a friendly name, and is not liable to explode at a moment's notice down the road.

dragonmux (changes requested):
Below is our preliminary review of this code. There are some interesting ideas here, and we'd definitely like to see support for this part land.

There are some concerning changes proposed here though which we've highlighted for discussion and potential remediation. We've not fine tooth combed through the change sets yet though and so there are likely a whole lot of constants needing `U` suffixes which we've not picked up, and such.

This should all be very solvable though and we look forward to how this evolves.

src/target/adiv5.c
hardesk: I think let's start from the root, ie. the data that is being read from the device. Here's the -tv5 output:

```
Black Magic Debug App v1.10.0-852-gf650fa8f-dirty
Using 1d50:6018 81D15AA3 Black Magic Debug
 Black Magic Probe v1.10.2
Running in Test Mode
Target voltage: 3.3V Volt
Speed set to 1.951MHz for SWD
Handling SWD multi-drop, TARGETID 0x1bb8202f
DP DPIDR 0x6ba02477 (v2 rev6) designer 0x43b partno 0xba
TARGETID 0x1bb8202f designer 0x17 partno 0xbb82
AP   0: IDR=84770001 CFG=00000000 BASE=f0000000 CSW=a3800040 (AHB3-AP var0 rev8)
Halt via DHCSR(01030003): success after 6ms
ROM: Table BASE=0xf0000000 SYSMEM=1, Manufacturer 43b Partno 4c1 (PIDR = 0x00000004000bb4c1)
ROM: Table BASE=0xe00ff000 SYSMEM=1, Manufacturer 43b Partno 4c0 (PIDR = 0x00000004000bb4c0)
 0 0xe000e000: Generic IP component - Cortex-M0 SCS (System Control Space) (PIDR = 0x00000004000bb008 DEVTYPE = 0x00 ARCHID = 0x0000)
 -> cortexm_probe
CPUID 0x410cc601 (M0+ var 0 rev 1)
Calling msp432p4_probe
Calling mspm0_probe
 1 0xe0001000: Generic IP component - Cortex-M0 DWT (Data Watchpoint and Trace) (PIDR = 0x00000004000bb00a DEVTYPE = 0x00 ARCHID = 0x0000)
 2 0xe0002000: Generic IP component - Cortex-M0 BPU (Breakpoint Unit) (PIDR = 0x00000004000bb00b DEVTYPE = 0x00 ARCHID = 0x0000)
 ROM: Table END
1 Entry 0x1002 -> Not present
2 Entry 0x50402002 -> Not present
3 Entry 0x10000002 -> Not present
ROM: Table END
AP   1: IDR=002e0001 (Unknown var0 rev0)
0 0x00000000: 0x00000000 <- does not match preamble (0xb105000d)
AP   2: IDR=002e0000 (Unknown var0 rev0)
0 0x00000000: 0x00000000 <- does not match preamble (0xb105000d)
AP   3: IDR=002e0003 (Unknown var0 rev0)
0 0x00000000: 0x01010101 <- does not match preamble (0xb105000d)
AP   4: IDR=002e0002 (Unknown var0 rev0)
0 0x00000000: 0x00000000 <- does not match preamble (0xb105000d)
***  1   MSPM0 M0+
RAM   Start: 0x20000000 length = 0x1000
Flash Start: 0x00000000 length = 0x10000 blocksize 0x400
```

Also the interesting part with -tv15
```
TARGETID 0x1bb8202f designer 0x17 partno 0xbb82
Write CTRL: 0x00000000
Read STATUS: 0x00000040
Write CTRL: 0x50000000
Read STATUS: 0xf0000040
Read AP 0 IDR: 0x84770001
AP   0: IDR=84770001Read AP 0 CFG: 0x00000000
Read AP 0 CSW: 0x03800052
Read AP 0 BASE: 0xf0000002
 CFG=00000000 BASE=f0000000 CSW=a3800040 (AHB3-AP var0 rev8)
Read STATUS: 0xf0000040
Write AP 0 CSW: 0xa3800042
Write AP 0 TAR: 0xe000edf0
Write AP 0 DRW: 0xa05f0001
Read RDBUFF: 0xf0000002
Write CTRL: 0xf0fff040
Write AP 0 DRW: 0xa05f0003
Read AP 0 DRW: 0xf0000002
Write CTRL: 0xf0fff040
Write AP 0 DRW: 0xa05f0003
```

The BASE.P bit is reported as zero. Although the address points to a valid ROM table and obviously skipping this AP makes the device unavailable. My experience with ADI etc. is just what I gathered while coding this. Reading the ADIv5 docs, the meaning of BASE.P field is sowewhat a mystery to me as the description in the docs is quite short. They talk about 'debug components' then say that the flag denotes the presence of 'debug entry' which.. is the same thing or not? So I will not argue that simply ignoring the bit is the correct way as the device might be buggy as well. But one thing is I wonder what other devices (with a ROM table) report. I noticed that this part of code was rewritten somewhat recently (at one point mspm0 support I wrote stopped working after a merge with the main) and wonder how much real-world testing it has gotten. The other thing is how to come around this issue if probing is supposed to ignore devices with zero BASE.P values. Check manufacturer/partnum and work around?

Somewhat unrelated, but doubt it warrants a separate issue. I noticed there's a bug in the original (unmodified) code. The function first discards lower bits of `ap.base` and later tests if address is equal to NOTPRESENT (in legacy format). This is not correct as ap.base is already mangled and will never be equal to 0xffff.ffff.

I think there should be a test right after reading BASE register, like so:
```
		ap.base = adiv5_ap_read(&ap, ADIV5_AP_BASE_LOW);
		if (!(ap.flags & ADIV5_AP_FLAGS_64BIT) && ap.base == ADIV5_AP_BASE_NOT_PRESENT) {
			DEBUG_INFO(" -> Not Present (Legacy)\n");
			return NULL;
		}
```

dragonmux: There's a lot here, so we're going to answer this in parts:

> The BASE.P bit is reported as zero. Although the address points to a valid ROM table and obviously skipping this AP makes the device unavailable. My experience with ADI etc. is just what I gathered while coding this. Reading the ADIv5 docs, the meaning of BASE.P field is sowewhat a mystery to me as the description in the docs is quite short. They talk about 'debug components' then say that the flag denotes the presence of 'debug entry' which.. is the same thing or not? So I will not argue that simply ignoring the bit is the correct way as the device might be buggy as well. But one thing is I wonder what other devices (with a ROM table) report. I noticed that this part of code was rewritten somewhat recently (at one point mspm0 support I wrote stopped working after a merge with the main) and wonder how much real-world testing it has gotten. The other thing is how to come around this issue if probing is supposed to ignore devices with zero BASE.P values. Check manufacturer/partnum and work around?

So, that bit in non-legacy mode performs the exact same function as `ADIV5_AP_BASE_NOT_PRESENT`, except with inverted logic - those bottom two bits of BASE being 2 means that the device is telling the debugger that there are no entries (no ROM table, no debug components, nothing) present on the AP. This would ordinarily be correct and is in fact exactly how parts like the STM32H5's and STM32H7's and the LPC4370 all tell us of APs that should be ignored. Failure to ignore those APs can very well crash the parts.

For example:
```
❯ ./blackmagic -s 8BB20695 -tpv 5
Black Magic Debug App v1.10.0-844-ga44f08695-dirty
Using 1d50:6018 8BB20695 Black Magic Debug
 Black Magic Probe v1.10.0-843-g070d59f21-dirty
Running in Test Mode
Target voltage: 3.2V Volt
Speed set to 1.951MHz for SWD
Handling SWD multi-drop, TARGETID 0x04830041
DP DPIDR 0x6ba02477 (v2 rev6) designer 0x43b partno 0xba
TARGETID 0x04830041 designer 0x20 partno 0x4830
AP   0: IDR=84770001 CFG=00000000 BASE=e00fe000 CSW=a3800040 (AHB3-AP var0 rev8)
Halt via DHCSR(00030003): success after 28ms
ROM: Table BASE=0xe00fe000 SYSMEM=1, Manufacturer 020 Partno 483 (PIDR = 0x00000000000a0483)
ROM: Table BASE=0xe00ff000 SYSMEM=1, Manufacturer 43b Partno 4c7 (PIDR = 0x00000004000bb4c7)
 0 0xe000e000: Generic IP component - Cortex-M4 SCS (System Control Space) (PIDR = 0x00000004000bb00c DEVTYPE = 0x00 ARCHID = 0x0000)
 -> cortexm_probe
CPUID 0x411fc272 (M7 var 1 rev 2)
Calling stm32f1_probe
Calling stm32f4_probe
Calling stm32h5_probe
Calling stm32h7_probe
 1 0xe0001000: Generic IP component - Cortex-M3 DWT (Data Watchpoint and Trace) (PIDR = 0x00000004000bb002 DEVTYPE = 0x00 ARCHID = 0x0000)
 2 0xe0002000: Generic IP component - Cortex-M7 FBP (Flash Patch and Breakpoint) (PIDR = 0x00000004000bb00e DEVTYPE = 0x00 ARCHID = 0x0000)
 3 0xe0000000: Generic IP component - Cortex-M3 ITM (Instrumentation Trace Module) (PIDR = 0x00000004000bb001 DEVTYPE = 0x00 ARCHID = 0x0000)
 4 Entry 0xfff41002 -> Not present
 5 Entry 0xfff42002 -> Not present
 ROM: Table END
1 0xe0041000: Debug component - Cortex-M7 ETM (Embedded Trace) (PIDR = 0x00000004001bb975 DEVTYPE = 0x13 ARCHID = 0x4a13)
2 0xe0043000: Debug component - CoreSight CTI (Cross Trigger) (PIDR = 0x00000004004bb906 DEVTYPE = 0x14 ARCHID = 0x0000)
3 Entry 0x1ff02002 -> Not present
ROM: Table END
AP   1: IDR=84770001 -> Not Present
AP   2: IDR=54770002 CFG=00000000 BASE=e00e0000 CSW=80000040 (APB2/3-AP var0 rev5)
ROM: Table BASE=0xe00e0000 SYSMEM=0, Manufacturer 020 Partno 483 (PIDR = 0x00000000000a0483)
0 0xe00e1000: 0x00000000000a0000 Non-ARM component ignored
  -> designer: 20, part no: 0
1 Entry 0x2002 -> Not present
2 0xe00e3000: Debug component - CoreSight SWO (Single Wire Output) (PIDR = 0x00000004002bb914 DEVTYPE = 0x11 ARCHID = 0x0000)
3 Entry 0x4002 -> Not present
4 Entry 0x5002 -> Not present
ROM: Table BASE=0xe00f0000 SYSMEM=0, Manufacturer 020 Partno 001 (PIDR = 0x00000000000a0001)
 0 0xe00f1000: Debug component - CoreSight CTI (Cross Trigger) (PIDR = 0x00000004005bb906 DEVTYPE = 0x14 ARCHID = 0x0000)
 1 Entry 0x2002 -> Not present
 2 0xe00f3000: Debug component - CoreSight CSTF (Trace Funnel) (PIDR = 0x00000004003bb908 DEVTYPE = 0x12 ARCHID = 0x0000)
 3 0xe00f4000: Debug component - CoreSight TMC (Trace Memory Controller) (PIDR = 0x00000004001bb961 DEVTYPE = 0x32 ARCHID = 0x0000)
 4 0xe00f5000: Debug component - CoreSight TPIU (Trace Port Interface Unit) (PIDR = 0x00000004005bb912 DEVTYPE = 0x11 ARCHID = 0x0000)
 5 Entry 0x6002 -> Not present
 6 Entry 0x7002 -> Not present
 7 Entry 0x8002 -> Not present
 ROM: Table END
ROM: Table END
```

For which that AP 1 entry:
```
Read AP 1 IDR: 0x84770001
AP   1: IDR=84770001
Read AP 1 CFG: 0x00000000
Read AP 1 CSW: 0x43800042
Read AP 1 BASE: 0x00000002
 -> Not Present
 ```

If BMD tries to read that AP, it crashes the part.

Now, in this case, because the BASEADDR is non-zero (even if a very strange value.. this part is doing a lot of weird things), we can probably get away with an errata check that base matches this part's address value, and skip the skip.

> Somewhat unrelated, but doubt it warrants a separate issue. I noticed there's a bug in the original (unmodified) code. The function first discards lower bits of ap.base and later tests if address is equal to NOTPRESENT (in legacy format). This is not correct as ap.base is already mangled and will never be equal to 0xffff.ffff.

Having just checked that logic, this is what we get for moving stuff around and having next to no parts to test that use the legacy non-presence value. You are correct this is wrong, what probably needs to happen is that
```c
		/* Make sure we only pay attention to the base address, not the presence and format bits */
		ap.base &= ADIV5_AP_BASE_BASEADDR;
```
gets moved down to just after the presence check, which then makes any tests done in the presence check if block a bit simpler.

Note that the comment inside that block documents another part (STM32MP15) which reads back with a non-legacy non-presence marker on one of its PAs, and another where reading that AP anyway precipitates a crash.

hardesk: Please see the latest (29e45dd) update. I think I fixed the legacy no-entry check and made so the discovery passes for TI parts when base-addr is non-zero. Don't really like using manufacturer number in that function, but there are similar ifs in a adiv5_new_ap-calling function, so maybe this is not too ugly. Should I check for partnums as well? But the issue may apply for the whole family (or may not, can't tell for other than MSPM0L1306 variant).

src/target/adiv5.c
```diff
@@ -766,18 +764,27 @@ adiv5_access_port_s *adiv5_new_ap(adiv5_debug_port_s *dp, uint8_t apsel)
...
+			* No debug entries... useless AP
+			* AP0 on STM32MP157C reads 0x00000002
+			* Unless it's a TI MSPM0, which reads present flag as no entry while reporting a ROM table
+			*/
+			if (!(dp->target_designer_code == JEP106_MANUFACTURER_TEXAS && ap.base != 0U)) {
+				DEBUG_INFO(" -> Not Present\n");
+				return NULL;
+			}
+		}
+
```
dragonmux, line 787: We're not especially a fan that this introduces a repeated return (this code is fine fwiw, we'd just like not to have the condition return repeated), perhaps this suggestion will work (also noting we don't find the inserted extra new line particularly helpful, so that's dropped in the suggestion)?

```suggestion
		/* Check the Debug Base Address register for not-present. See ADIv5 Specification C2.6.1 */
		if (base_flags == (ADIV5_AP_BASE_FORMAT_ADIV5 | ADIV5_AP_BASE_PRESENT_NO_ENTRY) ||
			(!(ap.flags & ADIV5_AP_FLAGS_64BIT) && (uint32_t)ap.base == ADIV5_AP_BASE_NOT_PRESENT)) {
			/*
			 * Debug Base Address not present in this MEM-AP
			 * No debug entries... useless AP
			 * AP0 on STM32MP157C reads 0x00000002
			 * 
			 * NB: MSPM0 parts erroneously set BASE.P = 0 despite there being
			 * valid debug components on AP0, so we have to have an exception
			 * for this part family.
			 */
			if (dp->target_designer_code != JEP106_MANUFACTURER_TEXAS || (ap.base & ADIV5_AP_BASE_BASEADDR) == 0U) {
				DEBUG_INFO(" -> Not Present\n");
				return NULL;
			}
		}
		/* Make sure we only pay attention to the base address, not the presence and format bits */
		ap.base &= ADIV5_AP_BASE_BASEADDR;
```

src/target/adiv5.c
dragonmux: The general idea looks sound and as you note, there's precedent for checking the part manufacturer code there so that's perfectly fine. The main thing we're suggesting below is a reduction in the repeated return condition by condensing all the checks. It is a suggestion only as technically the code you've come up with works.

src/target/adiv5.c
hardesk: Now it's only almost correct. This probably won't ever happen in real life as legacy is legacy, but with the proposed change if the BASE is in legacy and equal to -1 (aka not present) and it's TI then testing a masked ap.base for non-null will succeed and such AP will be deemed as present. Culling out this case (legacy not present) early and forgetting about it seems easier albeit with an additional return.

One goto could solve the additional return problem though ;)

In my opinion the proposed version makes the code harder on the logic like multiple tests for the same flags (64-bit) and masking ap.base twice, but if the goal is to optimize for the lowest return statements possible, then this is the way to go. Then because legacy not-present and ADIv5 no-entry checks are lumped together, would need to add check against 64bit and ADIV5_AP_BASE_NOT_PRESENT in the if that checks againt TI designer code. I'm not too fond of it.

dragonmux: We'd really like to avoid a goto because that turns the codegen to absolute slop; Given this very specific case and that we know the expected base value is `0xf0000002U`, and you indeed pointing out we got the logic wrong there for legacy.. that would make the inner guard:
```c
if (dp->target_designer_code != JEP106_MANUFACTURER_TEXAS || ap.base != 0xf0000002U) {
```

That helps to key directly to this specific exception without breaking the check for legacy not-present, or over-complicating the checks etc. Does that work for you?

hardesk: Right, it does create a few more lines vertical code, but it simplifies the conditions which I would argue increases the maintainability, not decreases it. I support the concern regarding forgetting to update the text if modifying the messages, but aside from increasing executable size if not updated in sync that is a minor issue. Speaking of size, your proposed is 26 bytes extra compared to my version (either goto or string duplication). The difference is only 4 bytes if TI check is removed.

Anyway, while I still think my version is simpler (and happens to be more compact) you're the maintainer so it's your decision. I'm fine making the code as you proposed and let's move on.

hardesk:
Thanks for all the feedback. With all of it I believe addressed there's that one issue left where code doesn't fit into 128k. Linux builds are failing. So what course of action would you suggest to tackle it?

dragonmux (commented):
Apologies we forgot to post this review. There are a couple of items found going back through this and please rebase this on `main`.

src/target/mspm0.c
```diff
@@ -0,0 +1,396 @@
...
+	mspm0_flash_s *mspm0_flash = (mspm0_flash_s *)target_flash;
+	unsigned sector = (addr - target_flash->start) / MSPM0_FLASH_SECTOR_SZ;
+
+	if (sector < 32U) {
+		uint32_t mask = ~(1U << sector);
+		target_mem32_write32(target_flash->t, MSPM0_FLASHCTL_CMDWEPROTA, mask);
+	} else if (sector < 256U) {
+		/* Depending whether main flash is multibank or not, the first 4 bits of PROTB either
+		start at 32k or overlay over PROTA responsible range. */
+		unsigned start_protb_sector = mspm0_flash->banks <= 1 ? 32U : 0U;
```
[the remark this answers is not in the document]
hardesk: Can you please explain how a possible conversion (from what into what) could happen in the next line?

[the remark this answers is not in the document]
hardesk: While I will change to uint32_t for the peace of mind, I do not see how the code could cause any issues. Both operands for subtraction are `unsigned int` so no integer promotion will happen as they are unsigned `ints` already. I guess they key point here is that the expression result is used as a shift amount value so any type ripple stops right there and does not affect `uint32_t mask`.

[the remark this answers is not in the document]
dragonmux: The `uint32_t` argument is `unsigned long` in this case, which is a different length when `int` is 16-bit on a platform, causing promotion which promotes to *signed* (your `unsigned` becomes `long`). Once that happens, the shifting is all done on *signed* numbers and you open the calculation to all of the pitfalls mentioned above. Remember that just because it's okay on one platform does not mean it's okay on all, because this is a platform-dependant issue.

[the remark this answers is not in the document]
hardesk: I guess what I fail to see is how the type of the shift amount can influence the resulting type of the shift operation (provided that the shift-by value calculates to a correct one). While the current type of `sector` is indeed problematic and should be kept as `target_addr_t`, or is uint32_t good enough?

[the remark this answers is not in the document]
dragonmux: `uint32_t` is sufficient. Because the target is a 32-bit target, that particular narrowing won't cause problems, and narrowing never signed-unsigned converts.

RX14:
I've tested this out using my stlinkv3 flashed with BMP firmware, and from what I can tell it all seems to work great. I haven't explored the reaches of functionality but it detects the target, loads code, runs breakpoints and all other functions I needed so far.

src/target/mspm0.c
```diff
@@ -0,0 +1,385 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <assert.h>
+#include <limits.h>
```
dragonmux, line 21: Due to some macros that `general.h` sets which drastically changes the compilation environment when these headers include the system header `features.h`, please put these after `general.h`

src/target/mspm0.c
```diff
@@ -0,0 +1,385 @@
...
+static command_s mspm0_cmds_list[] = {
+	{"dump_factory", mspm0_dump_factory_config, "Display FACTORY registers"},
+	{"dump_bcr", mspm0_dump_bcr_config, "Display NONMAIN (BCR/BSL) registers"},
+	{NULL, NULL, NULL},
+};
+
+typedef struct conf_register {
+	uint16_t reg_offset;
+	uint16_t size_words;
+	char const *id;
```
dragonmux, line 113: Please use east `const` here, `const char *` - that is how pointers are written in this code base.

src/target/mspm0.c
```diff
@@ -0,0 +1,385 @@
...
+	{0x148U, 1U, "SBLADDRESS"},
+	{0x14cU, 1U, "BSLAPPVER"},
+	{0x150U, 1U, "BSLCONFIG1"},
+	{0x154U, 1U, "BSLCRC"},
+	{0U, 0U, NULL},
+};
+
+static void mspm0_dump_regs(target_s *const target, const conf_register_s *regs, uint32_t base)
+{
+	for (conf_register_s const *r = regs; r->id; ++r) {
```
dragonmux, line 162: Ditto for this, along with a non-single-letter variable such as `reg` (ie `const conf_register_s *reg = regs;`)

src/target/mspm0.c
```diff
@@ -0,0 +1,385 @@
...
+
+bool mspm0_probe(target_s *const target)
+{
+	uint32_t const deviceid = target_mem32_read32(target, MSPM0_FACTORYREGION_DEVICEID);
+
+	uint32_t manufacturer = (deviceid & MSPM0_DEVICEID_MANUFACTURER_MASK) >> MSPM0_DEVICEID_MANUFACTURER_SHIFT;
+	if (manufacturer != JEP106_MANUFACTURER_TEXAS)
+		return false;
+
+	uint32_t partnum = (deviceid & MSPM0_DEVICEID_PARTNUM_MASK) >> MSPM0_DEVICEID_PARTNUM_SHIFT;
```
dragonmux, line 216: If these two variables can be made `const`, that would be good please.

dragonmux (commented):
There are a couple of items we spotted in doing a re-review, but nothing arduous. With them handled, and confirmation from another user that this support works, we'll be happy to merge this. Please rebase on `main` and we'll aim to have it merged in the next couple of days.

dragonmux (approved):
This looks good now, merging. Thank you for the contribution! We'll worry about the two build failures in a follow-up PR as those are both on community supported platforms and suggest those platforms need their targets configurations massaging.

PR #1852, opened 2024-06-27
Feature: AT32F435/437 Option Bytes support

ALTracer:
## Detailed description

* This is technically a new feature in a driver I contributed.
* The existing problem is at32f43x.c not supporting option bytes, which are required for e.g. RDP1 and EOPB0 modification.
* This PR solves it by porting driver code from stm32f1.c and adapting where necessary.

ArteryTek MCU AT32F435, F437 chips contain 512 byte User System Data regions (in SKUs with 2048 byte flash page size), or 4096 byte (for 4096-byte page size), that is 256 and 2048 halfwords [byte-invbyte]. The first 8 are very similar to STM32F1 in function, like 0x5AA5 for RDP1, but during testing via BMDA I encountered some functional differences. Also it's too long to overwrite all of them (and print), so some shortcuts were taken. Compared to #1803 my intent is to get this merged, because I don't have to tread carefully over legacy & widely-used stm32f1.c code. We needed to touch Userdata0-Userdata1 bytes, and on first pass the writes sticked, so I consider this somewhat working.

Remaining issue: I easily managed to escalate from unprotected to RDP1, but couldn't degrade back to unprotected, BMDA kept giving write errors. This needs to be investigated and resolved for soft-unbricking of targets. Observation: under RDP1 neither Flash & USD nor Unique ID is readable by debugger, which poses a problem for efficient identification (just PID+UID+F_SIZE).

If this gets blocked by `native` flash utilization, then I might have to do something about part_id jumptables, but the question about detecting RDP1-locked parts remains. If I can refactor this driver to take both `0x40023c00` and `0x40022000` FLASH_REG_BASE, then it could support __all__ AT32F4 parts in a generic fashion (which also helps containing those jumptables and dropping one vendor-specific driver from Meson/Makefiles config, disentangling from other stm32f1 targets).

## Closing issues

Not self-opening one, but there could be a checklist for features (sLib, EPP/WRP, encryption etc.)

### DIFF
--- a/src/target/at32f43x.c
+++ b/src/target/at32f43x.c
@@ -420,9 +420,14 @@ static bool at32f43_option_write_erased(target_s *const target, const size_t off
 
 	const uint32_t addr = AT32F43x_USD_BASE + (offset * 2U);
 	DEBUG_TARGET("%s: 0x%08" PRIX32 " <- 0x%04X\n", __func__, addr, value);
+	const uint32_t time_start = platform_time_ms();
 	target_mem32_write16(target, addr, value);
 
 	const bool result = at32f43_flash_busy_wait(target, 0, NULL);
+	const uint32_t time_end = platform_time_ms();
+	const uint32_t time_spent = time_end - time_start;
+	if (time_spent > 20U)
+		DEBUG_TARGET("%s: took %" PRIu32 " ms\n", __func__, time_spent);
 	if (result || offset != 0U)
 		return result;
 
--- a/src/target/at32f43x.c
+++ b/src/target/at32f43x.c
@@ -83,7 +83,8 @@ static bool at32f43_mass_erase(target_s *target);
 #define AT32F43x_USD_BASE 0x1fffc000U
 /* First option byte value for disabled read protection: 0x00a5 */
 #define AT32F43x_USD_RDP_KEY 0x5aa5U
-#define AT32F43x_OB_COUNT    256U
+#define AT32F43x_2K_OB_COUNT 256U
+#define AT32F43x_4K_OB_COUNT 2048U
 
 #define DBGMCU_IDCODE 0xe0042000U
 
@@ -430,8 +431,9 @@ static bool at32f43_option_write_erased(target_s *const target, const size_t off
 
 static bool at32f43_option_write(target_s *const target, const uint32_t addr, const uint16_t value)
 {
-	/* Arterytek F435/F437 has 256 option byte halfwords (512 bytes) */
-	const uint16_t ob_count = AT32F43x_OB_COUNT;
+	/* Arterytek F435/F437 has either 512 bytes or 4 KiB worth of USD */
+	const target_flash_s *target_flash = target->flash;
+	const uint16_t ob_count = target_flash->blocksize == 4096U ? AT32F43x_4K_OB_COUNT : AT32F43x_2K_OB_COUNT;
 
 	const uint32_t index = (addr - AT32F43x_USD_BASE) >> 1U;
 	/* If this underflows, then address is out of USD range */
@@ -439,7 +441,24 @@ static bool at32f43_option_write(target_s *const target, const uint32_t addr, co
 		return false;
 
 	bool erase_needed = false;
-	uint16_t opt_val[ob_count]; // FIXME: big VLA, replace with malloc+check
+	uint16_t opt_val_single = target_mem32_read16(target, addr);
+	/* No change pending */
+	if (opt_val_single == value)
+		return true;
+	/* Check whether erase is needed */
+	if (opt_val_single != 0xffffU)
+		erase_needed = true;
+	/* Flip single pair-of-bytes from 0xffff to desired value and exit */
+	if (!erase_needed)
+		return at32f43_option_write_erased(target, index, value);
+
+	uint16_t *const opt_val = calloc(ob_count, sizeof(uint16_t));
+	if (!opt_val) {
+		DEBUG_ERROR("calloc: failed in %s\n", __func__);
+		return false;
+	}
+	DEBUG_TARGET("%s: full overwrite triggered\n", __func__);
+
 	/* Save current values */
 	for (size_t i = 0U; i < ob_count * 2U; i += 4U) {
 		const size_t offset = i >> 1U;
@@ -447,32 +466,27 @@ static bool at32f43_option_write(target_s *const target, const uint32_t addr, co
 		opt_val[offset] = val & 0xffffU;
 		opt_val[offset + 1U] = val >> 16U;
 	}
-
-	/* No change pending */
-	if (opt_val[index] == value)
-		return true;
-
-	/* Check whether erase is needed */
-	if (opt_val[index] != 0xffffU)
-		erase_needed = true;
-
 	/* Update requested entry locally */
 	opt_val[index] = value;
 
-	/* Flip single pair-of-bytes from 0xffff to desired value and exit */
-	if (!erase_needed)
-		return at32f43_option_write_erased(target, index, value);
-
 	/* Wipe everything and write back. Writing matching values without an erase raises a PRGMERR. */
-	if (!at32f43_option_erase(target))
-		return false;
-	/* Write changed values using 16-bit accesses. */
-	for (size_t i = 0U; i < ob_count; ++i) {
-		if (!at32f43_option_write_erased(target, i, opt_val[i]))
-			return false;
-	}
+	bool result = true;
+	do {
+		if (!at32f43_option_erase(target)) {
+			result = false;
+			break;
+		}
+		/* Write changed values using 16-bit accesses. */
+		for (size_t i = 0U; i < ob_count; ++i) {
+			if (!at32f43_option_write_erased(target, i, opt_val[i])) {
+				result = false;
+				break;
+			}
+		}
+	} while (false);
 
-	return true;
+	free(opt_val);
+	return result;
 }
 
 static bool at32f43_cmd_option(target_s *target, int argc, const char **argv)
@@ -505,15 +519,22 @@ static bool at32f43_cmd_option(target_s *target, int argc, const char **argv)
 		/* Try and program the new option value to the requested option byte */
 		if (!at32f43_option_write(target, addr, val))
 			return false;
+		/* Display only changes */
+		const uint16_t val_new = target_mem32_read16(target, addr);
+		tc_printf(target, "0x%08" PRIX32 ": 0x%04X\n", addr, val_new);
+		return true;
 	} else
 		tc_printf(target, "usage: monitor option erase\nusage: monitor option <addr> <value>\n");
 
 	/* When all gets said and done, display the current option bytes values */
-	for (size_t i = 0U; i < AT32F43x_OB_COUNT * 2U; i += 4U) {
+	const target_flash_s *target_flash = target->flash;
+	const uint16_t ob_count = target_flash->blocksize == 4096U ? AT32F43x_4K_OB_COUNT : AT32F43x_2K_OB_COUNT;
+	uint16_t values[8] = {0};
+	for (size_t i = 0U; i < ob_count * 2U; i += 16U) {
 		const uint32_t addr = AT32F43x_USD_BASE + i;
-		const uint32_t val = target_mem32_read32(target, addr);
-		tc_printf(target, "0x%08X: 0x%04X\n", addr, val & 0xffffU);
-		tc_printf(target, "0x%08X: 0x%04X\n", addr + 2U, val >> 16U);
+		target_mem32_read(target, values, addr, 8 * sizeof(uint16_t));
+		tc_printf(target, "0x%08" PRIX32 ": 0x%04X 0x%04X 0x%04X 0x%04X 0x%04X 0x%04X 0x%04X 0x%04X\n", addr, values[0],
+			values[1], values[2], values[3], values[4], values[5], values[6], values[7]);
 	}
 
 	return true;
--- a/src/target/at32f43x.c
+++ b/src/target/at32f43x.c
@@ -438,6 +438,7 @@ static bool at32f43_option_write(target_s *const target, const uint32_t addr, co
 	if (index > ob_count - 1U)
 		return false;
 
+	bool erase_needed = false;
 	uint16_t opt_val[ob_count]; // FIXME: big VLA, replace with malloc+check
 	/* Save current values */
 	for (size_t i = 0U; i < ob_count * 2U; i += 4U) {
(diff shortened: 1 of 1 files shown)


src/target/at32f43x.c
```diff
@@ -304,6 +324,8 @@ static bool at32f43_flash_erase(target_flash_s *target_flash, target_addr_t addr
 	}
 
 	at32f43_flash_clear_eop(target, bank_reg_offset);
+	DEBUG_TARGET("%s: 0x%08" PRIX32 "+%" PRIu32 " reg_base 0x%08" PRIX32 "\n", __func__, addr, (uint32_t)len,
+		bank_reg_offset + AT32F43x_FLASH_REG_BASE);
```
dragonmux, line 328: This and the change in the Flash write routine are unrelated to the option bytes support, so should perhaps be dropped from this PR. A version of them does seem potentially very useful though.

src/target/at32f43x.c
```diff
@@ -364,3 +388,150 @@ static bool at32f43_mass_erase(target_s *target)
...
+
+#define STM32F1_OB_COUNT     8U
+#define AT32F43x_2K_OB_COUNT 256U
+#define AT32F43x_4K_OB_COUNT 2048U
+
+static bool at32f43_option_write(target_s *const target, const uint32_t addr, const uint16_t value)
+{
+	/* Arterytek F435/F437 has either 512 bytes or 4 KiB worth of USD */
+	const target_flash_s *target_flash = target->flash;
+	const uint16_t ob_count = (target_flash->blocksize == 4096U) ? AT32F43x_4K_OB_COUNT : AT32F43x_2K_OB_COUNT;
```
dragonmux, line 436: You can drop the parens here as they aren't disambiguating anything - the comparison condition binds tighter than `?` which binds tighter than `=`.

src/target/at32f43x.c
```diff
@@ -364,3 +388,150 @@ static bool at32f43_mass_erase(target_s *target)
...
+
+	/* For error on offset 0, that is RDP byte, signal back the failure to erase RDP (?) */
+	const uint8_t status =
+		target_mem32_read32(target, AT32F43x_FLASH_STS) & (AT32F43x_FLASH_STS_PRGMERR | AT32F43x_FLASH_STS_EPPERR);
+	return status == AT32F43x_FLASH_STS_PRGMERR;
+}
+
+#define STM32F1_OB_COUNT     8U
+#define AT32F43x_2K_OB_COUNT 256U
+#define AT32F43x_4K_OB_COUNT 2048U
```
dragonmux, line 430: We're glad to see these defined like this, just they should up at the top with the other defines please.

src/target/at32f43x.c
```diff
@@ -364,3 +388,150 @@ static bool at32f43_mass_erase(target_s *target)
...
+	/* No change pending */
+	if (opt_val_single == value)
+		return true;
+	/* Check whether erase is needed */
+	if (opt_val_single != 0xffffU)
+		erase_needed = true;
+	/* Flip single pair-of-bytes from 0xffff to desired value and exit */
+	if (!erase_needed) {
+		return at32f43_option_write_erased(target, index, value);
+	}
```
dragonmux, line 454: Extra braces, please drop them.

src/target/at32f43x.c
```diff
@@ -364,3 +388,150 @@ static bool at32f43_mass_erase(target_s *target)
...
+			break;
+		}
+		/* Write changed values using 16-bit accesses. */
+		for (size_t i = 0U; i < ob_count; ++i) {
+			if (!at32f43_option_write_erased(target, i, opt_val[i])) {
+				result = false;
+				break;
+			}
+		}
+	} while (false);
```
dragonmux, line 484: Is there a particular reason this is wrapped in a `do`-`while (false)`? We don't mind if there is a good reason, it should just be noted because this pattern is otherwise extremely surprising and eyebrow-raising.

src/target/at32f43x.c
```diff
@@ -364,3 +388,150 @@ static bool at32f43_mass_erase(target_s *target)
...
+		tc_printf(target, "usage: monitor option erase\nusage: monitor option <addr> <value>\n");
+
+	/* When all gets said and done, display the current option bytes values */
+	const target_flash_s *target_flash = target->flash;
+	const uint16_t ob_count = (target_flash->blocksize == 4096U) ? AT32F43x_4K_OB_COUNT : AT32F43x_2K_OB_COUNT;
+	for (size_t i = 0U; i < ob_count * 2U; i += 4U) {
+		const uint32_t addr = AT32F43x_USD_BASE + i;
+		const uint32_t val = target_mem32_read32(target, addr);
+		tc_printf(target, "0x%08X: 0x%04X\n", addr, val & 0xffffU);
+		tc_printf(target, "0x%08X: 0x%04X\n", addr + 2U, val >> 16U);
```
dragonmux, line 533: `PRIX32` for the `uint32_t`'s here please (and same in other prints above)

dragonmux (changes requested):
This is looking good. There are a few small stylistic items and a couple of queries, once they're taken care of this should be about ready to merge.

src/target/at32f43x.c
ALTracer: It used to early-return, now I can't leak the calloc pointer. Using goto cleanup for free sounded worse.

src/target/at32f43x.c
ALTracer: tc_printf attribute format is not merged, so I didn't notice this while copying. I thought to re-enable it locally, but then I'd get lots of noise from all the other TUs. Also I don't like the 2048 line long list, is there another way to present it?

dragonmux: The next best way we can think to present that would be to have lines of 8 16-bit values, which does complicate the printing logic, but which then improves display.. in the short term, you could move the high 16 bits display into the same print as the low so that at least it halves the number of lines to 1024, which is still a lot but less.

src/target/at32f43x.c
ALTracer: Assume I forgot to add this quality of life feature back when I wrote the driver. stm32f1.c has it, as does stm32h7.c fwiw (the two familiar to me drivers). If I drop this change, then either it makes for its own small PR, or gets postponed to the flash_reg_base refactor. I chose lesser of evils. Still, it's in its own separate commit. Should I add similar log statements to option_write_erased/option_erase?

src/target/at32f43x.c
ALTracer: Block refactored, please review again. Used PRIX32 for uint32_t; as values[] are uint16_t, 0x%04X flies for this.

src/target/at32f43x.c
dragonmux: Oh, that's fair then. Ok, we'll allow that despite being technically unrelated. Not sure whether to bother in the option bytes side of things.. though, given how option bytes is usually the sorest spot in the STM32 support implementations, perhaps that is a good idea to do.

If you think you can add reasonable and useful logging into the option bytes erase and write routines, then yes, please do so.

dragonmux (commented):
Fixes and clean-up so far looks good - leaves just a couple of items still to deal with and then hopefully we can merge. Nice work!

src/target/at32f43x.c
ALTracer: Added logging. Detected always full erases. Noticed a bug on line 444, I added an extra offset in fast path, making it never succeed and go slow path.

src/target/at32f43x.c
ALTracer: Dropped parens in both cases of this ternary.

ALTracer:
The good news is this approach technically can work. I've tested it and successfully unlocked Flash Access Protection using Black Magic Debug, without relying on OpenOCD fork or JLink tools or ICP. The bad news is my target, AT32F437ZM, has Flash (M) 4032 KiB and takes 14.4 seconds (typ) to mass-erase. The `*0x1fffc000=0x5aa5` write which degrades from RDP to Unlocked state, does not complete at SW-DP level, giving WAIT Acks *all that time* (size-specific).

CMSIS-DAP backend throws a bunch off timeout errors and gives up, but the mass erase completes and I could reattach (without reset) and write option bytes a second time. BMP-HL backend returns an error in 250 ms, writes DP ABORT under the hood, halting the mass erase. BMP non-HL backend I could quickly patch to wait 15000 ms and that completed faithfully poking SW-DP with RDBUFF for ~15 seconds. This may need some workaround.

I attached fx2la to vendor's In-Circuit Programmer over AT-Link EZ onboard and ran their command for Disabling RDP1, which seemed to perform roughly the same FPEC writes and then polled for 14 seconds RDBUFF-WAIT, drawing a GUI progress bar, after which it notified me "Read protection disabled!". As it's their custom flash reprogramming tool which does not need to do debug, they can do whatever via CMSIS-DAP API and JLink API.

Normal option bytes erasing and writing works, I've never faced an unexpected RDP1 or lost data -- I had to perform `(gdb) monitor option 0x1fffc000 0xffff` to specifically ask BMD to erase USD and *not* set the RDP-disabled value, then issue a hardware system reset (or a power-on reset). The added bonus is this command allows flipping an erased halfword 0xFFFF to arbitrary value once quickly, skipping the full save-erase-restore for 2000 halfwords.

dragonmux:
> CMSIS-DAP backend throws a bunch off timeout errors and gives up, but the mass erase completes and I could reattach (without reset) and write option bytes a second time. BMP-HL backend returns an error in 250 ms, writes DP ABORT under the hood, halting the mass erase. BMP non-HL backend I could quickly patch to wait 15000 ms and that completed faithfully poking SW-DP with RDBUFF for ~15 seconds. This may need some workaround.

We think it might be for a follow-up PR and particularly as this is going to impact every single STM32 target support implementation to some extent.. but our thought here is to take up on your own suggestion of the no-check ADIv5 write in some capacity and then poll every 10ms or so for WAIT to stop being returned.

Failing that being practical due to the need to handle other architectures such as in the STM32F1 support which needs to deal with RISC-V for the GD32VF103, another approach would be to introduce some way to request a custom timeout for a write we know is going to take a long time to complete, so we can use the current machinery but have it take longer to fail out via a DP ABORT.

As long as you're happy that it'll be a known issue and need fixing in another PR, then we're happy to move on with this PR and review the new changes with the intent to get this merged. May we suggest a rebase on `main` if you're happy with that path forward?

src/target/at32f43x.c
```diff
@@ -364,3 +393,155 @@ static bool at32f43_mass_erase(target_s *target)
...
+	if (value == 0xffffU)
+		return true;
+
+	at32f43_flash_clear_eop(target, 0);
+
+	/* Enable writing User System Data */
+	target_mem32_write32(target, AT32F43x_FLASH_CTRL, AT32F43x_FLASH_CTRL_USDPRGM | AT32F43x_FLASH_CTRL_USDULKS);
+
+	const uint32_t addr = AT32F43x_USD_BASE + (offset * 2U);
+	DEBUG_TARGET("%s: 0x%08X <- 0x%04X\n", __func__, addr, value);
```
dragonmux, line 422: `addr` is a uint32_t, so please use `PRIX32` or `PRIx32` for it here.

src/target/at32f43x.c
dragonmux: This smells like it should be refactored to become a full fat function then, but as this works and we're more interested at the moment in getting this support merged, we'll let that be as it is.

dragonmux (commented):
Only spotted the one item that needs tidying up in a review, so with that sorted and the branch rebased on `main`, we're happy to approve this and merge.

src/target/at32f43x.c
ALTracer: Yes, thanks. Verified with enabled DEBUG_TARGET, it used to fail the firmware build (-Wformat). Fixed.

src/target/at32f43x.c
ALTracer: It could be a real function. This idiom is effectively an anonymous function here, a poor man's lambda from C++ (or small try-catch exception replacement). For something like a string command parser it's more justified. Here I'd need to pass `*target`, `*opt_val`, `ob_count`, and return true. It could work. But OTOH I haven't tested whether calloc for 4096 flies on stlink/swlink with smaller heap (mitigate with PR1529).

ALTracer:
Rebased to main (past my own 1854). Build overflows for native with makefiles, but not with meson (gcc-12.2.rel1). This driver has grown twice its size with this new feature. I cannot reuse stm32f1.c code here because a) FPEC_BASE is different, b) allocations are larger, c) EOPB0 and other Arterytek-specific handling is required. OTOH I can drop `at32f43x.c` from src/Makefile and it still compiles, because I'm not actually relying on any stm32 code from adjacent files. Can you make it a separate target meson option?
To monitor the unprotecting mass-erase, I added another log statement in the top commit, which I picked from my stm32f1.c experiments. As long as timeouts in adiv5_swd.c, adiv5_jtag.c are bumped from 0.25 to 15 seconds, this should complete and report to BMDA stdout or BMF diagnostic log channel (also enable DEBUG_TARGET).

src/target/at32f43x.c
ALTracer: Inserted commit with suggested change. It doesn't matter because everything here is marked `static inline` and at -Os inlined properly. Please tell if I should squash it into the previous one for a cleaner history.

src/target/at32f43x.c
```diff
@@ -364,3 +393,164 @@ static bool at32f43_mass_erase(target_s *target)
...
+	return status == AT32F43x_FLASH_STS_PRGMERR;
+}
+
+static bool at32f43_option_overwrite(target_s *const target, const uint16_t *const opt_val, const uint16_t ob_count)
+{
+	if (!at32f43_option_erase(target))
+		return false;
+
+	/* Write changed values using 16-bit accesses. */
+	for (size_t i = 0U; i < ob_count; ++i)
```
dragonmux, line 446: The outer loop here should have braces added as, while this is legal, the `if` statement inside reads as a multi-line loop body and so the braces should be included in this situation.

src/target/at32f43x.c
dragonmux: Looks good - much cleaner and easier to understand 👍🏼.

src/target/at32f43x.c
ALTracer: The rule per ~~src/target/HACKING.md~~ multiple PR feedback seems to be
```c
if (expr)
    statement1;
else
    statement2;
```
and
```c
for (;;)
    statement-iter;
```
but
```c
for (;;) {
    if (expr)
        statement1;
}
```
Maybe add this to the codestyle doc? Or there already is one? pre-commit/clang-format don't catch this, and I don't (yet) use clang-tidy.

dragonmux: Correct. It's something where we started to deal with working through the kernel's and trying to write one but got stuck dealing with the super bombastic and kinda horrible way the kernel style guide is written so lost steam.. it's something we have on our list to deal with though as yeah, need a style guide.

dragonmux (approved):
LGTM, thank you for the contribution! Merging as you've shown on Discord that this works.

PR #1853, opened 2024-06-28
CtxLink_usb

sidprice:
## Detailed description

This is an initial PR to bring ctxLink with basic USB functionality into the upstream repository.

TRACESWO is not tested/implemented in this firmware. It will be added  in the future when the ongoing F4 TRACESWO work is merged into main.

### DIFF
--- a/src/platforms/ctxlink/Makefile.inc
+++ b/src/platforms/ctxlink/Makefile.inc
@@ -28,11 +28,7 @@ SRC +=               \
 	timing.c	\
 	timing_stm32.c	\
 
-all:	blackmagic.bin blackmagic_dfu.bin blackmagic_dfu.hex
-
-blackmagic_dfu.elf: usbdfu.o dfucore.o dfu_f1.o serialno.o
-	@echo "  LD      $@"
-	$(Q)$(CC) $^ -o $@ $(LDFLAGS_BOOT)
+all:	blackmagic.bin
 
 host_clean:
 	-$(Q)$(RM) blackmagic.bin blackmagic_dfu blackmagic_dfu.bin blackmagic_dfu.hex
--- a/src/platforms/ctxlink/Makefile.inc
+++ b/src/platforms/ctxlink/Makefile.inc
@@ -8,7 +8,7 @@ CFLAGS += -Istm32/include -mcpu=cortex-m4 -mthumb \
 	-DSTM32F4 -I../deps/libopencm3/include \
 	-Iplatforms/common/stm32
 
-LINKER_SCRIPT=platforms/f4discovery/f4discovery.ld
+LINKER_SCRIPT=platforms/ctxlink/ctxlink.ld
 
 LDFLAGS_BOOT = -lopencm3_stm32f4 \
 	-Wl,-T,$(LINKER_SCRIPT) -nostartfiles -lc -lnosys \
--- a/src/platforms/ctxlink/Makefile.inc
+++ b/src/platforms/ctxlink/Makefile.inc
@@ -8,10 +8,9 @@ CFLAGS += -Istm32/include -mcpu=cortex-m4 -mthumb \
 	-DSTM32F4 -DCTX_LINK -I../deps/libopencm3/include \
 	-Iplatforms/common/stm32
 
-LINKER_SCRIPT=platforms/ctxlink/ctxlink.ld
-
 LDFLAGS_BOOT = -lopencm3_stm32f4 \
-	-Wl,-T,$(LINKER_SCRIPT) -nostartfiles -lc -lnosys \
+	-Tctxlink.ld -nostartfiles -lc -lnosys \
+	-Lplatforms/ctxLink \
 	-Wl,-Map=mapfile -mthumb -mcpu=cortex-m4 -Wl,-gc-sections \
 	-mfloat-abi=hard -mfpu=fpv4-sp-d16 \
 	-L../deps/libopencm3/lib
@@ -33,8 +32,7 @@ SRC +=               \
 	traceswo.c	\
 	serialno.c	\
 	timing.c	\
-	timing_stm32.c	\
-	ctxLink_gdb_if.c
+	timing_stm32.c
 
 ifneq ($(BMP_BOOTLOADER), 1)
 all:	blackmagic.bin
--- a/src/platforms/ctxlink/Makefile.inc
+++ b/src/platforms/ctxlink/Makefile.inc
@@ -1,0 +1,38 @@
+CROSS_COMPILE ?= arm-none-eabi-
+CC = $(CROSS_COMPILE)gcc
+OBJCOPY = $(CROSS_COMPILE)objcopy
+
+CFLAGS += -Istm32/include -mcpu=cortex-m3 -mthumb \
+	-DSTM32F1 -DBLACKMAGIC -I../deps/libopencm3/include \
+	-Iplatforms/common/stm32 -DDFU_SERIAL_LENGTH=9
+
+LDFLAGS_BOOT := $(LDFLAGS) --specs=nano.specs -lopencm3_stm32f1 \
+	-Wl,-T,platforms/ctxLink/ctxLink.ld -nostartfiles -lc \
+	-Wl,-Map=mapfile -mthumb -mcpu=cortex-m3 -Wl,-gc-sections \
+	-L../deps/libopencm3/lib
+LDFLAGS = $(LDFLAGS_BOOT) -Wl,-Ttext=0x8002000
+
+ifeq ($(ENABLE_DEBUG), 1)
+LDFLAGS += --specs=rdimon.specs
+else
+LDFLAGS += --specs=nosys.specs
+endif
+
+VPATH += platforms/common/stm32
+
+SRC +=               \
+	platform.c \
+	traceswodecode.c \
+	traceswo.c	\
+	serialno.c	\
+	timing.c	\
+	timing_stm32.c	\
+
+all:	blackmagic.bin blackmagic_dfu.bin blackmagic_dfu.hex
+
+blackmagic_dfu.elf: usbdfu.o dfucore.o dfu_f1.o serialno.o
+	@echo "  LD      $@"
+	$(Q)$(CC) $^ -o $@ $(LDFLAGS_BOOT)
+
+host_clean:
+	-$(Q)$(RM) blackmagic.bin blackmagic_dfu blackmagic_dfu.bin blackmagic_dfu.hex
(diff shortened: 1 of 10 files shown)


ALTracer:
To facilitate bringup of TraceSWO Async support via in-tree drivers, please document relevant chip resources in `platform.h`. I understand that it'll be PC7, `USART6_RX`, DMA2 Channel 5 Stream 1/2, and some SRAM. Possibly `TIM3_CH2` (for Manchester) on same pin. Is `JTDO_PIN` PC6 or PC7 on all PCB revisions (green LED1 idle/run)? Otherwise one could use HDSEL as single-wire half-duplex receiver from `USART6_TX` pin but it's a hack -- full pinswap is only available in newer IP versions.
I expect this F401RE to behave in a similar way to F411CE, that is 84MHz APB2 and OVER8 gives 10.5 Mbaud; OVER16 supported by libopencm3 gives 5.25 Mbaud, or with a BRR of 21 an even 4Mbaud. If you tested practical limits with your out-of-tree fork, it would be nice to have them recorded here in the platform header or `README.md`.

sidprice:
@ALTracer Thanks for the comments, once we have the basic USB PR handled I will revisit your comments,
Sid

src/platforms/ctxlink/Makefile.inc
```diff
@@ -0,0 +1,47 @@
...
+	-Wl,-Map=mapfile -mthumb -mcpu=cortex-m4 -Wl,-gc-sections \
+	-mfloat-abi=hard -mfpu=fpv4-sp-d16 \
+	-L../deps/libopencm3/lib
+
+ifeq ($(BMP_BOOTLOADER), 1)
+$(info  Load address 0x08004000 for BMPBootloader)
+LDFLAGS = $(LDFLAGS_BOOT) -Wl,-Ttext=0x8004000
+CFLAGS += -DDFU_SERIAL_LENGTH=9
+else
+LDFLAGS += $(LDFLAGS_BOOT)
```
dragonmux, line 23: `LDFLAGS` should not be defined already in this branch (and if it is, wants replacing) so we suggest this be switched from `+=` to just `=` same as the first branch.

src/platforms/ctxlink/README.md
```diff
@@ -0,0 +1,49 @@
...
+ * PC9 -> LED3 -> Green - ctxLink Mode
+
+## Connections
+
+* PA3: TDI
+* PA4: TMS/SWDIO
+* PA5: TCK/SWCLK
+* PC7: TDO/TRACESWO
+* PA2: TRST
+* PA2: nRST
```
dragonmux, line 18: This lists PA2 twice, is that correct or a typo? (especially given nRST is not nTRST)

src/platforms/ctxlink/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## How to Flash with DFU
+
+After build:
+
+1) `apt install dfu-util`
+2) Force ctxLink into system bootloader mode by holding down the Mode switch while
+   pressing reset. Release reset followed by Mode. System bootloader should appear.
+3) `dfu-util -a 0 --dfuse-address 0x08000000 -D blackmagic_ctxlink_firmware.elf`
```
dragonmux, line 27: `dfu-util` needs the .bin here as it throws the contents of w/e file given verbatim at the device.

dragonmux (changes requested):
We've completed a first pass review and this is looking great. There are a few things to clean up in platform.c in particular, but with them taken care of this looks about ready to merge.

Really nicely done!

.gitignore
```diff
@@ -26,3 +26,4 @@ src/ftd2xx.dll
 scripts/swolisten
 
 .direnv/
+STM32F401.svd
```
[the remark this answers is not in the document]
sidprice: This file is used for my debugging, are you suggesting I drop the file or the ignore?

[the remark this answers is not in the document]
dragonmux: The ignore - as much as the file shouldn't get committed, it's going to be decidedly specific to your setup, unlike the other files that are listed in this file which are things that get generated with common IDEs that shouldn't be committed, and things which doing a build generates.

src/platforms/ctxlink/platform.c
```diff
@@ -0,0 +1,305 @@
...
+	RCC_CR |= (uint32_t)0x00000082;
+	/* Reset CFGR register */
+	RCC_CFGR = 0x00000000;
+
+	/* Disable all interrupts */
+	RCC_CIR = 0x00000000;
+
+	FLASH_ACR = 0;
+	__asm__ volatile("isb");
+	__asm__ volatile("dsb");
```
ALTracer, line 252: Note that there are primitives for this in a different platform, stlinkv3, used in Cortex-M7 cache setup, also written differently:
https://github.com/blackmagic-debug/blackmagic/blob/bbc1ba51b7a152835e6fe7444416875d8249a320/src/platforms/stlinkv3/platform.c#L61-L73
I'd add DMB if MPU setup was required, per `<mpu_v7.h>`, but your usage seems to do the Jump to MaskROM without rebooting, which is also valid. If this works better for your platform than the double-reboot I implemented/debugged and tested for blackpill-f4, then nevermind. Does it exit from MaskROM bootloader into BMF firmware/application without problems?

sidprice: I just ran a firmware update to check the exit from bootloader to BMF and it requires a reset to run the BMF firmware.

Dealing with right now would be a distraction so I am going to note your comment and address it later, assuming, with your approach the reset is not required.
Thanks.

ALTracer: Unfortunately this reset is also required in my experience. I'm likely missing some extra deinit/reset pokes on the USBOTG. Just so you know I might look into this or test suggestions. Yeah, get the USB-only platform in-tree, I won't distract further.

ALTracer: Ping, see #1861 and report if that works for your, F401RE, I assume. It was unrelated to OTGFS.

sidprice: @ALTracer , I am confused by the comment in your patch. 

I am using the ST System bootloader, are you suggesting adding your path before jumping to the bootloader? If so, your comment says "Unmap ST MaskROM and map back to Internal Flash"

If I am jumping to the bootloader, why would I unmap it?
:)

dragonmux: As best as we understand from the explanation and where the remap calls are made - it's when jumping back (ST bootloader -> BMD) so that when the firmware tries to engage interrupt handling, VTOR points at the right vector table.

ALTracer: No-no, the recent development is that I **have to** unmap System bootloader when it exits back into Blackmagic Debug Firmware -- I have tested this to work well with either raw BMF or BMF+BMPBootloader (I like the latter because it flashes faster). Note that the patch adds code *after* any old jumping code. You still can enter the ST MaskROM bootloader via enabling Remap and branching (or resetting core), it's the way back that was broken -- the infamous hang and having to press reset button (if it even exists).
You can review that there if you have more questions.

sidprice: So, this needs to be done early in the ctxlink application code, is that correct?

ALTracer: Correct. In the PR description I wrote why, but tl;dr is you have to do that after SYSCFG is clocked and before any interrupts are unblocked, because the first interrupt will vector you to 0 (or whatever SCB_VTOR) and there might be MaskROM sitting due to remap. `platform_init` is a good place for it, because BMF enables SysTick and USB and USART and whatnot.

sidprice: `	rcc_clock_setup_pll(&rcc_hse_8mhz_3v3[RCC_CLOCK_3V3_84MHZ]);
	/*
	 * Ensure the ST BOOTloader is unmapped
	 *
	 * Ref: https://github.com/blackmagic-debug/blackmagic/pull/1861
	 *
	 */
	if ((SYSCFG_MEMRM & 3U) == 1U)
		SYSCFG_MEMRM &= ~3U;

`
No, still have to press reset after bootloader.

ALTracer: +`rcc_periph_clock_enable(RCC_SYSCFG);` or something. Per RM it might be unclocked, see default values of RCC_APB1EN register etc. You would be reading zeroes but it's actually 0b01.

sidprice: I added the rcc_periph_clock_enable and still not working.

Also, the value in SYSCFG_MEMRM should be 0b00, according to the TRM:

syscfg_memrm

src/platforms/ctxlink/platform.c
```diff
@@ -0,0 +1,307 @@
...
+
+#define CTXLINK_BATTERY_INPUT        0 // ADC Channel for battery input
+#define CTXLINK_TARGET_VOLTAGE_INPUT 8 // ADC Chanmel for target voltage
+
+#define CTXLINK_ADC_BATTERY 0
+#define CTXLINK_ADC_TARGET  1
+
+static uint32_t inputVoltages[2] = {0};
+
+static uint8_t adcChannels[] = {CTXLINK_BATTERY_INPUT, CTXLINK_TARGET_VOLTAGE_INPUT}; /// ADC channels used by ctxLink
```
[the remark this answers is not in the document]
dragonmux: These two still need moving above the first function defined in the file on line 51, but otherwise 👍🏼 looks much better.

src/platforms/ctxlink/platform.c
```diff
@@ -0,0 +1,307 @@
...
+
+void platform_adc_read(void)
+{
+	adc_set_regular_sequence(ADC1, 1, &(adcChannels[CTXLINK_ADC_BATTERY]));
+	adc_start_conversion_regular(ADC1);
+	/* Wait for end of conversion. */
+	while (!adc_eoc(ADC1))
+		;
```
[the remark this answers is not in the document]
dragonmux: You got the second instance and missed the one on what is now line 77 😅  We'll leave this unresolved till that's taken care of.

src/platforms/ctxlink/platform.c
```diff
@@ -0,0 +1,307 @@
...
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Warray-bounds"
+
+void platform_request_boot(void)
+{
+	typedef void (*pFunction)(void);
+	const uint32_t ApplicationAddress = 0x1fff0000;
+	register uint32_t JumpAddress = 0;
+	register uint32_t addr = 0;
+	static pFunction Jump_To_Application;
```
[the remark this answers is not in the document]
dragonmux: This is looking much better, only item missing is dropping the extra `static` so you can drop gloals usage by one unnecessary one. Might suggest making it `register` instead like you've done for the new stack pointer.

dragonmux (commented):
This is looking much better. There are still a couple of small tweaks needed, noted by the non-resolved thread items from the last review, and then we'll be comfortable to merge this. Nice work!

dragonmux (approved):
Code-wise this all LGTM now. Only thing left to sort are some commit flow issues and this can be merged. We'll work with you to fix those though.

dragonmux (approved):
LGTM, all review items have now been addressed. Merging. Thank you for the contribution!

PR #1856, opened 2024-07-07
Fix gdb idling too much on DWC

ALTracer:
* Not a new feature.
* The existing problem is 2x reduced performance in `bmpflash read -b int file.bin`, `blackmagic -r file.bin` BMDA memory dumping, and less obvious in other actions, on `blackpill-f4` (and possibly `stlinkv3` F723, `f4discovery` F407).
* This PR solves it by correcting the check in USB interrupt callback code which controls WFI entry.

Follow-up to #1689. Because I added the check on `if(count_new == 0) __WFI();` but by the time of check `count_new` was either zeroed by previous lines of code, or already zero at callback entry, this added an extra sleep, reducing how often USB packets could be fetched from Host OUT endpoint. What I should've done instead is save the `count_new` value at entry and use that to dispatch the WFI. Now this works as I expect it to. Despite that, `0.26% gdb_set_noackmode`. Do with that what you will.

Tested on `blackpill-f411ce` to restore onboard w25q64 read data rate from 123-129 back to 230-257 KiB/s. Monitored with orbtop to sleep 10% of PC samples during the readout. It still sleeps 99% on `gdb_if` port closed, and on `ttyBmpGdb` open but no activity (pre-PR1689 it would not sleep with this port open at all). `ttyBmpTarg` state has no effect on the sleep, Aux UART is IRQ+DMA driven, as is Trace capture (Host IN-only), if only on max packet rate (because of extra Naks).
I think stm32f1 platforms are not affected by both of these changes (thanks to preprocessing macros).

## Closing issues

Fixes #1689.

### DIFF
--- a/src/platforms/common/stm32/gdb_if.c
+++ b/src/platforms/common/stm32/gdb_if.c
@@ -90,6 +90,8 @@ static void gdb_if_update_buf(void)
 #else
 	cm_disable_interrupts();
 	__asm__ volatile("isb");
+	/* count_new will become 0 by the time of decision to WFI, so save a copy at entry */
+	const uint32_t count_new_saved = count_new;
 	if (count_new) {
 		memcpy(buffer_out, double_buffer_out, count_new);
 		count_out = count_new;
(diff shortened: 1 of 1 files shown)


dragonmux (approved):
LGTM! Seen your data testing this on Discord, so as this improves throughput without causing any other problems, we'll get this merged. Nicely done!

PR #1861, opened 2024-07-11
Fix `blackpill-f4` requiring a reset after exiting ST MaskROM

ALTracer:
## Detailed description

* This is an old feature ported between platforms.
* The existing problem is `blackpill-f411ce` and likely other variants hanging in either BMF or BMD bootloader upon exiting ST MaskROM USB DFU -- mitigated by pressing reset button (after the initial flashing of BMD bootloader, or after `dfu-util -s :leave` for project bootloader-less builds.
* This PR finally solves it by copying the register pokes for restoring STM32F4 memory remap.

Armed with inception debugging, knowledge of adjacent platforms, and dfu-util on Linux amd64 I dived into this once more and noticed something weird. After ST MaskROM USB DFU (located at 0x1fff_0000) performs a leave into application (located at 0x0800_0000), it keeps itself remapped to 0. If BMD bootloader is present, it then updates SCB->VTOR to 0x0000_4000 and jumps to BMF at +16KiB. Finally, this enables interrupts and ends up in MaskROM again because that's what the vector table essentially is at this point.

There are at least two solutions to this. I opted for restoring SYSCFG->MEMRMP bits. Other platforms which have no ST remap functionality, at least always manually update their SCB->VTOR value to point to their vector tables in flash. I can do this on Cortex-M4. *Note that it's impossible on F072-IF which has a Cortex-M0, so remap has to be used.*

Another note is that I tried linking the entire BMF not into 0x0800_0000, but into 0x0, and it runs okay assuming remap is valid. The refman states this may increase performance by replacing S-bus fetch with I-Code bus and D-Code bus fetch in <0x2000_0000. I assume it's only useful for Boot from SRAM to reduce S-bus contention and/or Load-Store Unit utilization -- both MaskROM and Internal Flash fall into I-Code addressable executable region. SRAM fetch does not affect ART or Flash prefetch.

* [x] ~~It builds for hardware native (see [Building the firmware](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-firmware))~~
* [x] ~~It builds as BMDA (see [Building the BMDA](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-app))~~

## Closing issues

Certainly could fix something. I consider this port much more ready for release now.

### DIFF
--- a/src/platforms/common/blackpill-f4/blackpill-f4.c
+++ b/src/platforms/common/blackpill-f4/blackpill-f4.c
@@ -72,6 +72,12 @@ void platform_init(void)
 		scb_reset_core();
 	}
 #endif
+
+	/* Unmap ST MaskROM and map back Internal Flash */
+	rcc_periph_clock_enable(RCC_SYSCFG);
+	if ((SYSCFG_MEMRM & 3U) == 1U)
+		SYSCFG_MEMRM &= ~3U;
+
 	rcc_clock_setup_pll(&rcc_hse_25mhz_3v3[PLATFORM_CLOCK_FREQ]);
 
 	/* Set up DM/DP pins. PA9/PA10 are not routed to USB-C. */
--- a/src/platforms/common/blackpill-f4/usbdfu.c
+++ b/src/platforms/common/blackpill-f4/usbdfu.c
@@ -23,6 +23,7 @@
 #include <libopencm3/cm3/scb.h>
 #include <libopencm3/stm32/rcc.h>
 #include <libopencm3/stm32/gpio.h>
+#include <libopencm3/stm32/syscfg.h>
 #include <libopencm3/usb/dwc/otg_fs.h>
 
 #include "usbdfu.h"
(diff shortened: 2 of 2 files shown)


src/platforms/common/blackpill-f4/blackpill-f4.c
```diff
@@ -72,6 +72,13 @@ void platform_init(void)
 		scb_reset_core();
 	}
 #endif
+
+	/* Unmap ST MaskROM and map back Internal Flash */
+	rcc_periph_clock_enable(RCC_SYSCFG);
+	if ((SYSCFG_MEMRM & 3U) == 1U) {
```
dragonmux, line 78: The braces are superfluous in this if statement, and the one in the bootloader, please drop them.

dragonmux (changes requested):
There's one small stylistic fix to be done and then this can be merged.

Very interesting how this explains so much of the weirdness seen on these platforms when bootloading them over the last few years. Glad that it's been figured out and solved at last!

dragonmux (approved):

ALTracer:
Rebased to main.
Revisit after ctxLink gets in-tree, reconsider using SCB_VTOR.
Deal with all the platforms doing their own things despite being STM32.

PR #1864, opened 2024-07-13
Feature: Calculate and report real USART baudrate

ALTracer:
## Detailed description

* This is a small new feature.
* The existing problem is BMP platforms parroting back whatever baudrate you set it to. The discrepancy gets significant at high baudrates, close to Pclk DIV16, effectively ruining communications to other boards +-5% (especially TraceSWO Async which is tuned by an integer(!) divisor off TraceClk).
* This PR makes TI ICDI and STM32 platforms recompute real baudrate by reading back the programmed USART BRR divisors. It's a function missing from libopencm3 API, but to upstream this there OVER8 support would be required.

As I was compile-testing this, I also noticed some stray hardcoded 0x85 endpoints in ICDI and replaced them with new compile-time macros.
Runtime tested on `blackpill-f411ce`, `swlink`-bluepillplus, stlinkv3-minie (f723). I needed this for better TraceSWO Async UART diagnostics, to ensure TraceClk/SWO_PR is close enough to Pclk/BRR. I don't know how to visibly test how this positively affects aux_serial behaviour, I didn't see a difference in minicom, ckermit, tio. Only tio can reach non-standard baudrates because it uses a different API to set them (kinda like orbtrace serialFeeder), minicom can't (it uses B2000000 macros etc.)
Minor problem with this is I don't like that neither BMF nor libopencm3 has guardrails limiting BRR to 16 or more, so I could "set" and "read back" a baudrate equal to Pclk, which does not work.

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -605,12 +605,10 @@ static bool cmd_rtt(target_s *t, int argc, const char **argv)
 #ifdef PLATFORM_HAS_TRACESWO
 static bool cmd_traceswo_enable(int argc, const char **argv)
 {
-#if TRACESWO_PROTOCOL == 2
-	uint32_t baudrate = SWO_DEFAULT_BAUD;
-#endif
 	uint32_t swo_channelmask = 0; /* swo decoding off */
 	uint8_t decode_arg = 1;
 #if TRACESWO_PROTOCOL == 2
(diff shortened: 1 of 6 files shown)


dragonmux (approved):
This all LGTM, so we're happy to merge. Thank you for the contribution!

dragonmux (approved):

PR #1865, opened 2024-07-13
Fix CDC-ACM on STLINK/V3

ALTracer:
## Detailed description

* No new features.
* The existing problem is Virtual COM port terribly broken on `stlinkv3` platform Blackmagic Debug firmware, whereas it works fine on original proprietary firmware.
* This PR solves it by bumping aux_serial.c buffer sizes to 2x512 for USB HS STM32F723, index type widths, and correcting USB Out endpoint size.

Tested on ST-LINK/V3E by performing a cbrake/linux-serial-test in external loopback mode at 12 Mbaud (Hclk DIV18 gives a round number, and I need OVER8 to go 432M/24=18M) with 1'200'000 chars/s reported in both directions and Wireshark usbmon displaying Bulk packets up to 511 bytes long.
May need independent testing both on FSDev STM32F1 and OTGFS STM32F4 (so an swlink and a blackpill-f4), but the code is macro-gated. I expect slightly better codegen and faster 32-bit ALU-native ops where 8-bit indices were previously used.

## Miscellaneous
* The CDC-ACM Out EP kludge looks ugly to me, but that has to deal with PMA utilization limitation of FSDev (512 bytes?), and I can't walk the private config with descriptors from usb_serial.c -- otherwise that patch becomes
```c
#include "usb_private.h"
const uint16_t uart_epout_size = usbdev->config.interface[UART_IF_NO+1].altsetting.endpoint.wMaxPacketSize;
```
* Initially I suspected some Cortex-M7 D-cache coherency problem, and went to set up the unused MPU with Shareable Non-cacheable region on SRAM2 for 16 KiB and moved said buffers via attribute section there (with an altered linkerscript), but actually DMA2 can reach default DTCM via Cortex-M7 AHBS port, so this ended up being unnecessary. You can still do this to reduce AHB contention between CM7 S-bus, DMA2 and AHB SRAM during high-baudrate VCP and TraceSWO UART stream activity, but assuming they are no faster than 18 Mbaud each, and AHB runs at 216 MHz, I think it's not a significant enough problem.
* CM7 has no cacheability and no waitstates over ITCM & DTCM, so any cache coherency is dealt with automatically. ART requires we link to 0x00220000 not 0x08020000 for ITCM Flash fetch, otherwise only I-cache works for AXI Flash fetch. I tried it and measured no perf benefit.

* [x] ~~It builds for hardware native (see [Building the firmware](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-firmware))~~
* [x] ~~It builds as BMDA (see [Building the BMDA](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-app))~~

### DIFF
--- a/src/platforms/common/aux_serial.c
+++ b/src/platforms/common/aux_serial.c
@@ -40,14 +40,14 @@ static uint32_t aux_serial_active_baud_rate;
 
 static char aux_serial_receive_buffer[AUX_UART_BUFFER_SIZE];
 /* Fifo in pointer, writes assumed to be atomic, should be only incremented within RX ISR */
-static uint8_t aux_serial_receive_write_index = 0;
+static uint16_t aux_serial_receive_write_index = 0;
 /* Fifo out pointer, writes assumed to be atomic, should be only incremented outside RX ISR */
-static uint8_t aux_serial_receive_read_index = 0;
+static uint16_t aux_serial_receive_read_index = 0;
 
 #if defined(STM32F0) || defined(STM32F1) || defined(STM32F3) || defined(STM32F4) || defined(STM32F7)
 static char aux_serial_transmit_buffer[2U][AUX_UART_BUFFER_SIZE];
-static uint8_t aux_serial_transmit_buffer_index = 0;
-static uint8_t aux_serial_transmit_buffer_consumed = 0;
+static uint16_t aux_serial_transmit_buffer_index = 0;
+static uint16_t aux_serial_transmit_buffer_consumed = 0;
 static bool aux_serial_transmit_complete = true;
 
 static volatile uint8_t aux_serial_led_state = 0;
--- a/src/platforms/common/usb_serial.c
+++ b/src/platforms/common/usb_serial.c
@@ -195,8 +195,12 @@ void usb_serial_set_config(usbd_device *dev, uint16_t value)
 #endif
 
 	/* Serial interface */
-	usbd_ep_setup(
-		dev, CDCACM_UART_ENDPOINT, USB_ENDPOINT_ATTR_BULK, CDCACM_PACKET_SIZE / 2U, debug_serial_receive_callback);
+#if defined(USB_HS)
+	const uint16_t uart_epout_size = CDCACM_PACKET_SIZE;
+#else
+	const uint16_t uart_epout_size = CDCACM_PACKET_SIZE / 2U;
+#endif
+	usbd_ep_setup(dev, CDCACM_UART_ENDPOINT, USB_ENDPOINT_ATTR_BULK, uart_epout_size, debug_serial_receive_callback);
 	usbd_ep_setup(dev, CDCACM_UART_ENDPOINT | USB_REQ_TYPE_IN, USB_ENDPOINT_ATTR_BULK, CDCACM_PACKET_SIZE,
 		debug_serial_send_callback);
 #if defined(STM32F4) && CDCACM_UART_NOTIF_ENDPOINT >= 4
--- a/src/platforms/common/aux_serial.h
+++ b/src/platforms/common/aux_serial.h
@@ -27,11 +27,14 @@
 
 #if defined(STM32F0) || defined(STM32F1) || defined(STM32F3) || defined(STM32F4) || defined(STM32F7)
 /* XXX: Does the st_usbfs_v2_usb_driver work on F3 with 128 byte buffers? */
-#if defined(STM32F1) || defined(STM32F3) || defined(STM32F4) || defined(STM32F7)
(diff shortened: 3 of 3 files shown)


src/platforms/common/usb_serial.c
```diff
@@ -195,8 +195,12 @@ void usb_serial_set_config(usbd_device *dev, uint16_t value)
 #endif
 
 	/* Serial interface */
-	usbd_ep_setup(
-		dev, CDCACM_UART_ENDPOINT, USB_ENDPOINT_ATTR_BULK, CDCACM_PACKET_SIZE / 2U, debug_serial_receive_callback);
+#if defined(USB_HS)
+	const uint16_t uart_epout_size = CDCACM_PACKET_SIZE;
```
dragonmux, line 199: We strongly suspect this winds up wasting half the buffer on the F7, but as long as you think that's ok, we're happy with this.

src/platforms/common/aux_serial.c
```diff
@@ -40,14 +40,14 @@ static uint32_t aux_serial_active_baud_rate;
 
 static char aux_serial_receive_buffer[AUX_UART_BUFFER_SIZE];
 /* Fifo in pointer, writes assumed to be atomic, should be only incremented within RX ISR */
-static uint8_t aux_serial_receive_write_index = 0;
+static uint32_t aux_serial_receive_write_index = 0;
```
dragonmux, line 43: Given how this expands static memory usage for SRAM, what's the justification here for going to uint32_t instead of uint16_t please?

dragonmux (commented):
This is looking good - we've just a couple of queries which once answered will clear the way for this to be merged. Note, these are queries and not change requests per se.

src/platforms/common/usb_serial.c
ALTracer: I assume this has to match the descriptors
https://github.com/blackmagic-debug/blackmagic/blob/c4c763e75ee09511621eb432618116e8ec2013c5/src/platforms/common/usb_descriptors.h#L200-L204
It does not waste half the Tx FIFO, that is, I think I saw usbmon packets for 511 bytes in both directions; otherwise I'd see 255 bytes in one of directions.
You're right in that I should be more careful with FIFO memory allocation to non-periodic buffers, this isn't Linux afterall.

src/platforms/common/aux_serial.c
ALTracer: My technical argument for this is when you're targeting a natively 32-bit ALU machine, such as Cortex-M3 ARMv7-M (with 16-bit instructions Thumb1), a 32-bit unsigned integer type has the fastest executing codegen. 16-bit or 8-bit involves a halfword or byte load/store and masking: uxth, uxtb. Needs Compiler Explorer for hard evidence. Also consider using atomics, ldrex/strex will happily operate on 32-bit variables. Oh, and the AHB is 32-bit, where the SRAM is, accesses steal a couple cycles for AHB handshaking anyways. 
+12 bytes is not much to pay for correctness. If you want me to make it 6 bytes plus masking, then I'll need to retest with uint16_t.

dragonmux: > 16-bit or 8-bit involves a halfword or byte load/store and masking: uxth, uxtb.

We'd expect `ldrh` and `ldrb` for those loads, not the masking instructions. This should make things atomic even with uint16_t's.

src/platforms/common/usb_serial.c
dragonmux: Correct, it has to match the descriptors. As long as this patch does indeed make them match (which we just checked, it does), that's fine with us. Have to admit we didn't check, which is part of the query. We'll consider this resolved.

src/platforms/common/aux_serial.c
ALTracer: You're right, loads and stores are equivalent. It's the increment that bothers me (and %=SIZE). I'll temporarily accept using uint16_t.
Thinking some more, an F723 at 216M Hclk with UART set up for DIV18, that is 12 Mbaud, will have to perform the buffer switch in 18 * 10 baud = 180 clocks of its CM7; for optimistic DIV12 (in OVER8, 18Mbaud), in 12 * 10 = 120 clocks, or about half a microsecond, to avoid DMA spilling into the next half of buffer. If code requires more overhead, then this should be stated in platform readme and I won't care anymore.
Similarly for F411 at 96M Hclk and UART at DIV16, so 6Mbaud, there's 16 * 10baud = 160 clocks to do this for FS.

dragonmux (approved):
With the two queries taken care of, this LGTM. Happy to merge once this is rebased on `main`. Thank you for the contribution!

PR #1867, opened 2024-07-13
Fix mis-identification of STM32MP15x_CM4 as STM32H74x_CM4

ALTracer:
## Detailed description

* There are minor features here, I think.
* The existing problem is `STM32MP15 M4` wrongly identified by BMD as `STM32H7 M4` since 0f226232 of #1819.
* This PR solves it by adding gated code to read the DBGMCU_IDC value as mandated by ST for silicon identification.

First thing I tried is simply reordering the probe calls, and sure enough, that made the two chips light up as `STM32MP15 M4` again but also `STM32MP15 M7` which is unheard of. Then I added the H74x, H7Bx short literals to H7 driver to see slightly more detail in reported target list -- those don't have the H723-like string in their Engineering bytes.
Then the logging of three variants of IDs. By staring at them long enough I figured we only need to read DBGMCU in the dubious case of seeing 0x450.
Tested on STM32MP157C and Nucleo-H743ZI2 to detect each properly and not upset the Cortex-M AHB-AP. Tested on MiniH7B0 to not even try to read IDC because it's 0x480, not 0x450. Similarly, I believe H72x detection should not regress.

## Closing issues

### DIFF
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -565,6 +565,8 @@ bool cortexm_probe(adiv5_access_port_s *ap)
 	if (conn_reset)
 		target_mem32_write32(target, CORTEXM_DEMCR, 0);
 
+	DEBUG_TARGET("%s: Examining Part ID 0x%04x, AP Part ID: 0x%04x\n", __func__, target->part_id, ap->partno);
+
 	switch (target->designer_code) {
 	case JEP106_MANUFACTURER_FREESCALE:
 		PROBE(imxrt_probe);
--- a/src/target/stm32h7.c
+++ b/src/target/stm32h7.c
@@ -260,6 +260,15 @@ bool stm32h7_probe(target_s *target)
 	/* Use the partno from the AP always to handle the difference between JTAG and SWD */
 	if (ap->partno != ID_STM32H72x && ap->partno != ID_STM32H74x && ap->partno != ID_STM32H7Bx)
 		return false;
+
+	/* By now it's established that this is likely an H7, but check that it's not an MP15x_CM4 with an errata in AP part code */
+	const uint32_t idcode = target_mem32_read32(target, DBGMCU_IDCODE);
+	const uint16_t dev_id = idcode & STM32H7_DBGMCU_IDCODE_DEV_MASK;
+	DEBUG_TARGET("%s: looking at device ID 0x%03x at 0x%08" PRIx32 "\n", __func__, dev_id, DBGMCU_IDCODE);
+	/* MP15x_CM4 errata: has a partno of 0x450. SoC DBGMCU says 0x500. */
+	if (dev_id != ID_STM32H72x && dev_id != ID_STM32H74x && dev_id != ID_STM32H7Bx)
+		return false;
+
 	target->part_id = ap->partno;
 
 	/* Save private storage */
--- a/src/target/stm32h7.c
+++ b/src/target/stm32h7.c
@@ -277,6 +277,12 @@ bool stm32h7_probe(target_s *target)
 		write_be4((uint8_t *)priv_storage->name, 5U, target_mem32_read32(target, STM32H7_CHIP_IDENT));
 		priv_storage->name[9] = '\0';
 		break;
+	case ID_STM32H74x:
+		memcpy(priv_storage->name + 5U, "H74x", 5U); /* H742/H743/H753/H750 */
+		break;
+	case ID_STM32H7Bx:
+		memcpy(priv_storage->name + 5U, "H7Bx", 5U); /* H7A3/H7B3/H7B0 */
+		break;
 	default:
 		memcpy(priv_storage->name + 5U, "H7", 3U);
 		break;
--- a/src/target/stm32mp15.c
+++ b/src/target/stm32mp15.c
@@ -91,6 +91,15 @@ static bool stm32mp15_ident(target_s *const target, const bool cortexm)
 		if (!cortexm || ap->partno != ID_STM32MP15x_ERRATA)
 			return false;
 	}
+
+	/* By now it's established that this is likely an MP15x_CM4, but check that it's not an H74x */
+	const uint32_t idcode = target_mem32_read32(target, DBGMCU_IDCODE);
+	const uint16_t dev_id = idcode & STM32MP15_DBGMCU_IDCODE_DEV_MASK;
+	DEBUG_TARGET("%s: looking at device ID 0x%03x at 0x%08" PRIx32 "\n", __func__, dev_id, DBGMCU_IDCODE);
+	/* If this probe routine ever runs ahead of stm32h7_probe, skip the H74x. */
+	if (dev_id != ID_STM32MP15x)
+		return false;
+
 	/*
 	 * We now know the part is either a Cortex-M core with the errata code, or matched the main ID code.
 	 * Copy the correct (AP) part number over to the target structure to handle the difference between


src/target/stm32h7.c
```diff
@@ -257,9 +257,20 @@ void stm32h7_configure_wdts(target_s *const target)
 bool stm32h7_probe(target_s *target)
 {
 	const adiv5_access_port_s *const ap = cortex_ap(target);
+	DEBUG_TARGET("%s: AP partno %X, part_id %X\n", __func__, ap->partno, target->part_id);
```
dragonmux, line 260: This reads like some residual debugging stuff?

If something like this is actually warranted, we'd suggest putting it into cortexm.c in the handling case for STM32 parts, and making the print `DEBUG_TARGET("%s: AP part id: %04x, part id: %04x\n", __func__, ap->partno, target->part_id);` or so.

Essentially, please use lower-case hex for code base consistency, and force the formatted part number to be 4 digits long to get a consistent output. Having the display for the STM32 parts is probably good insurance given all the problems with their part numbering inconsistencies, but it should then be in the Cortex-M handling layer, rather than duplicated across all the target support implementations.

src/target/stm32h7.c
```diff
@@ -257,9 +257,20 @@ void stm32h7_configure_wdts(target_s *const target)
...
+
 	/* Use the partno from the AP always to handle the difference between JTAG and SWD */
 	if (ap->partno != ID_STM32H72x && ap->partno != ID_STM32H74x && ap->partno != ID_STM32H7Bx)
 		return false;
+
+	/* By now it's likely this is ST silicon, either H7 or MP15x_CM4 */
```
dragonmux, line 266: ST silicon is established by the manufacturer code in cortexm.c, so we'd suggest this read more like `/* By now it's established this is likely a H7, but check that it's not a MP15 with a bugged AP part code */` as that's what this is aiming to check.

src/target/stm32h7.c
```diff
@@ -257,9 +257,20 @@ void stm32h7_configure_wdts(target_s *const target)
...
+	const uint32_t idcode = target_mem32_read32(target, DBGMCU_IDCODE);
+	const uint16_t dev_id = idcode & STM32H7_DBGMCU_IDCODE_DEV_MASK;
+	DEBUG_TARGET("%s: looking at device ID 0x%3X at 0x%08X\n", __func__, dev_id, DBGMCU_IDCODE);
```
dragonmux, line 269: This reads like it's debugging?

src/target/stm32mp15.c
```diff
@@ -85,12 +85,23 @@ static void stm32mp15_detach(target_s *target);
 static bool stm32mp15_ident(target_s *const target, const bool cortexm)
 {
 	const adiv5_access_port_s *const ap = cortex_ap(target);
+	DEBUG_TARGET("%s: AP partno %X, part_id %X\n", __func__, ap->partno, target->part_id);
```
dragonmux, line 88: This and the block below get similar comments to their counterparts in the STM32H7 support, rewording accordingly for MP15x instead of H7.

dragonmux (changes requested):
The general gist of this patch looks good. There are a couple of times that could do with clarification or rewording for comments and prints, and with those taken care of, we're happy to merge this.

src/target/stm32h7.c
ALTracer: Yes, it's not essential to driver operation, but... it was essential to telling the chips apart and coding the mis-attach guardrails appropriately. I can post the logs this produced. It should not take up flash space in release firmware, but get optimized out.

src/target/cortexm.c
```diff
@@ -464,6 +464,8 @@ bool cortexm_probe(adiv5_access_port_s *ap)
 		target->designer_code = ap->designer_code;
 		target->part_id = ap->partno;
 	}
+	DEBUG_TARGET("%s: Designer 0x%3x, part_id 0x%4x, AP partno 0x%4x\n", __func__, target->designer_code,
+		target->part_id, ap->partno);
```
dragonmux, line 468: The format strings should include a 0 term to get them properly suffixed - eg, `%03x`, `%04x`. The designer code is only useful in the event of part identification failure in the error at the bottom of the function, so we'd drop that as duplicate. `part_id` and `partno` are internal names, which we'd like to avoid using in this message - hence the suggestion of `AP part id: %04x, part id: %04x` in the original review note, as this insulates us from internal terminology changes better.

It's an intriguing idea putting this here 🤔 - when we made the original review suggestion, we were thinking it'd go on line 580 (just inside the `case JEP106_MANUFACTURER_STM:` block) given it's only been ST parts that have done bad with their part codes, but it's probably a fair point that more parts deserve this as a just in case. Given that, please move it down to just above `switch (target->designer_code) {`.

src/target/stm32h7.c
dragonmux: We understand that it won't impact binary size, we're mostly thinking about how this'd look in the logs and how useful it could be (or not be) - we appreciate the usefulness to your figuring out this patch, but we're trying to weigh up the future usefulness of it. After all, it's not something a typical user of even BMDA is going to want to see in the logs, but rather only someone trying to dig into part identification things, noting that the target read gets logged at the `DEBUG_PROTO` level, allowing the value to be read in that case albeit at a slightly higher amount of effort.

We think that if this is going to be kept because you feel it does have utility, then please do two small things for code base consistency: use lower-case `x` for the device ID along with making sure it's suffixed with 0 as in `%03x`, and use PRIx32 for the ID code so its printing is better guaranteed to work as in `%08" PRIx32 "`. The rest of the code base uses lower-case hex for formatted values and literals for debug, so please let's not introduce a competing different way of doing that with this.

ALTracer: In my defense, `stm32l4_probe()` does a `DEBUG_INFO("ID Code: %08" PRIx32 "\n", device_id)` for DPv2/AP partno or `DBGMCU_IDCODE` of STM32L55. Other drivers don't. I'd like to keep it in some form, possibly rewording the logged string.

src/target/cortexm.c
dragonmux: That's a fair point, though that is because of using ARM's nomenclature for the TARGETID register, rather than internal nomenclature (if memory serves us correctly) - so in that case, `partno` is acceptable.. `part_id` though is internal nomenclature so should be avoided.

src/target/stm32h7.c
dragonmux: This is actually a print we intend on eliminating as it's both in the wrong debug level, and quite distracting in the logs due to it printing for a fair few ST devices that aren't actually handled by the L4 support, resulting in confusing and unhelpful logs with spurious extra lines. That's part of why we're concerned about continuing utility, because that log line from 8845a222 / a6a8606e does not have utility and introduces a spurious source of confusion in the logs

src/target/stm32h7.c
```diff
@@ -257,9 +257,20 @@ void stm32h7_configure_wdts(target_s *const target)
 bool stm32h7_probe(target_s *target)
 {
 	const adiv5_access_port_s *const ap = cortex_ap(target);
+	DEBUG_TARGET("%s: AP partno 0x%04x, part_id 0x%04x\n", __func__, ap->partno, target->part_id);
```
dragonmux, line 260: This and the matching DEBUG_TARGET in stm32mp15.c should be removed as it duplicates the one in cortexm.c

src/target/cortexm.c
```diff
@@ -565,6 +565,9 @@ bool cortexm_probe(adiv5_access_port_s *ap)
 	if (conn_reset)
 		target_mem32_write32(target, CORTEXM_DEMCR, 0);
 
+	DEBUG_TARGET("%s: Designer 0x%03x, selecting Part ID 0x%04x, AP Part ID: 0x%04x\n", __func__, target->designer_code,
```
dragonmux, line 568: We still maintain that the designer code is not useful here, bloating the string needlessly - in the case of an unrecognised part, the designer gets printed in the below DEBUG_ERROR; and otherwise, the list of probe routines tried is printed when INFO is enabled and which gives which designer code path is taken. The designer codes are (unlike the part number) displayed above in the output as well, which helps.

"selecting" is going to be real odd in the output to a user, perhaps "Examining"?

ALTracer: Applied.
The two significant numbers may differ for DPv2 but may match for conforming DPv1 implementations. It made more sense in STM32-specific code for the confused DPv2 targets. Keeping until logs seen.

src/target/stm32h7.c
ALTracer: Dropped intermediate commit. I could've reverted it and squashed into cortexm, but then it becomes less pretty.

dragonmux (approved):
All review items have been addressed, LGTM - merging. Thank you for the contribution!

PR #1869, opened 2024-07-13
Feature: watchdog & sleep management on ArteryTek AT32F43x

ALTracer:
## Detailed description

* This is a new feature.
* The existing problem is AT32F437 target watchdogs, WDT & WWDT, potentially resetting the chip during halting debug with BMD; and WFI sleep states interfering with halt-polling (and possibly RTT).
* This PR solves that by writing bits in DEBUG peripheral (analogous to ST DBGMCU) which control Sleep Emulation and Watchdog timer auto-freeze, with no save-restore but via read-modify-write.

WDT/WFI handling is required for feature parity with stm32f1.c and other target drivers.

## Closing issues

### DIFF
--- a/src/target/at32f43x.c
+++ b/src/target/at32f43x.c
@@ -89,14 +89,26 @@ static bool at32f43_mass_erase(target_s *target);
 #define AT32F43x_2K_OB_COUNT 256U
 #define AT32F43x_4K_OB_COUNT 2048U
 
-#define AT32F43x_DEBUG            0xe0042000U
-#define DBGMCU_IDCODE             (AT32F43x_DEBUG + 0U)
-#define AT32F43x_DEBUG_CTRL       (AT32F43x_DEBUG + 4U)
-#define AT32F43x_DEBUG_APB1_PAUSE (AT32F43x_DEBUG + 8U)
+/*
+ * refman: DEBUG has 5 registers, of which CTRL, APB1_PAUSE, APB2_PAUSE are
+ * "asynchronously reset by POR Reset (not reset by system reset). It can be written by the debugger under reset."
+ * Note that it has no TRACE_IOEN and SWO is controlled by GPIO IOMUX (AF) instead.
+ */
+#define AT32F43x_DBGMCU_BASE       0xe0042000U
+#define AT32F43x_DBGMCU_IDCODE     (AT32F43x_DBGMCU_BASE + 0x00U)
+#define AT32F43x_DBGMCU_CTRL       (AT32F43x_DBGMCU_BASE + 0x40U)
+#define AT32F43x_DBGMCU_APB1_PAUSE (AT32F43x_DBGMCU_BASE + 0x80U)
+#define AT32F43x_DBGMCU_APB2_PAUSE (AT32F43x_DBGMCU_BASE + 0x0cU)
+#define AT32F43x_DBGMCU_SER_ID     (AT32F43x_DBGMCU_BASE + 0x20U)
 
-#define AT32F43x_DEBUG_CTRL_SLEEP      0x7U
-#define AT32F43x_DEBUG_APB1_PAUSE_WWDT (1U << 11U)
-#define AT32F43x_DEBUG_APB1_PAUSE_WDT  (1U << 12U)
+#define AT32F43x_DBGMCU_CTRL_SLEEP_DEBUG     (1U << 0U)
+#define AT32F43x_DBGMCU_CTRL_DEEPSLEEP_DEBUG (1U << 1U)
+#define AT32F43x_DBGMCU_CTRL_STANDBY_DEBUG   (1U << 2U)
+#define AT32F43x_DBGMCU_CTRL_SLEEP_MASK \
+	(AT32F43x_DBGMCU_CTRL_SLEEP_DEBUG | AT32F43x_DBGMCU_CTRL_DEEPSLEEP_DEBUG | AT32F43x_DBGMCU_CTRL_STANDBY_DEBUG)
+
+#define AT32F43x_DBGMCU_APB1_PAUSE_WWDT (1U << 11U)
+#define AT32F43x_DBGMCU_APB1_PAUSE_WDT  (1U << 12U)
 
 #define AT32F4x_IDCODE_SERIES_MASK 0xfffff000U
 #define AT32F4x_IDCODE_PART_MASK   0x00000fffU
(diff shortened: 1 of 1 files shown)


src/target/at32f43x.c
```diff
@@ -89,7 +89,14 @@ static bool at32f43_mass_erase(target_s *target);
 #define AT32F43x_2K_OB_COUNT 256U
 #define AT32F43x_4K_OB_COUNT 2048U
 
-#define DBGMCU_IDCODE 0xe0042000U
+#define AT32F43x_DEBUG            0xe0042000U
+#define DBGMCU_IDCODE             (AT32F43x_DEBUG + 0U)
+#define AT32F43x_DEBUG_CTRL       (AT32F43x_DEBUG + 4U)
+#define AT32F43x_DEBUG_APB1_PAUSE (AT32F43x_DEBUG + 8U)
+
+#define AT32F43x_DEBUG_CTRL_SLEEP      0x7U
```
dragonmux, line 97: For consistency with other implementations like the [STM32H5 implementation](https://github.com/blackmagic-debug/blackmagic/blob/a4934aed4fdf61f764c93b2065111e34c0cf603e/src/target/stm32h5.c#L101-L117) that's coming in an overhaul of the watchdogs situation, please express this as the 3 different sleep control bits (so they have names) and them `|` them together below.

src/target/at32f43x.c
```diff
@@ -89,7 +89,14 @@ static bool at32f43_mass_erase(target_s *target);
 #define AT32F43x_2K_OB_COUNT 256U
 #define AT32F43x_4K_OB_COUNT 2048U
 
-#define DBGMCU_IDCODE 0xe0042000U
+#define AT32F43x_DEBUG            0xe0042000U
+#define DBGMCU_IDCODE             (AT32F43x_DEBUG + 0U)
+#define AT32F43x_DEBUG_CTRL       (AT32F43x_DEBUG + 4U)
+#define AT32F43x_DEBUG_APB1_PAUSE (AT32F43x_DEBUG + 8U)
```
dragonmux, line 95: Similar to the first comment, please prefix all of these with `AT32F43x_` and call the base address `AT32F43x_DBGMCU_BASE`, using the peripheral name as a secondary suffix for all of these. This is part of a drive to try and unify nomenclature across the code base and be more consistent about this.

dragonmux (changes requested):
There are a couple of nomenclature items to address, and then we're happy to merge this PR. It feels like a lighter touch version of our WDTs fix branch which is pretty spot on. The main reason the fix branch does more is so that the WDTs get halt-inhibited during scan so they can't so easily cause problems there and while waiting for the user to attach.

src/target/at32f43x.c
ALTracer: Detailed as three control bits with names from refman.

src/target/at32f43x.c
ALTracer: Renamed everything DEBUG-related to DBGMCU. Added a short doc block.

ALTracer:
Tested working on AT-START-F437 via its AT-Link EZ (CMSIS-DAP v2 FS) and via bluepillplus swlink older BMF, using BMDA.
Programmed option byte nWDT_ATO_EN to autostart watchdog upon exiting reset, and normal `main` branch BMD fails to `load` a somewhat big (256 KiB) binary with WDT previously set to 4 seconds (40kHz LICK, div64, 0x9c3).
Used branch BMDA which sets APB1 freeze WDT bit in attach, confirmed via an extra logging statement that the register contents stick across the sysreset applied pre-reprogramming.

Then rebased to main and added a couple comments deailng with review feedback. That should result in equivalent codegen and behaviour.

src/target/at32f43x.c
```diff
@@ -89,7 +89,26 @@ static bool at32f43_mass_erase(target_s *target);
...
+#define AT32F43x_DBGMCU_IDCODE     (AT32F43x_DBGMCU_BASE + 0x00U)
+#define AT32F43x_DBGMCU_CTRL       (AT32F43x_DBGMCU_BASE + 0x40U)
+#define AT32F43x_DBGMCU_APB1_PAUSE (AT32F43x_DBGMCU_BASE + 0x80U)
+#define AT32F43x_DBGMCU_APB2_PAUSE (AT32F43x_DBGMCU_BASE + 0x0cU)
+#define AT32F43x_DBGMCU_SER_ID     (AT32F43x_DBGMCU_BASE + 0x20U)
+
+#define AT32F43x_DBGMCU_CTRL_SLEEP_DEBUG     (1U << 0U)
+#define AT32F43x_DBGMCU_CTRL_DEEPSLEEP_DEBUG (1U << 1U)
+#define AT32F43x_DBGMCU_CTRL_STANDBY_DEBUG   (1U << 2U)
+#define AT32F43x_DBGMCU_CTRL_SLEEP_Msk \
```
dragonmux, line 107: Please, if you're going to define a constant like this, use `_MASK` not `_Msk`.

src/target/at32f43x.c
```diff
@@ -129,6 +148,48 @@ static void at32f43_add_flash(target_s *const target, const target_addr_t addr,
...
+{
+	/* Collect target_mem errors by |= */
+	bool error = false;
+	/*
+	 * Enable sleep state emulation (clocks fed by HICK)
+	 * and make both watchdogs pause during core halts so that they
+	 * don't issue extra resets when we're doing e.g. flash reprogramming
+	 */
+	const uint32_t dbgmcu_ctrl = target_mem32_read32(target, AT32F43x_DBGMCU_CTRL);
+	if ((dbgmcu_ctrl & AT32F43x_DBGMCU_CTRL_SLEEP_Msk) == 0U)
```
dragonmux, line 161: It doesn't matter if any of the bits are already set, we need to ensure all of them are so please either check that the value is not already the mask value, or elide the `if`. We prefer elision of the `if` for code size and complexity reduction.

Ditto for the pause mask check below.

With the if elided, we would prefer to see each of the debug bits being set individually as in `target_mem32_write32(target, AT32F43x_DBGMCU_CTRL, target_mem32_read32(target, AT32F43x_DBGMCU_CTRL) | AT32F43x_DBGMCU_CTRL_SLEEP_DEBUG | AT32F43x_DBGMCU_CTRL_DEEPSLEEP_DEBUG | AT32F43x_DBGMCU_CTRL_STANDBY_DEBUG);` so it's a little clearer what exactly is getting set here.

src/target/at32f43x.c
ALTracer: Renamed.

src/target/at32f43x.c
ALTracer: Oh, I did the check correctly for APB1_PAUSE, but forgot about CTRL. Enhanced the comparison.
Note that there are only these three significant bits in CTRL register. I intended to elide the SWD writes, which could be about 50 microseconds long each (assuming two transactions, the second checking/clearing error).

src/target/at32f43x.c
```diff
@@ -129,6 +148,46 @@ static void at32f43_add_flash(target_s *const target, const target_addr_t addr,
...
+	if (!cortexm_attach(target))
+		return false;
+
+	at32f43_configure_dbgmcu(target);
+	return true;
+}
+
+static void at32f43_detach(target_s *target)
+{
+	uint32_t dbgmcu_ctrl = target_mem32_read32(target, AT32F43x_DBGMCU_CTRL);
```
dragonmux, line 181: If you are going to express this as reads into locals, please at least make them `const`. The modification can be inlined into the write parameter parens, which allows this. That is less error-prone than this form.

ALTracer: Accepted.

dragonmux (approved):
We're still not sold on this read-compare-modify-write pattern choice as it does increase Flash footprint significantly and is more difficult to understand + maintain, however.. we're going to approve and merge this because we'd like to see the core bug fixed either way.

The normal code base pattern of `target_mem32_write32(target, addr, target_mem32_read32(target, addr) | mod);` generates very clean assembly and requires much less thinking about (+ can't leak values from the operation) so it remains our preferred pattern for doing RMW cycles where the target does not support an atomic bitset/bitclear/xor alternate address for registers (RP2040 and RP2350 have this which is why the mention).

This is still a big improvement over the pre-PR WDT and sleep behaviour however so thank you for the contribution!

PR #1870, opened 2024-07-13
Fix: Use unbuffered stdout on Windows

ALTracer:
## Detailed description

* Not a feature.
* The existing problem is BMDA on Windows defaulting to 4 KiB block buffering on stdout. Users don't see output progress (which is line-buffered on Linux) until enough text appears, or BMDA exits.
* This PR switches stdout to unbuffered mode for WIN32 and Cygwin.

Initially I tried _IOLBF, but it returns some error from libc internals.
I don't feel like dealing with placing strategic fflush(stdout) calls in BMDA to massage it into printing info interactively.
Tested on Msys2 ucrt64 to remove delays between activity and printed output. Also tested in Qt Creator Debug to corrupt (mix) Program output significantly.

* [x] ~~It builds for hardware native (see [Building the firmware](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-firmware))~~ and should not affect it

## Closing issues
Not self-opening one.

### DIFF
--- a/src/platforms/hosted/platform.c
+++ b/src/platforms/hosted/platform.c
@@ -114,6 +114,10 @@ void platform_init(int argc, char **argv)
 {
 #if defined(_WIN32) || defined(__CYGWIN__)
 	SetConsoleOutputCP(CP_UTF8);
+	if (setvbuf(stdout, NULL, _IONBF, 0) < 0) {
+		int err = errno;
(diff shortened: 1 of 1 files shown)


dragonmux (approved):
LGTM - the string optimisation being done with that second parameter is not necessary but also doesn't hurt anything here, so we'll allow it and merge. Thank you for the contribution!

PR #1875, opened 2024-07-16
Fix: CMSIS-DAP SWD data shift UB

dragonmux:
## Detailed description

This PR addresses a UBSAN runtime notice in the SWD support for CMSIS-DAP adaptors. This is caused by an improper widening via integer promotion (which discards unsigned-ness during widening) of sequence data in `dap_swd_seq_in_parity()`. When this widening takes place and a shift right that would change the sign bit occurs, such as shifting 128 up 24 bits on a platform where `int` is 32-bit, this invokes undefined behaviour.

Before this PR, users would see a message such as `../src/platforms/hosted/dap_swd.c:170:39: runtime error: left shift of 255 by 24 places cannot be represented in type 'int'` from BMDA when built with UBSAN and this buggy behaviour triggers. After, no UBSAN errors are reported and SWD comms proceed normally.


esden (approved):
LGTM

PR #1878, opened 2024-07-22
Feature: Adapt traceswoasync (F103) to F4 API

ALTracer:
## Detailed description

* This is a new feature.
* The existing problem is traceswoasync.c driver existing and working for F103 but incompatible with F4.
* This PR solves it by porting some of the macros from aux_serial.c driver.

Required for enabling/switching to Async capture on F4 platforms. Split from #1529.
Can only test the combination with platform-specific enablement PRs (pending). Previously tested builds from non-PR branches. The macros expected/required from platforms will be provided in a separate follow-up PR.

## Closing issues

### DIFF
--- a/src/platforms/common/stm32/traceswoasync.c
+++ b/src/platforms/common/stm32/traceswoasync.c
@@ -44,6 +44,20 @@
 #include <libopencm3/stm32/usart.h>
 #include <libopencm3/stm32/dma.h>
 
+#if defined(DMA_STREAM0)
+#define dma_channel_reset(dma, channel)   dma_stream_reset(dma, channel)
+#define dma_enable_channel(dma, channel)  dma_enable_stream(dma, channel)
+#define dma_disable_channel(dma, channel) dma_disable_stream(dma, channel)
+
+#define DMA_PSIZE_8BIT DMA_SxCR_PSIZE_8BIT
+#define DMA_MSIZE_8BIT DMA_SxCR_MSIZE_8BIT
+#define DMA_PL_HIGH    DMA_SxCR_PL_HIGH
+#else
+#define DMA_PSIZE_8BIT DMA_CCR_PSIZE_8BIT
+#define DMA_MSIZE_8BIT DMA_CCR_MSIZE_8BIT
+#define DMA_PL_HIGH    DMA_CCR_PL_HIGH
+#endif
+
 static volatile uint32_t write_index; /* Packet currently received via UART */
 static volatile uint32_t read_index;  /* Packet currently waiting to transmit to USB */
 /* Packets arrived from the SWO interface */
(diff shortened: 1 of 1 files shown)


src/platforms/common/stm32/traceswoasync.c
```diff
@@ -109,6 +130,22 @@ void traceswo_setspeed(uint32_t baudrate)
...
+	}
+	if (dma_get_interrupt_flag(SWO_DMA_BUS, SWO_DMA_CHAN, DMA_TCIF)) {
+		dma_clear_interrupt_flags(SWO_DMA_BUS, SWO_DMA_CHAN, DMA_TCIF);
+		memcpy(
+			&trace_rx_buf[write_index * TRACE_ENDPOINT_SIZE], &pingpong_buf[TRACE_ENDPOINT_SIZE], TRACE_ENDPOINT_SIZE);
+	}
+	write_index = (write_index + 1U) % NUM_TRACE_PACKETS;
+	trace_buf_drain(usbdev, TRACE_ENDPOINT | USB_REQ_TYPE_IN);
+}
+#else
```
dragonmux, line 148: This isn't a request to change any of this new code, but more that it would be really nice if these two interrupt handler implementations could be merged - raw register accesses replaced by the `dma_*` calls. It would be nice not to have to maintain two copies of the same routine after all.

dragonmux (commented):
We've only got the one review note, and once it's been discussed/addressed we are happy to get this merged.

src/platforms/common/stm32/traceswoasync.c
ALTracer: I considered unifying the implementations, but also I like to code all interrupt handlers using direct CMSIS Device register accesses for performance reasons. This DMA Stream F4 implementation is harder to inline manually because of flags being scattered across `DMA_HISR/DMA_LISR` and `DMA_HIFCR/DMA_LIFCR`, depending on which stream a given platform uses (0-3, 4-7) for USART Rx. Jumping to these functions may not be expensive thanks to ART cache, but this fires once per 64 bytes. I doubt we should fork these two functions from libopencm3 into static inline impls in headers for BMD platforms.
STM32F103, however, will suffer from similar jumps and so it benefits from coded direct accesses to `DMA_ISR` and `DMA_IFCR`. GD32F103 may not care thanks to 0WS.
The generic memcpy call TBH should be replaced with a faster memcpy4 or `__aeabi_memcpy4` intrinsic for speed, both buffers are aligned by at least 4 bytes (malloc and linkerscript) and are at least 64 bytes long (packet size). But I digress.

dragonmux: > but also I like to code all interrupt handlers using direct CMSIS Device register accesses for performance reasons.

This handler will be (is) dominated by the `memcpy()` calls, the difference in maintainability here outweighs any small performance losses.

> The generic memcpy call TBH should be replaced with a faster memcpy4 or __aeabi_memcpy4 intrinsic for speed

`memcpy()` is a pseudofunction, always. It will already always be being replaced by `__builtin_memcpy()` by the compiler. This doesn't change that copying 64 bytes of memory is still the heavier operation than the register pokes, but does mean that it'll already be the most performant it can be. If the compiler can reason that it can replace the builtin with what the function does as inline asm, it'll even do that.

ALTracer: Thanks for reminding me that F103 may also jump to the 28-byte memcpy function, that is a significant enough technical reason for me to not bother any further. Note however that this instance of memcpy seems to get inlined/substituted, I see no function calls in objdump/puncover other than `trace_buf_drain()` (which calls `usbd_ep_write_packet()` and so on), the `dma1_channel5_isr` handler itself is only 136 bytes.
When I apply the fix, it increases to 144 and adds two get_/clear_flag function calls which were referenced by aux_serial.c anyways. So the change is no longer status quo on flash size; for maintainability reasons, I guess.

dragonmux: Aye, in this instance while smaller Flash utilisation would be nice, maintainability is a more pressing goal here in our opinion. If we turn out to be wrong? We can always re-split the IRQ handler back out and optimise, so no real loss there.

dragonmux (approved):
All LGTM now, merging. Thank you for the contribution!

PR #1881, opened 2024-07-26
Feature: Add platform-specific commands support

Misaka0x2730:
## Detailed description

I'm currently porting this project to my custom platform and at one point I needed to add a custom command for my platform.
My port uses this project as a submodule, so I thought it would be great to be able to add platform-specific commands without patching command.c file.

## Closing issues

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -125,6 +125,10 @@ const command_s cmd_list[] = {
 	{NULL, NULL, NULL},
 };
 
+#ifdef PLATFORM_HAS_CUSTOM_COMMANDS
+extern const command_s platform_cmd_list[];
+#endif
+
 bool connect_assert_nrst;
 #if defined(PLATFORM_HAS_DEBUG) && PC_HOSTED == 0
 bool debug_bmp;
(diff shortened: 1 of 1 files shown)


src/command.c
```diff
@@ -161,6 +165,13 @@ int command_process(target_s *const t, char *const cmd_buffer)
 			return !cmd->handler(t, argc, argv);
 	}
 
+#ifdef PLATFORM_HAS_CUSTOM_COMMANDS
+	for (const command_s *cmd = platform_cmd_list; cmd->cmd; ++cmd) {
+		if ((argc == 0) || !strncmp(argv[0], cmd->cmd, strlen(argv[0])))
```
dragonmux, line 170: The `argc == 0` term here can be dropped as the only reason this exists in the block above is to handle `monitor` (no arguments) being given so that the version command fires to handle that.

src/command.c
```diff
@@ -125,6 +125,10 @@ const command_s cmd_list[] = {
 	{NULL, NULL, NULL},
 };
 
+#ifdef PLATFORM_HAS_CUSTOM_COMMANDS
+extern const command_s platform_cmd_list[];
```
dragonmux, line 129: Given how this is defined, and as it decays to a pointer immediately, we're going to suggest defining this here as `extern const command_s *platform_cmd_list;` as that's a little more truthful to what's actually going to go on here.

dragonmux (changes requested):
We've got just a couple of review notes. The idea presented here is quite intriguing, and as this simplifies your maintenance burden we'd be happy to bring it in tree once the review items are addressed.

src/command.c
Misaka0x2730: Removed.

src/command.c
Misaka0x2730: Fixed

Misaka0x2730:
> We've got just a couple of review notes. The idea presented here is quite intriguing, and as this simplifies your maintenance burden we'd be happy to bring it in tree once the review items are addressed.

Thank you, I've addressed all items.

dragonmux (approved):
LGTM, we'll get this merged as soon as the builds complete. Thank you for the contribution!

PR #1882, opened 2024-07-29
Feature: STM32 watchdog timer control

### LINKED ISSUE #117: Watchdog not paused when debugging

dragonmux:
## Detailed description

In this PR we sort out support for and deal with WDTs support across the whole STM32 range supported. Included in this are fixes for issues such as bad Flash maps on some families, which were necessary to properly deal with differences in part addresses such as on the STM32L1 family where the address the DBGMCU is at changes based on which part it is in the family.

The approach used here needs work for handling what happens if a user attaches into a part, changes the code in Flash on the part, runs that new code which then modifies certain DBGMCU registers such as to set up tracing, and then detaches. Even with this limitation/bug in place, this greatly improves the state of debug across these families, by properly handling the WDTs during processor halt, and by properly setting up handling for WFI and WFE instructions encountered while debugging, solving a very long-standing problem with STM32 support.

This PR has been tested as much as possible on STM32F1/GD32F1, GD32VF1, STM32F4, STM32H5, STM32MP15 and STM32H7 parts, but we do not have the parts to test this on other series. We've done our best not to break anything however, so the chances of a regression are low. That said, user testing by people who do have parts, and preferably by people using the WDTs or WFI/WFE instructions, would be much appreciated.

## Closing issues

Fixes #117 (at least for STM32; LPC parts are generally already ok; i.MXRT is already taken care of; other parts are unknowns, but should probably be their own separate PRs and issues)

### DIFF
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -585,6 +585,7 @@ bool cortexm_probe(adiv5_access_port_s *ap)
 		PROBE(stm32h7_probe);
 		PROBE(stm32mp15_cm4_probe);
 		PROBE(stm32l0_probe);
+		PROBE(stm32l1_probe);
 		PROBE(stm32l4_probe);
 		PROBE(stm32g0_probe);
 		break;
--- a/src/target/riscv_debug.c
+++ b/src/target/riscv_debug.c
@@ -142,6 +142,7 @@ static const char *const riscv_gpr_names[RV_GPRS_COUNT] = {
 };
 
 // clang-format on
+
 typedef struct riscv_csr_descriptor {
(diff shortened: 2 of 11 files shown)


esden (approved):
LGTM! Merging so that it gets more testing from users. It should improve the status quo.

PR #1886, opened 2024-08-05
Feature: RISC-V Debug remote protocol acceleration

dragonmux:
## Detailed description

In this PR we implement acceleration of the RISC-V Debug protocol when using BMDA with the BMD remote protocol with a probe. This dramatically improves the speed of RISC-V communications, bringing the 3s discovery time for the GD32VF103 down to just under 0.5s.

This acceleration makes use of the protocol v4 framework for optional accelerations, allowing it to be introduced in an entirely backwards and forwards compatible way. It is gated behind a new Meson compilation option (`enable_riscv_accel`) so the Flash usage it causes is fully opt-in.

Additionally, this PR introduces RISC-V specific acceleration expansion, allowing other protocols for the DTM to be implemented later without bumping the protocol version, due to the supported RISC-V DTM protocols being enumerated in the same manner as acceleration protocols are.

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -112,3 +112,9 @@ option(
 	type: 'feature',
 	value: 'auto'
 )
+option(
+	'enable_riscv_accel',
+	type: 'boolean',
+	value: false,
+	description: 'Enable firmware-side protocol acceleration of RISC-V Debug'
+)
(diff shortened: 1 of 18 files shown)


esden (approved):
Nice addition! Thanks!

PR #1893, opened 2024-08-13
cortexar: fixed CPSR mapping for GDB

litui:
Prerequisite to further work on resuming from a load to RAM.

## Detailed description

* Solves issue where CPSR couldn't be written at the end of a load or with a manual "set $cpsr="
* Macro defines 25U as the remapped position in GDB and accounts for the remapping in `cortexar_reg_ptr` and `cortexar_reg_width`

### DIFF
--- a/src/target/cortexar.c
+++ b/src/target/cortexar.c
@@ -194,6 +194,9 @@ typedef struct cortexar_priv {
 #define CORTEXAR_CPSR_MODE_SYS  0x0000001fU
 #define CORTEXAR_CPSR_THUMB     (1U << 5U)
 
+/* CPSR remap position for GDB XML mapping */
+#define CORTEXAR_CPSR_GDB_REMAP_POS 25U
+
 /* Banked register offsets for when using using the DB{0,3} interface */
 enum {
 	CORTEXAR_BANKED_DTRTX,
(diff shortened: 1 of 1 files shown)


src/target/cortexar.c
```diff
@@ -1758,8 +1761,19 @@ static size_t cortexar_build_target_description(char *const buffer, size_t max_l
...
+		if (i == 3U)
+			/* Remap CPSR in particular */
+			snprintf(tag_closure, regnum_length, " regnum=\"%u\"/>", CORTEXAR_CPSR_GDB_REMAP_POS);
+		else
+			strncpy(tag_closure, "/>", regnum_length);
+
+		offset += snprintf(buffer + offset, print_size, "<reg name=\"%s\" bitsize=\"32\"%s%s", name,
+			gdb_reg_type_strings[type], tag_closure);
+
+		free(tag_closure);
```
dragonmux, line 1776: This change seems to produce the same result as before, just via an extra memory allocation?

Think it might be better to add a comment in front of the original snprintf() in that case to have it say what the `regnum="25"` stuff is about so it's not just a random condition in the code.

dragonmux (commented):
Only see the one small thing that could potentially do with adjusting, but with that taken care of we'll be happy to merge this.

src/target/cortexar.c
litui: Okay, fair. I was just aiming to reduce the number of places in which the 25 remap is defined. Will restore.

litui: Amended the commit.

dragonmux (approved):

PR #1896, opened 2024-08-15
Fix: renesas_rz: autodetect RAM size based on RAM reads

litui:
## Detailed description

* Previously, RAM was statically set to size 0x200000U but the RZ/A1L and RZ/A1LU have an additional 1MB of RAM. This code adds a function to determine RAM size ~~for a given part_id~~ based on auto-detection (there may be other examples of this in the RZ series).
* Code currently iterates through RZ-series RAM at probe time at 512kB intervals and sets the RAM size to the maximum error-free read position. If that fails, it sets RAM to the maximum known for the RZ series (at present, 10MB).

### DIFF
--- a/src/target/renesas_rz.c
+++ b/src/target/renesas_rz.c
@@ -73,7 +73,7 @@
  * except for pnrs stored in 'FIXED_PNR1', where the code is stored in reverse order (but the last 3 bytes are still 0x20 aka ' ')
  */
 
-/* Base address for the RZ-series, including mirror */
+/* Base address for the OCRAM regions, including their mirrors (including RETRAM) */
 #define RENESAS_OCRAM_BASE        0x20000000U
 #define RENESAS_OCRAM_MIRROR_BASE 0x60000000U
 
@@ -191,7 +191,7 @@ bool renesas_rz_probe(target_s *const target)
 		renesas_rz_add_flash(target);
 
 	/* Determine RAM size by attempting to read in 512MB increments */
-	uint32_t ram_size = renesas_rz_ram_size(target);
+	const uint32_t ram_size = renesas_rz_ram_size(target);
 	target_add_ram32(target, RENESAS_OCRAM_BASE, ram_size);
 	target_add_ram32(target, RENESAS_OCRAM_MIRROR_BASE, ram_size);
 	return true;
@@ -216,9 +216,9 @@ static const char *renesas_rz_part_name(const uint32_t part_id)
  */
 static uint32_t renesas_rz_ram_size(target_s *target)
 {
-	uint32_t result = 0;
 	for (uint32_t addr = RENESAS_OCRAM_BASE; addr < RENESAS_OCRAM_BASE + RENESAS_OCRAM_MAX_SIZE;
 		 addr += RENESAS_OCRAM_SCAN_INCREMENT) {
+		uint32_t result;
 		/* Read ahead by one scan increment and if there's an error, return the current size. */
 		if (target_mem32_read(target, &result, addr + RENESAS_OCRAM_SCAN_INCREMENT - 8U, sizeof(result)))
 			return addr - RENESAS_OCRAM_BASE;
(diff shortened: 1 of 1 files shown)


src/target/renesas_rz.c
```diff
@@ -77,6 +77,7 @@
 #define RENESAS_OCRAM_BASE        0x20000000U
 #define RENESAS_OCRAM_MIRROR_BASE 0x60000000U
 #define RENESAS_OCRAM_SIZE        0x00200000U
+#define RENESAS_OCRAM_SIZE_A1L    0x002fffffU
```
dragonmux, line 80: It would be unusual for the amount of RAM to be non-power-of-two - note this isn't the address of the last addressable byte, but how many there are, so a region defined as 0x20000000-0x202fffff has 0x00300000 bytes in it. Otherwise you'd be defining the region to be 0x20000000-0x202ffffe with this. Are you sure this isn't `0x00300000U`?

dragonmux (changes requested):
We only spot the one item in review that's suspect - with that taken care of though, we'll be happy to merge this.

src/target/renesas_rz.c
litui: ah good catch! Missed that. Will fix.

litui:
Please hold off on merging this. The RZ/A1H (10MB RAM) detects as the same cpu currently.

dragonmux:
No worries, just let us know when you're happy that the patch is correct and works and we'll re-review.

litui:
Good to go for a review. 👍🏻

src/target/renesas_rz.c
```diff
@@ -73,10 +73,16 @@
  * except for pnrs stored in 'FIXED_PNR1', where the code is stored in reverse order (but the last 3 bytes are still 0x20 aka ' ')
  */
 
-/* Base address and size for the 4 OCRAM regions + their mirrors (includes RETRAM) */
+/* Base address for the RZ-series, including mirror */
```
dragonmux, line 76: Grammatical nitpicks: `/* Base address for the OCRAM regions, including their mirrors */` - the sentence needs an object for what the subject is about, and as the object us plural the follow on must be too.

If memory serves correctly, this still includes the RETRAM region so probably worth noting that.

src/target/renesas_rz.c
```diff
@@ -183,8 +190,10 @@ bool renesas_rz_probe(target_s *const target)
 	if (boot_mode == RENESAS_BSCAN_BOOT_MODE_SPI)
 		renesas_rz_add_flash(target);
 
-	target_add_ram32(target, RENESAS_OCRAM_BASE, RENESAS_OCRAM_SIZE);
-	target_add_ram32(target, RENESAS_OCRAM_MIRROR_BASE, RENESAS_OCRAM_SIZE);
+	/* Determine RAM size by attempting to read in 512MB increments */
+	uint32_t ram_size = renesas_rz_ram_size(target);
```
dragonmux, line 194: This can be `const` so let's make it so.

src/target/renesas_rz.c
```diff
@@ -202,6 +211,21 @@ static const char *renesas_rz_part_name(const uint32_t part_id)
 	return "Unknown";
 }
 
+/*
+ * Function to read RAM incrementally to determine the size of it for the given model.
+ */
+static uint32_t renesas_rz_ram_size(target_s *target)
+{
+	uint32_t result = 0;
```
dragonmux, line 219: Given this is only used in the loop and is otherwise not read, perhaps move this to just inside the loop and drop the init? Should improve codegen a touch and improves scoping.

dragonmux (changes requested):
Couple of small things, mostly just comment grammar, but nothing show stopping. With those taken care of, please *rebase* this on `main` (`git rebase -i main`) to eliminate the merge commit and we'll get this merged.

dragonmux (approved):
All LGTM now, merging. Thank you for the contribution!

PR #1897, opened 2024-08-16
Fix addresses on STM32H5 (DBGMCU, SRAM alias)

ALTracer:
## Detailed description

* No new features.
* The existing problems are `mon rev` not working on H503, and SRAM at 0x20000000 not declared by BMD.
* This PR solves them by moving the DBGMCU base from AP0 to AP1 viewpoint, and adding more RAM regions.

First part tested on WeAct Studio STM32H503CBU board using up-to-date BMDA via an "swlink". Second part is more trivial, but I could test it on request (and expand to H563 without testing).
DBGMCU accesses now stick, but I didn't verify their effect other than looking at these bytes in gdb.
Factory firmware (VCP and blinky) was linked to 0x20000000 address of SRAM, but BMP and BMDA reported only 0x0a000000. I get that it's a Secure-aware chip, and it also has a Secure counterpart alias at 0x0e000000 to Non-secure, but it used to be normal to simply link to 0x20000000, and I think it's still valid to. Otherwise GDB shouldn't let me read variables from .data/.bss or unwind stack (I didn't try with a custom firmware I'd have symbols to yet).

## Closing issues

Not self-reporting this as an issue.


dragonmux (approved):
LGTM, and you've shown the DBGMCU change working in Discord, so merging. Thank you for the contribution!

PR #1900, opened 2024-08-17
Fix: s32k3xx: Remove the single assert 

ALTracer:
## Detailed description

* No new features.
* The existing problem is that the S32K3xx driver, introduced in #1652, references `assert()` which pulls `fiprintf()` and `_vfiprintf_r` again, resulting in extra 1.5 KiB of newlib stdio.
* This PR solves it by ~~replacing the assert with a logged runtime error.~~ IIUC target flash layer should call it with a `len=writesize`. Depending on review and API guarantees I can drop even that. Update: dropped.

I expect a significant size-diff from CI. This should restore the effect of my previous #1513.
Cannot test on hardware. CC @via.

## Closing issues

### DIFF
--- a/src/target/s32k3xx.c
+++ b/src/target/s32k3xx.c
@@ -23,8 +23,6 @@
  * the XML memory map and Flash memory programming.
  */
 
-#include <assert.h>
-
 #include "command.h"
 #include "general.h"
 #include "target.h"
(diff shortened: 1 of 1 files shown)


dragonmux (approved):
Given the target Flash API layer guarantees this property already - LGTM, merging once the builds complete. Thank you for the contribution!

PR #1903, opened 2024-08-18
Fix: Enable parallel make workers on GH Actions CI

ALTracer:
## Detailed description

* This can sound as a minor new feature.
* The existing problem is suboptimal utilisation of Free runners, which are 4-core, and hence longer build times (4 minutes) for Makefile workflows.
* This PR solves it by reintroducing -j option and nproc applet (coreutils). Not touching Windows/mingw or MacOS.

Tested on GH CI Free in fork by manually triggering a workflow from branch, the speedup is there (2m20s).

There was a known problem with log clobbering, where multiple `make` jobserver workers attempt to pass compiler errors up simultaneously. Make 4.3 now supports `--output-sync` option, and I used `-Otarget` here to make it buffer and sync output up to one build target (object). https://www.gnu.org/software/make/manual/html_node/Parallel-Output.html Did not test this yet, I'd need to see (or introduce) errors in multiple sources of one probe platform for this.

I would also like to apply -k, --keep to both make and meson/ninja, so that they attempt to compile everything and gather up all diagnostics, and not simply fail early on first build error -- this project is relatively small in terms of build time.

## Closing issues

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -97,7 +97,7 @@ jobs:
 
       # Build the firmware for all platform variants and BMDA in BMP-only mode
       - name: Build all platform variants firmware and BMP only BMDA
-        run: make all_platforms HOSTED_BMP_ONLY=1
+        run: make all_platforms HOSTED_BMP_ONLY=1 -j$(nproc) -Otarget
 
       - name: Clean
         run: make clean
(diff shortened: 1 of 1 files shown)


dragonmux (approved):

Regarding `-k`, we can see the appeal but think that's an option better used locally once CI highlights an issue.

PR #1905, opened 2024-08-18
Feature: Update swolisten to support walking USB descriptors

ALTracer:
## Detailed description

* There is a new feature.
* The existing problem is in-tree swolisten.c falling outdated, meanwhile the hardcoded endpoint was moved.
* This PR solves it by adding required code to fetch libusb config and walk it.

New code is extracting interface number (5) and endpoint number (0x85 or 0x83), matching on interface string descriptor.
Note that this source file is relatively cold-blooded software, and is ignored by clang-format. I attempted to add minimal changes and follow its tab format. I guess it's not even checked by CI.
`gcc -Wall -Og -g -I/usr/include/libusb-1.0 swolisten.c /usr/lib/x86_64-linux-gnu/libusb-1.0.so.0 -o swolisten`
`gcc -Wall -Og -g $(pkg-config --cflags libusb-1.0) swolisten.c $(pkg-config --libs libusb-1.0) -o swolisten`

With old swolisten.c and recent firmware, Linux kernel is angry in dmesg with lines like `usbfs: process N (swolisten) did not claim interface 2 before use`. With updated swolisten.c it should open both old and recent firmware.

## Closing issues


scripts/swolisten.c
```diff
@@ -435,10 +435,36 @@ int usbFeeder(void)
       if (!(dev = libusb_get_device(handle)))
 	continue;
 
-      if (libusb_claim_interface (handle, INTERFACE)<0)
+      uint8_t bmp_interface = 0;
+      unsigned char bmp_endpoint = 0;
```
dragonmux, line 439: We would suggest using `uint8_t` for this variable too as that then matches to the type for bEndpointAddress and removes an item needing modernisation.

scripts/swolisten.c
```diff
@@ -435,10 +435,36 @@ int usbFeeder(void)
...
+
+      if (libusb_get_active_config_descriptor(dev, &config) < 0)
+	continue;
+
+      int interface_found = 0;
+      for (int if_num = 0; if_num < config->bNumInterfaces && !interface_found; if_num++) {
```
dragonmux, line 445: Given `bNumInterfaces` is a uint8_t in the config descriptors structure, please use `uint8_t` for `if_num` here.

dragonmux (changes requested):
There's only two small things we see focusing only on your changes; with those addressed, we're happy to merge this and see swolisten fixed

scripts/swolisten.c
ALTracer: Accepted. I was first looking at signature of `libusb_bulk_transfer()`, and it's `unsigned char endpoint` there.
```c
int LIBUSB_CALL libusb_bulk_transfer(libusb_device_handle *dev_handle,
	unsigned char endpoint, unsigned char *data, int length,
	int *actual_length, unsigned int timeout);
```

scripts/swolisten.c
ALTracer: Accepted. It's not like STM32F1 fsdev or STM32F4 dwc2 OTG_FS can support more than 8 endpoints, so I can't imagine >256 interfaces being plausible.

dragonmux (approved):
LGTM, Merging. thank you for the contribution!

PR #1854, opened 2024-06-30
Fix stlinkv3 platform bits

ALTracer:
## Detailed description

* There is a small new feature.
* Existing problems are `(gdb) compare-sections` potentially not working on STLINK/V3, and SWO capture baudrate too low.
* This PR solves these by enabling RCC_CRC and switching UART5 clocksource from APB1 Pclk to Sysclk.

Also I add a couple paragraphs of documentation to platform README.md with info from ST manuals and libopencm3.
Testing on stlinkv3 required prior to merge.

## Your checklist for this pull request

* [x] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md)
* [x] I've read the [guidelines for contributing](https://github.com/blackmagic-debug/blackmagic/blob/main/CONTRIBUTING.md) to this repository
* [x] It builds for hardware native (see [Building the firmware](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-firmware)) *and does not affect it*
* [x] It builds as BMDA (see [Building the BMDA](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-app)) *and does not affect it*
* [ ] I've tested it to the best of my ability
* [x] My commit messages provide a useful short description of what the commits do

## Closing issues
None reported.

PR #1908, opened 2024-08-22
Feature: Enable USART OVER8 as needed

ALTracer:
## Detailed description

* This is a new non-breaking feature.
* The existing problem is low max baudrates on in-tree platforms with STM32 USART IP which supports OVER8.
* This PR solves it by adding a register poke to unlock higher baudrates in DIV16..DIV8 range, switching to hardware oversampling x8.

F103 platforms should not be affected, the new code is gated by a macro.
F4 platforms should reach double the baudrates on all ports. This is a logical follow-up to #1864 and whatever went in libopencm3.
F723, likewise, should be able to reach double the baudrates (especially when coupled with #1854), becoming the fastest UART of in-tree platforms.
Whether enabling CR3 ONEBIT mode for better clock deviation tolerance (2%->3%) is also required -- is a good discussion point.

I tried drying the implementation since April 27, and now there's a single `bmp_usart_set_baudrate()` function called by aux_serial.c and traceswoasync{,_f723}.c; depending on review it can be reworked into a different distribution/form.

## Closing issues

### DIFF
--- a/src/platforms/common/aux_serial.c
+++ b/src/platforms/common/aux_serial.c
@@ -87,6 +87,39 @@ static char aux_serial_transmit_buffer[AUX_UART_BUFFER_SIZE];
 #define usart_set_parity(uart, parity)     uart_set_parity(uart, parity)
 #endif
 
+void bmd_usart_set_baudrate(uint32_t usart, uint32_t baud_rate)
+{
+	/* If the new baud rate is out of supported range for a given USART, then keep previous */
+#if defined(LM4F)
+	/* Are we running off the internal clock or system clock? */
+	const uint32_t clock = UART_CC(usart) == UART_CC_CS_PIOSC ? 16000000U : rcc_get_system_clock_frequency();
+#else
+	const uint32_t clock = rcc_get_usart_clk_freq(usart);
+#endif
+	const uint32_t baud_lowest = clock / 65535U;
+	const uint32_t baud_highest_16x = clock / 16U;
+#if defined(USART_CR1_OVER8)
+	const uint32_t baud_highest_8x = clock / 8U;
+
+	/* Four-way range match */
+	if (baud_rate < baud_lowest) /* Too low */
+		return;
+	/* less-than-or-equal: Prefer OVER16 at exactly /16 */
+	else if (baud_rate > baud_lowest && baud_rate <= baud_highest_16x)
+		usart_set_oversampling(usart, USART_OVERSAMPLING_16);
+	else if (baud_rate > baud_highest_16x && baud_rate <= baud_highest_8x)
+		usart_set_oversampling(usart, USART_OVERSAMPLING_8);
+	else /* Too high */
+		return;
+#else
+	/* STM32F103 (16-bit wide BRR) and TM4C123 PL011 (16-bit wide IBRD) */
+	if (baud_rate < baud_lowest || baud_rate > baud_highest_16x)
+		return;
+#endif
+
+	usart_set_baudrate(usart, baud_rate);
+}
+
 #if defined(STM32F0) || defined(STM32F1) || defined(STM32F3) || defined(STM32F4) || defined(STM32F7)
 void aux_serial_init(void)
 {
@@ -96,7 +129,7 @@ void aux_serial_init(void)
 
 	/* Setup UART parameters */
 	UART_PIN_SETUP();
-	usart_set_baudrate(USBUSART, 38400);
+	bmd_usart_set_baudrate(USBUSART, 38400);
 	usart_set_databits(USBUSART, 8);
 	usart_set_stopbits(USBUSART, USART_STOPBITS_1);
 	usart_set_mode(USBUSART, USART_MODE_TX_RX);
@@ -218,7 +251,7 @@ void aux_serial_set_encoding(const usb_cdc_line_coding_s *const coding)
 	/* Some devices require that the usart is disabled before
 	 * changing the usart registers. */
 	usart_disable(USBUSART);
-	usart_set_baudrate(USBUSART, coding->dwDTERate);
+	bmd_usart_set_baudrate(USBUSART, coding->dwDTERate);
 
 #if defined(STM32F0) || defined(STM32F1) || defined(STM32F3) || defined(STM32F4) || defined(STM32F7)
 	if (coding->bParityType != USB_CDC_NO_PARITY)
--- a/src/platforms/common/traceswo.h
+++ b/src/platforms/common/traceswo.h
@@ -29,6 +29,7 @@
 void traceswo_init(uint32_t baudrate, uint32_t swo_chan_bitmask);
 void traceswo_deinit(void);
 uint32_t traceswo_get_baudrate(void);
+void bmd_usart_set_baudrate(uint32_t usart, uint32_t baud_rate);
 #else
 void traceswo_init(uint32_t swo_chan_bitmask);
 #endif
--- a/src/platforms/stlinkv3/README.md
+++ b/src/platforms/stlinkv3/README.md
@@ -29,15 +29,27 @@ some situations, Romtable access may also fail on STM32 device and so a debugger
 warm plug will fail. Cold plug should work with any STM32 device.
 
 ## Performance
-Stock firmware has three settings: High-performance (192M), Standard frequency (96M), Low-consumption (48M).
-SWD up to 24000 kHz, JTAG up to 21333 kHz (SPI), SWO up to 16 Mbaud, VCP 732-16000000 baud.
-Suffix 7 devices are rated for -40..+125 deg C but at 200 MHz frequency maximum.
-Other mentioned limits from DS11853 are 144M, 168M, 180M for scales 3-2-1 and overdrive on/off.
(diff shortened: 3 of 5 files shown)


src/platforms/common/traceswo.h
```diff
@@ -29,6 +29,7 @@
 void traceswo_init(uint32_t baudrate, uint32_t swo_chan_bitmask);
 void traceswo_deinit(void);
 uint32_t traceswo_get_baudrate(void);
+void bmp_usart_set_baudrate(uint32_t usart, uint32_t baud_rate);
```
dragonmux, line 32: Given you're suffixing this (good), please use `bmd` not `bmp` - it's not a function specific to `native`, so should not use `native`'s name shortening but rather then project's. This is a point of consistency that needs work elsewhere in the code base still with this being the intended direction.

src/platforms/common/aux_serial.c
```diff
@@ -87,6 +87,23 @@ static char aux_serial_transmit_buffer[AUX_UART_BUFFER_SIZE];
...
+void bmp_usart_set_baudrate(uint32_t usart, uint32_t baud_rate)
+{
+#if defined(USART_CR1_OVER8)
+	const uint32_t clock = rcc_get_usart_clk_freq(usart);
+	/* Autoselect default OVER16 unless a very high baudrate requested */
+	if (baud_rate <= clock / 16U) {
+		usart_set_oversampling(usart, USART_OVERSAMPLING_16);
+	} else if ((clock / 16U < baud_rate) && (baud_rate <= clock / 8U)) {
+		usart_set_oversampling(usart, USART_OVERSAMPLING_8);
+	} else {
```
dragonmux, line 99: The braces around the sections of this `if` block are unnecessary as are the parens in the `else if` clause, please drop them.

src/platforms/common/aux_serial.c
```diff
@@ -87,6 +87,23 @@ static char aux_serial_transmit_buffer[AUX_UART_BUFFER_SIZE];
 #define usart_set_parity(uart, parity)     uart_set_parity(uart, parity)
 #endif
 
+void bmp_usart_set_baudrate(uint32_t usart, uint32_t baud_rate)
+{
+#if defined(USART_CR1_OVER8)
+	const uint32_t clock = rcc_get_usart_clk_freq(usart);
+	/* Autoselect default OVER16 unless a very high baudrate requested */
+	if (baud_rate <= clock / 16U) {
```
dragonmux, line 95: This could be fine, but please check if a BRR of 1 is allowed. Otherwise this will need to check for `clock / 32U` or possibly even higher to check for a large enough dividend to work.

dragonmux (changes requested):
There are a couple of stylistic items that need addressing, but with them taken care of and the one query/request answered (even if it does not need code changed) we'll be happy to merge this - looks good principally.

src/platforms/common/traceswo.h
ALTracer: I was looking at https://github.com/blackmagic-debug/blackmagic/blob/11d804d1fc5a094a5cf82978be1184a680484d7f/src/platforms/common/stm32/gpio.h#L28 for `bmp_gpio_set()`. You're right per `s/bmp_/bmd_/g`, just like in `-Dbmd_bootloader=true` case. Accepting.

src/platforms/common/aux_serial.c
ALTracer: RM0383 tells the highest legal baudrate is 10.5Mbaud for 84MHz Pclk in OVER8 when a fixed-point number 1.000 is "programmed in the baud rate register" (Table 82, end of example baudrate tables).
I understand the refman such that DIV_Mantissa has to be at least 1, and DIV_Fraction can be anything. So I limit to /16 normally, Q12.4 = 1+0/16, and to /8 specially, Q12.3 = 1+0/8. It corresponds to BRR of 16 (0x10) and 16 (0x10), not 1 (0x01). This is an improvement to previous behaviour where BMF would silently set any BRR, including illegal values (1..7, i.e. 0+1/16 .. 0+7/16 etc.)

Also see 19.3.3 Receiver, Selecting the proper oversampling method, 
> Depending on the application:
•select oversampling by 8 (OVER8=1) to achieve higher speed (up to fPCLK/8). ...
•select oversampling by 16 (OVER8=0) to increase the tolerance of the receiver to clock deviations. In this case, the maximum speed is limited to maximum fPCLK/16.

src/platforms/common/aux_serial.c
dragonmux: This does include the else block too, we forgot to make clear.

src/platforms/common/aux_serial.c
dragonmux: Long version short: Because of matching code in libopencm3 which generates a BRR of 0x0010U when `baud_rate == clock / 16U` it's a sufficient check? (trying to summarise what you're saying)

If that is the case then we're happy and can mark this review note respolved.

src/platforms/common/aux_serial.c
ALTracer: Removed all braces.

src/platforms/common/aux_serial.c
ALTracer: Expanded check to take care of four ranges. Added in-range check for F103.

src/platforms/stlinkv3/README.md
```diff
@@ -29,15 +29,27 @@ some situations, Romtable access may also fail on STM32 device and so a debugger
...
-SWO capture is implemented as UART5 Rx DMA. 3375k (4500k with OVER8) is the limit.
-13.5 Mbaud (18M with OVER8) could be derived from Sysclk, 216/N where N>=16 (OVER8: 2*216/N where 24<=N<=31).
+(UM2448) Stock firmware has three settings: High-performance (192M), Standard frequency (96M), Low-consumption (48M).
+SWD up to 24000 kHz (then 8000 kHz, 3333, ...); JTAG up to 21333 kHz (both likely via SPI5, SPI3), has a return clock pin, does not support daisy chaining of targets.
+
+SWO up to 16 Mbaud, VCP 732-16000000 baud.
+See 14.2 Baud rate computing: 192/N where 16<=N<=65535 (OVER8: 2*192/N where 24<=N<=31). Substitute 96 or 48 correspondingly.
+
+(RN0093) "On STLINK-V3 boards in high-performance mode, the minimum baud rate for Virtual COM port is 2931 bauds."
+2931*65535~=192M, or rather 192M/65535~=2929.7, which also confirms the first system frequency.
```
dragonmux, line 39: is that 192M in MHz - if so, please include the full units (in all relevant places).

src/platforms/stlinkv3/README.md
```diff
@@ -29,15 +29,27 @@ some situations, Romtable access may also fail on STM32 device and so a debugger
...
+(DS11853 F723) Suffix 7 devices are rated for -40..+125 deg C but at 200 MHz frequency maximum.
+Other mentioned limits from datasheet are 144M, 168M, 180M for scales 3-2-1 and overdrive on/off.
+
+This firmware, in contrast, always runs at 216 MHz (suffix 6). AHB clock = 216M, APB1 = 54M, APB2 = 108M.
+SWD can be around 12-18 MHz (no SPI) when size-optimized, JTAG is slower (no SPI). CRC32 offloading enabled.
+
+USART1, USART6 on APB2 (108M Pclk) are 6.75Mbaud, others (including UART5) on APB1, 3375000 baud.
+SWO capture is implemented as UART5 Rx DMA in particular.
+
+Recent versions of firmware switch `uart_ker_ck` to Sysclk, bumping them to 13.5Mbaud, and enable OVER8 as needed, allowing 216M/8=27Mbaud in theory.
```
dragonmux, line 50: `sysclk` is not capitalised (for consistency) we suggest.

dragonmux (commented):
There are some stylistic items to be dealt with (also, apologies for how long this review took to get done), but once they're handled, we're happy to merge this. It looks functionally good.

src/platforms/common/aux_serial.c
```diff
@@ -87,6 +87,39 @@ static char aux_serial_transmit_buffer[AUX_UART_BUFFER_SIZE];
...
+#if defined(USART_CR1_OVER8)
+	const uint32_t baud_highest_8x = clock / 8U;
+
+	/* Four-way range match */
+	if (baud_rate < baud_lowest) /* Too low */
+		return;
+	/* less-than-or-equal: Prefer OVER16 at exactly /16 */
+	else if ((baud_lowest < baud_rate) && (baud_rate <= baud_highest_16x))
+		usart_set_oversampling(usart, USART_OVERSAMPLING_16);
+	else if ((baud_highest_16x < baud_rate) && (baud_rate <= baud_highest_8x))
```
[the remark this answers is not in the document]
ALTracer: Downgraded range check to suggested form, placing `baud_rate` as first terms.

src/platforms/stlinkv3/README.md
ALTracer: Expanded to MHz or millions everywhere in changed snippets.

dragonmux (approved):

PR #1910, opened 2024-08-25
Feature: ADIv6 support

dragonmux:
## Detailed description

In this PR we introduce support for ADIv6 and associated remote protocol acceleration so it runs at a reasonable pace. This is a precursor to RP2350 support which is an ADIv6 device.

ARM pivoted how the debug infrastructure looks from the debugger's perspective from ADIv5.2, changing the DP from directly connecting to up to 256 APs and using an index number to determine which to talk to. Instead in the new architecture, the DP connects to a debug resources bus which may contain a ROM table and any number of debug resources (these are not required to be all APs!). The registers for the AP are now memory mapped into a 4kiB block on this bus instead of being banked as in the old model. There are also more registers possible now w/ the original base set mapped at +0xd00 in the new memory space.

This requires new routines for accessing AP registers, and new remote protocol acceleration for delivering the requisite information to the probe to act on. We have aimed to keep the amount of extra code and work as thin as possible however, to keep extra Flash utilisation down by maximally re-using the existing infrastructure and code. Once on an APv2 (what ADIv6 calls these new APs), the previous ADIv5 component probe logic can take over and work as normal, as well all the memory I/O logic on an AP remains the same.

## Known Issues

* ST-Link protocol handling of these new DPs and APs is unknown.


src/target/adiv6.c
```diff
@@ -0,0 +1,373 @@
...
+		return entry_lower;
+	const uint32_t entry_upper = adiv5_ap_read(&rom_table->base, ADIV5_AP_REG(entry_offset + 4U));
+	return ((uint64_t)entry_upper << 32U) | (uint64_t)entry_lower;
+}
+
+static bool adiv6_parse_coresight_v0_rom_table(
+	adiv6_access_port_s *const rom_table, const target_addr64_t base_address, const uint64_t pidr)
+{
+	/* Extract the designer code and part number from the part ID register */
+	const uint16_t designer_code = adiv5_designer_from_pidr(pidr);
```
esden, line 167: If the functions are generic between adiv6 and adiv5, should the naming of those functions be updated to reflect the fact that they are not specific to the ADI version?

dragonmux: We wondered about that but weren't sure about doing so and the resulting churn - but this would definitely be a good idea. We suggest moving those functions and the declarations into adi.c and adi.h as it wouldn't make much sense to leave them in adiv5.{c,h} as a result. If you agree we'll write a short series appended to this to do those moves.

dragonmux: Presuming that moving the parts out into a new file is correct, we have made the suggested adjustments.

esden: Yeah, moving them to the non versioned files is the correct thing to do. :)

esden (approved):
LGTM

PR #1912, opened 2024-08-27
Fix: Removed GDB_MAX_PACKET_SIZE

Misaka0x2730:
## Detailed description

Removed GDB_MAX_PACKET_SIZE because there is already GDB_PACKET_BUFFER_SIZE define for gdb buffer size.

### DIFF
--- a/src/include/gdb_main.h
+++ b/src/include/gdb_main.h
@@ -23,6 +23,7 @@
 
 #include "target.h"
 
+/* Allow override in other platforms if needed */
 #ifndef GDB_PACKET_BUFFER_SIZE
 #define GDB_PACKET_BUFFER_SIZE 1024U
 #endif


src/include/gdb_main.h
```diff
@@ -23,7 +23,9 @@
 
 #include "target.h"
 
+#ifndef GDB_PACKET_BUFFER_SIZE
```
dragonmux, line 26: There shouldn't be a need to guard this define - it is always only defined here and never in the build system.

If this is because of something you want to do externally to change the buffer sizing, please add a comment before this `#ifndef` explaining its usage (when it is used, how, that 1024 is the absolute lower bound due to the BMP remote protocol) so it's documented and we'll know not to remove the logic here for an external platform.

dragonmux (commented):
Got one query, but with that answered then we're happy and  will approve for merge.

dragonmux (approved):

PR #1913, opened 2024-08-27
Fix: imxrt: Cortex-M4 core detection for RT1176

litui:
## Detailed description

* The i.MX RT1176 includes both an M7 and M4 core but only the M7 was being detected by BMP / BMDA. This fix enables detection of the M4 core (a separate AP entry) under the same target part_id as the M7.
* The generic RAM mapping provided in imxrt.c for all i.MXRT devices was inaccurate for the RT117x and did not distinguish between the memory available to the M7 core vs. the M4 core. This fix maps the RAM based on the target->part_id and core and includes a fairly comprehensive mapping for the RT117x.
* Flash detection logic did not require changing (both cores map the FlexSPI controller at the same location).
* The i.MX RT117x supports register-configurable SDRAM and SRAM but the addresses for this are not fixed and the same controller can be used for flash memory as well. These ranges have been excluded from this fix.
* The [i.MX RT1170 Processor Reference Manual rev 2](https://www.nxp.com/webapp/Download?colCode=IMXRT1170RM) (sign-in required) was used as reference.

### DIFF
--- a/src/target/imxrt.c
+++ b/src/target/imxrt.c
@@ -50,10 +50,34 @@
 #define IMXRT_SRC_BOOT_MODE1 (IMXRT_SRC_BASE + 0x004U)
 #define IMXRT_SRC_BOOT_MODE2 (IMXRT_SRC_BASE + 0x01cU)
 
-#define IMXRT_OCRAM1_BASE  UINT32_C(0x20280000)
-#define IMXRT_OCRAM1_SIZE  0x00080000U
-#define IMXRT_OCRAM2_BASE  UINT32_C(0x20200000)
-#define IMXRT_OCRAM2_SIZE  0x00080000U
+#define IMXRT_OCRAM1_BASE UINT32_C(0x20280000)
+#define IMXRT_OCRAM1_SIZE 0x00080000U
+#define IMXRT_OCRAM2_BASE UINT32_C(0x20200000)
+#define IMXRT_OCRAM2_SIZE 0x00080000U
+
+#define IMXRT117x_PART_ID 0x88c6U
+
+#define IMXRT117x_ITCM_BASE       0x00000000U /* FlexRAM */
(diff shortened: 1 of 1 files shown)


litui:
Tested using a Google Coral Micro dev board running its `multi_core_hello` example (the M4 core has to be switched on for it to be detected).

src/target/imxrt.c
```diff
@@ -50,10 +50,34 @@
...
+#define IMXRT117x_PART_ID 0x88c6U
+
+#define IMXRT117x_ITCM_BASE       0x00000000U /* FlexRAM */
+#define IMXRT117x_ITCM_SIZE       0x00080000U
+#define IMXRT117x_DTCM_BASE       0x20000000U /* FlexRAM */
+#define IMXRT117x_DTCM_SIZE       0x00080000U
+#define IMXRT117x_CODE_TCM_BASE   0x1ffe0000U
+#define IMXRT117x_CODE_TCM_SIZE   0x00020000U
+#define IMXRT117x_SYS_TCM_BASE    0x20000000U
+#define IMXRT117x_SYS_TCM_SIZE    0x00020000U
```
dragonmux, line 67: Due to how the TCMs work, we had intentionally omitted them from the memory maps before. Are they necessary for breakpoints to work properly, or can they remain omitted on account of faulting till enabled and not being readable in all situations when they are?

dragonmux (changes requested):
This is looking very good. We have a couple of queries, but once they've been addressed we'll be happy to get this merged.

src/target/imxrt.c
litui: Probably not necessary for now. Removed TCMs.

dragonmux (dismissed):
LGTM, merging once the build failing item is fixed. Thank you for the contribution!

dragonmux (approved):

PR #1915, opened 2024-08-30
Meson_build_option_for_trace_protocol_selection

sidprice:
## Detailed description

This PR introduces a new meson build option, "trace_protocol."

This option is used to select the protocol used by a platform for trace SWO.

## Closing issues

Platforms will need testing to verify the changes made.

### DIFF
--- a/src/platforms/native/meson.build
+++ b/src/platforms/native/meson.build
@@ -40,8 +40,8 @@ probe_native_args = [
 ]
 
 trace_protocol = get_option('trace_protocol')
-if trace_protocol == '1'
-	probe_native_args += [f'-DTRACESWO_PROTOCOL=@trace_protocol@']
+if trace_protocol in ['1', '3']
+	probe_native_args += [f'-DTRACESWO_PROTOCOL=1']
 	probe_native_dependencies = fixme_platform_stm32_traceswo
 else
 	error('Unsupported SWO mode requested')
--- a/src/platforms/native/meson.build
+++ b/src/platforms/native/meson.build
@@ -41,7 +41,7 @@ probe_native_args = [
 
 trace_protocol = get_option('trace_protocol')
 if trace_protocol == '1'
-probe_native_args += [f'-DTRACESWO_PROTOCOL=@trace_protocol@']
+	probe_native_args += [f'-DTRACESWO_PROTOCOL=@trace_protocol@']
 	probe_native_dependencies = fixme_platform_stm32_traceswo
 else
 	error('Unsupported SWO mode requested')
--- a/src/platforms/native/meson.build
+++ b/src/platforms/native/meson.build
@@ -39,6 +39,14 @@ probe_native_args = [
 	'-DBLACKMAGIC',
 ]
 
+trace_protocol = get_option('trace_protocol')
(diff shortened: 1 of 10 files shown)


src/platforms/native/platform.h
```diff
@@ -27,7 +27,6 @@
 #include "timing.h"
 #include "timing_stm32.h"
 
-#define TRACESWO_PROTOCOL 1U /* 1 = Manchester, 2 = NRZ / async */
```
dragonmux, line 30: This change needs to be matched with the introduction of this define into the Makefile for native.

src/platforms/stlinkv3/platform.h
```diff
@@ -81,7 +81,6 @@ extern bool debug_bmp;
 
 #define PLATFORM_HAS_TRACESWO 1
 #define NUM_TRACE_PACKETS     (16)
-#define TRACESWO_PROTOCOL     2 /* 1 = Manchester, 2 = NRZ / async */
```
dragonmux, line 84: Ditto the native platform.h change.

dragonmux (changes requested):
Save for a couple of Makefile changes that are needed to match the header changes up fully with the build system changes, this looks great. With those two issues taken care of (and the Makefile builds getting fixed as a result), we're happy to merge this.

dragonmux (approved):
Looks much better, however something is going on in native that needs a little more investigation.

src/platforms/native/meson.build
```diff
@@ -39,6 +39,14 @@ probe_native_args = [
 	'-DBLACKMAGIC',
 ]
 
+trace_protocol = get_option('trace_protocol')
+if trace_protocol == '1'
```
dragonmux, line 43: Due to the default being '3' and this intending to be "whatever the platform's default would be", this check needs to change slightly to `trace_protocol in ['1', '3']` to accommodate that and result in a net-zero change on the OOBE (Out Of Box Experience).

We suggest indenting the line below and changing the f-string to having a constant `=1` as a result.

sidprice: I agree the OOB experience is paramount!

Isn't the default trace protocol OOB better set in the crossfile?

dragonmux: No, as the cross-file is only a suggestion - there's absolutely nothing stopping a user from doing `meson setup build --cross-file=cross-file/arm-none-eabi.ini -Dprobe=native` and that needs to work, so all defaults need encoding directly in the build system. This is related to the same problem as a user reconfiguring an existing build directory to a different probe (eg, `meson setup build --reconfigure -Dprobe=stlink` having been configured for native) - this is a perfectly reasonable thing to want to do and needs to work as well but never re-reads any cross files to change configuration options.

sidprice (commented):
I couldn't get the default to work for native, then I just spotted, the messon_options file had '2' as the default! Will get all the patches tested later after I correct that :)

dragonmux:
Sounds good, apparently we missed that during review too so 😅 oops! Glad it's been found first though.

dragonmux (approved):
All looks good now, merging. Thank you for the contribution!

PR #1918, opened 2024-09-03
Feature: Cortex-M TrustZone support

### LINKED ISSUE #849: Cortex-M33 (Armv8-M) has more registers
As some core registers in V8 have a secure and non-secure incarnation, the announced register set needs to be updated.

dragonmux:
## Detailed description

In this PR we deal with some issues with core register I/O for ARM targets, and with TrustZone support between GDB and BMD for Cortex-M23 and Cortex-M33 targets. This is split into four main parts as described below.

The first part of this PR deals with some lints and code quality issues. The second deals with a residual ADIv6 support bug whereby because the register I/O code for both Cortex-M and Cortex-A/R support was directly writing the DP's SELECT register and presuming ADIv5, that was unintentionally deselecting the AP under ADIv6 and causing the reads and writes to go to an arbitrary point on the DP's debug resource bus.

The third part of this PR deals with some code organisation consistency problems, moving all the target.xml builder code for Cortex-M to the bottom of the file to match the Cortex-A/R and RISC-V support. This also sees corrections to the top of function comments and function nomenclature for the XML builders.

The fourth part implements support for GDB's secext block to read out the secure and non-secure stack pointers which are 4 new registers to handle. This involves generation of the XML for the feature block, handling the new registers in the mass read/write functions, and handling them in the individual read/write functions. Care must be taken, however, not to attempt reading the identity register on ARMv6-M cores as it's an invalid register there; and accessing the extra registers on ARMv7-M cores as their IDs are reserved values for DCRSR there.

## Closing issues

Fixes #849

### DIFF
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -549,7 +549,6 @@ bool cortexm_probe(adiv5_access_port_s *ap)
 				/* Go on and try to detect the target anyways */
 				break;
 			}
-			continue;
 		}
 	}
 
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -1212,36 +1212,38 @@ static size_t cortexm_build_target_fpu_description(char *const buffer, const siz
  * function will return the amount of bytes that would be necessary to create this string.
  *
  * The string it creates is XML-equivalent to the following:
- * "<?xml version=\"1.0\"?>"
- * "<!DOCTYPE target SYSTEM \"gdb-target.dtd\">"
- * "<target>"
- * "  <architecture>arm</architecture>"
- * "  <feature name=\"org.gnu.gdb.arm.m-profile\">"
- * "    <reg name=\"r0\" bitsize=\"32\"/>"
- * "    <reg name=\"r1\" bitsize=\"32\"/>"
- * "    <reg name=\"r2\" bitsize=\"32\"/>"
- * "    <reg name=\"r3\" bitsize=\"32\"/>"
- * "    <reg name=\"r4\" bitsize=\"32\"/>"
- * "    <reg name=\"r5\" bitsize=\"32\"/>"
- * "    <reg name=\"r6\" bitsize=\"32\"/>"
- * "    <reg name=\"r7\" bitsize=\"32\"/>"
- * "    <reg name=\"r8\" bitsize=\"32\"/>"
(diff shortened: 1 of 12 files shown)


src/target/cortexm.c
```diff
@@ -86,19 +86,24 @@ typedef struct cortexm_priv {
...
-	0x40, 0x41, 0x42, 0x43, 0x44, 0x45, 0x46, 0x47, /* s0-s7 */
-	0x48, 0x49, 0x4a, 0x4b, 0x4c, 0x4d, 0x4e, 0x4f, /* s8-s15 */
-	0x50, 0x51, 0x52, 0x53, 0x54, 0x55, 0x56, 0x57, /* s16-s23 */
-	0x58, 0x59, 0x5a, 0x5b, 0x5c, 0x5d, 0x5e, 0x5f, /* s24-s31 */
+	0x21U,                                                  /* fpscr */
+	0x40U, 0x41U, 0x42U, 0x43U, 0x44U, 0x45U, 0x46U, 0x47U, /* s0-s7 */
+	0x48U, 0x49U, 0x4aU, 0x4bU, 0x4cU, 0x4dU, 0x4eU, 0x4fU, /* s8-s15 */
+	0x50U, 0x51U, 0x52U, 0x53U, 0x54U, 0x55U, 0x56U, 0x57U, /* s16-s23 */
+	0x58U, 0x59U, 0x5aU, 0x5bU, 0x5cU, 0x5dU, 0x5eU, 0x5fU, /* s24-s31 */
 };
```
ALTracer, line 107: Is there a good reason to keep the core register indices 32-bit unsigned? Last three times I checked, these were only used as `uint8_t`. This will reduce `.rodata` flash occupation by (20+33)*(4-1)=53*3=159 bytes. Well, plus four TZ SP regs.

dragonmux: The way those registers are encoded and used here needs an overhaul and checking that nothing still tries to `sizeof()` these arrays to compute a value that goes back to GDB. If that constraint has now been fully broken, then yeah, we can shrink them to `uint8_t` arrays.

dragonmux: Just done this check and yeah, that works out - so we've added one more commit to the patch series to address this. Thank you!

esden (approved):
LGTM

PR #1922, opened 2024-09-04
Feature: RP2350 support

dragonmux:
## Detailed description

In this PR we introduce support for the ARM core part of the RP2350. This includes full support for the Flash controller (the QMI) and all modes of operation save for from POR state which is currently very slightly broken and needs more work.

Steps that still remain for this target: RISC-V support via the DMI bus MEM-AP which requires ADI DTM bindings for the RISC-V architecture implementation; testing that access all still works the same from the RISC-V cores; Flash stubs for both core types to make Flashing go faster.

Despite these missing components and the POR state drawback (which, admittedly, is a very minimal problem because it can only be encountered with connect-under-reset or when re-attaching to the part immediately after mass-erasing it), it feels prudent to get the ARM support into `main` so it can be more broadly tested and a follow-up done for the RISC-V support.


esden (approved):
LGTM! Thanks for all the hard work! Great to see this land!

PR #1924, opened 2024-09-07
Feature: RP2350 RISC-V core support

dragonmux:
## Detailed description

In this PR we implement support for RISC-V Debug over ADI (so dubbed the "RISC-V ADI DTM"), and enable support for the RP2350's RISC-V cores.

Also included, due to the limitations on the Hazard3 core's implementation of the abstract Access Register command, are fallbacks for reading CSRs via use of the program buffer and a scratch GPR.

BMD will automatically skip cores marked "unavailable" in the DM which streamlines the switching between the RISC-V and ARM cores for the RP2350. The resulting support is not especially fast as using ADI as a DTM is not particularly fast, but it is functional and reliable.

## Future work

The CSR access fallback path clobbers register a0 (x10) and to maintain debugging homeostasis then has to repeatedly restore the clobbered value. BMD should cache the register on the first read after a halt, and only restore it as it goes to resume the core, invalidating the cache. This should deliver a speed-up in CSR access of about a factor of 1.5 to 2 (1-2 accesses, depending on read vs write, of 4 being able to be eliminated).


esden (approved):
LGTM

PR #1925, opened 2024-09-08
Fix: Broken CMSIS-DAP SWD sequence

dragonmux:
## Detailed description

In this PR we correct an issue pointed out by both @litui on the MCULink adaptors, and by @asumagic on debugprobe firmware running on a RPi Pico. When using the logic from ARM for handling these transactions, they come out wrong (1 bit shifted due to all the in sequences capturing data one cycle too early).

We adjust for this with a new quirk flag when we detect this bugged behaviour during DP IDR readout, and post-processing the results with a corrective loop.

Tested working with both adaptors, allowing the debugprobe firmware to work flawlessly with BMD and the MCULink firmware to get a lot further with the MCXC041 before hitting a different issue (FAULT response handling during memory accesses setup). This latter issue will be dealt with in a separate PR.

### DIFF
--- a/src/platforms/hosted/dap.h
+++ b/src/platforms/hosted/dap.h
@@ -67,6 +67,7 @@ typedef enum dap_led_type {
 
 #define DAP_QUIRK_NO_JTAG_MUTLI_TAP          (1U << 0U)
 #define DAP_QUIRK_BAD_SWD_NO_RESP_DATA_PHASE (1U << 1U)
+#define DAP_QUIRK_BROKEN_SWD_SEQUENCE        (1U << 2U)
 
 extern uint8_t dap_caps;
 extern dap_cap_e dap_mode;
--- a/src/platforms/hosted/dap_command.c
+++ b/src/platforms/hosted/dap_command.c
@@ -372,14 +372,14 @@ bool perform_dap_swd_sequences(dap_swd_sequence_s *const sequences, const uint8_
 
 	DEBUG_PROBE("-> dap_swd_sequence (%u sequences)\n", sequence_count);
 	/* 47 is 2 + (5 * 9) where 9 is the max length of each sequence request */
-	uint8_t request[47] = {
+	uint8_t request[47U] = {
 		DAP_SWD_SEQUENCE,
 		sequence_count,
 	};
 	/* Encode the transfers into the buffer */
 	size_t offset = 2U;
 	size_t result_length = 0U;
-	for (uint8_t i = 0; i < sequence_count; ++i) {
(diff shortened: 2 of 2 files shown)


asumagic (commented):
Can confirm this gets the BMD app far enough for me to flash a M33 program on the RP2350 and debug it, using debugprobe-on-pico (on a stock pi pico board).

esden (approved):
LGTM

bugadani:
My understanding is that BMDA does not account for the turnaround bit when it queues up a read after the initial write. I can't find either of the references pointing out incorrect behaviour, as they are not linked to this PR, so I can't tell if the adaptors are supposed to discard one bit of turnaround for each direction change, or if BMDA introduced a bug with this patch. Without documentation stating otherwise, I would consider the reference CMSIS-DAP implementation correct here, this hack be removed and replaced by reading back and discarding the turnaround bit when emulating a complete read sequence.

bugadani:
Hmm I wrote my comment based on https://github.com/blackmagic-debug/blackmagic/blob/f1212c88a0f9248403f0ed0fbe77517650ac973f/src/platforms/hosted/dap_swd.c#L225 which 
 has been changed since.

bugadani:
For context, this came up because someone suggested I should work around the quirk detection by shifting the data I return for SWD_Sequence by one bit, similar to https://github.com/devanlai/dap42/commit/d67619bca30125713bf425efe1a48e8cfb3f2bc5 . In my opinion that would have made my firmware non-compliant, so I started looking for why it's needed, found that seemingly incorrect DPIDR read in this PR's revision, and wrote my comment.

I'll have to check whether my probe is still deemed as "buggy" in the next release candidate.

dragonmux:
> Hmm I wrote my comment based on
> 
> https://github.com/blackmagic-debug/blackmagic/blob/f1212c88a0f9248403f0ed0fbe77517650ac973f/src/platforms/hosted/dap_swd.c#L225
> which has been changed since.

If one actually changes that 3 to a 4, all probes will fail and give improper responses.. this is the "damned either way, but at least this way it works" approach that went on here. It's kind of maddening 🙃 - but yeah, on the face of it that looks like it should be wrong..

bugadani:
Something doesn't add up there, but I guess if the current state works I really shouldn't be lobbying for changing it. Thanks for your input :)

bugadani:
I don't mean to necromance this any longer, but I wanted to state that testing the latest build available in the github actions, neither my unmodified raspberry pi debugprobe, nor my alternative firmware written for it is deemed as "buggy" by BDMA, which is nice :)

PR #1927, opened 2024-09-09
Initial merge of Wi-Fi support code for ctxLink.

sidprice:
## Detailed description

This PR brings an early version of Wi-Fi support for ctxLink. Initially, it attempts to update the souce code to fit more closely with the style of BMF source code as definied by Clang-Tidy.

The built firmware, while offering initiall Wi-Fi connectivity and operation, is not full featured. The intent of this PR is to allow outstanding features in smaller PRs, thus reducing the burden on the author and the reviewer.

## Closing issues
None

### DIFF
--- a/src/platforms/ctxlink/WiFi_Server.c
+++ b/src/platforms/ctxlink/WiFi_Server.c
@@ -1,0 +1,1495 @@
+//
+//	File:	WiFi_Server.c
+//
+//	TCP/IP client/server for WBMP
+//
+
+#include "general.h"
+#include "morse.h"
+
+#include <libopencm3/stm32/f4/rcc.h>
+#include <libopencm3/cm3/scb.h>
+#include <libopencm3/cm3/nvic.h>
+#include <libopencm3/stm32/exti.h>
+#include <libopencm3/stm32/timer.h>
+#include <libopencm3/stm32/usart.h>
+#include <libopencm3/stm32/syscfg.h>
+#include <libopencm3/usb/usbd.h>
+
+#include <libopencm3/stm32/f4/adc.h>
+#include <libopencm3/stm32/f4/spi.h>
+
+#include "WiFi_Server.h"
+
+#include "platform.h"
+
+#include "ctxLink_mode_led.h"
+
+#include "winc1500_api.h"
+#include "winc1500_driver_api_helpers.h"
+#include "wf_socket.h"
+
+void traceSendData(void);
+
+#define GDBServerPort        2159
+#define UART_DebugServerPort 2160
+#define SWO_TraceServerPort  2161
+
+#define INPUT_BUFFER_SIZE 2048
+static unsigned char inputBuffer[INPUT_BUFFER_SIZE] = {0}; ///< The input buffer[ input buffer size]
+static volatile u_int32_t uiInputIndex = 0;                ///< Zero-based index of the input
+static volatile u_int32_t uiOutputIndex = 0;               ///< Zero-based index of the output
+static volatile u_int32_t uiBufferCount = 0;               ///< Number of buffers
+
+static unsigned char localBuffer[INPUT_BUFFER_SIZE] = {0}; ///< The local buffer[ input buffer size]
+
+static bool g_driverInitComplete = false;    ///< True to driver initialize complete
+static bool g_wifi_connected = false;        ///< True if WiFi connected
+static bool g_ipAddressAssigned = false;     ///< True if IP address assigned
+static bool g_dnsResolved = false;           ///< True if DNS resolved
+static bool g_gdbClientConnected = false;    ///< True if client connected
+static bool g_gdbServerIsRunning = false;    ///< True if server is running
+static bool g_newGDBClientConnected = false; ///< True if new client connected
+
+static SOCKET gdbServerSocket = SOCK_ERR_INVALID; ///< The main gdb server socket
+static SOCKET gdbClientSocket = SOCK_ERR_INVALID; ///< The gdb client socket
+
+#define UART_DEBUG_INPUT_BUFFER_SIZE 32
+static unsigned char localUartDebugBuffer[UART_DEBUG_INPUT_BUFFER_SIZE] = {0}; ///< The local buffer[ input buffer size]
+
+static SOCKET uartDebugServerSocket = SOCK_ERR_INVALID;
+static SOCKET uartDebugClientSocket = SOCK_ERR_INVALID;
+static bool g_uartDebugClientConnected = false;
+static bool g_userConfiguredUart = false;
+static bool g_uartDebugServerIsRunning = false;
+static bool g_newUartDebugClientconncted = false;
+
+static SOCKET swoTraceServerSocket = SOCK_ERR_INVALID;
+static SOCKET swoTraceClientSocket = SOCK_ERR_INVALID;
+static bool g_swoTraceClientConnected = false;
+static bool g_swoTraceServerIsRunning = false;
+static bool g_newSwoTraceClientconncted = false;
+
+#define SWO_TRACE_INPUT_BUFFER_SIZE 32
+static unsigned char localSwoTraceBuffer[SWO_TRACE_INPUT_BUFFER_SIZE] = {0}; ///< The local buffer[ input buffer size]
+
+#define WPS_LOCAL_TIMEOUT 30 // Timeout value in seconds
+
+//
+// Sign-on message for new UART data clients
+//
+static char uartClientSignon[] = "\r\nctxLink UART connection.\r\nPlease enter the UART setup as baud, bits, parity, "
+								 "stop.\r\ne.g. 38400,8,N,1\r\n\r\n";
+
+typedef enum WiFi_APP_STATES {
+	APP_STATE_WAIT_FOR_DRIVER_INIT,         ///< 0
+	APP_STATE_READ_MAC_ADDR,                ///< 1
+	APP_STATE_CONNECT_TO_WIFI,              ///< 2
+	APP_STATE_WAIT_WIFI_DISCONNECT_FOR_WPS, ///< 3
+	APP_STATE_WAIT_WIFI_DISCONNECT_FOR_HTTP,
+	APP_STATE_CONNECT_WPS,           ///< 5
+	APP_STATE_WAIT_WPS_EVENT,        ///< 6
+	APP_STATE_HTTP_PROVISION,        ///< 7
+	APP_STATE_WAIT_PROVISION_EVENT,  ///< 8
+	APP_STATE_WAIT_FOR_WIFI_CONNECT, ///< 9
+	APP_STATE_START_SERVER,          ///< 10
+	APP_STATE_WAIT_FOR_SERVER,       ///< 11
+	APP_STATE_ERROR,                 ///< 12
+	APP_STATE_CHECK_DEFAULT_CONN,    ///< 13
+	APP_STATE_SPIN                   ///< 14
+} APP_STATES;
+
+APP_STATES appState; ///< State of the application
+/*
+ * Define the send queue, this is used in the socket event callback
+ * to correctly process output and sync with ACKs
+ */
+#define SEND_QUEUE_SIZE        4
+#define SEND_QUEUE_BUFFER_SIZE 1024
+
+#define BUTTON_PRESS_WPS               2500 // Enter WPS mode if 2.5 Seconds > 2.5S < 5 Seconds
+#define BUTTON_PRESS_HTTP_PROVISIONING 5000
+#define BUTTON_PRESS_MODE_CANCEL       7500
+
+typedef struct tagSendQueueEntry {
+	unsigned char packet[SEND_QUEUE_BUFFER_SIZE]; ///< The packet[ send queue buffer size]
+	unsigned len;                                 ///< The length
+} SEND_QUEUE_ENTRY;
+
+SEND_QUEUE_ENTRY gdbSendQueue[SEND_QUEUE_SIZE] = {0}; ///< The send queue[ send queue size]
+unsigned volatile uiGDBSendQueueIn = 0;               ///< The send queue in
+unsigned volatile uiGDBSendQueueOut = 0;              ///< The send queue out
+unsigned volatile uiGDBSendQueueLength = 0;           ///< Length of the send queue
+void DoGDBSend(void);
+
+SEND_QUEUE_ENTRY uartDebugSendQueue[SEND_QUEUE_SIZE] = {0}; ///< The send queue[ send queue size]
+unsigned volatile uiUartDebugSendQueueIn = 0;               ///< The send queue in
+unsigned volatile uiUartDebugSendQueueOut = 0;              ///< The send queue out
+unsigned volatile uiUartDebugSendQueueLength = 0;           ///< Length of the send queue
+void DoUartDebugSend(void);
+
+SEND_QUEUE_ENTRY swoTraceSendQueue[SEND_QUEUE_SIZE] = {0}; ///< The send queue[ send queue size]
+unsigned volatile uiSwoTraceSendQueueIn = 0;               ///< The send queue in
+unsigned volatile uiSwoTraceSendQueueOut = 0;              ///< The send queue out
+unsigned volatile uiSwoTraceSendQueueLength = 0;           ///< Length of the send queue
+void DoSwoTraceSend(void);
+
+static bool pressActive = false; ///< True to press active
+bool wpsActive = false;          ///< True to wps active
+bool httpActive = false;         ///< True when HTTP provisioning is active
+bool waitingAccessPoint =
+	false; ///< Set true when http provisioning is started, this allows ignoring of that connection event
+
+void exti9_5_isr(void)
+{
+	//
+	// Is it EXTI9?
+	//
+	if (exti_get_flag_status(EXTI9) == EXTI9) {
+		// Reset the interrupt state
+		exti_reset_request(EXTI9);
+		m2m_EintHandler();
+	}
+}
+
+//
+// TIMER2 is used to provide a 1mS tick
+//
+#define TIMER2_COMPARE_VALUE 50
+
+//
+// Flag used to run the MODE LED state machine
+//
+static bool runModeLedTask = false; ///< True to run mode LED task
+
+static uint32_t pressTimer = 0; ///< The press timer
+
+void startPressTimer(void)
+{
+	timer_disable_irq(TIM2, TIM_DIER_CC1IE);
+	pressTimer = 0; // Reset the press timer
+	timer_enable_irq(TIM2, TIM_DIER_CC1IE);
+}
+
+uint32_t getPressTimer(void)
+{
+	timer_disable_irq(TIM2, TIM_DIER_CC1IE);
+	const uint32_t tmp = pressTimer;
+	timer_enable_irq(TIM2, TIM_DIER_CC1IE);
+	return tmp;
+}
+
+static uint32_t timeoutSeconds = 0;     // Used to perform a countdown in seconds for app tasks
+static uint32_t timeoutTickCounter = 0; // counts Timer1 ticks (1mS) for an active timeout
+static bool fTimeout = false;           // Asserted at end of timeout
+#define TIMEOUT_TICK_COUNT 1000
+
+void tim2_startSecondsTimeout(uint32_t uiSeconds)
+{
+	timer_disable_irq(TIM2, TIM_DIER_CC1IE);
+	timeoutTickCounter = TIMEOUT_TICK_COUNT;
+	timeoutSeconds = uiSeconds;
+	fTimeout = false;
+	timer_enable_irq(TIM2, TIM_DIER_CC1IE);
+}
+
+void tim2_cancelSecondsTimeout(void)
+{
+	timer_disable_irq(TIM2, TIM_DIER_CC1IE);
+	timeoutTickCounter = 0;
+	timeoutSeconds = 0;
+	fTimeout = false;
+	timer_enable_irq(TIM2, TIM_DIER_CC1IE);
+}
+
+bool tim2_isSecondsTimeout(void)
+{
+	return fTimeout;
+}
+
+void tim2_isr(void)
+{
+	if (timer_get_flag(TIM2, TIM_SR_CC1IF)) {
+		/* Clear compare interrupt flag. */
+		timer_clear_flag(TIM2, TIM_SR_CC1IF);
+
+		/*
+		 * Get current timer value to calculate next
+		 * compare register value.
+		 */
+		const uint16_t compare_time = timer_get_counter(TIM2);
+
+		/* Calculate and set the next compare value. */
+		const uint16_t frequency = TIMER2_COMPARE_VALUE;
+		const uint16_t new_time = compare_time + frequency;
+
+		timer_set_oc_value(TIM2, TIM_OC1, new_time);
+		//timer_set_counter (TIM2, 0);
+		m2m_TMR_ISR();
+		runModeLedTask = true;
+		pressTimer++;
+	}
+	/*
+		Do we have a seconds timeout setup
+	*/
+	if (timeoutSeconds != 0) {
+		if (--timeoutTickCounter == 0) {
+			// Decrement seconds
+			if (--timeoutSeconds == 0) {
+				fTimeout = true;
+			} else {
+				timeoutTickCounter = TIMEOUT_TICK_COUNT;
+			}
+		}
+	}
+}
+
+void timer_init(void)
+{
+	/* Enable TIM2 clock. */
+	rcc_periph_clock_enable(RCC_TIM2);
+
+	/* Enable TIM2 interrupt. */
+	nvic_enable_irq(NVIC_TIM2_IRQ);
+
+	/* Reset TIM2 peripheral to defaults. */
+	rcc_periph_reset_pulse(RST_TIM2);
+
+	/* Timer global mode:
+	 * - No divider
+	 * - Alignment edge
+	 * - Direction up
+	 * (These are actually default values after reset above, so this call
+	 * is strictly unnecessary, but demos the api for alternative settings)
+	 */
+	timer_set_mode(TIM2, TIM_CR1_CKD_CK_INT, TIM_CR1_CMS_EDGE, TIM_CR1_DIR_UP);
+
+	/*
+	 * Please take note that the clock source for STM32 timers
+	 * might not be the raw APB1/APB2 clocks.  In various conditions they
+	 * are doubled.  See the Reference Manual for full details!
+	 * In our case, TIM2 on APB1 is running at double frequency, so this
+	 * sets the prescaler to have the timer run at 50kHz
+	 */
+	timer_set_prescaler(TIM2, ((rcc_apb1_frequency * 2) / 50000));
+
+	/* Disable preload. */
+	timer_disable_preload(TIM2);
+	timer_continuous_mode(TIM2);
+
+	/* count full range, as we'll update compare value continuously */
+	timer_set_period(TIM2, 65535);
+
+	/* Set the initual output compare value for OC1. */
+	timer_set_oc_value(TIM2, TIM_OC1, TIMER2_COMPARE_VALUE);
+
+	/* Counter enable. */
+	timer_enable_counter(TIM2);
+
+	/* Enable Channel 1 compare interrupt to recalculate compare values */
+	timer_enable_irq(TIM2, TIM_DIER_CC1IE);
+}
+
+static enum WiFi_TCPServerStates {
+	SM_HOME = 0,  ///< State of the TCP server
+	SM_LISTENING, ///< An enum constant representing the sm listening option
+	SM_CLOSING,   ///< An enum constant representing the sm closing option
+	SM_IDLE,      ///< An enum constant representing the sm idle option
+} GDB_TCPServerState = SM_IDLE, UART_DEBUG_TCPServerState = SM_IDLE, SWO_TRACE_TCPServerState = SM_IDLE;
+
+struct sockaddr_in gdb_addr = {0};
+
+void doWiFiDisconnect(void)
+{
+	//
+	// Clients
+	//
+	if (gdbClientSocket != SOCK_ERR_INVALID) {
+		close(gdbClientSocket);
+		g_gdbClientConnected = false;
+		gdbClientSocket = SOCK_ERR_INVALID;
+	}
+	if (uartDebugClientSocket != SOCK_ERR_INVALID) {
+		close(uartDebugClientSocket);
+		g_uartDebugClientConnected = false;
+		uartDebugClientSocket = SOCK_ERR_INVALID;
+	}
+	if (swoTraceClientSocket != SOCK_ERR_INVALID) {
+		close(swoTraceClientSocket);
+		g_swoTraceClientConnected = false;
+		swoTraceClientSocket = SOCK_ERR_INVALID;
+	}
+	//
+	// Servers
+	//
+	if (gdbServerSocket != SOCK_ERR_INVALID) {
+		GDB_TCPServerState = SM_IDLE;
+		g_gdbServerIsRunning = false;
+		close(gdbServerSocket);
+		gdbServerSocket = SOCK_ERR_INVALID;
+	}
+	if (uartDebugServerSocket != SOCK_ERR_INVALID) {
+		UART_DEBUG_TCPServerState = SM_IDLE;
+		g_uartDebugServerIsRunning = false;
+		close(uartDebugServerSocket);
+		uartDebugServerSocket = SOCK_ERR_INVALID;
+	}
+	if (swoTraceServerSocket != SOCK_ERR_INVALID) {
+		SWO_TRACE_TCPServerState = SM_IDLE;
+		g_swoTraceServerIsRunning = false;
+		close(swoTraceServerSocket);
+		swoTraceServerSocket = SOCK_ERR_INVALID;
+	}
+}
+
+void GDB_TCPServer(void)
+{
+	switch (GDB_TCPServerState) {
+	case SM_IDLE:
+		break; // Startup and testing do nothing state
+	case SM_HOME:
+		//
+		// Allocate a socket for this server to listen and accept connections on
+		//
+		gdbServerSocket = socket(AF_INET, SOCK_STREAM, 0);
+		if (gdbServerSocket < SOCK_ERR_NO_ERROR) {
+			return;
+		}
+		//
+		// Bind the socket
+		//
+		gdb_addr.sin_addr.s_addr = 0;
+		gdb_addr.sin_family = AF_INET;
+		gdb_addr.sin_port = _htons(GDBServerPort);
+		int8_t result = bind(gdbServerSocket, (struct sockaddr *)&gdb_addr, sizeof(gdb_addr));
+		if (result != SOCK_ERR_NO_ERROR) {
+			return;
+		}
+		GDB_TCPServerState = SM_LISTENING;
+		break;
+
+	case SM_LISTENING:
+		//
+		// No need to perform any flush.
+		// TCP data in TX FIFO will automatically transmit itself after it accumulates for a while.
+		// If you want to decrease latency (at the expense of wasting network bandwidth on TCP overhead),
+		// perform and explicit flush via the TCPFlush() API.
+		break;
+
+	case SM_CLOSING:
+		// Close the socket connection.
+		close(gdbServerSocket);
+		gdbServerSocket = SOCK_ERR_INVALID;
+		GDB_TCPServerState = SM_IDLE;
+		break;
+	}
+}
+
+struct sockaddr_in uart_debug_addr = {0};
+struct sockaddr_in swo_trace_addr = {0};
+
+void DATA_TCPServer(void)
+{
+	//
+	// UART Server
+	//
+	switch (UART_DEBUG_TCPServerState) {
+	case SM_IDLE: {
+		break; // Startup and testing do nothing state
+	}
+	case SM_HOME: {
+		//
+		// Allocate a socket for this server to listen and accept connections on
+		//
+		uartDebugServerSocket = socket(AF_INET, SOCK_STREAM, 0);
+		if (uartDebugServerSocket < SOCK_ERR_NO_ERROR) {
+			return;
+		}
+		//
+		// Bind the socket
+		//
+		uart_debug_addr.sin_addr.s_addr = 0;
+		uart_debug_addr.sin_family = AF_INET;
+		uart_debug_addr.sin_port = _htons(UART_DebugServerPort);
+		int8_t result = bind(uartDebugServerSocket, (struct sockaddr *)&uart_debug_addr, sizeof(uart_debug_addr));
+		if (result != SOCK_ERR_NO_ERROR) {
+			return;
+		}
+		UART_DEBUG_TCPServerState = SM_LISTENING;
+		break;
+	}
+	case SM_LISTENING: {
+		//
+		// No need to perform any flush.
+		// TCP data in TX FIFO will automatically transmit itself after it accumulates for a while.
+		// If you want to decrease latency (at the expense of wasting network bandwidth on TCP overhead),
+		// perform and explicit flush via the TCPFlush() API.
+		break;
+	}
+	case SM_CLOSING: {
+		// Close the socket connection.
+		close(uartDebugServerSocket);
+		uartDebugServerSocket = SOCK_ERR_INVALID;
+		UART_DEBUG_TCPServerState = SM_IDLE;
+	} break;
+	}
+	//
+	// SWO Trace Data server
+	//
+	switch (SWO_TRACE_TCPServerState) {
+	case SM_IDLE: {
+		break; // Startup and testing do nothing state
+	}
+	case SM_HOME: {
+		//
+		// Allocate a socket for this server to listen and accept connections on
+		//
+		swoTraceServerSocket = socket(AF_INET, SOCK_STREAM, 0);
+		if (swoTraceServerSocket < SOCK_ERR_NO_ERROR) {
+			return;
+		}
+		//
+		// Bind the socket
+		//
+		swo_trace_addr.sin_addr.s_addr = 0;
+		swo_trace_addr.sin_family = AF_INET;
+		swo_trace_addr.sin_port = _htons(SWO_TraceServerPort);
+		int8_t result = bind(swoTraceServerSocket, (struct sockaddr *)&swo_trace_addr, sizeof(swo_trace_addr));
+		if (result != SOCK_ERR_NO_ERROR) {
+			return;
+		}
+		SWO_TRACE_TCPServerState = SM_LISTENING;
+		break;
+	}
+	case SM_LISTENING: {
+		//
+		// No need to perform any flush.
+		// TCP data in TX FIFO will automatically transmit itself after it accumulates for a while.
+		// If you want to decrease latency (at the expense of wasting network bandwidth on TCP overhead),
+		// perform an explicit flush via the TCPFlush() API.
+		break;
+	}
+	case SM_CLOSING: {
+		// Close the socket connection.
+		close(swoTraceServerSocket);
+		swoTraceServerSocket = SOCK_ERR_INVALID;
+		SWO_TRACE_TCPServerState = SM_IDLE;
+		break;
+	}
+	}
+}
+
+void WiFi_setupSwoTraceServer(void)
+{
+	//
+	// Is there a UART client connected?
+	//
+	if (g_uartDebugClientConnected) {
+		// Close the connection to the client
+		close(uartDebugClientSocket);
+		uartDebugClientSocket = SOCK_ERR_INVALID; // Mark socket invalid
+		g_uartDebugClientConnected = false;       // No longer connected
+		g_userConfiguredUart = false;
+	}
+	//
+	// If the UART server is up, close it timeDown
+	//
+	if (uartDebugServerSocket != SOCK_ERR_INVALID) {
+		close(uartDebugServerSocket);
+		uartDebugServerSocket = SOCK_ERR_INVALID;
+	}
+	//
+	// Set up the SWO Trace Server
+	//
+	SWO_TRACE_TCPServerState = SM_HOME;
+}
+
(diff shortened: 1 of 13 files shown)


dragonmux (changes requested):
Done an initial review, there are still several more files we need to look at but this gives some easy first items to deal with.

src/platforms/ctxlink/platform.c
```diff
@@ -67,26 +71,79 @@ static void adc_init(void)
 		__asm__("nop");
 }
 
+//
+// Use the passed string to configure the USB UART
+//
+// e.g. 38400,8,N,1
+bool platform_configure_uart(char *configurationString)
+{
+	bool fResult;
```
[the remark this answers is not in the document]
sidprice: Changed as requested

src/platforms/ctxlink/platform.c
```diff
@@ -67,26 +71,79 @@ static void adc_init(void)
...
+	uint32_t baudRate;
+	uint32_t bits;
+	uint32_t stopBits;
+	char parity;
+	uint32_t count;
```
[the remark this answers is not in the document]
sidprice: Changed as requested

src/platforms/ctxlink/platform.c
```diff
@@ -67,26 +71,79 @@ static void adc_init(void)
...
+	if (strlen(configurationString) > 5) {
+		count = sscanf(configurationString, "%ld,%ld,%c,%ld", &baudRate, &bits, &parity, &stopBits);
```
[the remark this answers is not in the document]
sidprice: Changed as requested

src/platforms/ctxlink/platform.c
```diff
@@ -67,26 +71,79 @@ static void adc_init(void)
...
+
 void platform_adc_read(void)
 {
 	adc_set_regular_sequence(ADC1, 1, &adc_channels[CTXLINK_ADC_BATTERY]);
 	adc_start_conversion_regular(ADC1);
 	/* Wait for end of conversion. */
 	while (!adc_eoc(ADC1))
 		continue;
-	input_voltages[CTXLINK_ADC_BATTERY] = adc_read_regular(ADC1);
+	input_voltages[CTXLINK_ADC_BATTERY] = ((adc_read_regular(ADC1) * 16) / 10);
```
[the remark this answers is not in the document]
sidprice: Changed as requested

src/platforms/ctxlink/platform.c
```diff
@@ -191,6 +248,25 @@ void platform_init(void)
...
+	wifi_init(); //	Setup the Wifi channel
+#ifdef WINC_1500_FIRMWARE_UPDATE
+	//
+	// ONLY for firmware update
+	//
+	// Perform WINC1500 reset sequence
+	//
+	m2mStub_PinSet_CE(M2M_WIFI_PIN_LOW);
+	m2mStub_PinSet_RESET(M2M_WIFI_PIN_LOW);
+	DelayMs(100);
```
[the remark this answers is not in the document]
sidprice: The delay routine is the one used in the WINC1500 driver source and this small piece of code was taken from their app note explaining how to update the WINC1500 firmware. CHanged to use the platform_delay, since the file it is in is outside of the WINC1500 driver.

src/platforms/ctxlink/platform.c
```diff
@@ -191,6 +248,25 @@ void platform_init(void)
...
+	m2mStub_PinSet_CE(M2M_WIFI_PIN_HIGH);
+	DelayMs(10);
+	m2mStub_PinSet_RESET(M2M_WIFI_PIN_HIGH);
+	DelayMs(10);
+	while (1) {
+		;
+	}
```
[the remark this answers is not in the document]
sidprice: Changed as requested

src/platforms/ctxlink/platform.c
```diff
@@ -213,7 +319,17 @@ bool platform_nrst_get_val(void)
 
 const char *platform_target_voltage(void)
 {
-	return "Unknown";
+	static char target[64] = {0};
+	uint32_t val = platform_target_voltage_sense();
+	target[0] = '0' + val / 1000U;
+	target[1] = '.';
+	val = val % 1000;
```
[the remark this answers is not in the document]
sidprice: Changed as requested

src/platforms/ctxlink/platform.c
```diff
@@ -245,6 +361,74 @@ void platform_target_clk_output_enable(bool enable)
...
+//  Note the battery voltage is divided by 2 with resistor divider
+//
+// No battery voltage 1 == 2.0v
+// No battery voltage 2 == 4.268v
+// Battery present (report voltage) < 4.268v
+// Low batter voltage == 3.6v
+//
+#define BATTERY_VOLTAGE_1 2000
+#define BATTERY_VOLTAGE_2 4268
+#define BATTERY_LOW       3600
```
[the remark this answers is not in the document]
sidprice: Changed as requested

src/platforms/ctxlink/platform.h
```diff
@@ -98,6 +144,20 @@
...
+#define SWO_UART        USART6
+#define SWO_UART_CR1    USART6_CR1
+#define SWO_UART_DR     USART6_DR
+#define SWO_UART_CLK    RCC_USART6
+#define SWO_UART_PORT   GPIOC
+#define SWO_UART_RX_PIN GPIO7
+#define SWO_UART_ISR    usart6_isr
+#define SWO_UART_IRQ    NVIC_USART6_IRQ
+
+#define TRACESWO_PROTOCOL 2 /* 1 = Manchester, 2 = NRZ / async */
```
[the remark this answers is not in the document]
sidprice: Removed as requested

src/platforms/ctxlink/platform.h
```diff
@@ -234,4 +296,10 @@
 		gpio_set_val(LED_PORT, LED_ERROR, state); \
 	}
 
+void platform_tasks(void);
+const char *platform_battery_voltage(void);
+bool platform_check_battery_voltage(void);
+bool platform_has_network_client(uint8_t *lpBuf_rx, uint8_t *lpBuf_rx_in, uint8_t *lpBuf_rx_out, unsigned fifoSize);
+bool platform_configure_uart(char *configurationString);
```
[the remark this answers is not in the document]
sidprice: Removed "platform_has_network_client", an unused function.
Renamed the parameter for "platform_configure_uart"

src/platforms/ctxlink/platform.c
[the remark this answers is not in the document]
dragonmux: Ahhh, okay, that makes sense of that. Thank you!

src/platforms/ctxlink/winc1500_driver_stub_api.c
```diff
@@ -0,0 +1,271 @@
...
+// ---------------------
+//    - The WINC1500 state machines require a timer with one millisecond resolution.
+//    - The timer is a 32-bit counter that counts up starting at 0x00000000 and
+//      wraps back to 0 after reaching 0xffffffff.
+//==============================================================================
+
+// A TMR peripheral can be set to interrupt every 1ms. The timer ISR increments
+// a global counter.
+
+static volatile uint32_t one_ms_counter = 0;
```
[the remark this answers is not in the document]
sidprice: Changed as requested, also simplified the access function "m2mStub_GetOneMsTimer"

src/platforms/ctxlink/winc1500_driver_stub_api.c
```diff
@@ -0,0 +1,271 @@
...
+// SPI Stub Functions:
+// ---------------------
+//    - The Host MCU will communicate to the WINC1500 via the SPI interface.
+//==============================================================================
+
+/* NOLINTNEXTLINE(readability-identifier-naming) */
+void m2mStub_SpiTxRx(uint8_t *p_txBuf, uint16_t txLen, uint8_t *p_rxBuf, uint16_t rxLen)
+{
+	uint16_t byteCount;
+	uint16_t i;
```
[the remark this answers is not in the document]
sidprice: Changed as requested.

src/platforms/ctxlink/winc1500_driver_stub_api.c
```diff
@@ -0,0 +1,271 @@
...
+	// The following are intermediate buffers used to ensure
+	// the TX and RX message sizes handked by the transfer routine
+	// are the same size.
+	//
+	uint8_t outputBuffer[256] = {0};
+	uint8_t inputBuffer[256] = {0};
+	/*
+     *	total number of byte to clock is whichever is larger, txLen or rxLen
+     */
+	byteCount = (txLen >= rxLen) ? txLen : rxLen;
```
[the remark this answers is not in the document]
sidprice: Parens removed and both input length variables checked for potential overrun.

src/platforms/ctxlink/winc1500_driver_stub_api.c
```diff
@@ -0,0 +1,271 @@
...
+	//
+	for (i = 0; i < byteCount; i++) {
+		inputBuffer[i] = spi_xfer(WINC1500_SPI_CHANNEL, outputBuffer[i]);
+	}
+	//
+	// If we ewxpected to receive bytes copy them to the rx buffer
+	//
+	if (rxLen > 0) {
+		memcpy(p_rxBuf, &inputBuffer[0], rxLen);
+	}
```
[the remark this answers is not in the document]
sidprice: Changed as requested

src/platforms/ctxlink/winc1500_driver_stub_api.c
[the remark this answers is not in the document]
sidprice: Indeed the function names are dictated by the WINC1500 driver. Other issues may be addressed in a followup review/PR

src/platforms/ctxlink/winc1500_driver_stub_api.c
```diff
@@ -0,0 +1,271 @@
...
+// ---------------------
+//    These are callback functions that the WINC1500 host driver calls to notify
+//    the application of events. There are four categories of events:
+//      - Wi-Fi events
+//      - Socket events
+//      - OTA (Over-The-Air) update events
+//      - Error Events
+//==============================================================================
+
+volatile tpfAppWifiCb app_wifi_cb = NULL;
```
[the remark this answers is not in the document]
sidprice: I see no reason ... removed.

src/platforms/ctxlink/platform.c
```diff
@@ -213,7 +330,17 @@ bool platform_nrst_get_val(void)
...
+	static char target[64] = {0};
+	uint32_t val = platform_target_voltage_sense();
+	target[0] = '0' + val / 1000U;
+	target[1] = '.';
+	val = val % 1000U;
+	target[2] = '0' + val / 100U;
+	val = val % 100U;
+	target[3] = '0' + val / 10U;
+	target[4] = 'V';
+	strcat(target, platform_battery_voltage());
```
dragonmux, line 342: Because of how the battery voltage part works, this may overflow the storage for this string. Please use `strncat()` for safety.

src/platforms/ctxlink/winc1500_driver_stub_api.c
```diff
@@ -0,0 +1,265 @@
...
+{
+	if (app_wifi_cb)
+		app_wifi_cb(eventCode, p_eventData);
+	else
+		DEBUG_WARN("STUB_WIFI_EVENT[%d]: Wi-Fi event handler not registered!\r\n", eventCode);
+}
+
+//             --------------- * end of wifi event block * ---------------
+
+volatile tpfAppSocketCb app_socket_cb = NULL;
```
dragonmux, line 198: Same comment as with the `app_wifi_cb` originally being volatile.

All global variable definitions like these should go at the top of the file please.

src/platforms/ctxlink/platform.c
```diff
@@ -67,26 +87,75 @@ static void adc_init(void)
...
+				break;
+			}
+
+			case 'E': {
+				parityValue = USART_PARITY_EVEN;
+				break;
+			}
+			}
+			usart_set_parity(USBUSART, parityValue);
+		}
```
dragonmux, line 126: This should probably have an `else` block that follows it, allowing this function to `return false;` when the incorrect number of things get parsed by sscanf(). It looks like that was the intent with the original logic save for `fResult` not being initialised when defined.

src/platforms/ctxlink/ctxLink_traceswo.c
```diff
@@ -0,0 +1,210 @@
...
+					output_buffer = 0;
+				}
+			}
+			send_swo_trace_data(&swo_data[0], dataCount);
+			__atomic_fetch_sub(&buffer_size, dataCount, __ATOMIC_RELAXED);
+		}
+	}
+}
+
+void _trace_buf_drain(usbd_device *dev, uint8_t ep)
```
dragonmux, line 87: Defining `_`-prefixed functions in user code invoke UB, please don't do that.

src/platforms/ctxlink/ctxLink_traceswo.c
```diff
@@ -0,0 +1,210 @@
...
+#include <libopencmsis/core_cm3.h>
+#include <libopencm3/cm3/nvic.h>
+#include <libopencm3/stm32/timer.h>
+#include <libopencm3/stm32/rcc.h>
+#include <libopencm3/stm32/usart.h>
+
+/* For speed this is set to the USB transfer size */
+#define FULL_SWO_PACKET (64)
+/* Default line rate....used as default for a request without baudrate */
+#define DEFAULTSPEED (2250000)
```
dragonmux, line 48: This is lifted into swo.c in the common tree, allowing it to be dropped here. The baud rate delivered to `swo_uart_init()` (the replacement of `traceswo_init()` below) is guaranteed to be a reasonable value.

src/platforms/ctxlink/ctxLink_traceswo.c
```diff
@@ -0,0 +1,210 @@
...
+
+#define BUFFER_SIZE 1024
+
+static volatile uint32_t input_buffer = 0;  // input buffer index
+static volatile uint32_t output_buffer = 0; // output buffer index
+volatile uint32_t buffer_size = 0;          // Number of bytes in the buffer
```
dragonmux, line 54: Rather than marking this as `volatile` and manually doing atomic operations on it, please use `_Atomic` instead. If you still wish to use `__ATOMIC_RELAXED` mode, then you can use [`atomic_load_explicit`](https://en.cppreference.com/w/c/atomic/atomic_load), [`atomic_fetch_sub`](https://en.cppreference.com/w/c/atomic/atomic_fetch_sub) and [`atomic_fetch_add`](https://en.cppreference.com/w/c/atomic/atomic_fetch_add) from stdatomic.h with `memory_order_relaxed` as the final parameter.

If you don't care about the memory ordering, then just reading the variable (`const uint32_t data_count = buffer_size;`) and then using `+=` and `-=` on the atomic will do the correct thing. That's the C11 standardised/sanctioned way of doing the atomic operations below.

src/platforms/ctxlink/ctxLink_traceswo.c
```diff
@@ -0,0 +1,210 @@
...
+		__atomic_load(&buffer_size, &dataCount, __ATOMIC_RELAXED);
+		if (dataCount >= FULL_SWO_PACKET) {
+			//
+			// Copy the data
+			//
+			for (uint32_t i = 0; i < dataCount; i++) {
+				swo_data[i] = trace_rx_buf[output_buffer++];
+				if (output_buffer >= BUFFER_SIZE) {
+					output_buffer = 0;
+				}
```
dragonmux, line 79: The extra braces should be dropped from this if statement please.

src/platforms/ctxlink/ctxLink_traceswo.c
```diff
@@ -0,0 +1,210 @@
...
+
+void _trace_buf_drain(usbd_device *dev, uint8_t ep)
+{
+	uint32_t outCount;
+	uint8_t *bufferPointer, *bufferStart;
+
+	__atomic_load(&buffer_size, &outCount, __ATOMIC_RELAXED);
+	if (outCount == 0) {
+		return;
+	}
```
dragonmux, line 95: Same comment for this and the other single-statement if blocks below about dropping the extra braces.

src/platforms/ctxlink/ctxLink_mode_led.c
```diff
@@ -0,0 +1,156 @@
...
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#include "general.h"
+#include "platform.h"
+#include "ctxLink_mode_led.h"
+
+mode_led_modes_e led_mode = mode_led_idle; // Initial state of the mode led
+mode_led_task_states_e mode_task_state = mode_led_idle_state;
+
+u_int32_t led_mode_timeout = 0;       // Used to time the on/off state of the led
```
dragonmux, line 14: Unsure where this type is getting defined, but the C11 one is `uint32_t` please. `U` suffixes on the values being assigned would suppress a signed-unsigned conversion.

src/platforms/ctxlink/ctxLink_mode_led.c
```diff
@@ -0,0 +1,156 @@
...
+	}
+	default: {
+		break;
+	}
+	}
+}
+
+bool mode_check_timeout()
+{
+	bool fResult = false;
```
dragonmux, line 59: Please write this return variable out - the if statement can just `return true;` in its body, with a following `return false;` below.

src/platforms/ctxlink/ctxLink_mode_led.c
```diff
@@ -0,0 +1,156 @@
...
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+static mode_led_modes_e saved_mode = mode_led_invalid;
+
+void mode_led_task(void)
+{
+	//
+	// Use this periodic task to check the battery voltage
+	//
+	if (platform_check_battery_voltage() == false) {
```
dragonmux, line 81: If this function already returns a boolean, there's no need to do a conversion, just `if (!platform_check_battery_voltage())` would be absolutely fine. Same for the explicit check against `true` below, but dropping the `!` (ie `if (mode_check_timeout())`).

src/platforms/ctxlink/WiFi_Server.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
```
dragonmux, line 1: This gets the same include guards and copyright notice comment as that other header please.

src/platforms/ctxlink/WiFi_Server.h
```diff
@@ -0,0 +1,27 @@
...
+void data_tcp_server(void);
+bool is_uart_client_connected(void);
+void send_uart_data(uint8_t *lpBuffer, uint8_t length);
+
+bool swo_trace_server_active(void);
+void wifi_setup_swo_trace_server(void);
+bool is_swo_trace_client_connected(void);
+void send_swo_trace_data(uint8_t *lpBuffer, uint8_t length);
+
+void wifi_gdb_putchar(unsigned char ch, int flush);
```
dragonmux, line 21: We believe that `flush` should be `bool` here - you'll need to `#include <stdbool.h>` at the top of the file for that but it should otherwise be good to go.

src/platforms/ctxlink/WiFi_Server.h
```diff
@@ -0,0 +1,27 @@
...
+#endif
+void app_initialize(void);
+void app_task(void);
+//
+void gdb_tcp_server(void);
+bool is_gdb_client_connected(void);
+
+void data_tcp_server(void);
+bool is_uart_client_connected(void);
+void send_uart_data(uint8_t *lpBuffer, uint8_t length);
```
dragonmux, line 14: `buffer` instead of `lpBuffer` please. Same for `send_swo_trace_data()`.

dragonmux (commented):
With this review, we have only the WiFi server implementation to review and we think we've covered most of the rest. Some of the comments likely apply across multiple files, so if you see, eg, `u_int32_t` in another file to the one commented in, then our `uint32_t` comment applies to it as well.

src/platforms/ctxlink/winc1500_driver_stub_api.c
sidprice: Changed as requested.

src/platforms/ctxlink/platform.c
```diff
@@ -245,6 +373,58 @@ void platform_target_clk_output_enable(bool enable)
...
+		battery_present = true;
+		//
+		// Is the voltage good?
+		//
+		/* NOLINTNEXTLINE(bugprone-branch-clone) */
+		if (input_voltages[CTXLINK_ADC_BATTERY] <= BATTERY_LOW) {
+			last_battery_state = fResult = false;
+		} else {
+			last_battery_state = fResult = true;
+		}
```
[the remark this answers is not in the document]
sidprice: Refactored the function.

src/platforms/ctxlink/platform.c
sidprice: Changed to use strncat.

src/platforms/ctxlink/platform.c
sidprice: Refactored the function to return false on failure.

src/platforms/ctxlink/ctxLink_traceswo.c
sidprice: Changes as requested.

src/platforms/ctxlink/ctxLink_traceswo.c
sidprice: This file is likely to have major changes when the Wi-Fi SWO feature is integrated with the new SWO structure in the base platform. This function is currently unused, therefore exckude from compilation, but left here for reference.

src/platforms/ctxlink/ctxLink_traceswo.c
sidprice: Changed variable to "_Atomic" and adjust usage sites

src/platforms/ctxlink/ctxLink_traceswo.c
sidprice: Remove the definition.
Again note, this file is currently not being compiled into the project, and will be under major review when Wi-Fi SWO is implemented

src/platforms/ctxlink/ctxLink_mode_led.c
sidprice: Changes as requested.

src/platforms/ctxlink/ctxLink_mode_led.c
sidprice: Changes as requested.

src/platforms/ctxlink/ctxLink_mode_led.c
sidprice: Like you I don't know where this is defined, I assumed it was related to the WINC1500 driver source, but a search there does not reveal it.
Have searched my source and replaced with uint32_t.

src/platforms/ctxlink/WiFi_Server.h
sidprice: This function is based up the common/stm32/gdb_if.c file, which as:
```void gdb_if_putchar(const char c, const int flush)```

File: src/platforms/ctxlink/WiFi_Server.h (hunk 29 above)
Comment (sidprice) [reply src/platforms/ctxlink/WiFi_Server.h:14]: Changed as requested.

File: src/platforms/ctxlink/platform.c [hunk 31]
```diff
@@ -245,6 +374,50 @@ void platform_target_clk_output_enable(bool enable)
...
+}
+
+bool platform_check_battery_voltage(void)
+{
+	platform_adc_read();
+	//
+	// Is battery connected?
+	//
+	if ((input_voltages[CTXLINK_ADC_BATTERY] <= BATTERY_VOLTAGE_1) ||
+		(input_voltages[CTXLINK_ADC_BATTERY] >= BATTERY_VOLTAGE_2)) {
```
dragonmux, line 408: The extra parens around the two conditions are superfluous and can be dropped.

src/platforms/ctxlink/winc1500_driver_stub_api.c
dragonmux, line 1: This file does still have globals being defined all down through the body rather than all at the top of file, however that's the only thing we see now that should ideally be addressed in this PR before the other things get addressed in a follow-up.

src/platforms/ctxlink/ctxLink_traceswo.c
dragonmux: That's noted 👍🏼 couldn't remember if this was being compiled in or not so reviewed as if it was.

src/platforms/ctxlink/ctxLink_mode_led.c
dragonmux: Think you got your `!` and no `!` the wrong way around - checking if a function returns false is `!func()` and if it returns true just `func()`.

src/platforms/ctxlink/platform.c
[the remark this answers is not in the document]
sidprice: I missed that ... fixed

src/platforms/ctxlink/winc1500_driver_stub_api.c
sidprice: I think I got the all static/global variables moved.

src/platforms/ctxlink/platform.c
sidprice: Fixed

dragonmux (changes requested):
In the interests of not being too repetitive in the review comments, we've only highlighted the first couple of cases of extra braces or truthy bool checks in WiFi_Server.c. Other than that one files though, this is all looking really good and we think this is probably one review cycle away from being merged.

.gitignore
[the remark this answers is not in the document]
sidprice: Dropped both of those commits, the Ozone one was from when I was ysing the Onone debugger.The folder had some configurattion data.

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1517 @@
...
+//
+static char uart_client_signon[] = "\r\nctxLink UART connection.\r\nPlease enter the UART setup as baud, bits, parity, "
+								   "stop.\r\ne.g. 38400,8,N,1\r\n\r\n";
+
+typedef enum wi_fi_app_states {
+	app_state_wait_for_driver_init,         ///< 0
+	app_state_read_mac_address,             ///< 1
+	app_state_connect_to_wifi,              ///< 2
+	app_state_wait_wifi_disconnect_for_wps, ///< 3
+	app_state_wait_wifi_disconnect_for_http,
```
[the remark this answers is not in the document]
sidprice: Added missing constant value.

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1517 @@
...
+#define SEND_QUEUE_SIZE        4
+#define SEND_QUEUE_BUFFER_SIZE 1024
+
+#define BUTTON_PRESS_WPS               2500 // Enter WPS mode if 2.5 Seconds > 2.5S < 5 Seconds
+#define BUTTON_PRESS_HTTP_PROVISIONING 5000
+#define BUTTON_PRESS_MODE_CANCEL       7500
+
+typedef struct {
+	unsigned char packet[SEND_QUEUE_BUFFER_SIZE]; ///< The packet[ send queue buffer size]
+	unsigned len;                                 ///< The length
```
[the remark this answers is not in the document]
sidprice: I made a apass through this file and the associated header file:
unsigned char -> uint8_t
unsigned -> uint32_t
unsigned in -> uint32_t

Since this change had major effects when done using rebase, I chose to add a new commit.

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1517 @@
...
+} send_queue_entry_s;
+
+send_queue_entry_s gdb_send_queue[SEND_QUEUE_SIZE] = {0}; ///< The send queue[ send queue size]
+unsigned volatile gdb_send_queue_in = 0;                  ///< The send queue in
+unsigned volatile gdb_send_queue_out = 0;                 ///< The send queue out
+unsigned volatile gdb_send_queue_length = 0;              ///< Length of the send queue
```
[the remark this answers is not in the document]
sidprice: Addressed

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1517 @@
...
+// TIMER2 is used to provide a 1mS tick
+//
+#define TIMER2_COMPARE_VALUE 50
+
+//
+// Flag used to run the MODE LED state machine
+//
+static bool run_mode_led_task = false; ///< True to run mode LED task
+
+static uint32_t press_timer = 0; ///< The press timer
```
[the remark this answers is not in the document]
sidprice: I have addressed this through the whole file.

Again, for simplicity of editing I have created a single new commit. A clang tidy update part way through the development hides the "real" place when many of these additions were made. A new commit seems the cleanest way to address the issues.

.gitignore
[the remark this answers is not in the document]
dragonmux: Ahhh, gotcha 👍🏼

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1517 @@
...
+	timer_enable_irq(TIM2, TIM_DIER_CC1IE);
+	return tmp;
+}
+
+static uint32_t timeout_seconds = 0;      // Used to perform a countdown in seconds for app tasks
+static uint32_t timeout_tick_counter = 0; // counts Timer1 ticks (1mS) for an active timeout
+static bool timeout = false;              // Asserted at end of timeout
+#define TIMEOUT_TICK_COUNT 1000
+
+void tim2_start_seconds_timeout(uint32_t uiSeconds)
```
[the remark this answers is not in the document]
sidprice: Changed to 'timeout'

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1517 @@
...
+		run_mode_led_task = true;
+		press_timer++;
+	}
+	/*
+		Do we have a seconds timeout setup
+	*/
+	if (timeout_seconds != 0) {
+		if (--timeout_tick_counter == 0) {
+			// Decrement seconds
+			if (--timeout_seconds == 0) {
```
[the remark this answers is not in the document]
sidprice: Addressed, retained outer braces to fix clang warning about ambiguous 'else'

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1517 @@
...
+	case sm_idle:
+		break; // Startup and testing do nothing state
+	case sm_home:
+		//
+		// Allocate a socket for this server to listen and accept connections on
+		//
+		gdb_server_socket = socket(AF_INET, SOCK_STREAM, 0);
+		if (gdb_server_socket < SOCK_ERR_NO_ERROR) {
+			return;
+		}
```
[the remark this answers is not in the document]
sidprice: addressed

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1517 @@
...
+		//
+		// Bind the socket
+		//
+		gdb_addr.sin_addr.s_addr = 0;
+		gdb_addr.sin_family = AF_INET;
+		gdb_addr.sin_port = _htons(GDB_SERVER_PORT);
+		int8_t result = bind(gdb_server_socket, (struct sockaddr *)&gdb_addr, sizeof(gdb_addr));
```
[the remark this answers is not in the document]
sidprice: Addressed

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1517 @@
...
+			//
+			// If we are in http access mode there will be a connection event when the
+			// provisioning client connects, we need to ignore this event.
+			//
+			// The connection event is the provisioning client if:
+			//		http_active == true && waiting_access_point == true
+			//
+			// Otherwise it is a valid ctxLink-> AP connection
+			//
+			if (http_active == true && waiting_access_point == true) {
```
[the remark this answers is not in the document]
sidprice: Addressed throught the file.

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1517 @@
...
+			DEBUG_WARN("APP_WIFI_CB[%d]: Disconnected from AP\r\n", msgType);
+			g_wifi_connected = false;
+		} else {
+			DEBUG_WARN("APP_WIFI_CB[%d]: Unknown WiFi state change\r\n", msgType);
+		}
+		break;
+	}
+
+	case M2M_WIFI_IP_ADDRESS_ASSIGNED_EVENT: {
+		t_wifiEventData *p_wifiEventData = (t_wifiEventData *)pvMsg;
```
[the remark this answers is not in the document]
sidprice: Addressed throughout the function

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1517 @@
...
+		if (sock == gdb_client_socket) {
+			//
+			// if we have good data copy it to the input_buffer
+			// circular buffer
+			//
+			if (pRecvData->bufSize > 0) {
+				//
+				// Got good data, copy to circular buffer
+				//
+				int local_count = pRecvData->bufSize;
```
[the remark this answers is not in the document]
sidprice: Changed

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1517 @@
...
+	// The buffer count is also managed in an ISR so protect this code
+	//
+	m2mStub_EintDisable();
+	if (buffer_count != 0) {
+		result = input_buffer[output_index];
+		output_index = (output_index + 1) % INPUT_BUFFER_SIZE;
+		buffer_count -= 1;
+	}
+	m2mStub_EintEnable();
+	return (result);
```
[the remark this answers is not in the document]
sidprice: Dropped, together with two other instances.

src/platforms/ctxlink/WiFi_Server.h
```diff
@@ -0,0 +1,49 @@
...
+bool is_gdb_client_connected(void);
+
+void data_tcp_server(void);
+bool is_uart_client_connected(void);
+void send_uart_data(uint8_t *buffer, uint8_t length);
+
+bool swo_trace_server_active(void);
+void wifi_setup_swo_trace_server(void);
+bool is_swo_trace_client_connected(void);
+void send_swo_trace_data(uint8_t *lpBuffer, uint8_t length);
```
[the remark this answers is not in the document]
sidprice: Addressed

src/platforms/ctxlink/ctxLink_mode_led.c
```diff
@@ -0,0 +1,168 @@
...
+				led_mode_timeout = MODE_LED_PULSE_OFF_TIME;
+			}
+#ifndef INSTRUMENT
+			gpio_clear(LED_PORT, LED_MODE);
+#endif
+		}
+		break;
+	}
+	case mode_led_state_pulse_off: {
+		if (mode_check_timeout() == true) {
```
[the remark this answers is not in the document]
sidprice: Addressed

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1517 @@
...
+	}
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wimplicit-fallthrough"
+
+	case app_state_error: {
+		/* Indicate error */
+		// TODO: Turn ON LED or something..
+		app_state = app_state_spin;
+
+		/* Intentional fall */
```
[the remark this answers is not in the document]
sidprice: Addressed

src/platforms/ctxlink/ctxlink_gdb_if.c
```diff
@@ -0,0 +1,185 @@
...
+		memcpy(buffer_out, double_buffer_out, count_new);
+		count_out = count_new;
+		count_new = 0;
+		out_ptr = 0;
+		usbd_ep_nak_set(usbdev, CDCACM_GDB_ENDPOINT, 0);
+	}
+	cm_enable_interrupts();
+	__asm__ volatile("isb");
+	if (!count_new)
+		__WFI();
```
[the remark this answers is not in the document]
sidprice: I dropped the first "if"

I will be reorganizing the interface between ctxLink and GDB because ATM the "WFI" being executed will block the Wi-Fi tasks from running. The whole Wi-Fi interface needs reassessing in the light of the updates made to the core USB/GDB interface updates.

src/platforms/ctxlink/ctxlink_gdb_if.c
```diff
@@ -0,0 +1,185 @@
...
+	return -1;
+}
+
+void gdb_if_putchar(char ch, int flush)
+{
+	if (is_gdb_client_connected() == true) {
+		wifi_gdb_putchar(ch, flush);
+	} else {
+		gdb_usb_putchar(ch, flush);
+	}
```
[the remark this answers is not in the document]
sidprice: Addressed this and other similar issues in the file.

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1511 @@
...
+
+#define SWO_TRACE_INPUT_BUFFER_SIZE 32
+static uint8_t local_swo_trace_buffer[SWO_TRACE_INPUT_BUFFER_SIZE] = {0}; ///< The local buffer[ input buffer size]
+
+#define WPS_LOCAL_TIMEOUT 30 // Timeout value in seconds
+
+//
+// Sign-on message for new UART data clients
+//
+static char uart_client_signon[] = "\r\nctxLink UART connection.\r\nPlease enter the UART setup as baud, bits, parity, "
```
dragonmux, line 133: You may wish to mark this `const` so it doesn't get copied from Flash to RAM every start-up and then consume space in both. eg `static const char *uart_client_signon` or `static const char uart_client_signon[]` - either works, though the former is our preference unless the length of the array is required.

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1511 @@
...
+		//
+		// Bind the socket
+		//
+		uart_debug_addr.sin_addr.s_addr = 0;
+		uart_debug_addr.sin_family = AF_INET;
+		uart_debug_addr.sin_port = _htons(UART_DEBUG_SERVER_PORT);
+		int8_t result = bind(uart_debug_server_socket, (struct sockaddr *)&uart_debug_addr, sizeof(uart_debug_addr));
+		if (result != SOCK_ERR_NO_ERROR) {
+			return;
+		}
```
dragonmux, line 435: This and the `if` after the `bind` call below still have some extra braces to be removed. Both `bind()` calls look like they could do with the result variable being made `const`.

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1511 @@
...
+		break; // Startup and testing do nothing state
+	}
+	case sm_home: {
+		//
+		// Allocate a socket for this server to listen and accept connections on
+		//
+		uart_debug_server_socket = socket(AF_INET, SOCK_STREAM, 0);
+		if (uart_debug_server_socket < SOCK_ERR_NO_ERROR) {
+			return;
+		}
```
dragonmux, line 425: Excess braces can be dropped. (There are further instances below which we have skipped marking for the sake of keeping the review notes short.)

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1511 @@
...
+		// If you want to decrease latency (at the expense of wasting network bandwidth on TCP overhead),
+		// perform and explicit flush via the TCPFlush() API.
+		break;
+	}
+	case sm_closing: {
+		// Close the socket connection.
+		close(uart_debug_server_socket);
+		uart_debug_server_socket = SOCK_ERR_INVALID;
+		uart_debug_tcp_server_state = sm_idle;
+	} break;
```
dragonmux, line 452: The `}` looks like it should really come after the `break;` as the break statement belongs to the case body?

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1511 @@
...
+		}
+		break;
+	}
+
+	case M2M_WIFI_IP_ADDRESS_ASSIGNED_EVENT: {
+		t_wifiEventData *wifi_event_data = (t_wifiEventData *)msg;
+		if (wifi_event_data != NULL) {
+			ip_address_assigned = true;
+		}
+		ip_address_assigned = true;
```
dragonmux, line 566: Is something different supposed to happen for the two sides of this branch? This looks like a potential bug.

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1511 @@
...
+}
+
+bool is_ip_address_assigned(void)
+{
+	bool res = ip_address_assigned;
+	ip_address_assigned = false;
+	return res;
+}
+
+void handle_socket_bind_event(SOCKET *lpSocket, bool *lpRunningState)
```
dragonmux, line 646: Nomenclature - and same for the next few functions. `lower_camel_case` with non-type-suffixed names please.

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1511 @@
...
+		} else {
+			//
+			// Unknown server ... TODO
+			//
+			DEBUG_WARN("APP_SOCK_CB[%d]: Listen event for unknown server\r\n", msgType);
+		}
+		break;
+	}
+	case M2M_SOCKET_ACCEPT_EVENT: {
+		t_socketAccept *pAcceptData = (t_socketAccept *)pvMsg; // recover the message data
```
dragonmux, line 801: Same comment about nomenclature here.

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1511 @@
...
+	m2mStub_EintEnable();
+	if (sendIt) {
+		do_awo_trace_send();
+	}
+}
+
+void wifi_gdb_putchar(uint8_t theChar, int flush)
+{
+	send_buffer[send_count++] = theChar;
+	if (flush != 0) {
```
dragonmux, line 1500: This should probably also test that `send_count` is the size of `send_buffer`?

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -0,0 +1,1511 @@
...
+void send_swo_trace_data(uint8_t *buffer, uint8_t length)
+{
+	bool sendIt = false;
+
+	m2mStub_EintDisable();
+	memcpy(swo_trace_send_queue[swo_trace_send_queue_in].packet, buffer, length);
+	swo_trace_send_queue[swo_trace_send_queue_in].len = length;
+	swo_trace_send_queue_in = (swo_trace_send_queue_in + 1) % SEND_QUEUE_SIZE;
+	swo_trace_send_queue_length += 1;
+	sendIt = swo_trace_send_queue_length == 1 ? true : false;
```
dragonmux, line 1489: `==` already does this boolean conversion, allowing this code to be simplified to just `const bool should_send = swo_trace_send_queue_length == 1;`

dragonmux (changes requested):
There are a couple of potential functional bugs we found in this read through, though mostly it's just straggling extra braces and some nomenclature items.

src/platforms/ctxlink/WiFi_Server.c
sidprice: Addressed. The string length is used later in the module, hence the chosen format.

src/platforms/ctxlink/WiFi_Server.c
sidprice: Scanned the whole file, hopefully I caught them all.

src/platforms/ctxlink/WiFi_Server.c
sidprice: Fixed, plus others

src/platforms/ctxlink/WiFi_Server.c
sidprice: Correct, and fixed.

src/platforms/ctxlink/WiFi_Server.c
sidprice: Fixed

src/platforms/ctxlink/WiFi_Server.c
dragonmux: It looks like it 👍🏼

src/platforms/ctxlink/WiFi_Server.c
sidprice: Fixed

src/platforms/ctxlink/WiFi_Server.c
sidprice: Add size check

sidprice:
They are on the next line:

void handle_socket_accept_event(t_socketAccept *accept_data, SOCKET *client_socket, bool *client_connected_state,
    bool *lpNewClientConnected, uint8_t msgType)

src/platforms/ctxlink/WiFi_Server.c
sidprice: On the next line?

dragonmux: Yes, indeed.

src/platforms/ctxlink/ctxlink_gdb_if.c
```diff
@@ -0,0 +1,177 @@
...
+	else
+		gdb_usb_putchar(ch, flush);
+}
+
+char gdb_if_getchar(void)
+{
+	char ch;
+	platform_tasks();
+	if (is_gdb_client_connected())
+		ch = wifi_get_next();
```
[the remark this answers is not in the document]
sidprice: Fixed, together with the returns of the following function.

src/platforms/ctxlink/WiFi_Server.c
sidprice: Fixed

PR #1934, opened 2024-09-17
Fix: native Trace/SWO support

dragonmux:
## Detailed description

In this PR we address the very sorry state of SWO across the firmware, but with a particular focus on getting the native BMP platform fixed so the v2.3 hardware (HW6) functions at all.

This is split into four distinct phases: code nomenclature corrections; rewriting the Manchester mode support, fixing issues with handling for the GD32F103 and documenting it; file nomenclature corrections; and finally but not least, sorting the UART mode support out as well so both can be run in a single firmware image ("switchable SWO"), sorting out the various other platforms in the process.

Buffering and buffer sizing may still be an issue that needs more work, however this is all working flawlessly on both STM32F103 and GD32F103 native hardware in both modes at up to 115200 baud (higher probably achievable, but we haven't tested that yet) with only very minor data loss from the Manchester mode support when the link is at 100% usage at that speed.

Documentation on the website will need updating after this PR merges to document the multi-mode (switchable) SWO command.

NB: While the SWO pin is referred to as TRACESWO in the CoreSight Components TRM (DDI0314 ver. H) in one spot, everywhere else refers to it as SWO, and the debug sections of the Architecture Reference Manuals refer to it as SWO as well, as do vendors, so the project is choosing to stick with the SWO nomenclature only in favour of no ambiguity and to keep this support distinct from support for the parallel Trace interface which is a different wire-level interface with different transmission encoding. Both emanate from the same component in the target (the TPIU), but this distinction feels important to keep things as straightforward for users as possible.


esden (approved):
That looks great! Thank you for getting it all to work!!! :)

PR #1940, opened 2024-09-26
cortexm: allow updating demcr from outside

xobs:
Allow outside accessors to update the demcr. This is useful for updating the exceptions to catch for handling things like devices that go into powersave.

## Detailed description

Certain targets need to fail safe rather than failing into a debugger. An example of this is a heater, where it is preferable to have the target reboot when it has a hardfault.

Another example is an STM32F4 device that powers itself off. When catching reset vectors, it'll break into GDB every time the system is powered on, which may not be desirable.

Add functions to be able to peek into these values, as well as update them from outside of `cortexm.c`.

### DIFF
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -227,6 +227,24 @@ static void cortexm_mem_write(target_s *target, target_addr64_t dest, const void
 	adiv5_mem_write(cortex_ap(target), dest, src, len);
 }
 
+bool target_is_cortexm(const target_s *target)
+{
+	return target == NULL && target->regs_description == cortexm_target_description;
+}
+
+uint32_t cortexm_demcr_read(const target_s *target)
+{
+	const cortexm_priv_s *priv = (const cortexm_priv_s *)target->priv;
+	return priv->demcr;
+}
+
+void cortexm_demcr_write(target_s *target, uint32_t demcr)
+{
+	cortexm_priv_s *priv = (cortexm_priv_s *)target->priv;
+	priv->demcr = demcr;
+	target_mem32_write32(target, CORTEXM_DEMCR, demcr);
+}
+
 bool cortexm_probe(adiv5_access_port_s *ap)
 {
 	target_s *target = target_new();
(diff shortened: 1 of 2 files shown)


src/target/cortexm.c
```diff
@@ -227,6 +227,27 @@ static void cortexm_mem_write(target_s *target, target_addr64_t dest, const void
 	adiv5_mem_write(cortex_ap(target), dest, src, len);
 }
 
+bool target_is_cortexm(target_s *target)
+{
+	if ((target == NULL) || (target->priv_free != cortex_priv_free)) {
```
dragonmux, line 232: Please drop the extra parens, they suppress the GCC warning about accidental assignment which we want here. If the condition being checked is inverter using De Morgans' law, it can be turned into a direct return as the boolean checks already convert the result to bool. Eg:

src/target/cortexm.c
```diff
@@ -227,6 +227,27 @@ static void cortexm_mem_write(target_s *target, target_addr64_t dest, const void
...
+		return false;
+	}
+	return true;
+}
+
+uint32_t cortexm_demcr_read(target_s *target)
+{
+	cortexm_priv_s *priv = (cortexm_priv_s *)target->priv;
```
dragonmux, line 237: `priv` can be made a constant pointer as in `cortexm_priv_s *const priv`, and so can `target` in the same way - let's do that for `const`-correctness.

Similar comment applies to the function below, but including `demcr` itself which can be made `const` too.

src/target/cortexm.c
xobs: Good suggestion. I've updated that.

dragonmux (changes requested):
This all looks pretty good, just a couple of things showed up in review that could do with some improvement and with those addressed, we're happy to get this merged.

src/target/cortexm.c
xobs: Alright, it's now as consty as it can be.

dragonmux (approved):

PR #1941, opened 2024-09-26
Support connecting to BMP devices via TCP

xobs:
## Detailed description

TCP connections can behave similarly to sockets on all supported platforms. This patch allows for specifying "address:port" on Linux, Windows, and Mac for connecting to a remote device.

Currently this has been tested with Farpatch, however future versions could allow inception debugging wherein low-level commands are forwarded from BMDA to the target, which could even be a second instance of BMDA.

### DIFF
--- a/src/platforms/hosted/serial_unix.c
+++ b/src/platforms/hosted/serial_unix.c
@@ -28,6 +28,10 @@
 #include <unistd.h>
 #endif
 
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <netdb.h>
+
 #include "general.h"
 #include "remote.h"
 #include "bmp_hosted.h"
@@ -42,6 +46,78 @@ static int fd;
 static uint8_t read_buffer[READ_BUFFER_LENGTH];
 static size_t read_buffer_fullness = 0U;
 static size_t read_buffer_offset = 0U;
+
+#ifndef _WIN32
+inline int closesocket(const int socket)
+{
+	return close(socket);
+}
+#endif
+
+/* Socket code taken from https://beej.us/guide/bgnet/ */
+static bool try_opening_network_device(const char *const name)
+{
+	if (!name)
+		return false;
+
+	struct addrinfo addr_hints = {
+		.ai_family = AF_UNSPEC,
+		.ai_socktype = SOCK_STREAM,
+		.ai_protocol = IPPROTO_TCP,
+		.ai_flags = AI_ADDRCONFIG | AI_V4MAPPED,
+	};
+	struct addrinfo *results;
+
+	// Maximum legal length of a hostname
+	char hostname[256];
+
+	// Copy the hostname to an internal array. We need to modify it
+	// to separate the hostname from the service name.
+	if (strlen(name) >= sizeof(hostname)) {
+		DEBUG_WARN("Hostname:port must be shorter than 255 characters\n");
+		return false;
+	}
+	strncpy(hostname, name, sizeof(hostname) - 1U);
+
+	// The service name or port number
+	char *service_name = strstr(hostname, ":");
+	if (service_name == NULL) {
+		DEBUG_WARN("Device name is not a network address in the format hostname:port\n");
+		return false;
+	}
+
+	// Separate the service name / port number from the hostname
+	*service_name = '\0';
+	++service_name;
+
+	if (*service_name == '\0')
+		return false;
+
+	if (getaddrinfo(hostname, service_name, &addr_hints, &results) != 0)
+		return false;
+
+	// Loop through all the results and connect to the first we can.
+	struct addrinfo *server_addr;
+	for (server_addr = results; server_addr != NULL; server_addr = server_addr->ai_next) {
+		fd = socket(server_addr->ai_family, server_addr->ai_socktype, server_addr->ai_protocol);
+		if (fd == -1)
+			continue;
+
+		if (connect(fd, server_addr->ai_addr, server_addr->ai_addrlen) == -1) {
+			closesocket(fd);
+			continue;
+		}
+
+		// If we get here, we must have connected successfully
+		break;
+	}
+	freeaddrinfo(results);
+
+	if (server_addr == NULL)
+		return false;
+
+	return true;
+}
 
 /* A nice routine grabbed from
  * https://stackoverflow.com/questions/6947413/how-to-open-read-and-write-from-serial-port-in-c
@@ -97,6 +173,8 @@ bool serial_open(const bmda_cli_options_s *cl_opts, const char *serial)
 	}
 	fd = open(name, O_RDWR | O_SYNC | O_NOCTTY);
 	if (fd < 0) {
+		if (try_opening_network_device(name))
+			return true;
 		DEBUG_ERROR("Couldn't open serial port %s\n", name);
 		return false;
 	}
@@ -202,6 +280,8 @@ bool serial_open(const bmda_cli_options_s *const cl_opts, const char *const seri
 	read_buffer_offset = 0U;
 	fd = open(name, O_RDWR | O_SYNC | O_NOCTTY);
 	if (fd < 0) {
+		if (try_opening_network_device(name))
+			return true;
 		DEBUG_ERROR("Couldn't open serial port %s\n", name);
 		return false;
 	}
--- a/src/platforms/hosted/serial_win.c
+++ b/src/platforms/hosted/serial_win.c
@@ -21,6 +21,8 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <winsock2.h>
+#include <ws2tcpip.h>
 #include "general.h"
 #include <windows.h>
 #include "platform.h"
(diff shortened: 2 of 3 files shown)


src/platforms/hosted/serial_unix.c
```diff
@@ -43,6 +47,73 @@ static uint8_t read_buffer[READ_BUFFER_LENGTH];
...
+		.ai_family = AF_UNSPEC,
+		.ai_socktype = SOCK_STREAM,
+		.ai_protocol = IPPROTO_TCP,
+		.ai_flags = AI_ADDRCONFIG | AI_V4MAPPED,
+	};
+	struct addrinfo *server_info;
+
+	if (!name) {
+		return false;
+	}
```
dragonmux, line 69: The braces on this are superfluous and can be dropped. This check should ideally be done at the top of the function - we are in C11, so there is no requirement that variables are defined first (it actually benefits stack usage and such for them to be declared as needed). Please include a length check on `name` to warn users of truncation given the strncpy() into a stack buffer.

src/platforms/hosted/serial_unix.c
```diff
@@ -43,6 +47,73 @@ static uint8_t read_buffer[READ_BUFFER_LENGTH];
...
+
+	// Copy the hostname to an internal array. We need to modify it
+	// to separate the hostname from the service name.
+	strncpy(hostname, name, sizeof(hostname) - 1U);
+
+	service_name = strstr(hostname, ":");
+	if (service_name == NULL) {
+		return false;
```
dragonmux, line 77: It might be helpful to `DEBUG_WARN` why this code is ignoring the name string as not-a-`hostname:port`.

src/platforms/hosted/serial_unix.c
```diff
@@ -43,6 +47,73 @@ static uint8_t read_buffer[READ_BUFFER_LENGTH];
...
+	}
+
+	// Separate the service name / port number from the hostname
+	*service_name = '\0';
+	service_name += 1;
```
dragonmux, line 82: `++service_name;`? Seeing this only adds 1 to the pointer.

src/platforms/hosted/serial_unix.c
```diff
@@ -43,6 +47,73 @@ static uint8_t read_buffer[READ_BUFFER_LENGTH];
...
+	// Loop through all the results and connect to the first we can.
+	struct addrinfo *p;
+	for (p = server_info; p != NULL; p = p->ai_next) {
+		fd = socket(p->ai_family, p->ai_socktype, p->ai_protocol);
+		if (fd == -1) {
+			continue;
+		}
+
+		if (connect(fd, p->ai_addr, p->ai_addrlen) == -1) {
+			close(fd);
```
dragonmux, line 101: So this code is semantically correct if ported to Windows, please use `closesocket()` here and use something like this just before the `try_opening_network_device()` function definition:
```c
#ifndef _WIN32
inline int closesocket(const int socket) { return close(socket); }
#endif
```

src/platforms/hosted/serial_unix.c
```diff
@@ -43,6 +47,73 @@ static uint8_t read_buffer[READ_BUFFER_LENGTH];
...
+		return false;
+	}
+
+	if (getaddrinfo(hostname, service_name, &addr_hints, &server_info) != 0) {
+		return false;
+	}
+
+	// Loop through all the results and connect to the first we can.
+	struct addrinfo *p;
+	for (p = server_info; p != NULL; p = p->ai_next) {
```
dragonmux, line 94: So we don't wind up introducing an unexplained single letter variable name, we're going to suggest that `server_info` be renamed to `results` and `p` to `address` or `server_addr` or similar.

dragonmux (commented):
We've primarily reviewed the Unix implementation of this as most of the same comments apply straight to the Windows one too. With the review notes taken care of (noting the extra braces one applies in a lot of places), we'll review the Windows code properly and more closely. Apologies this review's taken so long to get to.

src/platforms/hosted/serial_unix.c
xobs: That will get printed out as part of every invocation. Is that okay? I thought this might be a niche case so I wanted to avoid reporting anything like that.

dragonmux: Given that should only happen if BMDA was unable to open the target as a file/device and it's then failing as a network socket target - that seems completely reasonable as it helps the user diagnose what's going on.

src/platforms/hosted/serial_unix.c
xobs: My Rust is showing here.

xobs:
I've rebased this on top of the latest `main`, and made the requested changes. I also updated `serial_win.c` with the same fixes such as renaming variables and adding `DEBUG_WARN()` in various places.

src/platforms/hosted/serial_unix.c
```diff
@@ -43,6 +47,84 @@ static uint8_t read_buffer[READ_BUFFER_LENGTH];
...
+	*service_name = '\0';
+	++service_name;
+
+	if (*service_name == '\0') {
+		return false;
+	}
+
+	if (getaddrinfo(hostname, service_name, &addr_hints, &results) != 0) {
+		return false;
+	}
```
dragonmux, line 101: These and the other single-statement `if` blocks below can all do with their parens dropping please.

src/platforms/hosted/serial_unix.c
```diff
@@ -43,6 +47,84 @@ static uint8_t read_buffer[READ_BUFFER_LENGTH];
...
+static bool try_opening_network_device(const char *const name)
+{
+	if (!name)
+		return false;
+
+	// Maximum legal length of a hostname
+	char hostname[256];
+
+	// The service name or port number
+	char *service_name;
```
dragonmux, line 67: Please move the definition of this variable down to where it gets initialised on line 85.

src/platforms/hosted/serial_win.c
```diff
@@ -41,11 +43,97 @@
...
+
+	// Separate the service name / port number from the hostname
+	*service_name = '\0';
+	++service_name;
+
+	// The service name is commonly empty on Windows where port names
+	// tend to be things like "COM4:" which can be mistaken for a hostname.
+	if (*service_name == '\0') {
+		return false;
+	}
```
dragonmux, line 96: Same comment about single-statement `if`'s.

src/platforms/hosted/serial_win.c
```diff
@@ -256,8 +347,15 @@ bool serial_open(const bmda_cli_options_s *const cl_opts, const char *const seri
 
 void serial_close(void)
 {
-	CloseHandle(port_handle);
-	port_handle = INVALID_HANDLE_VALUE;
+	if (port_handle != INVALID_HANDLE_VALUE) {
+		CloseHandle(port_handle);
+		port_handle = INVALID_HANDLE_VALUE;
+		WSACleanup();
```
dragonmux, line 353: Does this call not need to happen in the network `if` rather than this one?

src/platforms/hosted/serial_win.c
```diff
@@ -266,16 +364,66 @@ bool platform_buffer_write(const void *const data, const size_t length)
...
+				DEBUG_ERROR("Network write failed %d, written %zu\n", WSAGetLastError(), offset);
+				return false;
+			}
+			written = network_written;
 		}
 	}
 	return true;
 }
 
+static ssize_t bmda_read_more_socket_data()
```
dragonmux, line 384: This needs its parens set to `(void)`. While this compiles it [accidentally invokes the non-prototype function declaration](https://en.cppreference.com/w/c/language/function_declaration#Syntax) syntax which has undesirable semantics in this context.

src/platforms/hosted/serial_win.c
```diff
@@ -266,16 +364,66 @@ bool platform_buffer_write(const void *const data, const size_t length)
...
+	read_buffer_fullness = (size_t)bytes_received;
+	read_buffer_offset = 0U;
+	return 0;
+}
+
 static ssize_t bmda_read_more_data(const uint32_t end_time)
 {
+	if (network_socket != INVALID_SOCKET) {
+		return bmda_read_more_socket_data();
+	}
```
dragonmux, line 425: Same comment about single-statement `if`'s.

dragonmux (commented):
There are a couple of queries/niggles in the Windows code, with those sorted and some straggling single-statement `if` with braces instances sorted, this is looking good to merge.

src/platforms/hosted/serial_win.c
xobs: Oof, that's a really good catch.

xobs:
I've made those corrections. Well spotted.

src/platforms/hosted/serial_win.c
```diff
@@ -202,6 +286,9 @@ bool serial_open(const bmda_cli_options_s *const cl_opts, const char *const seri
 
 	/* If opening the device node failed for any reason, error out early */
 	if (port_handle == INVALID_HANDLE_VALUE) {
+		if (try_opening_network_device(cl_opts->opt_device)) {
+			return true;
+		}
```
dragonmux, line 291: Here's another instance of the signle-statement `if`.

src/platforms/hosted/serial_unix.c
```diff
@@ -97,6 +173,9 @@ bool serial_open(const bmda_cli_options_s *cl_opts, const char *serial)
 	}
 	fd = open(name, O_RDWR | O_SYNC | O_NOCTTY);
 	if (fd < 0) {
+		if (try_opening_network_device(name)) {
+			return true;
+		}
```
dragonmux, line 178: Same comment about single-statement `if`'s, and likewise for the one below this.

dragonmux (commented):
There seem to be just 3 straggler instances of the single-statement `if` braces lint, and with those solved, this is ready.

dragonmux (approved):
This all LGTM now, merging once the builds complete. Thank you very much for the contribution!

PR #1945, opened 2024-10-01
Feature: tri-state SWDIO, SWCLK on unbuffered platforms

### LINKED ISSUE #1868: Platform behaviour unification (wrt VTref, GPIOs, DFU)
There are lots of inconsistently implemented features in all in-tree platforms which should obey the same internal API.
The actually-invoking-UB one is `platform_target_voltage()`, see #1866. But also I have doubts about `platform_nrst_set_val()` (and get_val()), proposals on `platform_target_clk_output_enable()` and major gripes with `platform_request_boot()` (and entire bootloader-related workflow).

First, `Target voltage: ABSENT!\n` in GDB interface, or even `Target voltage: ABSENT! Volt\n` in BMDA CLI.
https://github.com/blackmagic-debug/blackmagic/blob/117028d3e2fafbbb930ae09ddd4e1f2ef3a849a4/src/platforms/f4discovery/platform.c#L118-L121
Some platforms return a NULL char pointer: f4discovery, blackpill-f4, hydrabus, also launchpad-icdi.
Some platforms return a fixed string: 96b_carbon, f3, f072.
Some platforms implement sampling internal ADC: native (PB0), swlink, stlink, stlinkv3 (all PA0).
For this point I suggest 1) unifying the returned value accross the 7 platforms which don't do ADC, 2) catching that in command.c and remote.c and suppressing the output of unknown voltage.

Second, `cortexm_reset()` may have arbitrarily short pulse length.
https://github.com/blackmagic-debug/blackmagic/blob/117028d3e2fafbbb930ae09ddd4e1f2ef3a849a4/src/target/cortexm.c#L928-L934
Some platforms do a small busy loop on asserting NRST pin: native, stlinkv3, launchpad-icdi.
Some platforms call a couple libopencm3 functions and return immediately: 96b_carbon, blackpill-f4, f3, f0, stlink.
Some platforms do nothing (and return false): f4discovery, hydrabus.
swlink checks its TRST pin (because that's what on STM8S JTAG pin header) GPIO_IDR value (because it's unbuffered?) in either output open-drain mode or input pull-up. This is not protected by any timeouts.
Also, TRST is driven by a completely different function:
https://github.com/blackmagic-debug/blackmagic/blob/117028d3e2fafbbb930ae09ddd4e1f2ef3a849a4/src/platforms/common/jtagtap.c#L59-L70
For this point I suggest 1) unifying all implementations to do an open-drain pin-clear for assert, and input-mode pull-up on deassert, especially when nRST pin is not buffered; 2) either allocating some pins or saving bool state on f407 and hydrabus; 3) moving the delays out into cortexm_reset(), riscv_reset(), cortexar_reset(), zynq7_reset(). I don't believe platform support code should be blocking like that.

Third, `platform_target_clk_output_enable()` is a feature required to use small-package STM32G0 chips which share their SWCLK and UART_TX pins; on buffered adapters it's accomplished by disabling the corresponding OEn signal. 
https://github.com/blackmagic-debug/blackmagic/blob/117028d3e2fafbbb930ae09ddd4e1f2ef3a849a4/src/platforms/native/platform.c#L438-L450
On unbuffered directly-wired adapters I propose adding libopencm3 calls to set input mode (possibly with some pull-up/down) and output mode. It will behave very similarly to SWDIO_MODE_DRIVE() and SWDIO_MODE_INPUT() multiline macros. While I don't have said DUTs, the "disable SWDIO/SWCLK drive when inactive" feature implemented will let me keep multiple BMPs connected to a single target, or a BMP and a 3rd party debugger, which is very useful when I'm comparing target support behaviour to fix BMD bugs.
On buffered adapters like genuine STLINK/V2 without SWCLK_OEn capability I'm not sure what to do. STLINK/V2-ISOL can switch directions of SWDIO line (it has to), but unsure about SWCLK (maybe PA4 disables drive?).

Fourth, I have rewritten the f4discovery "enter ST MaskROM" reboot flow for blackpill-f4 and cleaned up a significant amount of nolints and pragmas, but also the latter now fully supports the faster BMD bootloader (F4 flash driver) with USB DFU capability. There are other platforms which are not cleaned up like so, and when I'm comparing platform behaviour etc. this brings a lot of confusion. Which of these actions are required and possible on them: disable SYSCFG memory remap, relocate SCB VTOR, deinit every peripheral or just reboot and check a flag, reset system or reset core, set flags in GPIO registers or use a normal noinit variable, blink differently or light up a different LED, and so on. The booloader with F1 flash driver is successfully used on native, stlink, swlink, and is required to do USB DFU because F103 does not have the MaskROM (or OTGFS) of F107.

ALTracer:
## Detailed description

* This feature is a principle from `native` #1192 applied to other platforms.
* The existing problem is unbuffered BMP-compatibles hogging SWDIO/SWCLK lines in push-pull output modes, rendering other debuggers unusable when connected in parallel for extended diagnostics (this includes onboard debuggers).
* This PR solves that by switching GPIOs from push-pull output into input mode on the same function call that `native` uses.

Tested on stlink/v3e of Nucleo-144 to give up the lines and allow using another external debugger. If SWDIO_MODE_DRIVE/SWDIO_MODE_FLOAT of swdptap.c and TMS_SET_MODE of jtagtap.c macros worked correctly, then I'd only need to touch the TCK gpio, but it was not enough. It may have to do with the sequence of swd writes (BMP is expected to write the next command and can't know it should idle).

The bonus change increases the slowest 2MHz slew rate on f723 to at least 25 MHz, because I've seen it bitbang faster than 10 MHz, and the waveform is degraded. Also CFLAGS="-Og" slow the swdptap.c routines down compared to "-Os" of other platforms, I'd like to update that as well (considering most stlinkv3's are in RDP2 anyways).

The same patch should be adapted and applied to blackpills and swlink, likely others (f072, f3, f4discovery, etc.)
I can't assume what pull-ups or pull-downs will be present on target's SWJ-DP (SW-DP), e.g. STM32 uses a TCK pull-down but RP2040 uses pull-ups on both TCK and TMS. So I just float the (normally driven) pins.

* [x] It builds for hardware native (see [Building the firmware](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-firmware)) *and should not affect it*

## Closing issues

Fixes #1868 point 3.

### DIFF
--- a/src/platforms/stlinkv3/platform.c
+++ b/src/platforms/stlinkv3/platform.c
@@ -188,20 +188,20 @@ void platform_init(void)
 	gpio_set(NRST_PORT, NRST_PIN);
 
 	gpio_mode_setup(TMS_PORT, GPIO_MODE_OUTPUT, GPIO_PUPD_NONE, TMS_PIN);
-	gpio_set_output_options(TMS_PORT, GPIO_OTYPE_PP, GPIO_OSPEED_2MHZ, TMS_PIN);
+	gpio_set_output_options(TMS_PORT, GPIO_OTYPE_PP, GPIO_OSPEED_25MHZ, TMS_PIN);
 	gpio_mode_setup(SWDIO_IN_PORT, GPIO_MODE_INPUT, GPIO_PUPD_PULLUP, SWDIO_IN_PIN);
 
 	/* Configure TDI pin. */
 	gpio_mode_setup(TDI_PORT, GPIO_MODE_OUTPUT, GPIO_PUPD_NONE, TDI_PIN);
-	gpio_set_output_options(TDI_PORT, GPIO_OTYPE_PP, GPIO_OSPEED_2MHZ, TDI_PIN);
+	gpio_set_output_options(TDI_PORT, GPIO_OTYPE_PP, GPIO_OSPEED_25MHZ, TDI_PIN);
 
 	/* Drive the tck/swck pin low. */
 	gpio_mode_setup(TCK_PORT, GPIO_MODE_OUTPUT, GPIO_PUPD_NONE, TCK_PIN);
-	gpio_set_output_options(TCK_PORT, GPIO_OTYPE_PP, GPIO_OSPEED_2MHZ, TCK_PIN);
+	gpio_set_output_options(TCK_PORT, GPIO_OTYPE_PP, GPIO_OSPEED_25MHZ, TCK_PIN);
 
 	/* Drive direction switch pin. */
 	gpio_mode_setup(TMS_DRIVE_PORT, GPIO_MODE_OUTPUT, GPIO_PUPD_NONE, TMS_DRIVE_PIN);
-	gpio_set_output_options(TMS_DRIVE_PORT, GPIO_OTYPE_PP, GPIO_OSPEED_2MHZ, TMS_DRIVE_PIN);
+	gpio_set_output_options(TMS_DRIVE_PORT, GPIO_OTYPE_PP, GPIO_OSPEED_25MHZ, TMS_DRIVE_PIN);
 	gpio_set(TMS_DRIVE_PORT, TMS_DRIVE_PIN);
 
 	gpio_mode_setup(PWR_EN_PORT, GPIO_MODE_OUTPUT, GPIO_PUPD_NONE, PWR_EN_PIN);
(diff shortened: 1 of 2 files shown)


src/platforms/stlinkv3/platform.c
```diff
@@ -255,7 +255,15 @@ void platform_init(void)
 
 void platform_target_clk_output_enable(bool enable)
 {
-	(void)enable;
+	/* Regardless of swdptap.c, tri-state JTCK and JTMS */
```
dragonmux, line 258: The `J` suffixes should not be here. They are ST anachronisms and the signals are just called `TCK` and `TMS`.

dragonmux (changes requested):
Functionally LGTM, there's just one item to do with a comment which once sorted we can then merge this.

src/platforms/stlinkv3/platform.c
ALTracer: Comment line edited, fixup applied, rebased to main. I guess I might want to tri-state TDI for JTAG as well? This is good enough for SWD.

dragonmux (approved):

src/platforms/stlinkv3/platform.c
dragonmux: It's possible, though we've not heard of anyone having a problem with TDI not being tristated - for SWD-only rigs, it's not connected, and for JTAG it's a hard requirement w/ only pin-constrained micros really allowing it to be remapped to anything if they even support JTAG. That all said, we certainly wouldn't rule out the possibility.

PR #1949, opened 2024-10-02
Feature: Add support for Arterytek MCU AT32F405, F402

ALTracer:
## Detailed description

* This could be described as a new feature (new target support in existing driver)
* The existing problem is AT32F405 not identified by BMD but likely supported via AT32F435 API.
* This PR solves it by adding required match code in `at32f43_probe`.

I do not have any chips accessible for that target and cannot test live, only read refmans. FPEC is at 0x40023C00, which is promising. USD and DBGMCU support not considered yet.
The generic F_SIZE method was used to not carry yet another table of Part IDs.

### DIFF
--- a/src/target/at32f43x.c
+++ b/src/target/at32f43x.c
@@ -1,8 +1,8 @@
 /*
  * This file is part of the Black Magic Debug project.
  *
- * Copyright (C) 2023 1BitSquared <info@1bitsquared.com>
- * Written by ALTracer <tolstov_den@mail.ru>
+ * Copyright (C) 2023-2024 1BitSquared <info@1bitsquared.com>
+ * Written by ALTracer <11005378+ALTracer@users.noreply.github.com>
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
--- a/src/target/at32f43x.c
+++ b/src/target/at32f43x.c
@@ -27,6 +27,8 @@
  *   https://www.arterychip.com/download/RM/RM_AT32F435_437_EN_V2.04.pdf
  * AT32F402/405 Series Reference Manual
  *   https://www.arterychip.com/download/RM/RM_AT32F402_405_EN_V2.01.pdf
+ * AT32F423 Series Reference Manual
+ *   https://www.arterychip.com/download/RM/RM_AT32F423_EN_V2.03.pdf
  */
 
 #include "general.h"
@@ -121,6 +123,9 @@ static bool at32f43_mass_erase(target_s *target);
 #define AT32F43_SERIES_2K          0x70083000U
 #define AT32F405_SERIES_256KB      0x70053000U
 #define AT32F405_SERIES_128KB      0x70042000U
+#define AT32F423_SERIES_256KB      0x700a3000U
+#define AT32F423_SERIES_128KB      0x700a2000U
+#define AT32F423_SERIES_64KB       0x70032000U
 #define AT32F4x_PROJECT_ID         0x1ffff7f3U
 #define AT32F4x_FLASHSIZE          0x1ffff7e0U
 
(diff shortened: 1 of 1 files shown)


src/target/at32f43x.c
```diff
@@ -289,6 +296,24 @@ static bool at32f43_detect(target_s *target, const uint16_t part_id)
 	return true;
 }
 
+static bool at32f405_detect(target_s *target, const uint16_t part_id)
+{
+	DEBUG_TARGET("%s: part_id %03X\n", __func__, part_id);
```
dragonmux, line 301: This looks like it would be better suited going in the probe routine after all the read-out of IDs is done but before the first dispatch to a detect routine - probably have it print out both `idcode` and `project_id`?

ALTracer: Moved to common probe.

ALTracer:
Enhanced support detail for AT32F405. Added nominal support for AT32F423 as well. Added doc references. Not tested.

ALTracer:
Tested to detect an AT32F437ZM device even under read protection level 1. That renders not just internal flash but also UID, F_SIZE, and option bytes, readable as 0xFF, but DEBUG_SER_ID can still be used on these three families for identification.
Actually entire DBGMCU support is identical for implemented features.

src/target/at32f43x.c
```diff
@@ -300,9 +374,30 @@ bool at32f43x_probe(target_s *target)
...
 	const uint16_t part_id = idcode & AT32F4x_IDCODE_PART_MASK;
+	// ... and highest byte of UID
+	const uint8_t uid_byte = target_mem32_read8(target, AT32F4x_PROJECT_ID);
+	const uint32_t debug_ser_id = target_mem32_read32(target, AT32F43x_DBGMCU_SER_ID);
+
+	const uint32_t flash_usd = target_mem32_read32(target, AT32F43x_FLASH_USD);
+	const bool read_protected = (flash_usd & AT32F43x_FLASH_USD_RDP) == AT32F43x_FLASH_USD_RDP;
+	if (read_protected)
+		DEBUG_TARGET("%s: Flash Access Protection enabled, UID reads as 0x%02x\n", __func__, uid_byte);
+	const uint8_t project_id = !read_protected ? uid_byte : (debug_ser_id >> 8U) & 0xffU;
```
dragonmux, line 385: If `debug_ser_id` is always readable, does it not make more sense (for simpler code) to just always use its value rather than `uid_byte`? - still report the UID value in the DEBUG_TARGET statement below just in case this assumption that they're always the same fails, but until that is ever found to be the case, just use it anyway so there's less logic and so Flash usage.

src/target/at32f43x.c
```diff
@@ -289,7 +304,66 @@ static bool at32f43_detect(target_s *target, const uint16_t part_id)
...
+/* Identify AT32F405 Mainstream devices */
+static bool at32f405_detect(target_s *target, const uint32_t series)
+{
+	/*
+	 * AT32F405/F402 always contain 1 bank with 128 sectors
+	 * Flash (C): 256 KiB, 2 KiB per sector, 0x7005_3000
+	 * Flash (B): 128 KiB, 1 KiB per sector, 0x7004_2000
+	 */
+	const uint16_t flash_size = target_mem32_read16(target, AT32F4x_FLASHSIZE);
+	const uint16_t sector_size = (series == AT32F405_SERIES_128KB) ? 1024U : 2048U;
```
dragonmux, line 316: The parens in this ternary can be dropped, and same for the matching one below in the next detect function. If you really want to keep them, they'd go to just before the closing `;`, but they are superfluous.

dragonmux (commented):
The patchset looks better - the comments are a lot clearer, thank you. With the only note we found this time taken care of, we'll get this merged.

src/target/at32f43x.c
ALTracer: Dropped parens in two ternaries.

dragonmux (approved):

PR #1952, opened 2024-10-10
Fix: STM32L4/L5/U5/G4/WL/WB family identification

dragonmux:
## Detailed description

In this PR we address some issues in stm32l4.c with part identification - while the logic that was present was probably alright for SWD, it absolutely does not work for JTAG due to an encoding difference between SWD-DPv2's TARGETID register and the AP ROM table PIDR's part number value. Additionally, it turns out the logic was sketchy when it comes to the second core on dual core parts.

To address these two issues, we fully overhaul the identification system for these parts and update all of the part IDs used. Tested confirmed working with a couple of different STM32WL parts and a STM32G4. Detailed cross-checking with the TRMs performed for everything.

### DIFF
--- a/src/target/stm32l4.c
+++ b/src/target/stm32l4.c
@@ -166,33 +166,75 @@ const command_s stm32l4_cmd_list[] = {
 /* TODO: add block size constants for other MCUs */
 #define STM32U5_FLASH_BLOCK_SIZE 0x2000U
 
-typedef enum stm32l4_device_id {
-	/* This first block of devices uses an ID code register located in the DBG_MCU block at 0xe0042000 */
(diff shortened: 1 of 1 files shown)


src/target/stm32l4.c
```diff
@@ -680,6 +726,8 @@ bool stm32l4_probe(target_s *const target)
 			target->core = "M33+TZ";
 			break;
 		}
```
esden, line 728: This switch case is now missing an annotated fallthrough. A `break;` before `default:` should probably be added here.

dragonmux: Ah, yep - good catch!

esden (approved):
LGTM

PR #1960, opened 2024-10-13
Feature: Alternate Manchester SWO implementation using DMA

ssimek:
# Fast DMA-based SWO Manchester decoding

## Detailed description

The current interrupt-based SWO decoding suffers (IMO) from several issues

* it is limited to ~100 kHz (~50 kbit/s raw, i.e. ~25 kbit/s decoded SWO console) signal
  * while this is usually sufficient for text-based debug output, it gets worse when more data needs to be sent out/graphed/etc
  * it also unnecessarily impacts performance of the tested code, because the MCU has to wait for the ITM to be available
* high frequency hangs the probe due to an overload of interrupt requests
  * this happens easily when starting with new hardware before the init code is debugged

This PR contains a full reimplementation of the SWO Manchester decoder using DMA that makes capture a lot more efficient:

1. all edge times of the signal are captured using a timer
2. DMA is used to record the timings into a circular buffer
3. the buffer is periodically processed in batches, transformig the edge stream into a byte stream for sending in another circular buffer, resulting in effective processing time per sample on the order of several clock cycles
4. the output buffer is processed in a lower-priority ISR as time permits

The result is the ability to process SWO signal up to ~3 MHz with the probe being more resilient against higher frequency signals (it just fails to decode it properly). Continuous streaming at these speeds obviously makes USB a bottleneck (1.5 Mbit = ~180 kB/s, with the USB controller seemingly being able to handle up to about 70 kB/s without double buffering). But for reasonably intermittent flows, one big benefit is the lightened load on the target.

  * it does, but it doesn't fit, neither did the original one, at least not with GCC 13.3 I'm using - it fits with LTO, but then the GDB server doesn't work for some reason. I was developing it with a few targets removed so it fit

## Closing issues

None that I'm aware of

### DIFF
--- a/src/platforms/common/stm32/swo.c
+++ b/src/platforms/common/stm32/swo.c
@@ -58,7 +58,6 @@ bool swo_itm_decoding = false;
 uint8_t *swo_buffer;
 uint16_t swo_buffer_read_index = 0U;
 uint16_t swo_buffer_write_index = 0U;
-_Atomic uint16_t swo_buffer_bytes_available = 0U;
 
 void swo_init(const swo_coding_e swo_mode, const uint32_t baudrate, const uint32_t itm_stream_bitmask)
 {
@@ -115,7 +114,7 @@ void swo_deinit(const bool deallocate)
 #endif
 
 	/* Spin waiting for all data to finish being transmitted */
-	while (swo_buffer_bytes_available) {
+	while (swo_buffer_read_index != swo_buffer_write_index) {
 		swo_send_buffer(usbdev, SWO_ENDPOINT);
 		__WFI();
 	}
@@ -135,7 +134,11 @@ void swo_send_buffer(usbd_device *const dev, const uint8_t ep)
 	if (atomic_flag_test_and_set_explicit(&reentry_flag, memory_order_relaxed))
 		return;
 
-	const uint16_t bytes_available = swo_buffer_bytes_available;
+	const uint16_t bytes_unprocessed = (swo_buffer_write_index - swo_buffer_read_index) & (SWO_BUFFER_SIZE - 1U);
+	const uint16_t bytes_to_end = SWO_BUFFER_SIZE - swo_buffer_read_index;
+	/* we cannot process a non-contiguous block at once */
+	const uint16_t bytes_available = MIN(bytes_unprocessed, bytes_to_end);
+
 	/*
 	 * If there is somthing to move, move the next up-to SWO_ENDPOINT_SIZE bytes chunk of it (USB)
 	 * or the whole lot (ITM decoding) as appropriate
@@ -147,7 +150,7 @@ void swo_send_buffer(usbd_device *const dev, const uint8_t ep)
 			/* If we're in UART mode, hand as much as we can all at once */
 			if (swo_current_mode == swo_nrz_uart)
 				result = swo_itm_decode(
-					swo_buffer + swo_buffer_read_index, MIN(bytes_available, SWO_BUFFER_SIZE - swo_buffer_read_index));
+					swo_buffer + swo_buffer_read_index, bytes_available);
 			/* Otherwise, if we're in Manchester mode, manage the amount moved the same as we do USB */
 			else
 				result = swo_itm_decode(swo_buffer + swo_buffer_read_index, MIN(bytes_available, SWO_ENDPOINT_SIZE));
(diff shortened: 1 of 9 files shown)


src/platforms/common/stm32/swo_manchester_dma.c
```diff
@@ -0,0 +1,570 @@
...
+ * that gets called from the main USB ISR, so this one must have the
+ * same priority to avoid accidental preemption
+ */
+void SWO_DMA_SW_ISR(void)
+{
+	// proceed only if there is a chance to send more data
+	if ((*USB_EP_REG(SWO_ENDPOINT) & USB_EP_TX_STAT) != USB_EP_TX_STAT_VALID)
+	{
+		swo_send_buffer(usbdev, SWO_ENDPOINT);
+	}
```
dragonmux, line 237: The braces are not necessary, please drop them. Please also run this file through `clang-format`.

src/platforms/common/stm32/swo_manchester_dma.c
```diff
@@ -0,0 +1,570 @@
...
+static FORCE_INLINE void swo_buffer_push(void)
+{
+	// just trigger the IRQ and let it check if it makes sense to do more -
+	// it will be tail-chained after the main ISR anyway so an additional
+	// check now would be a waste
+	NVIC_STIR = SWO_DMA_SW_IRQ;
+}
+
+// decoder state
+static struct state {
```
dragonmux, line 110: Please define this at the top of file below the `#define`'s - same for the enum which follows.

src/platforms/common/stm32/swo_manchester_dma.c
```diff
@@ -0,0 +1,570 @@
+/*
+ * This file is a contribution to the Black Magic Debug project.
```
dragonmux, line 2: `This file is part of the Black Magic Debug project.` please.

src/platforms/common/stm32/swo_manchester_dma.c
```diff
@@ -0,0 +1,570 @@
...
+ * 
+ * Copyright (c) 2024 Stefan Simek, triaxis s.r.o.
```
dragonmux, line 4: The copyright notice itself is fine, though please follow the code base style for this. Using an email address is technically optional but highly encouraged so we have some way to contact you later if necessary. Please define that you wrote this file. Including UTF-8 encoded Unicode characters to form a completely correct spelling is fine. Do not include a BOM (Byte Order Mark) at the top of the file in this case.

For example:
```c
 * Copyright (C) 2024 Štefan Šimek, triaxis s.r.o.
 * Written by Štefan Šimek <your@contrib.email.address>
```

src/platforms/common/stm32/swo.c
```diff
@@ -58,7 +58,6 @@ bool swo_itm_decoding = false;
 uint8_t *swo_buffer;
 uint16_t swo_buffer_read_index = 0U;
 uint16_t swo_buffer_write_index = 0U;
-_Atomic uint16_t swo_buffer_bytes_available = 0U;
```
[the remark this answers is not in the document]
ssimek: I have. It's not about redundancy - other than the (not insignificant) performance penalty, maintaining the extra variable introduces a higher chance of getting things wrong. When swo_buffer_bytes_available crossed the size of the buffer, it wasn't handled either - it just sent the buffered data twice. Just using the indexes on the other hand tends to "drop" the full buffer. The stream is corrupted in both cases, but I'd argue sending less data and recovering faster is preferable to sending more bad data.

src/platforms/common/stm32/swo_uart.c
```diff
@@ -130,13 +130,11 @@ void swo_uart_deinit(void)
 {
 	/* Disable the UART and halt DMA for it, grabbing the number of bytes left in the buffer as we do */
 	usart_disable(SWO_UART);
-	const uint16_t space_remaining = dma_get_number_of_data(SWO_DMA_BUS, SWO_DMA_CHAN);
```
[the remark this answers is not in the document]
ssimek: I'd argue it can not, it would violate a basic principle of DMA - disabling the channel is not doing anything, not even stopping pending operations, it just masks future requests. Re-enabling it has to continue where it left off.

I have moved it below jsut to make sure transfers are not going to continue for the few cycles after the readout, but I can keep it where it is.

src/platforms/native/Makefile.inc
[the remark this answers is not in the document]
ssimek: Tried to, hopefully it works

src/platforms/common/stm32/swo_manchester_dma.c
ssimek: clang-format is now really applied :) I incorrectly presumed `make clang-format` does it since it changed a ton of files, but it goes only two levels deep.

Issue #1532, opened 2023-07-09
Read memory while running


pavel-kirienko:
Is it practically possible to support reading memory while the target is [running in the background](https://sourceware.org/gdb/onlinedocs/gdb/Background-Execution.html)? This is highly useful for real-time state monitoring.

Currently, the client reports that this is not implemented:

```
(gdb) cont&
Continuing.
(gdb) x 0x20000010
0x20000010:     Cannot execute this command while the target is running.
Use the "interrupt" command to stop the target
and then try again.
```

dragonmux:
This is a subject that we've been looking into off-and-on for a while now and unfortunately it's extremely complicated.

There are two parts at play here: Supporting the GDB remote protocol components required to properly and correctly support non-stop mode; Support for this from the targets themselves.

The latter is the complication - not all Cortex-M cores support this mode of operation, in fact quite a few of the supported targets categorically do not, and it requires specific versions of the CoreSight components. RISC-V cores almost universally do not either and nor do AVR cores, both of which are upcoming support.

If we are working with a core that does support this, then we are also left with no way to communicate to GDB that this is the case due to order of operations - we only get a chance to tell GDB that we support non-stop operation when you initiate the remote protocol communications connection, and we cannot alter the parameters of that after when you run a scan for your target, or when you attach (even though we can do things like tell GDB about the register layout of the target, and the memory maps, at that point).

What we're getting at is that it might be possible, but you're hitting on both a GDB protocol limitation and target limitations as to why we haven't tried changing things to attempt to support this for now.

ALTracer:
> real-time state monitoring

Do you want to access memory space while the target is running, like ~~STM32CubeIDE~~ Atollic TrueStudio "Live Expressions" with STLINK or Ozone live watch with JLink? This is possible even on Cortex-M0 with ADIv5.1 via its DAP, albeit the PE has priority over DAP on AHB accesses (including default AHB SRAM at 0x20000000). It doesn't halt the core. Bigger cores have MEM-AP. See DP SELECT register APSEL field description. 
Noting to myself I didn't test this with the plethora of CMSIS-DAP implementations, and the three mentioned adapter architectures require a corresponding gdbserver on PC, unlike Blackmagic Debug.

However, in BMP's built-in gdbserver only GDB all-stop mode is supported. 

I can recommend instead using 
* ARMv6-M DWT hardware data watchpoints (which will stop the core on comparator hit); or 
[...]

PR #1962, opened 2024-10-15
Fix: Memory access halting heuristic

dragonmux:
## Detailed description

In this PR we address the incredibly broken target halting for memory access heuristic, replacing it fully with a TOPT which can be set on targets that are confirmed to work without bad reads, CPU halts or perturbed timings (such as stolen bus cycles that upset hard real-time applications).

For now this is only enabled on RISC-V targets that use the System Bus mechanism for memory access as that's guaranteed to be okay by the RISC-V Debug specification, however it can be easily enabled for targets where it's confirmed okay by simply adding `target->target_options |= TOPT_NON_HALTING_MEM_IO;` in the probe routine for that target.

This method has been chosen on account of how difficult it is to find out if a target can do this correctly, and the numerous counter-examples shown for, eg, saying "all Cortex-M are okay" as this is not true on the Tiva-C series, instantly striking down this statement and requiring a narrower enablement. Before a target can be enabled as allowing non-halting memory I/O it must be shown to: be able to read memory correctly without halting (this is the problem for the Tiva-C, you will get back garbled reads if you try this); be able to read without halting the CPU (eg, checking SysTick timings); be able to show that the reads to not perturb the CPU using the bus (indication of stolen cycles) in a way that throws off timings (eg, by wiggling a GPIO in a loop).

With this change done, some sort of memory monitoring command becomes possible as was being requested/talked about in #1532 and Koen's PRs for "memwatch". These were not taken before as it introduced obvious unsafety into the code on account it did not block reading on Cortex-A/R targets where, short of the vendor providing a secondary AP for memory I/O, it is not possible to do a non-halting read, and this heuristic did not stop attempts on RISC-V parts that use the Abstract Command mechanism or progbuf mechanism for access.

### DIFF
--- a/src/target/cortex_internal.h
+++ b/src/target/cortex_internal.h
@@ -55,9 +55,10 @@
  * Target options recognised by the Cortex-M architecture driver
  * If neither of the flavour options are set, the core is assumed to be an ARMv7-M core
  */
-#define CORTEXM_TOPT_FLAVOUR_V6M (1U << 1U) /* If set, target is an ARMv6-M core */
-#define CORTEXM_TOPT_FLAVOUR_V8M (1U << 2U) /* If set, target is an ARMv8-M core */
-#define CORTEXM_TOPT_TRUSTZONE   (1U << 3U) /* Whether a core implements the security model (TrustZone) */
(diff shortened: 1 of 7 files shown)


esden (approved):
LGTM, great to see that we are getting closer to having non halting memory access. :)

PR #1967, opened 2024-10-17
Remove the makefile based build system

esden:
## Detailed description

As we are already using meson for a while with pretty good success and the gnu makefile based system is starting to require maintenance effort. It is time to retire the makefiles.

If it turns out that we need to bring them back for some reason it should still be relatively easy to do.

Still this is an end of an era. :)

As a side-effect this PR also deletes the libopencm3 submodule as it is not needed any more.


src/target/flashstub/Makefile
dragonmux, line 1: This one Makefile needs to remain for now to allow rebuilding the Flash stubs which is only done on demand (manually). Aki actually had an idea on how to replace this nicely yesterday, which we'll do in the upcoming feature/better-targets-control as that'll be a welcome removal and allow the stubs to automatically be rebuilt as necessary.

dragonmux (changes requested):
There's one Makefile that's being removed which should not be for the moment and isn't called by the outgoing build system (it was always only ever used manually). Beyond that, this is all looking good and while melancholy to see the old system go, it is indeed starting to cause some aggravating problems and becoming a painful maintenance burden.

src/target/flashstub/Makefile
esden: Ahh yeah... good point. Sorry, was bit too eager removing the Makefiles. I will bring it back.

sidprice:
(Almost) End of an era! But what an era!

src/target/flashstub/Makefile
esden: Ok, I brought back the stub makefile. It should be removed as part of your better-targets-control PR. :)

dragonmux (approved):
This all LGTM, merging.

PR #1981, opened 2024-10-29
Feature: ADIv6 JTAG support

### LINKED ISSUE #1970: Cortex M55/adiv6 support
### Discussed in https://github.com/blackmagic-debug/blackmagic/discussions/1969

<div type='discussions-op-text'>

<sup>Originally posted by **florg-32** October 18, 2024</sup>
Hey guys,

I have a Cortex M55 chip I'd like to debug with my BMP. Should this be supported already (on nightly builds)? 

My gdb `monitor jtag_scan` does not find the device, voltage is detected correctly though and resetting works. The debug app outputs the following, which I don't really understand; shouldn't it be ADIv6?

```
❯ blackmagic -tjv8
Black Magic Debug App d896778
Using 1d50:6018 81D749A3 Black Magic Debug
 Black Magic Probe d896778
Running in Test Mode
Target voltage: 1.8V
Speed set to 1.951MHz for JTAG
Resetting TAP
Change state to Shift-DR
Scanning out ID codes
Return to Run-Test/Idle
Change state to Shift-IR
Scanning out IRs
Return to Run-Test/Idle
Change state to Shift-DR
Return to Run-Test/Idle
ID code 0x4ba06477: ADIv5 JTAG-DP port.
Abort: 00000004
Failed to read DPIDR
Given target number 1 not available max 0
```
Thanks,
Florian</div>

dragonmux:
## Detailed description

In this PR we address JTAG-based ADIv6 support, as we didn't have an example of a JTAG target for this when doing the initial support, but one has turned up via a user, allowing the ADIv6 support to be sorted out and completed.

This has seen a couple of challenges to get it to a usable state - the first of which is that the JTAG ACK values change between ADIv5 and ADIv6, with the latter splitting the OK and FAULT ACKs into different values (yay!). This means that the JTAG low-level code for ADI now has version-dependant behaviour, requiring a new remote protocol packet to properly handle.

The second challenge is that the JTAG IDCode version bits now have additional valid values, which it turns out come from the CoreSight SoC TRMs, that determine if the TAP is for a JTAG-DPv0 or a JTAG-DPv1. The new JTAG-DP implements the DPIDR field but the DP registers do not reset with the state machine, so requiring some initial setup to ensure we read the right thing to determine ID. This also comes with needing to deal with devices that implement the Dormant State part of the JTAG <-> SWD switch-over state machine.

This PR includes the changes needed to correctly identify the Cortex-M55 debug components and some more of the CoreSight SoC-600(M) debug components and allow debugging Cortex-M55 processors.

## Closing issues

Fixes #1970


esden (approved):
LGTM

PR #1983, opened 2024-10-30
fix/ctxlink/platform

sidprice:
This PR addresses a crash when reporting the battery voltage in the response to the "mon swd" command. The buffer being used was not initailized before each use and multiple reports were being concatinated.


dragonmux (approved):
Ah, good catch - this all LGTM. The only issue we see is the commit message needs a small adjustment for contribution guidelines - `ctxlink/platform: Fix crash in platform_target_voltage() due to the buffer not getting cleared` for example.

PR #1988, opened 2024-11-01
Feature: Implement support for CoreSight ROM table format

marysaka:
## Detailed description

This is partially based on adiv6 CoreSight ROM parser

This was tested against a [Solitude AML-S905D3-CC](https://libre.computer/products/aml-s905d3-cc/).


src/target/adi.c
```diff
@@ -586,14 +591,147 @@ static void adi_parse_adi_rom_table(adiv5_access_port_s *const ap, const target_
...
+	adiv5_access_port_s *const ap, const uint8_t rom_format, const target_addr64_t entry_address)
+{
+	const uint32_t entry_lower = adi_mem_read32(ap, entry_address);
+	if (rom_format == ADIV5_ROM_CLASS9_DEVID_FORMAT_32BIT)
+		return entry_lower;
+	const uint32_t entry_upper = adi_mem_read32(ap, entry_address + 4U);
+	return ((uint64_t)entry_upper << 32U) | (uint64_t)entry_lower;
+}
+
+static void adi_parse_adi_rom_table_class9(adiv5_access_port_s *const ap, const target_addr64_t base_address,
```
dragonmux, line 646: This isn't an ADI ROM table, but a CoreSight ROM table - please use the nomenclature `adi_parse_coresight_rom_table` (this also removes the need to rename the other parser function above).

While the ADIv6 spec defines how these tables work, the documentation in question is lifted from the CoreSight specification and TRMs, hence why this particular choice in nomenclature.

src/target/adi.c
```diff
@@ -565,7 +570,7 @@ static void adi_parse_adi_rom_table(adiv5_access_port_s *const ap, const target_
 	}
 
 	/* Check SYSMEM bit */
-	const bool memtype = adi_mem_read32(ap, base_address + ADIV5_ROM_MEMTYPE) & ADIV5_ROM_MEMTYPE_SYSMEM;
+	const bool memtype = adi_mem_read32(ap, base_address + ADIV5_ROM_CLASS1_MEMTYPE) & ADIV5_ROM_CLASS1_MEMTYPE_SYSMEM;
```
dragonmux, line 573: As per the note below, we advise using the nomenclature `ADIV5_ROM_ADI_MEMTYPE` here, with `ADIV5_ROM_CORESIGHT_` used for the parser below. "Class 1" ROMs are only defined by the debug interface spec and are not CoreSight ROMs, while "Class 9" are CoreSight ROMs and they have a version number associated with them too (v0 right now, but ARM left space for a v1, v2, etc).

src/target/adi.c
```diff
@@ -699,5 +837,21 @@ void adi_ap_component_probe(
 				break;
 			}
 		}
+
+		/* Start out by making a fake AP to use for all the reads */
+		adiv5_access_port_s base_ap = {
+			.dp = ap->dp,
+			.base = base_address,
+		};
```
dragonmux, line 845: You should be able to skip this (it is used in ADIv6 to simulate an AP for the debug resource bus accesses that that parser has to make) and just pass `ap` to the parse function below.

dragonmux (changes requested):
This is looking pretty good, though there are a couple of things that need addressing along with the conversation we've had on Discord about code sharing between ADIv5 and v6 where possible. We look forward to seeing the refresh of this.

src/target/adi.c
```diff
@@ -103,6 +103,7 @@ static const arm_coresight_component_s arm_component_lut[] = {
 	{0x4c4, 0x00, 0, aa_nosupport, cidc_unknown, ARM_COMPONENT_STR("Cortex-M4 ROM", "(Cortex-M4 ROM)")},
 	{0x4c7, 0x00, 0, aa_nosupport, cidc_unknown, ARM_COMPONENT_STR("Cortex-M7 PPB", "(Cortex-M7 PPB ROM Table)")},
 	{0x4c8, 0x00, 0, aa_nosupport, cidc_unknown, ARM_COMPONENT_STR("Cortex-M7 ROM", "(Cortex-M7 ROM)")},
+	{0x4e4, 0x00, 0x0af7, aa_nosupport, cidc_dc, ARM_COMPONENT_STR("CoreSight ROM", "(ROM Table)")},
```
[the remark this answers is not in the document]
marysaka: It still does show up in my logs :thinking:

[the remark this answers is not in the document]
dragonmux: Ahh, just spotted why too - hmm 🤔 actually, thinking about it we won't include a review note for why because it probably actually makes some sense to display that we found a ROM Table like any other component.. there might need to be a consistency change for ADIv6's version of the parser too then to make it make more sense.

src/target/adi.c
```diff
@@ -589,17 +595,150 @@ static void adi_parse_adi_rom_table(adiv5_access_port_s *const ap, const target_
...
+			}
+		}
+	}
+	/* Regardless of what happened, extract whether system reset is supported this way */
+	if (pridr0 & CORESIGHT_ROM_PRIDR0_HAS_SYS_RESET_REQ)
+		ap->flags |= ADIV6_DP_FLAGS_HAS_SYSRESETREQ;
+	return true;
+}
+
+static inline uint64_t adi_read_rom_class_9_entry(
```
dragonmux, line 640: This function needs its nomenclature updated - `adi_read_coresight_rom_entry`?

marysaka: Oops, should be fixed :+1:

dragonmux (approved):
LGTM, merging once rebased on `main`. Thank you for the contribution!

PR #1989, opened 2024-11-02
Fix: Old firmware BMP identification via libusb

dragonmux:
## Detailed description

In this PR we address an issue uncovered while poking at a BMP running pre-v1.7 firmware where the version string is not embedded in the product string. With probes in this configuration, BMDA will correctly find them but then display "Probe" as the version string and "Black Magic" as the product string due to getting handling of the versionless string wrong.

We address this by porting the Windows BMP-only probe discovery product string parsing to the libusb probe discovery logic, which also allows us to properly address probes that use alternative platforms such as Blue Pills and Black Pills and which may not have a version string if their firmware is too old, thus solving the problem.


esden (approved):
LGTM

PR #1992, opened 2024-11-04
Feature/misc meson improvements

perigoso:
## Detailed description

This is a minor cleanup/simplification of the meson configuration around the target enabling

This should result in *no* behavior changes, except for a small, but very intentional one.

Right now when *manually* configuring the build, i.e. not using the pre-made cross-files for the probes, the default behaviour was to enable *all* targets by default, resulting in the need to safe-guard against special cases for certain probes to reduce user confusion.

This changes things such that by default when *manually* configuring the build *no* targets are enabled by default, a helpful error message should be displayed about it (improvement over the previous compilation error).

I think this makes sense because this is to me an "advanced user" usage and reduces the maintenance burden on the build system.


perigoso:
Whoops, i missed that this needed rebasing

PR #1996, opened 2024-11-10
Fix: AT32F43x cleanup

ALTracer:
## Detailed description

* The `bank_split` field of `struct at32f43_flash` can be eliminated from `_probe`/`_add_flash` if I can use a different method of accessing second bank in `_mass_erase`, checking that `target->flash->next` list has a non-NULL entry. No other functions need it because BMD API guarantees no operations span multiple banks. *This change was rebased past AT32F405 merge.*
* Also, hook the STM32-generic UID printing command to this driver and introduce a meson dependency between these TUs. Move the macros for "engineering bytes"/UID into its own group.

Verified on AT-START-F437 to not break flashloading, a firmware spanning two banks was used.
This should bring a minor reduction to target_flash priv size, and it's a preparation to refactoring this driver to carry USD_BASE and FPEC_BASE in that priv such that targets in stm32f1.c can be migrated here.

## Closing issues

### DIFF
--- a/src/target/at32f43x.c
+++ b/src/target/at32f43x.c
@@ -35,11 +35,14 @@
 #include "target.h"
 #include "target_internal.h"
 #include "cortexm.h"
+#include "stm32_common.h"
 
 static bool at32f43_cmd_option(target_s *target, int argc, const char **argv);
+static bool at32f43_cmd_uid(target_s *target, int argc, const char **argv);
 
 const command_s at32f43_cmd_list[] = {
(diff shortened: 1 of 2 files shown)


dragonmux (approved):
This all looks reasonable, merging once rebased on `main`. In regards the mass erase point, once Perigoso's #1619 is merged, you'll be able to specify the mass erase function as `flash->mass_erase`, setting it to the bank mass erase function, and eliminating the custom mass erase implementation. That'll solve the bank split headache properly.

PR #1998, opened 2024-11-10
crc32: Buffer up log output to avoid prefix-chunking, fix `func`

ALTracer:
## Detailed description

* This is a fix to one of my edited/contributed features regarding `compare-sections` via `qCRC`.
* There are three existing problems, which are not critical to upstream BMD here.
* This PR updates crc32 dispatch code to deal with them.

1. Because of #1713 and requirements to support Windows and MacOS (cross-platform portable software), the `__func__` portion of log became identical as collateral damage. This change restores the behaviour intended in #1708, so that we can tell apart speed reports from BMF+`ENABLE_DEBUG` and BMDA.
2. On out-of-tree platforms like Farpatch which use BMD as a git submodule directly, custom logger implementations format every `DEBUG_INFO()` call etc. by prepending loglevel and timestamp, so the 2-3 piece logmessage gets broken and mixed. This change buffers the optional output pieces into a zeroed stack buffer via `snprintf()` to avoid that.
3. Xtensa Farpatch crashes on every divide-by-zero, which can happen here when <1 milliseconds are reported elapsed by corresponding platform timing code. BMF on Cortex-M3 does not trap integer divide-by-zero into UsageFault, so it was never affected. BMDA usually gets more than 1ms due to USB FS delays and all the layer overheads. This change skips the speed report by checking for both length and time to be big enough.

## Closing issues
Fixes some issues in Farpatch.


src/crc32.c
```diff
@@ -193,20 +193,27 @@ bool bmd_crc32(target_s *const target, uint32_t *const result, const uint32_t ba
...
+	loglen = snprintf(&logbuf[loglen], logcap, "%s: 0x%08" PRIx32 "+%" PRIu32 " -> %" PRIu32 "ms", func, base,
+		(uint32_t)len, time_elapsed);
+	if (len >= 512U && time_elapsed > 0U) {
 		const uint32_t speed = len * 1000U / time_elapsed / 1024U;
-		DEBUG_INFO(", %" PRIu32 " KiB/s", speed);
+		loglen += snprintf(&logbuf[loglen], logcap - loglen, ", %" PRIu32 " KiB/s", speed);
 	}
-	DEBUG_INFO("\n");
+	loglen += snprintf(&logbuf[loglen], logcap - loglen, "\n");
+	DEBUG_INFO("%s", logbuf);
```
xobs, line 216: Could you roll the "\n" into the DEBUG_INFO() here? I could see there being a case where we replace all instances of `DEBUG_x("...\n"` with something else, assuming BMP ever moves to a logging infrastructure that adds newlines.

xobs:
I like this change! It improves readability when doing checksum benchmarking.

src/crc32.c
ALTracer: Sounds good, applied. So log entries would stored in some circular dmesg buffer, separated with `\0` not `\n`, and the consuming function would format numeric millisecond timestamps and potentially multiline messages with e.g. `\r\n` if needed? I don't remember how NuttX ramlog driver works, but the formatting is up to configurable `syslog()`, not callsites; and in BMF we didn't have that.

dragonmux:
Please could you rebase this on `main` and we'll look at getting this reviewed as soon as we can.

src/crc32.c
```diff
@@ -193,20 +193,28 @@ bool bmd_crc32(target_s *const target, uint32_t *const result, const uint32_t ba
...
+	/* "generic_crc32: 0x08000110+75272 -> 1353ms, 54 KiB/s" */
+	/* "stm32_crc32: 0x08000110+75272 -> 237ms, 310 KiB/s" */
 	const uint32_t end_time = platform_time_ms();
 	const uint32_t time_elapsed = end_time - start_time;
-	DEBUG_INFO("%s: 0x%08" PRIx32 "+%" PRIu32 " -> %" PRIu32 "ms", __func__, base, (uint32_t)len, time_elapsed);
-	if (len >= 512U) {
+	char logbuf[64] = {0};
+	const size_t logcap = 64;
+	size_t loglen = 0;
+	loglen = snprintf(&logbuf[loglen], logcap, "%s: 0x%08" PRIx32 "+%" PRIu32 " -> %" PRIu32 "ms", func, base,
```
dragonmux, line 209: because of how this is used, we would strongly suggest expressing the first parameter as `logbuf + loglen` please - it won't affect the operation performed but is a little less brain-melting to follow.

Do note that the result of `snprintf()` is `int`/`ssize_t`, so the result *must* be cast to `size_t` to prevent a signed-unsigned conversion warning. The conversion is fine here, this is just to fix the lint this introduces.

src/crc32.c
```diff
@@ -193,20 +193,28 @@ bool bmd_crc32(target_s *const target, uint32_t *const result, const uint32_t ba
...
+		(uint32_t)len, time_elapsed);
+	if (len >= 512U && time_elapsed > 0U) {
 		const uint32_t speed = len * 1000U / time_elapsed / 1024U;
-		DEBUG_INFO(", %" PRIu32 " KiB/s", speed);
+		loglen += snprintf(&logbuf[loglen], logcap - loglen, ", %" PRIu32 " KiB/s", speed);
```
dragonmux, line 213: Same comments on this one.

src/crc32.c
```diff
@@ -193,20 +193,28 @@ bool bmd_crc32(target_s *const target, uint32_t *const result, const uint32_t ba
...
-	/* "stm32_crc32: 08000110+75272 -> 237ms, 310 KiB/s" */
+	/* "generic_crc32: 0x08000110+75272 -> 1353ms, 54 KiB/s" */
+	/* "stm32_crc32: 0x08000110+75272 -> 237ms, 310 KiB/s" */
 	const uint32_t end_time = platform_time_ms();
 	const uint32_t time_elapsed = end_time - start_time;
-	DEBUG_INFO("%s: 0x%08" PRIx32 "+%" PRIu32 " -> %" PRIu32 "ms", __func__, base, (uint32_t)len, time_elapsed);
-	if (len >= 512U) {
+	char logbuf[64] = {0};
+	const size_t logcap = 64;
+	size_t loglen = 0;
```
dragonmux, line 208: This variable declaration can be merged into the line below for what it's worth

src/crc32.c
```diff
@@ -193,20 +193,28 @@ bool bmd_crc32(target_s *const target, uint32_t *const result, const uint32_t ba
...
-	/* "generic_crc32: 08000110+75272 -> 1353ms, 54 KiB/s" */
-	/* "stm32_crc32: 08000110+75272 -> 237ms, 310 KiB/s" */
+	/* "generic_crc32: 0x08000110+75272 -> 1353ms, 54 KiB/s" */
+	/* "stm32_crc32: 0x08000110+75272 -> 237ms, 310 KiB/s" */
 	const uint32_t end_time = platform_time_ms();
 	const uint32_t time_elapsed = end_time - start_time;
-	DEBUG_INFO("%s: 0x%08" PRIx32 "+%" PRIu32 " -> %" PRIu32 "ms", __func__, base, (uint32_t)len, time_elapsed);
-	if (len >= 512U) {
+	char logbuf[64] = {0};
+	const size_t logcap = 64;
```
dragonmux, line 207: Perhaps define this in terms of `ARRAY_LENGTH()`? that way if logbuf ever changes, this'll automatically update.

dragonmux (commented):
Couple of notes in doing a review - but with them fixed, this looks basically ready to merge. Good job!

src/crc32.c
ALTracer: Applied. What does GDB send in qCRC for 64-bit addresses? BMD only supports 32-bit, per function signatures up this callsite. Maybe increase the stack buffer size?... but stack depth of this function (and its callee which has a 1024 byte stack buffer) is pretty big in BMF already.

src/crc32.c
ALTracer: `snprintf` returns int, so maybe use `ssize_t` in the first place for loglen, logcap?

src/crc32.c
ALTracer: If I do that here, then I hit `-Wuninitialized`:
`	ssize_t loglen = snprintf(logbuf + loglen, logcap, ...);`
but the first print starts from beginning of string, so I can omit the addition.
Likewise, cppcheck complains that second assignment is unreadVariable, which is also fair -- do I even need to capture the result?

src/crc32.c
dragonmux: This is a good question - we don't actually know; there's quite likely a bug here in that scenario to be dealt with.

src/crc32.c
dragonmux: With the second `+=`, it is technically, however we suggest leaving it in so further checking can be done down the road if it turns out to be required.

ALTracer: clang-tidy also says about snprintf that `warning: the value returned by this function should be used [cert-err33-c]` so I'll just store it and (void) it later -- keeping as-is for now.

dragonmux (approved):

While there's a known issue with 64-bit targets here, we're not going to have this block merging this as that's a separate issue that'll take a pile more work to fix and so should be a separate PR.

PR #2000, opened 2024-11-12
puya: add support for PY32F002B

silbe:
## Detailed description

The PY32F002B series uses different DBGMCU IDCODE values and there does not seem to be a register containing RAM and flash size.

While the procedure for preparing flash access (copying parameters from factory-programmed EPPARAx registers to flash peripheral registers) is the same, the bit allocation inside the registers is slightly different.

The structure and values of the DBGMCU IDCODE register are undocumented but the vendor SDK splits it into DEV_ID and REV_ID fields. We use the DEV_ID fields of the IDCODE values observed on PY32F002AW15U and PY32F002BF15P6 to distinguish between the two families. An internet search shows that at least the PY32F002BW15 uses the same value as the PY32F002BF15P6. The full IDCODE values are retained as comments to make it easier to fix the code later if it turns out the DEV_ID/REV_ID split is incorrect.


silbe:
Rebased on `main`, fixed style issue shown by clang-format. No functional code changes; test still successful.

src/target/puya.c
```diff
@@ -114,18 +131,32 @@ bool puya_probe(target_s *target)
...
+	}
+	case PUYA_DEV_ID_PY32F002B:
+		/*
+		 * 0x1fff0ffc contains 0; did not find any other location that looks like it might contain the flash
+		 * and RAM sizes. We'll hard-code the datasheet values for now. Both flash size and RAM size actually
+		 * match the datasheet value, unlike PY32F002A which (sometimes?) has more RAM and flash than
+		 * documented.
+		 */
+		flash_size = 24 * 1024;
+		ram_size = 3 * 1024;
```
dragonmux, line 151: Please suffix all these constants with `U` so they are defined unsigned and don't do a signed-unsigned conversion on assignment.

src/target/puya.c
```diff
@@ -99,6 +100,22 @@
...
+#define PUYA_DBG_IDCODE_REV_ID_SHIFT 16U
+#define PUYA_DBG_IDCODE_REV_ID_MASK  0xffffU
+
+/* On PY32F002AW15U an IDCODE value of 0x60001000 was observed */
+#define PUYA_DEV_ID_PY32F002A 0x000
+/*
+ * On PY32F002BF15P an IDCODE value of 0x20220064 was observed. Internet search shows the same value is used on
+ * PY32F002BW15.
+ */
+#define PUYA_DEV_ID_PY32F002B 0x064
```
dragonmux, line 118: Please suffix these with `U` so they are defined unsigned.

src/target/puya.c
```diff
@@ -151,33 +182,63 @@ static bool puya_flash_prepare(target_flash_s *flash)
...
+	case PUYA_DEV_ID_PY32F002B:
+		target_mem32_write32(flash->t, PUYA_FLASH_TS0, eppara0 & 0x1ffU);
+		target_mem32_write32(flash->t, PUYA_FLASH_TS1, (eppara0 >> 18U) & 0x3ffU);
+		target_mem32_write32(flash->t, PUYA_FLASH_TS3, (eppara0 >> 9U) & 0x1ffU);
+		target_mem32_write32(flash->t, PUYA_FLASH_TS2P, eppara1 & 0x1ffU);
+		target_mem32_write32(flash->t, PUYA_FLASH_TPS3, (eppara1 >> 16U) & 0xfffU);
+		target_mem32_write32(flash->t, PUYA_FLASH_PERTPE, eppara2 & 0x3ffffU);
+		target_mem32_write32(flash->t, PUYA_FLASH_SMERTPE, eppara3 & 0x3ffffU);
+		target_mem32_write32(flash->t, PUYA_FLASH_PRGTPE, eppara4 & 0xffffU);
+		target_mem32_write32(flash->t, PUYA_FLASH_PRETPE, (eppara4 >> 16U) & 0x3fffU);
```
dragonmux, line 239: With the number of magic numbers involved, it might make sense to name them via `#defines`'s.

dragonmux (commented):
We love the reduction in unnamed magic numbers! Very curious that they chose to rework several registers' layouts like that too.. seems very odd. Our main review notes are on some signed-but-should-be-unsigned constants and the other is on a further reduction of magic numbers. With those taken care of we're happy to merge this.

src/target/puya.c
silbe: Fixed locally; will push together with the other changes once they are done.

src/target/puya.c
silbe: Fixed locally; will push together with the other changes once they are done.

src/target/puya.c
silbe: In progress. While doing so I noticed that the existing code uses additional bits from EPPARA4 for PRETPE (datasheet says 26:16 are PRETPE so the mask should be 0x7ff but the code uses 0x3fff). @ArcaneNibble was this on purpose? If not, would you be willing to test my changes if I fix this to match the datasheet values? I don't have a PY32F002A myself.

silbe: Changed locally to use defines named like `PY32F002A_EPPARA0_TS0_SHIFT` / `…_MASK` and tested successfully on PY32F002B. Waiting for reply from @ArcaneNibble before pushing.

dragonmux: Excellent, let's hope that they get back to you on this query fairly soon then 🙂

silbe: It's been a week without feedback from @ArcaneNibble so we'll probably need to continue without their assistance.

I've had a closer look at this and it's quite a mess. The PY32F002A Reference Manual says EPPARA4 26:16 (11 bit) contain the `PRETPE` value but that it's 12 bit wide (11:0). Also `FLASH_PRETPE` is 14 bit wide (13:0) and must be initialised from the value in `EPPARA4`. So there are 3 different bit widths documented for what should be the same value. In addition the PY32F002A flash driver (`PY32F0xx_20.FLM`) contained in the [Keil Device Family Pack v1.2.1](https://www.puyasemi.com/download_path/%E5%B7%A5%E5%85%B7/MCU%20%E5%BE%AE%E5%A4%84%E7%90%86%E5%99%A8/PY32F0xx_Keil_DFP_V1.2.1.zip) ([archive.org copy](https://web.archive.org/web/20230205071913if_/https://www.puyasemi.com/uploadfiles/2022/11/PY-MCU%E8%B5%84%E6%96%99-20221117.rar)) provided by Puya stores the full upper 16 bits of `EPPARA4` into `FLASH_PRETPE`. So I'm going to assume that the reserved bits are 0 and will keep the 0x3fff mask which matches the documented `FLASH_PRETPE` size (14 bits) and has been tested before.

dragonmux: Given the conflicting documentation and such, that seems like a solid plan. If it turns out to be wrong in some way, a further PR can be opened and merged to fix it - so no worries there.

ArcaneNibble: Hi, our personal life has been a bit on fire. Apologies. I hope to review this in a few hours.

ArcaneNibble: cc our partner @jediminer543 to see if she has any spoons to look at it

ArcaneNibble: Okay, we've gone and double-checked:

* the mask 0x3fff (14 bits) is what the vendor HAL uses (in `py32f0xx_hal_flash.h`)
* the mask 0xffff (16 bits) is what the vendor DFP uses (in `FlashPrg.c`), but we did not look in this code or notice it when originally implementing support
* the "PY32F030 Reference Manual V1.1" (English) indeed has all of the inconsistencies pointed out above. This appears to be the only source floating around on my hard drive which indicates anything fewer than 14 bits
* the "PY32F002A 参考手册 V1.1" (Chinese) is internally consistent, indicating 14 bits everywhere

This appears... most consistent with keeping the mask as 0x3fff for now?

<hr>

As a separate issue, the PY32F002A is widely suspected to contain the same die as the PY32F030. It's not clear whether or not the 002A and the 030 would have different `DEV_ID` in the `IDCODE` (i.e. does `IDCODE` identify the die? the final packaged part? something else entirely?). I wouldn't be entirely confident in asserting that `0x60001000` is *definitely* the 002A, but I don't know how you'd want to handle this without additional information (I don't have any 030 around)

Incidentally, the 002A seems to have disappeared from the vendor's website entirely (the 002B is still there, but the 002B is suspected to be a different die). I can't find any PCNs regarding it (or any PCNs at all), so I don't know what to make of this.

ArcaneNibble: Some hardware tests on a PY32F002A that happened to be on hand:

* `FLASH_PRETPE` contains 14 settable bits, with the rest being ignored
```
(gdb) set *(int*)0x40022120 = 0xffffffff
(gdb) x/1xw 0x40022120
0x40022120:	0x00003fff
```
* the calibration value at `0x1fff0f7c` (aka `EPPARA4` for 24 MHz) contains the value `0x12c05dc0`. This means that 11 or 12 bits is *probably not correct*

In conclusion, *14 bits* is likely the correct width for the `PRETPE` calibration value.

dragonmux: That's excellent - hopefully that gets silbe sorted out on what needs to happen here so this PR can get finished up then 😄

silbe: @ArcaneNibble Thanks a lot for spending the time and all the information! Good to know that there are differences between the English and Chinese versions of the datasheet even regarding bit assignments in the registers. I just checked for the PY32F002B and indeed the Chinese version is more consistent. :confused: 

Agree that we should stick with 14 bits for the PY32F002A. Regarding the chip/driver identification string, until somebody tries this with a PY32F030 we won't know a) what its ID code is and b) whether it will actually work with the current code. So displaying "PY32F002A" would be the best match for now IMO (but I don't feel strongly about it). It also has the highest chance of getting reports from users when PY32F030 happens to work (which means we can ask them for the ID code that was matched).

silbe: FYI: After adding the comment regarding `EPPARA4` for PY32F002A I retested mostly pro forma and ran into lots of trouble. Even the previous version which worked fine last week now shows intermittent failures. Of course, the less debug output is provided, the more likely it is to fail, with direct usage of the "serial" interface from GDB failing quite often (so I guess it's timing related in some way).
Configuring the SWD frequency to around 100kHz (actual frequency around 190kHz) did not help.
I also tried not writing the timing parameters in `puya_flash_prepare()` as both `PY32F002Bxx_24.FLM` and `CMSIS/Flash/PY32F002Bxx/FlashPrg.c` in the Puya Keil DFP don't do this either, but the effects are the same.

Guess I need to attach a logic analyser to find out what failed. It'll probably be a while until I get around to that so I will mark this MR as draft and push the current code in case somebody else wants to pick it up in the meantime.

silbe: BTW, thanks for the review so far!

PR #2005, opened 2024-11-22
Fix: Shrink SRAM sizes of STM32F0

ALTracer:
## Detailed description

* Not a feature.
* The existing problem is incorrect SRAM size declared in stm32f1.c for STM32F0 families, 20 KiB, which matches neither of them (but might have been copied from STM32F103). This interacts badly with RTT auto search.
* This PR adjusts ram_size values to whatever documented in reference manuals, just like flash size.

Tested on 32F072B-Disco to no longer trigger lots of `SWD access resulted in fault, retrying` every time mainloop calls `poll_rtt()` and that crosses 0x20004000 boundary of a 16 KiB SRAM F072RB device to "search" for control blocks up to 0x20005000.
I think also the DUT BMF (which I ported there as a f072-if derivative) got stuck when this bug was triggering, judging by the morse blinky stopping.
My problem is specifically advertising 20 KiB on a 16 KiB device; for 32 KiB devices it's probably less of an issue (like stack not readable). I know there is `mon rtt ram 0x20000000 0x20004000` as a workaround, and it indeed works, but BMD is supposed to be automatic, with no configuration.

However, on per grid pics from ST website there are more variations on SRAM/Flash than 2 TRMs contain, and I did not feel like downloading all 12 datasheets to try to match the 5 part IDs to them, but note that there is `F_SIZE` at 0x1FFF_F7CC for a 16-bit flash size in kibibytes that someone could try to read as a better alternative to static flash capacity associations.

Size increase on `native` is +40 bytes because the constants are no longer identical: 4,6,8,16,32 not 20.

## Closing issues

### DIFF
--- a/src/target/stm32f1.c
+++ b/src/target/stm32f1.c
@@ -965,35 +965,35 @@ bool stm32f1_probe(target_s *target)
 
 	case 0x444U: /* STM32F03 RM0091 Rev. 7, STM32F030x[4|6] RM0360 Rev. 4 */
 		target->driver = "STM32F03";
-		ram_size = 0x5000;
+		ram_size = 0x1000;
 		flash_size = 0x8000;
 		block_size = 0x400;
 		break;
 
 	case 0x445U: /* STM32F04 RM0091 Rev. 7, STM32F070x6 RM0360 Rev. 4 */
 		target->driver = "STM32F04/F070x6";
-		ram_size = 0x5000;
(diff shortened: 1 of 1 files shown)


dragonmux (approved):
We presume these were taken from the TRMs, so looks good to us - merging once this has been rebased on `main`. Thank you for the contribution!

PR #2007, opened 2024-11-27
ctxlink/wifi: fix the speed issues with the Wi-Fi interface

sidprice:
The reading of the ADC to acquire the target and battery
voltages was being done in a blocking manner and at
a place in the code that blocked the timely servicing
of the Wi-Fi channel.

In this updated code the ADC is read in the Timer service of the Wi-Fi Server.
There are two channels and each channel conversion is started on one timer service and read on the next timer service. Giving four phases of operation.

Also, the ADC reading code is only run every 100ms.

## Closing issues
None.

### DIFF
--- a/src/platforms/ctxlink/platform.c
+++ b/src/platforms/ctxlink/platform.c
@@ -29,6 +29,7 @@
 #include "morse.h"
 #include "exception.h"
 #include <string.h>
+#include <stdatomic.h>
 
 #include <libopencm3/stm32/rcc.h>
 #include <libopencm3/cm3/scb.h>
@@ -67,7 +68,13 @@
 bool last_battery_state = true;
 bool battery_present = false;
 
-static uint32_t input_voltages[2] = {0};
+#define READ_ADC_PERIOD 100 //< ms count
+static uint32_t read_adc_period = READ_ADC_PERIOD;
+
+#define MAX_ADC_PHASE 4
+static uint8_t adc_phase;
+
+_Atomic uint32_t input_voltages[2] = {0};
 static uint8_t adc_channels[] = {CTXLINK_BATTERY_INPUT, CTXLINK_TARGET_VOLTAGE_INPUT}; /// ADC channels used by ctxLink
 
 typedef void (*irq_function_t)(void);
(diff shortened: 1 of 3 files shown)


src/platforms/ctxlink/platform.c
```diff
@@ -139,25 +145,11 @@ void wifi_init(void)
...
-	while (!adc_eoc(ADC1))
-		continue;
-	input_voltages[CTXLINK_ADC_TARGET] = (adc_read_regular(ADC1) * 16U) / 10U;
-}
-
 uint32_t platform_target_voltage_sense(void)
 {
-	return input_voltages[CTXLINK_ADC_TARGET];
+	uint32_t voltage;
+	__atomic_load(&input_voltages[CTXLINK_ADC_TARGET], &voltage, __ATOMIC_RELAXED);
```
dragonmux, line 151: Please include `<stdatomic.h>` and use the [non-underscored](https://en.cppreference.com/w/c/atomic/atomic_load) version of this function which is standardised. It is called in a very similar way, but does not depend on compiler specifics. In this context it would be used as something like:
```c
return atomic_load(input_voltages[CTXLINK_ADC_TARGET], memory_order_relaxed);
```

src/platforms/ctxlink/platform.c
```diff
@@ -330,6 +322,50 @@ bool platform_nrst_get_val(void)
...
+		BMD_FALLTHROUGH;
+	}
+	case 0: {
+		adc_set_regular_sequence(ADC1, 1, &adc_channels[CTXLINK_ADC_TARGET]);
+		adc_start_conversion_regular(ADC1);
+		break;
+	}
+	case 1: {
+		adc_value = (adc_read_regular(ADC1) * 16U) / 10U;
+		__atomic_load(&adc_value, &input_voltages[CTXLINK_ADC_TARGET], __ATOMIC_RELAXED);
```
dragonmux, line 353: This looks like it should be atomic_store() to write `adc_value` into `input_voltages[CTXLINK_ADC_TARGET]`? Same comment about using the `<stdatomic.h>` call. Likewise the one on line 363.

src/platforms/ctxlink/platform.c
```diff
@@ -401,12 +437,12 @@ const char *platform_battery_voltage(void)
...
 bool platform_check_battery_voltage(void)
 {
-	platform_adc_read();
 	//
 	// Is battery connected?
 	//
-	if (input_voltages[CTXLINK_ADC_BATTERY] <= BATTERY_VOLTAGE_1 ||
-		input_voltages[CTXLINK_ADC_BATTERY] >= BATTERY_VOLTAGE_2) {
+	uint32_t voltage;
+	__atomic_load(&input_voltages[CTXLINK_ADC_BATTERY], &voltage, __ATOMIC_RELAXED);
```
dragonmux, line 444: Same comment as line 151.

dragonmux (changes requested):
This definitely gets most of the way there and we like what we see! There are only two major comments here which are about using the GCC-specific compiler intrinsics rather than the stdlib functions, and as a result a couple of parameters winding up backwards which might compile with the builtins, but which shouldn't as it does alter the meaning of the code. Using the stdlib functions prevents this due to how they're typed.

src/platforms/ctxlink/platform.c
sidprice: Nice catches all around, I had forgotten how to use 'atomics' and research took me down the wrong path.

src/platforms/ctxlink/platform.c
sidprice: Fixed

src/platforms/ctxlink/platform.c
sidprice: fixed

src/platforms/ctxlink/platform.c
dragonmux: All good - nice job with the cleanup/fixups!

dragonmux (approved):
This all LGTM, merging. Thank you for the contribution and congrats on the speed doubling (if not slightly more) of the WiFi interface on your ctxLink!

PR #2008, opened 2024-11-30
Cortex-M: Temptative fix for watchpoints so they work on ARMv8-M cores (eg Cortex-M33)

mean00:
## Detailed description

The DWT block is different on Arm V8m compared to older revisions.  As a result the watchpoint(s) function are not working. 
The patch adds a second path dealing with the ArmV8m  .
Tested on cortex m33, check setup watch some_variable  (such as xTickCount on freeRtos)
- without the patch, does not stop
- with the patch, does stop as expected

It was partially tested, i dont have other chip with that DWT block. Only test write watchpoints.
As far as i can tell, watchpoints are still working with older chips (cortex M4)

### DIFF
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -1063,6 +1063,25 @@ static uint32_t cortexm_dwt_func(target_s *target, target_breakwatch_e type)
 	}
 }
 
+static uint32_t cortexm_dwtv2_func(target_breakwatch_e type, size_t len)
+{
+	uint32_t value = CORTEXM_DWTv2_FUNC_ACTION_DEBUG_EVENT | CORTEXM_DWTv2_FUNC_LEN_VALUE(len);
+	switch (type) {
+	case TARGET_WATCH_WRITE:
+		value |= CORTEXM_DWTv2_FUNC_MATCH_WRITE;
+		break;
+	case TARGET_WATCH_READ:
+		value |= CORTEXM_DWTv2_FUNC_MATCH_READ;
+		break;
+	case TARGET_WATCH_ACCESS:
+		value |= CORTEXM_DWTv2_FUNC_MATCH_ACCESS;
+		break;
+	default:
+		return 0U;
+	}
+	return value;
+}
+
 static int cortexm_breakwatch_set(target_s *target, breakwatch_s *breakwatch)
 {
 	cortexm_priv_s *priv = target->priv;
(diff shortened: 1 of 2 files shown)


src/target/cortexm.c
```diff
@@ -1062,6 +1062,25 @@ static uint32_t cortexm_dwt_func(target_s *target, target_breakwatch_e type)
 	}
 }
 
+static uint32_t cortexm_dwt_func_v8m(target_s *target, target_breakwatch_e type, size_t len)
+{
+	uint32_t x = CORTEXM_DWT_FUNC_ACTION_DEBUG_EVENT_V8M;
```
dragonmux, line 1067: Please don't use single letter variable names - something like `result` or `func` would be better here. Thank you for using `uint32_t` though!

src/target/cortexm.c
```diff
@@ -1062,6 +1062,25 @@ static uint32_t cortexm_dwt_func(target_s *target, target_breakwatch_e type)
...
+	case TARGET_WATCH_WRITE:
+		x |= CORTEXM_DWT_FUNC_MATCH_WRITE_V8M;
+		break;
+	case TARGET_WATCH_READ:
+		x |= CORTEXM_DWT_FUNC_MATCH_READ_V8M;
+		break;
+	case TARGET_WATCH_ACCESS:
+		x |= CORTEXM_DWT_FUNC_MATCH_ACCESS_V8M;
+	default:
+		return -1;
```
dragonmux, line 1078: This does an implementation defined conversion, please use `UINT32_MAX` instead or a result like `0U` for less code generation.

src/target/cortexm.c
```diff
@@ -1062,6 +1062,25 @@ static uint32_t cortexm_dwt_func(target_s *target, target_breakwatch_e type)
 	}
 }
 
+static uint32_t cortexm_dwt_func_v8m(target_s *target, target_breakwatch_e type, size_t len)
```
dragonmux, line 1065: Similar comment as in the header about `_v8m` vs `dwtv2` - please prefer the latter.

dragonmux (changes requested):
This is looking really good and complements our own work on ARMv8-M debug on the fix branch `fix/adi-structural-cleanup`. We'd love to get this merged for v2.0 so if you can take care of the small number of review notes (most are to do with nomenclature and we don't think will take that long to deal with as it's mostly just find-replace stuff), we'll get this merged ASAP.

src/target/cortexm.c
```diff
@@ -1062,6 +1062,25 @@ static uint32_t cortexm_dwt_func(target_s *target, target_breakwatch_e type)
...
+	switch (type) {
+	case TARGET_WATCH_WRITE:
+		x |= CORTEXM_DWT_FUNC_MATCH_WRITE_V8M;
+		break;
+	case TARGET_WATCH_READ:
+		x |= CORTEXM_DWT_FUNC_MATCH_READ_V8M;
+		break;
+	case TARGET_WATCH_ACCESS:
+		x |= CORTEXM_DWT_FUNC_MATCH_ACCESS_V8M;
+	default:
```
dragonmux, line 1077: There's a missing `break;` before this, which is what's got CI hung up.

mean00:
Thank you for the comments, will update it as soon as i can

mean00:
I think i took all the comments into account.
If it's ok i'll squash the commits
Thank you

dragonmux (approved):
Yeah, looks good to us. Please let us know once you've squashed the commits down (noting that there should not be a space between the suffix and its' `:`, so `cortexm:` not `cortexm :`) and we'll get this merged.

Thank you for the contribution! Nicely done.

mean00:
Should be okay now
Thank you

dragonmux (approved):

PR #2009, opened 2024-12-01
Add support for nRF54L

zyp:
## Detailed description

This PR adds support for nRF54L.

The functionality of this PR has been tested through an automated test script published [here](https://github.com/zyp/gdb_autotest):

image

## Closing issues

N/A


src/target/nrf54l.c
```diff
@@ -0,0 +1,166 @@
...
+	f->write = rram_write;
+	f->prepare = rram_prepare;
+	f->done = rram_done;
+	f->erased = 0xff;
+	target_add_flash(t, f);
+}
+
+bool nrf54l_probe(target_s *target)
+{
+	DEBUG_INFO("nrf54l_probe:\n");
```
dragonmux, line 84: This should already be output from `cortexm_probe()` as a [`DEBUG_TARGET()` level](https://github.com/blackmagic-debug/blackmagic/blob/b0eb8d02b551e4bcb5d36120dbe126117d9d866f/src/target/target_probe.h#L30) message.

src/target/nrf54l.c
```diff
@@ -0,0 +1,166 @@
...
+
+	adiv5_access_port_s *ap = cortex_ap(target);
+
+	if (ap->dp->version < 2U)
+		return false;
+
+	switch (ap->dp->target_partno) {
+	case 0x1c0:
```
dragonmux, line 92: Please name this constant (having suffixed it to make it unsigned) - something like `ID_NRF45L` would fit with the nomenclature for part numbers.

dragonmux (changes requested):
We've already had a bit of a chat on Discord about the adi.c and adiv5.c changes and how to evolve those, so this review is purely on the new nRF54 support. There are a couple of things but nothing major 🙂 - we look forward to landing this!

zyp:
Review comments addressed, and I also improved the RRAM routines to use the write buffer, and added a benchmark to my test script to measure the performance:
image

src/target/nrf54l.c
```diff
@@ -0,0 +1,189 @@
...
+	(void)len;
+	// RRAM doesn't need to be erased before being written, so we just return ok.
+	return true;
+}
+
+static bool rram_write(target_flash_s *flash, target_addr_t dest, const void *src, size_t len)
+{
+	while (target_mem32_read32(flash->t, NRF54L_RRAMC_BUFSTATUS_READYNEXT) != 1U) {
+		// Wait for rram to be ready for next write.
+	}
```
dragonmux, line 51: Please drop the braces in preference for `continue;` as the loop body as is the style in the rest of the code base. eg:

```suggestion
	// Wait for RRAM to be ready for next write.
	while (target_mem32_read32(flash->t, NRF54L_RRAMC_BUFSTATUS_READYNEXT) != 1U)
		continue;
```

It would be useful to include a `#define` above for what the magic number being checked means.

Same request for `rram_done()` which has the same pattern but for a different register.

src/target/nrf54l.c
```diff
@@ -0,0 +1,189 @@
...
+		target->target_options |= TOPT_INHIBIT_NRST;
+		break;
+	default:
+		return false;
+	}
+
+	target->mass_erase = rram_mass_erase;
+
+	uint32_t info_ram = target_mem32_read32(target, NRF54L_FICR_INFO_RAM);
+	uint32_t info_rram = target_mem32_read32(target, NRF54L_FICR_INFO_RRAM);
```
dragonmux, line 125: These look like they can be made `const`, so lets do so.

src/target/nrf54l.c
```diff
@@ -0,0 +1,189 @@
...
+bool nrf54l_ctrl_ap_probe(adiv5_access_port_s *ap)
+{
+	switch (ap->idr) {
+	case NRF54L_CTRL_AP_IDR_VALUE:
+		break;
+	default:
+		return false;
+	}
+
+	target_s *t = target_new();
```
dragonmux, line 145: It seems this one got missed in all the fixes for single letter variable names. Please use `target` here.

src/target/adi.c
```diff
@@ -380,7 +381,13 @@ bool adi_configure_mem_ap(adiv5_access_port_s *const ap)
 			 * valid debug components on AP0, so we have to have an exception
 			 * for this part family.
 			 */
-		if (ap->dp->target_designer_code != JEP106_MANUFACTURER_TEXAS || ap->base != 0xf0000002U) {
+		if (ap->dp->target_designer_code == JEP106_MANUFACTURER_TEXAS && ap->base == 0xf0000002U)
+			ignore_not_present = true;
+
+		if (ap->dp->target_designer_code == JEP106_MANUFACTURER_NORDIC && ap->base != 0x00000002U)
```
dragonmux, line 387: This should probably be made an `else` to the above `if` (as in `else if`) - we see why you've reworked the logic in this way given the multiple different conditions to test for vendors having done different silly things here, but as this list could get quite long a cascading if-else check would allow the compiler a touch more efficiency in the long term.

dragonmux (commented):
This is definitely a step in the right direction, nice work! Most of our review notes this round are stylistic in nature.

Please `target/` from your commit messages, as per the contribution guidelines, it's not to be included in the suffix.

After those things have been taken care of we think this'll be ready for merge so we can get early nRF54L support in 😄

dragonmux (approved):

PR #2013, opened 2024-12-09
Fix: STM32 DBGMCU attach/detach behaviour

dragonmux:
## Detailed description

In #1882 we introduced control of the WDT halt state to prevent debug sessions getting clobbered by them while the part is intentionally halted. In that PR we stored the state of the DBGMCU on attach, and restored it on detach - which on its face seems like a reasonable heuristic, however.. we worried at the time and have since had that fear confirmed that this doesn't work well because both the user code and the debugger can change the state of the DBGMCU register in question itself to do things like enabling tracing, and users want to be able to detach from the target to leave tracing running.

In the scenario outlined above, however, BMD will disable tracing by restoring the pre-image of the DBGMCU control register, and undo any other freeze configuration by restoring the freeze register for the WDTs and this breaks setups and re-disables tracing. This PR fixes that by always doing a RMW cycle on the registers regardless and dropping the stored state component.

We also include a small fix for some Flash size issues to restore the build on the swlink platform.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -80,7 +80,7 @@ jobs:
 
       # Install BMDA's deps (libftdi1, hidapi-hidraw)
       - name: Install BMDA dependencies
-        run: sudo apt-get -y install libftdi1-dev libhidapi-dev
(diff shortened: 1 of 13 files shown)


esden (approved):
LGTM

PR #2014, opened 2024-12-10
Feature/gdb packet optimization

perigoso:
## Detailed description

This tries to improve GDB packet transmission by removing dynamic allocations thus reduncing memory fragmentation potential.

Proper testing is still required.

### DIFF
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -426,8 +426,8 @@ static void exec_q_rcmd(const char *packet, const size_t length)
 	else if (result == 0)
 		gdb_put_packet_ok();
 	else {
-		const char *const response = "Failed\n";
-		gdb_put_packet_hex(response, strlen(response));
+		static const char response[] = "Failed\n";
+		gdb_put_packet_hex(response, ARRAY_LENGTH(response));
 	}
 }
 
@@ -863,8 +863,8 @@ static void handle_v_packet(const gdb_packet_s *const packet)
 	 * The vMustReplyEmpty is used as a feature test to check how gdbserver handles
 	 * unknown packets, don't print an error message for it.
 	 */
-	static const char *const must_reply_empty = "vMustReplyEmpty";
-	if (strncmp(packet->data, must_reply_empty, strlen(must_reply_empty)) != 0)
+	static const char must_reply_empty[] = "vMustReplyEmpty";
+	if (strncmp(packet->data, must_reply_empty, ARRAY_LENGTH(must_reply_empty) - 1U) != 0)
 		DEBUG_GDB("*** Unsupported packet: %s\n", packet->data);
 #endif
 	gdb_put_packet_empty();
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -83,9 +83,9 @@ target_s *last_target;
 bool gdb_target_running = false;
 static bool gdb_needs_detach_notify = false;
 
-static void handle_q_packet(char *packet, size_t len);
-static void handle_v_packet(char *packet, size_t len);
-static void handle_z_packet(char *packet, size_t len);
+static void handle_q_packet(const char *packet, size_t len);
+static void handle_v_packet(const char *packet, size_t len);
+static void handle_z_packet(const char *packet, size_t len);
 static void handle_kill_target(void);
 
 static void gdb_target_destroy_callback(target_controller_s *tc, target_s *t)
@@ -114,7 +114,8 @@ target_controller_s gdb_controller = {
 };
 
 /* execute gdb remote command stored in 'pbuf'. returns immediately, no busy waiting. */
-int32_t gdb_main_loop(target_controller_s *tc, char *pbuf, size_t pbuf_size, size_t size, bool in_syscall)
+int32_t gdb_main_loop(target_controller_s *const tc, const char *const pbuf, const size_t pbuf_size, const size_t size,
+	const bool in_syscall)
 {
 	bool single_step = false;
 	const char *rest = NULL;
@@ -382,7 +383,7 @@ int32_t gdb_main_loop(target_controller_s *tc, char *pbuf, size_t pbuf_size, siz
 	return 0;
 }
 
-static bool exec_command(char *packet, const size_t length, const cmd_executer_s *exec)
+static bool exec_command(const char *const packet, const size_t length, const cmd_executer_s *exec)
 {
 	while (exec->cmd_prefix) {
 		const size_t prefix_length = strlen(exec->cmd_prefix);
@@ -605,7 +606,7 @@ static void handle_kill_target(void)
 	}
 }
 
-static void handle_q_packet(char *packet, const size_t length)
+static void handle_q_packet(const char *const packet, const size_t length)
 {
 	if (exec_command(packet, length, q_commands))
 		return;
@@ -613,7 +614,7 @@ static void handle_q_packet(char *packet, const size_t length)
 	gdb_putpacketz("");
 }
 
-static void exec_v_attach(const char *packet, const size_t length)
+static void exec_v_attach(const char *const packet, const size_t length)
 {
 	(void)length;
 
@@ -839,7 +840,7 @@ static const cmd_executer_s v_commands[] = {
 	{NULL, NULL},
 };
 
-static void handle_v_packet(char *packet, const size_t plen)
+static void handle_v_packet(const char *const packet, const size_t plen)
 {
 	if (exec_command(packet, plen, v_commands))
 		return;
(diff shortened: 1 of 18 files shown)


src/gdb_packet.c
```diff
@@ -62,6 +71,43 @@ void gdb_set_noackmode(bool enable)
 	noackmode = enable;
 }
 
+#if ENABLE_DEBUG == 1
```
dragonmux, line 74: Might be worth guarding this with `#ifndef DEBUG_GDB_IS_NOOP` specifically

src/gdb_packet.c
```diff
@@ -209,165 +245,210 @@ size_t gdb_getpacket(char *const packet, const size_t size)
...
+	 * This is safe because the escaping is done during the actual packet transmission
+	 * but it will result in a packet larger than what we told GDB we could handle
+	 */
+	if (preamble != NULL && preamble_size > 0) {
+		preamble_size = MIN(preamble_size, GDB_PACKET_BUFFER_SIZE);
+		memcpy(packet_buffer.data, preamble, preamble_size);
+		packet_buffer.size = preamble_size;
+	}
+
+	/* Add the data to the packet buffer and tranform it if needed */
```
dragonmux, line 357: `transform`?

src/gdb_packet.c
```diff
@@ -209,165 +245,210 @@ size_t gdb_getpacket(char *const packet, const size_t size)
...
+     * Program console output packet
+     * See https://sourceware.org/gdb/current/onlinedocs/gdb.html/Stop-Reply-Packets.html#Stop-Reply-Packets
+     * 
+     * Format; ‘O XX…’
+     * ‘XX…’ is hex encoding of ASCII data, to be written as the program’s console output.
+     * 
+     * Can happen at any time while the program is running and the debugger should continue to wait for ‘W’, ‘T’, etc.
+     * This reply is not permitted in non-stop mode.
+     */
+	gdb_putpacket("O", 1U, str, strnlen(str, GDB_OUT_PACKET_MAX_SIZE), true);
```
dragonmux, line 427: Might be worth putting this in a loop to chew through GDB_OUT_PACKET_MAX_SIZE bytes at a time, so we can never wind up with a truncated string sent, something like this:

```suggestion
	const size_t str_len = strlen(str);
	for (size_t offset = 0; offset < str_len; offset += GDB_OUT_PACKET_MAX_SIZE) {
		const size_t amount = MIN(str_len - offset, GDB_OUT_PACKET_MAX_SIZE);
		gdb_putpacket("O", 1U, str + offset, amount, true);
	}
```

dragonmux (commented):
This is looking awesome - really excited to see how this runs and performs. We've got just a handful of things we noticed reading through this PR and with them addressed and some testing done, we'll be happy to merge this PR. This solves some mighty nasty issues with the GDB interface logic and with its stack and heap usage!

src/gdb_main.c
```diff
@@ -361,28 +361,28 @@ int32_t gdb_main_loop(target_controller_s *tc, char *pbuf, size_t pbuf_size, siz
 
 	case 'Q': /* General set packet */
 	case 'q': /* General query packet */
-		handle_q_packet(pbuf, size);
+		handle_q_packet(packet->data, packet->size);
```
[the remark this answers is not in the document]
perigoso: Yep, I did start doing that but in the interest of quickness I left if for later as it was not *required*, might do it if I find a little time

src/gdb_main.c
```diff
@@ -413,9 +413,7 @@ static void exec_q_rcmd(const char *packet, const size_t length)
 		gdb_putpacketz("OK");
 	else {
 		const char *const response = "Failed\n";
-		const size_t response_length = strlen(response);
-		char *pbuf = alloca(response_length * 2 + 1);
-		gdb_putpacket(hexify(pbuf, response, response_length), 2 * response_length);
+		gdb_putpacketx(response, strlen(response));
```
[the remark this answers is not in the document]
perigoso: I'll look into it, did not question what was there before

src/gdb_main.c
```diff
@@ -607,15 +605,15 @@ static void handle_kill_target(void)
...
 }
 
-static void handle_q_packet(char *packet, const size_t length)
+static void handle_q_packet(const char *const packet, const size_t length)
 {
 	if (exec_command(packet, length, q_commands))
 		return;
 	DEBUG_GDB("*** Unsupported packet: %s\n", packet);
-	gdb_putpacket("", 0);
+	gdb_putpacketz("");
```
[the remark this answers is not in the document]
perigoso: The flush Is done in the checksum part of the packet transmission and is not affected by how much data is sent, a zero length packet is valid and transmitted like any other.
The start flag, end flag, and checksum are transmitted independent of the amount of data and the flush *always* happens after the last byte is sent which is the checksum lower nibble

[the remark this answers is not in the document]
dragonmux: Ah, spot on - so in other words, the checksum handling code ensures that the buffer is non-empty so the flush condition can never occur when empty 👍🏼

src/gdb_main.c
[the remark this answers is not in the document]
dragonmux: All good 😁 we'll look forward to the follow-up then once this is merged!

src/gdb_packet.c
perigoso: That is a good idea that I too had cross my mind.
For `gdb_out` it's easy enough to do as the string is static and already stored somewhere, but for `gdb_voutf`/`gdb_outf` it's not so simple, we need to allocate space for the formatted string somewhere, and right now its thankfully bounded to half the size of `GDB_OUT_PACKET_MAX_SIZE` but if we allow arbitrarily sized strings there we would need dynamic memory again.
My idea for removing the need for the scratch buffer wouldn't work either as I don't think it's easy to partially format strings (If we hooked into the "write" call of the print f we could do it, but sounds like more trouble than it's worth).

perigoso: I wouldn't do it in one but not the others as that would probably be confusing, having functions that appear similar on the surface have subtly different behaviors

src/gdb_main.c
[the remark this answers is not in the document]
perigoso: This is correct, we may respond with intermediate "O" output packets but the final "actual" response is a hex encoded string response.

src/gdb_packet.c
dragonmux: 🤔 The target.xml flows via the `gdb_putpacket*()` functions, so it doesn't hurt that.. this only impacts strings being sent to GDB's console.. oki!

Please make a note of this limitation then in the header as it applies to all GDB console output functions, and note the default limit in the process please so it's explicit, and we'll mark this resolved.

src/gdb_main.c
```diff
@@ -408,14 +422,12 @@ static void exec_q_rcmd(const char *packet, const size_t length)
 
 	const int result = command_process(cur_target, data);
 	if (result < 0)
-		gdb_putpacketz("");
+		gdb_put_packet_empty();
 	else if (result == 0)
-		gdb_putpacketz("OK");
+		gdb_put_packet_ok();
 	else {
 		const char *const response = "Failed\n";
```
dragonmux, line 429: Given how this gets used, it probably makes sense to make this a `static const char []` and then use `ARRAY_LENGTH()` in the put packet call (trimming 1 from the result to account for the '\0' implied at the end of the string).

src/gdb_main.c
```diff
@@ -841,53 +853,54 @@ static const cmd_executer_s v_commands[] = {
...
 
+#ifndef DEBUG_GDB_IS_NOOP
 	/*
 	 * The vMustReplyEmpty is used as a feature test to check how gdbserver handles
 	 * unknown packets, don't print an error message for it.
 	 */
-	if (strcmp(packet, "vMustReplyEmpty") != 0)
-		DEBUG_GDB("*** Unsupported packet: %s\n", packet);
-	gdb_putpacketz("");
+	static const char *const must_reply_empty = "vMustReplyEmpty";
```
dragonmux, line 866: Perhaps use `static const char []` here to leverage `ARRAY_LENGTH()` in the strncmp()?

dragonmux (commented):
Couple of things we spotted in the new code from the last review.. we'll get this compiled and thrown on a probe in a bit though and give it a spin with a known good target or two and see what happens 😄

src/gdb_main.c
perigoso: The idea being removing the need for strlen?

dragonmux: yep, exactly - it's a static string, so let's get as close to a std::string_view as we can in C so we're not wasting cycles on information we already have.

perigoso: I would wager it gets optimized out, but using ARRAY_LENGTH doesn't hurt

dragonmux (approved):
This all LGTM. We'll test this tomorrow and let you know if anything falls out that needs fixing. If nothing does, we'll just merge it. Thank you for the contribution!

dragonmux (approved):

dragonmux:
Just tested these changes and they appear to be working wonderfully. Merging - thank you for this wonderful contribution!

PR #2016, opened 2024-12-11
Fix/windowscompile

zhangmx:
## Detailed description

During the compilation of the project in a Windows MSYS2 environment, I encountered two warning messages and made some adjustments to address these issues. One of the problems was related to the const qualifier, which caused the compilation failure. I resolved this issue by using a type cast to ensure the pointer could be passed correctly.

### DIFF
--- a/src/platforms/hosted/bmp_libusb.c
+++ b/src/platforms/hosted/bmp_libusb.c
@@ -245,6 +245,8 @@ void stlinkv2_read_serial(libusb_device_descriptor_s *device_descriptor, libusb_
 }
 
 #if defined(_WIN32) || defined(__CYGWIN__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wcast-qual"
 static probe_info_s *process_ftdi_probe(void)
 {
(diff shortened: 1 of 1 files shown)


src/platforms/hosted/windows/ftdi.c
```diff
@@ -149,11 +149,22 @@ int ftdi_read_data(struct ftdi_context *ftdi, unsigned char *buf, const int size
 
 int ftdi_write_data(struct ftdi_context *ftdi, const unsigned char *buf, int size)
 {
-	(void)ftdi;
-	DWORD bytes_written;
-	if (FT_Write(ftdi_handle, (unsigned char *)buf, size, &bytes_written) != FT_OK)
```
dragonmux, line 154: This line should only need wrapping with
```c
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wcast-qual"
...
#pragma GCC diagnostic pop
```
Not sure the rest of this change should be necessary and it massively pessimises performance if it is, so we want to avoid it if possible.

dragonmux (changes requested):

PR #2017, opened 2024-12-13
target: Support 64-bit address for break/watch point

marysaka:
This change breakwatch::addr to be 64-bit as this will be required for ARMv8 AArch64 support.

This also change the layout of breakwatch to avoid extraneous padding in 64-bit and reduce the size of the reserved array to 64 bits.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -160,6 +160,12 @@ if is_firmware_build and get_option('print_memory_usage')
 	add_project_link_arguments('-Wl,--print-memory-usage', language: 'c')
 endif
 
+# We need to know the size of a pointer for the machine running the code
+add_project_link_arguments('-DCONFIG_POINTER_SIZE=@0@'.format(cc_host.sizeof('void *')), language: 'c')
+if is_cross_build
+	add_project_link_arguments('-DCONFIG_POINTER_SIZE=@0@'.format(cc_native.sizeof('void *')), language: 'c', native: true)
+endif
+
 subdir('src')
 
 ## Black Magic Firmware (BMF) targets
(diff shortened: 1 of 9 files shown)


src/target/target_internal.h
```diff
@@ -103,12 +103,11 @@ struct target_command {
 typedef struct breakwatch breakwatch_s;
 
 struct breakwatch {
-	/* XXX: This needs adjusting for 64-bit operations */
 	breakwatch_s *next;
-	target_breakwatch_e type;
-	target_addr32_t addr;
+	target_addr64_t addr;
```
dragonmux, line 107: This reordering needs to be gated on `sizeof(void *)` - for the firmware, pointers are 32-bit so this mis-aligns the 64-bit int of `addr`, while this correctly aligns it on 64-bit platforms. Conditionally reordering the `addr` and `type` members is probably the right way forwards.

dragonmux (changes requested):
We only have one review note and it's to do with the ordering of members in the `breakwatch` structure, nothing too strenuous though and with that addressed we're happy to merge this.

meson.build
```diff
@@ -160,6 +160,13 @@ if is_firmware_build and get_option('print_memory_usage')
 	add_project_link_arguments('-Wl,--print-memory-usage', language: 'c')
 endif
 
+# We need to know the size of a pointer for the target running the code
```
dragonmux, line 163: Think you mean "machine" rather than "target" here.

You will need to make a call for checking `sizeof(void *)` for both `cc_host` and `cc_native`, otherwise one of those two is likely going to be wrong for one of either the host or the probe. Use `cc_native` in the `is_cross_build` block.

dragonmux (commented):
There's one small issue we spotted in this round, and then we think this PR is there and can be merged. Good work!

dragonmux (approved):

PR #2024, opened 2024-12-15
gdb_packet: allow for external packet buffer

xobs:
## Detailed description

Allow supplying a packet buffer from an external source. This allows the GDB packet functions to be thread-safe, assuming the packet buffer comes from some form of thread-local storage.

Guard this function with a macro to allow keeping the previous behavior.

This allows platforms such as Farpatch to support multiple clients. A simple lock around accessing gdb provides some measure of concurrency management. When built without the macro, the function is marked as "static", in an effort to induce the compiler to inline the function to get the benefits of the new approach.

### DIFF
--- a/src/gdb_packet.c
+++ b/src/gdb_packet.c
@@ -41,14 +41,23 @@ typedef enum packet_state {
 
 static bool noackmode = false;
 
+#ifdef EXTERNAL_PACKET_BUFFER
+extern gdb_packet_s *gdb_full_packet_buffer(void);
+#else
 /* This has to be aligned so the remote protocol can re-use it without causing Problems */
 static gdb_packet_s BMD_ALIGN_DEF(8) packet_buffer;
+
+static inline gdb_packet_s *gdb_full_packet_buffer(void)
+{
+	return &packet_buffer;
+}
 
 char *gdb_packet_buffer(void)
 {
 	/* Return the static packet data buffer */
 	return packet_buffer.data;
 }
+#endif /* EXTERNAL_PACKET_BUFFER */
 
 /* https://sourceware.org/gdb/onlinedocs/gdb/Packet-Acknowledgment.html */
 void gdb_set_noackmode(bool enable)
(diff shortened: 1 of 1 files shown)


src/gdb_packet.c
```diff
@@ -167,18 +176,19 @@ gdb_packet_s *gdb_packet_receive(void)
 {
 	packet_state_e state = PACKET_IDLE; /* State of the packet capture */
 	uint8_t rx_checksum = 0;
+	gdb_packet_s *packet_buffer = gdb_full_packet_buffer();
```
dragonmux, line 179: To prevent confusion between this and the global, perhaps rename the local to `packet`?

src/gdb_packet.c
```diff
@@ -41,14 +41,23 @@ typedef enum packet_state {
 
 static bool noackmode = false;
 
+#ifdef EXTERNAL_PACKET_BUFFER
+extern gdb_packet_s *gdb_full_packet_buffer(void);
+#else
 /* This has to be aligned so the remote protocol can re-use it without causing Problems */
 static gdb_packet_s BMD_ALIGN_DEF(8) packet_buffer;
 
+static gdb_packet_s *gdb_full_packet_buffer(void)
```
dragonmux, line 50: This function could do with being marked `inline` to strongly suggest to the compiler that it can, well, inline it.

src/gdb_packet.c
```diff
@@ -332,13 +342,15 @@ void gdb_packet_send(const gdb_packet_s *const packet)
 
 void gdb_put_packet(const char *preamble, size_t preamble_size, const char *data, size_t data_size, bool hex_data)
 {
+	gdb_packet_s *packet_buffer = gdb_full_packet_buffer();
```
dragonmux, line 345: Same comment about the local here as above (and in the functions below)

dragonmux (changes requested):
This is looking good, the only comments are on naming and compiler hints. Functionally this looks good 👍🏼

src/gdb_packet.c
xobs: Sure. During development I actually used the names `_packet_buffer` and `packet_buffer`. I'll change the name to what you suggest.

src/gdb_packet.c
xobs: Done

src/gdb_packet.c
xobs: Also done

dragonmux (approved):

PR #2028, opened 2024-12-19
Feature/formally parse riscv isa

perigoso:
## Detailed description

This is a cherry-pick of a couple commits from #1399 that stand on their own as general improvements.

This adds more compreensive parsing of the RISC-V ISA from the misa register when available, this is used to show a more complete ISA when a target is probed.

Example with a RP2350 which is the only RISC-V target I have on hand:

before:
```
Available Targets:
No. Att Driver
 1      RP2350 rv32
 2      RP2350 rv32
```

after:
```
Available Targets:
No. Att Driver
 1      RP2350 rv32imac
 2      RP2350 rv32imac
```

The RP2350's formal ISA according to the datasheet is `rv32imac_zicsr_zifencei_zba_zbb_zbs_zbkb` (or `rv32ima_zicsr_zifencei_zba_zbb_zbs_zbkb_zca_zcb_zcmp` with the most recent interpretation), so the BMP's parsing is correct.

It's important to note, this only interprets the ISA standard *ratified* extensions, reserved extension bits are ignored for now, and since Z/X/S extensions cannot easily be detected (There's no CSR for them) they are "ignored" too. 

Technically the presence of the 'B' extension can be interpreted as the `Zba`, `Zbb`, and `Zbs` extensions being available, but in the interest of a cleaner ISA string we ignore it too.

The 'I' is not directly intepreted because it is a base ISA, and is implicit unless the 'E' base ISA is set which we do check for. The 'E' bit is always the complement of 'I' only one is ever set, and one must be set.

This implementation could be cleaned up and be made much more efficient, but it works as is and can be improved later, submitting as is in the interest of getting merged before the v2.0 release

### DIFF
--- a/src/include/buffer_utils.h
+++ b/src/include/buffer_utils.h
@@ -84,4 +84,11 @@ static inline uint64_t read_be8(const uint8_t *const buffer, const size_t offset
 		((uint64_t)buffer[offset + 6] << 8U) | buffer[offset + 7];
 }
 
+static inline size_t write_char(char *const buffer, const size_t buffer_size, const size_t offset, const char c)
+{
+	if (buffer && offset < buffer_size)
+		buffer[offset] = c;
+	return offset + 1U;
+}
+
 #endif /*INCLUDE_BUFFER_UTILS_H*/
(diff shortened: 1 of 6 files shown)


dragonmux (approved):
This all looks reasonable to us. We'll run a test with it on a couple of the RISC-V targets we have here, but please rebase this on `main` and we'll get it merged presuming that all goes smoothly.

PR #2040, opened 2025-01-06
Fix: target discovery logic in stm32l4

ALTracer:
## Detailed description

* This is a couple bugfixes.
* The existing problem exhibits as invalid CCM RAM size on STM32G474CE.
* The PR solves it by changing the `target->priv` field behind typecasts to `target->target_storage` correct one.

Dereferencing that `priv` and typecasting it to stm32l4_priv_s results in unpredictable CCM sizes in BMDA and large sizes in BMF. It's for `cortexm_priv_s`, and the rest of this driver uses `target_storage`, so fix the last use of it.
Other dereferences of both priv and target_storage in other TUs seem to be okay per Qt Creator lookups (check them!)

Also, `*device` is not going to be NULL, it can be the 18th sentinel entry, but its `device_id` can be zero -- so correct the no-match branch condition which should trigger a second LUT walk.

Tested on gd32f103cb bluepillplus as `bluepill.ini`/`stlink` platform and stm32g474ceu target, BMDA as well. Gentoo gcc-13, crossdev. In particular, RTT autosearch no longer hangs up faulting in nonexistent CCM and retrying for ages.


dragonmux (approved):

PR #2041, opened 2025-01-06
Fix: unexpected E02 from gdb m-packets

ALTracer:
## Detailed description

* This is a minor but important bugfix.
* The existing problem is gdbserver returning unexpected read errors E02 `Cannot access memory at address` to otherwise normal accesses.
* The PR solves it by fixing a mistranslated bounds-check introduced in ca. v1.10.0-1567-g de645593 of #2014.

Tested on gd32f103cb bluepillplus running bluepill.ini/stlink BMF; and BMDA; against stm32g474ce. At first I expected wrong RCC/Flash config or invalid SRAM mapping, but interactive debugging of BMDA and extended logging proved otherwise.
Note: `(gdb) maintenance packet m08070000,201` triggers a 513 byte read which is rejected by BMD. Only <=512 is good because of 1024 byte RSP packet buffer.


dragonmux (approved):
LGTM, nice work finding this bug - merging. Thank you for the contribution!

perigoso:
Thanks for the cleanup, good catch!

I see this uses alloca a couple lines down, and I do see a way we can optimize that out!

PR #2055, opened 2025-01-15
Feature: enable noackmode by default

perigoso:
## Detailed description

Enabling noackmode advertisement tells GDB we support and prefer noackmode, who will then enable it by default. Right now the mode is supported but not advertised, it needs to be explicitly enabled by the user.

This should make the protocol more efficient even if by negligible amount without any drawbacks, as we have a reliable data connection making the checksums and acknowledgements redundant.

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -107,7 +107,7 @@ option(
 option(
 	'advertise_noackmode',
 	type: 'boolean',
-	value: false,
+	value: true,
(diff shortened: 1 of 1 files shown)


dragonmux (approved):
LGTM, given how well this has been working for us since we enabled it on native out the box, we'll merge this. Thank you for the contribution!

PR #2059, opened 2025-01-19
Fix: WiFi configuration issues via USB (HTTP/WPS provisioning sometimes unreliable)

sidprice:
This PR adds Wi-Fi management for ctxLink using GDB and a new monitor command.

With the original releases of ctxLink some users had issues using the HTTP and WPS provisioning features of ctxLink.

Also, after provisioning some users found it hard to discover the IP address of ctxLink, this is required to connect to the probe.

The new monitor command is "wifi".

To inspect the current Wi-Fi status the following command is used:

mon wifi

This will display either "Not Connected," or the SSID, Signal Strength (RSSI), and IP address of the connected Access Point.

To connect to a Wi-Fi Access Point the following command is used, substituting the SSID and PassPhrase settings of the AP:

mon wifi SSID,Passphrase

Spaces are accepted for these two parameters, however, there should be no space before or after the ",".

## Closing issues

None.

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -113,7 +113,7 @@ const command_s cmd_list[] = {
 	{"battery", cmd_target_battery, "Reads the battery state"},
 #endif
 #ifdef PLATFORM_HAS_WIFI
-	{"wifi", cmd_wifi, "Read/Set Wi-Fi connection [AP name, passphrase]"},
+	{"wifi", cmd_wifi, "Show/Set Wi-Fi connection [-forget] [AP name,passphrase]"},
 #endif
 #ifdef ENABLE_RTT
 	{"rtt", cmd_rtt,
--- a/src/command.c
+++ b/src/command.c
@@ -551,7 +551,8 @@ static bool cmd_target_battery(target_s *t, int argc, const char **argv)
 static bool cmd_wifi(target_s *t, int argc, const char **argv)
 {
 	(void)t;
-	return platform_wifi_state(argc, argv);
+	gdb_out(platform_wifi_state(argc, argv));
+	return true;
 }
 #endif
 
--- a/src/command.c
+++ b/src/command.c
@@ -113,7 +113,7 @@ const command_s cmd_list[] = {
 	{"battery", cmd_target_battery, "Reads the battery state"},
 #endif
 #ifdef PLATFORM_HAS_WIFI
-	{"wifi", cmd_wifi, "Reads the wi-fi state"},
+	{"wifi", cmd_wifi, "Read/Set Wi-Fi connection [AP name, passphrase]"},
 #endif
 #ifdef ENABLE_RTT
 	{"rtt", cmd_rtt,
--- a/src/command.c
+++ b/src/command.c
@@ -113,7 +113,7 @@ const command_s cmd_list[] = {
 	{"battery", cmd_target_battery, "Reads the battery state"},
 #endif
 #ifdef PLATFORM_HAS_WIFI
-	{"wifi", cmd_wifi, "Show/Set Wi-Fi connection [-forget] [AP name,passphrase]"},
+	{"wifi", cmd_wifi, "Show/Set Wi-Fi connection [AP name,passphrase]"},
 #endif
 #ifdef ENABLE_RTT
 	{"rtt", cmd_rtt,
--- a/src/command.c
+++ b/src/command.c
@@ -70,6 +70,9 @@ static bool cmd_target_power(target_s *target, int argc, const char **argv);
 #ifdef PLATFORM_HAS_BATTERY
 static bool cmd_target_battery(target_s *t, int argc, const char **argv);
 #endif
+#ifdef PLATFORM_HAS_WIFI
+static bool cmd_wifi(target_s *t, int argc, const char **argv);
+#endif
 #ifdef PLATFORM_HAS_TRACESWO
 static bool cmd_swo(target_s *target, int argc, const char **argv);
 #endif
@@ -108,6 +111,9 @@ const command_s cmd_list[] = {
 #endif
 #ifdef PLATFORM_HAS_BATTERY
 	{"battery", cmd_target_battery, "Reads the battery state"},
+#endif
+#ifdef PLATFORM_HAS_WIFI
+	{"wifi", cmd_wifi, "Reads the wi-fi state"},
 #endif
 #ifdef ENABLE_RTT
 	{"rtt", cmd_rtt,
@@ -541,6 +547,14 @@ static bool cmd_target_battery(target_s *t, int argc, const char **argv)
 	return true;
 }
 #endif
+#ifdef PLATFORM_HAS_WIFI
+static bool cmd_wifi(target_s *t, int argc, const char **argv)
+{
+	(void)t;
+	return platform_wifi_state(argc, argv);
+}
+#endif
+
 #ifdef ENABLE_RTT
 static const char *on_or_off(const bool value)
 {
--- a/src/platforms/ctxlink/WiFi_Server.c
+++ b/src/platforms/ctxlink/WiFi_Server.c
@@ -651,6 +651,17 @@ bool is_ip_address_assigned(void)
 	return res;
 }
 
+void wifi_get_ip_address(char *buffer, uint32_t size)
+{
+	if (g_wifi_connected)
+		snprintf(buffer, size, "%d.%d.%d.%d\n", conn_info.acSSID[3], conn_info.acSSID[2], conn_info.acSSID[1],
+			conn_info.acSSID[0]);
+	else {
+		memset(buffer, 0x00, size - 1);
+		memcpy(buffer, "Not connected\n", strlen("Not connected"));
+	}
+}
+
 void handle_socket_bind_event(SOCKET *sock, bool *running_state)
 {
 	if (m2m_wifi_get_socket_event_data()->bindStatus == 0)
--- a/src/platforms/ctxlink/WiFi_Server.c
+++ b/src/platforms/ctxlink/WiFi_Server.c
@@ -666,7 +666,7 @@ void wifi_get_ip_address(char *buffer, uint32_t size)
 		snprintf(local_buffer, sizeof(local_buffer), "SSID = %s\n", conn_info.acSSID);
 		strncpy(buffer, local_buffer, size);
 		snprintf(local_buffer, sizeof(local_buffer), "RSSI = %d\n", conn_info.s8RSSI);
-		strncpy(buffer, local_buffer, size);
+		strncat(buffer, local_buffer, size);
 		snprintf(local_buffer, sizeof(local_buffer), "IP = %d.%d.%d.%d\n", conn_info.au8IPAddr[0],
 			conn_info.au8IPAddr[1], conn_info.au8IPAddr[2], conn_info.au8IPAddr[3]);
 		strncat(buffer, local_buffer, size);
--- a/src/platforms/ctxlink/WiFi_Server.c
+++ b/src/platforms/ctxlink/WiFi_Server.c
@@ -703,9 +703,7 @@ void wifi_connect(size_t argc, const char **argv, char *buffer, uint32_t size)
 	char ssid[64] = {0};
 	char pass_phrase[64] = {0};
 	char *output_buffer = ssid;
-	char *delimeter;
-	bool add_space = false;
-	uint32_t transfer_count = 0;
+	bool first_element = true;
 	memset(buffer, 0x00, size);
 	//
 	// Iterate over the arguments received to build the SSID and passphrase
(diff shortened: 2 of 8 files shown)


sidprice:
@Misaka0x2730 Hello!

This was previously suggested by @ALTracer, and is a good idea. Since this PR is a feature update it is awaiting review and merging until after the next release. Hopefully, during this period I can adopt the suggestion.

sidprice:
I am not sure this is such a good idea since it prevents grouping the custom commands logically in the command list. Plus, there are quite a few other custom commands already in the command list, e.g. TraceSWO & RTT.

I think unless the maintainers (@dragonmux ) feel it is required I am going to leave the added command for this PR as it is.

Misaka0x2730:
I agree that @dragonmux is the one who’ll ultimately decide here.  
However, I want to clarify my thoughts on “custom” commands. For instance, TraceSWO and RTT are supported by many (if not most) platforms—so they’re hardly “custom” in the sense of belonging to just one.  
By contrast, Wi-Fi is specific to your platform, which I believe aligns more with the idea of platform-specific commands.

perigoso:
I agree that these should be platform commands at least for now, though not having control over the grouping is unfortunate.
I have some ideas to rework to the command system but until then we are limited with what we can do. Do note that this only affects the print on the help command and doesn't affect functionality

sidprice:
@perigoso Thanks for your input. I would be happy to review the platform commands should the command table/architecture change in the future.

dragonmux:
Having had a conversation with Sid on Discord, we are switching this to being a fix PR rather than a feature, and promoting it into the v2.0 track for after rc2's release.

src/command.c
```diff
@@ -70,6 +70,9 @@ static bool cmd_target_power(target_s *target, int argc, const char **argv);
 #ifdef PLATFORM_HAS_BATTERY
 static bool cmd_target_battery(target_s *t, int argc, const char **argv);
 #endif
+#ifdef PLATFORM_HAS_WIFI
+static bool cmd_wifi(target_s *t, int argc, const char **argv);
+#endif
```
dragonmux, line 75: Not for this PR, but for a future one: These should probably both become platform commands. That way you don't need to export as much platform state into the rest of the code base, and they get grouped under the platform commands heading in the `mon help` output. We can look at improving this in v2.1 though so this is fine as is for now.

src/platforms/ctxlink/platform.c
```diff
@@ -156,6 +156,20 @@ int platform_hwversion(void)
 	return 0;
 }
 
+static char parameters[128] = {0};
+
+const char *platform_wifi_state(int argc, const char **argv)
+{
+	(void)argv;
```
dragonmux, line 163: This parameter is used in the `if` block below, so please drop this line.

src/platforms/ctxlink/ctxLink_mode_led.c
```diff
@@ -108,9 +108,9 @@ void mode_led_task(void)
 			 */
 		if (led_mode != mode_led_idle) {
 			/*
-				 * Set up the led control registers according to the requested mode
-				 */
-			mode_task_state = mode_led_state_on;
+						 * Set up the led control registers according to the requested mode
+						 */
```
dragonmux, line 112: This comment looks over-indented - think some tabs need removing here

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -531,6 +536,12 @@ static void app_wifi_callback(uint8_t msg_type, void *msg)
 		break;
 	}
 
+	case M2M_WIFI_CONN_INFO_RESPONSE_EVENT: {
+		tstrM2MConnInfo *connection_info = (tstrM2MConnInfo *)msg;
+		memcpy(&conn_info, connection_info, sizeof(tstrM2MConnInfo));
```
dragonmux, line 541: For what it's worth, `memcpy()` takes a void pointer for the pointer parameters so you could simplify this to `memcpy(&conn_info, msg, sizeof(tstrM2MConnInfo));` and forgo the pointer conversion and local.

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -558,10 +569,14 @@ static void app_wifi_callback(uint8_t msg_type, void *msg)
 	}
 
 	case M2M_WIFI_IP_ADDRESS_ASSIGNED_EVENT: {
-		t_wifiEventData *wifi_event_data = (t_wifiEventData *)msg;
-		if (wifi_event_data != NULL)
+		tstrM2MIPConfig *ip_config = (tstrM2MIPConfig *)msg;
```
dragonmux, line 572: Beyond checking the pointer is non-NULL, is this local used for anything here? If not then the `if` should probably become `if (msg != NULL)`

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -641,6 +655,130 @@ bool is_ip_address_assigned(void)
...
+// Format the current connection data for display to user
+//
+//	SSID = 'name'
+//	RSSI = xx
+//	ip = xxx.xxx.xxx.xxx
+//
+void wifi_get_ip_address(char *buffer, uint32_t size)
+{
+	char local_buffer[64] = {0};
+	memset(buffer, 0x00, size - 1);
```
dragonmux, line 668: `size - 1`? Surely this leaves the last character in the buffer untouched and thus breaks NUL character termination rules.. That said, if correct, please suffix the `1` with a U so the calculation is carried out unsigned for signed-unsigned continuity to `memset()` (3rd parameter is a `size_t`)

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -641,6 +655,130 @@ bool is_ip_address_assigned(void)
...
+	}
+}
+
+//
+// Wait for app_state to spin with timeout
+//
+void app_task_wait_spin(void)
+{
+	uint32_t wait_timeout = 2000U;
+	while (1) {
```
dragonmux, line 688: Please use `true` in place of `1` here for consistency with the rest of the code base.

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -641,6 +655,130 @@ bool is_ip_address_assigned(void)
...
+		platform_tasks();
+		if (app_state == app_state_spin)
+			break;
+		platform_delay(1);
+		if (wait_timeout-- == 0)
```
dragonmux, line 693: The constants here both need suffixing with `U` to not do signed-unsigned conversions.

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -641,6 +655,130 @@ bool is_ip_address_assigned(void)
...
+			break;
+	}
+}
+
+//
+// Using the passed arguments, attempt to connect to a Wi-Fi AP
+//
+void wifi_connect(int argc, const char **argv, char *buffer, uint32_t size)
```
dragonmux, line 701: As you check `argc` is at least 1 in platform_wifi_state(), please switch the type of the parameter here to `size_t` and in the loop below to avoid potential boundary condition problems in the `loop < argc` check. In practice this should never be a problem, but C doesn't make it a guarantee unless you're using unsigned types due to overflow UB on signed ones.

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -641,6 +655,130 @@ bool is_ip_address_assigned(void)
...
+	//
+	// The remaining arguments are then concatenated into the passphrase with
+	// an space added between them.
+	//
+	for (int loop = 1; loop < argc; loop++) {
+		delimeter = strchr(argv[loop], ',');
+		if (delimeter == NULL) {
+			if (add_space) {
+				strcat(output_buffer, " ");
+				add_space = true;
```
dragonmux, line 732: This looks to be wrong logic? Surely this should be assigning `false` to this variable having performed the action the variable signals (to reset for next loop).

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -641,6 +655,130 @@ bool is_ip_address_assigned(void)
...
+			//
+			output_buffer = pass_phrase;
+			add_space = true;
+			strcat(output_buffer, delimeter + 1);
+		}
+	}
+	//
+	// If we have both SSID and Passphrase attempt to connect
+	//
+	if (ssid[0] != 0x00 && pass_phrase[0] != 0x00) {
```
dragonmux, line 756: As you are comparing char's here, please use `'\0'` instead of `0x00` so the types agree.

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -1244,7 +1395,16 @@ void app_task(void)
 	}
 	case app_state_wait_for_server: {
 		if (is_gdb_server_running())
```
dragonmux, line 1397: This `if` becomes multi-line with the inner `if` addition - please either add parens, or.. due to the construction of the inner.. combine as in `if (is_gdb_server_running() && is_ip_address_assigned())`.

dragonmux (changes requested):
Apologies getting to review this took so long. Most of the items are types related (signed-unsigned conversions mostly) and should all be pretty simple fixes. With them taken care of we'll get this merged so this issue can be fixed for the full v2.0 release.

src/command.c
sidprice: Agreed as previously discussed.

src/platforms/ctxlink/WiFi_Server.c
sidprice: Agreed, however the local aids with debugging, that is why it was added.

src/platforms/ctxlink/WiFi_Server.c
sidprice: Like the above comment, the local is a useful debugging addition.

src/platforms/ctxlink/WiFi_Server.c
sidprice: Fixed.

src/platforms/ctxlink/WiFi_Server.c
sidprice: Fixed.

src/platforms/ctxlink/WiFi_Server.c
sidprice: Fixed.

src/platforms/ctxlink/WiFi_Server.c
sidprice: Fixed.

src/platforms/ctxlink/platform.c
sidprice: Fixed.

src/platforms/ctxlink/WiFi_Server.c
sidprice: Fixed.

src/platforms/ctxlink/WiFi_Server.c
sidprice: Fixed.

src/platforms/ctxlink/ctxLink_mode_led.c
sidprice: Fixed.

src/platforms/ctxlink/WiFi_Server.c
sidprice: INdeed, not only the wrong logic, also this parser did not handle multiple spaces in the arguments to the "wifi" monitor command.

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -641,6 +655,123 @@ bool is_ip_address_assigned(void)
...
+	// It does this initially into the SSID name, up until an argument is found
+	// that contains a comma. The string before the comma is concatenated with
+	// the SSID and the string after the comma is used as the first element of
+	// the passphrase.
+	//
+	// The remaining arguments are then concatenated into the passphrase with
+	// an space added between them.
+	//
+	for (size_t loop = 1; loop < argc; loop++) {
+		delimeter = strchr(argv[loop], ',');
```
dragonmux, line 727: As this variable is only used inside the loop, we suggest moving its definition to this line, allowing it to become a little more `const` - `char *const delimeter = strchr(argv[loop], ',');`

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -641,6 +655,123 @@ bool is_ip_address_assigned(void)
...
+		if (delimeter == NULL) {
```
dragonmux, line 728: As passphrases may contain embedded commas too, we suggest this boundary condition should be `if (delimeter == NULL || output_buffer == pass_phrase)` so the split logic doesn't re-trigger in that case

src/platforms/ctxlink/WiFi_Server.c
```diff
@@ -844,7 +975,7 @@ static void app_socket_callback(SOCKET sock, uint8_t msg_type, void *msg)
 				// Copy data to circular input buffer
 				//
 				for (int16_t i = 0; local_count != 0;
-					 i++, local_count--, input_index = (input_index + 1) % INPUT_BUFFER_SIZE) {
+					i++, local_count--, input_index = (input_index + 1) % INPUT_BUFFER_SIZE) {
```
dragonmux, line 978: `clang-format` wants its space back here - likely this is older vs newer clang-format and a bug being triggered in the older. We'll get the version of `clang-format` upgraded in CI once v2.0 is released.

dragonmux (commented):
This is looking much better - good work! Spotted a couple of small things in the rewritten `wifi_connect()` logic, but with them taken care of we're happy to merge this.

src/platforms/ctxlink/WiFi_Server.c
sidprice: Fixed as per suggestion.

src/platforms/ctxlink/WiFi_Server.c
sidprice: While this is fixed, it is bound to reoccur if I work on this file again. When I save the file clang-format runs and re-fixes it for me. So, have the CI scripts deal with it correctly is kind of important.

src/platforms/ctxlink/WiFi_Server.c
sidprice: Excellent catch, fixed.

dragonmux (approved):
LGTM, merging - nice work cleaning things up so quickly. Thank you for the contribution!

PR #2063, opened 2025-01-25
Add ESP32-P4 support

orsolic:
Here is commit with start of support for ESP32-P4.

I have started working on flashing support but I need to shelve that plan for the time being.
I am pushing the things which seems to work - patch is rough, but more will hopefully come later.

Tested with BMA and ESP-prog (FT2232H) on FreeBSD with Waveshare ESP32-P4 board with flashed IDF example.

Works:
- JTAG scan
- attach to target
- run program
- reset
- hardware breakpoint
- continue
- printing of var in SRAM
```
>>> p var
$4 = 123
>>> p &var
$5 = (uint32_t *) 0x4ff26334
>>> x /d 0x4ff26334
0x4ff26334:     123
```
- changing value of var:
```
>>> set var var = 100
>>> p var
$8 = 100
>>> x /d 0x4ff26334
0x4ff26334:     100
```

What doesn't work:
- software breakpoint (needs flash support)
- stepping
- single instruction stepping

### DIFF
--- a/src/target/jtag_devs.c
+++ b/src/target/jtag_devs.c
@@ -413,6 +413,32 @@ const jtag_dev_descr_s dev_descr[] = {
 #endif
 		.handler = riscv_jtag_dtm_handler,
 	},
+	{
+		.idcode = 0x00012c25U,
+		.idmask = 0x0fffffffU,
+#if ENABLE_DEBUG == 1
+		.descr = "ESP32-P4",
+#endif
+		.handler = riscv_jtag_dtm_handler,
+		.ir_quirks =
+			{
+				.ir_length = 5U,
+				.ir_value = 5U,
+			},
+	},
+	{
+		.idcode = 0x00012c25U,
+		.idmask = 0x0fffffffU,
+#if ENABLE_DEBUG == 1
+		.descr = "ESP32-P4",
+#endif
+		.handler = riscv_jtag_dtm_handler,
+		.ir_quirks =
+			{
+				.ir_length = 5U,
+				.ir_value = 1U,
+			},
+	},
 #endif
 #if defined(CONFIG_CORTEXAR) // && defined(ENABLE_SITARA)
 	{
--- a/src/target/jtag_scan.c
+++ b/src/target/jtag_scan.c
@@ -210,9 +210,17 @@ static void jtag_display_idcodes(void)
 
 static jtag_ir_quirks_s jtag_device_get_quirks(const uint32_t idcode)
 {
+	static size_t previous_idx = 0;
+
 	for (size_t idx = 0; dev_descr[idx].idcode; ++idx) {
 		if ((idcode & dev_descr[idx].idmask) == dev_descr[idx].idcode)
+		{
+			/* workaround for ESP32-P4 which has 2 different ir_value */
+			if (idcode == 0x12c25U && previous_idx == idx)
+				idx++;
+			previous_idx = idx;
 			return dev_descr[idx].ir_quirks;
+		}
 	}
 	return (jtag_ir_quirks_s){0};
 }
--- a/src/target/riscv_debug.c
+++ b/src/target/riscv_debug.c
@@ -304,6 +304,30 @@ void riscv_dmi_init(riscv_dmi_s *const dmi)
 	do {
 		/* Read out the DM's status register */
 		uint32_t dm_status = 0;
+
+		/* Turn on DM before trying to read version  */
+		if (!riscv_dmi_write(dmi, base_addr + RV_DM_CONTROL, RV_DM_CTRL_ACTIVE)) {
+			DEBUG_ERROR("error turning on DM!\n");
+			return;
+		}
+
+		/* After changing the value of dm_active, the debugger must poll dmcontrol
+		 * until dm_active has taken the requested value */
+		bool dm_active = false;
+		uint32_t dm_control = 0;
+		while (!dm_active) {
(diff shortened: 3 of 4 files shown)


src/target/jtag_scan.c
```diff
@@ -210,9 +210,17 @@ static void jtag_display_idcodes(void)
...
 {
+	static size_t previous_idx = 0;
+
 	for (size_t idx = 0; dev_descr[idx].idcode; ++idx) {
 		if ((idcode & dev_descr[idx].idmask) == dev_descr[idx].idcode)
+		{
+			/* workaround for ESP32-P4 which has 2 different ir_value */
+			if ((idcode == 0x12c25) && (previous_idx == idx))
+				idx++;
+			previous_idx = idx;
```
dragonmux, line 221: Do please note that this will only work for a single scan - the next run through will break. This is because `previous_idx` is `static` and so only initialised once on program start and is unable to be reset to `0`.

It might be better to make the previous index used a return parameter to this function stored as a local in the calling context, this way it can be properly reinitialised on each new scan.

src/target/jtag_scan.c
```diff
@@ -210,9 +210,17 @@ static void jtag_display_idcodes(void)
 
 static jtag_ir_quirks_s jtag_device_get_quirks(const uint32_t idcode)
 {
+	static size_t previous_idx = 0;
+
 	for (size_t idx = 0; dev_descr[idx].idcode; ++idx) {
 		if ((idcode & dev_descr[idx].idmask) == dev_descr[idx].idcode)
+		{
+			/* workaround for ESP32-P4 which has 2 different ir_value */
+			if ((idcode == 0x12c25) && (previous_idx == idx))
```
dragonmux, line 219: Please drop the extra parens - they are not only not required, they actively suppress a warning about accidental assignment. Please suffix the integer constant with `U` to make it unsigned.

src/target/riscv_debug.c
```diff
@@ -304,6 +304,30 @@ void riscv_dmi_init(riscv_dmi_s *const dmi)
...
+
+		/* Turn on DM before trying to read version  */
+		if (!riscv_dmi_write(dmi, base_addr + RV_DM_CONTROL, RV_DM_CTRL_ACTIVE)) {
+			DEBUG_ERROR("error turning on DM!\n");
+			return;
+		}
+
+		/* After changing the value of dm_active, the debugger must poll dmcontrol
+		 * until dm_active has taken the requested value */
+		while (dm_active == false) {
```
dragonmux, line 319: `bool` is a true boolean type - this check can and should be simplified to `while (!dm_active)`

src/target/riscv_debug.c
```diff
@@ -304,6 +304,30 @@ void riscv_dmi_init(riscv_dmi_s *const dmi)
 	do {
 		/* Read out the DM's status register */
 		uint32_t dm_status = 0;
+		uint32_t dm_control = 0;
+		bool dm_active = false;
+		static uint8_t counter = 0;
```
dragonmux, line 309: Why is this `static`? This reads like it should be reset each entry to this function. Otherwise, after 100 scans this function permanently fails.

dragonmux, line 309: Please moves these down to where they're actually used - `dm_control` inside the loop just before the `riscv_dmi_read()` call, `dm_status` down to its `riscv_dmi_read()` call, etc.

src/target/jtag_devs.c
```diff
@@ -413,6 +413,32 @@ const jtag_dev_descr_s dev_descr[] = {
...
+		.idcode = 0x00012c25U,
+		.idmask = 0x0fffffffU,
+#if ENABLE_DEBUG == 1
+		.descr = "ESP32-P4",
+#endif
+		.handler = riscv_jtag_dtm_handler,
+		.ir_quirks =
+			{
+				.ir_length = 5,
+				.ir_value = 5,
```
dragonmux, line 426: Please suffix these and the two in the block below with `U` to make them unsigned.

dragonmux (changes requested):
This looks interesting - there are some issues though, particular involving repeated scans within a session. Please see our review notes for details.

src/target/jtag_scan.c
orsolic: It works for multiple scans - first scan:
jtag_device_get_quirks() returning idx: 40
jtag_device_get_quirks() returning idx: 41

Second scan:
jtag_device_get_quirks() returning idx: 40
jtag_device_get_quirks() returning idx: 41

If it isn't static it will fail on the 1st scan:
jtag_device_get_quirks() returning idx: 40
jtag_device_get_quirks() returning idx: 40
jtag_scan: IR does not match the expected value, bailing out

orsolic:
I've accidentally pushed the wrong branch (the one with included FreeBSD build fix), then deleted it, which caused this PR to be closed, and now I can't reopen it.
Can you reopen it? Or, should I create a new PR?

orsolic:
Reopen? Reopen.

esden:
The problem is that you deleted the branch on your repository, so this PR does not have anything to reference to. I think you will have to open a new pull request.

src/target/jtag_scan.c
dragonmux: We are not convinced, in part because consider: what happens if a scan fails having found the first TAP but before it can find the second, and then a user tries re-scanning? Please lift the previous index value out into the calling function as a local, and pass a pointer to it to this function. This hidden global is way too brittle.

PR #2064, opened 2025-01-28
Add ESP32-P4 support

orsolic:
Continuation of mistakenly closed #2063

Here is commit with start of support for ESP32-P4.

I have started working on flashing support but I need to shelve that plan for the time being.
I am pushing the things which seems to work - patch is rough, but more will hopefully come later.

Tested with BMA and ESP-prog (FT2232H) on FreeBSD with Waveshare ESP32-P4 board with flashed IDF example.

Works:
- JTAG scan
- attach to target
- run program
- reset
- hardware breakpoint
- continue
- printing of var in SRAM
```
>>> p var
$4 = 123
>>> p &var
$5 = (uint32_t *) 0x4ff26334
>>> x /d 0x4ff26334
0x4ff26334:     123
```
- changing value of var:
```
>>> set var var = 100
>>> p var
$8 = 100
>>> x /d 0x4ff26334
0x4ff26334:     100
```

What doesn't work:
- software breakpoint (needs flash support)
- stepping
- single instruction stepping

### DIFF
--- a/src/target/riscv_debug.c
+++ b/src/target/riscv_debug.c
@@ -304,6 +304,30 @@ void riscv_dmi_init(riscv_dmi_s *const dmi)
 	do {
 		/* Read out the DM's status register */
 		uint32_t dm_status = 0;
+
+		/* Turn on DM before trying to read version  */
+		if (!riscv_dmi_write(dmi, base_addr + RV_DM_CONTROL, RV_DM_CTRL_ACTIVE)) {
+			DEBUG_ERROR("error turning on DM!\n");
+			return;
+		}
+
+		/* After changing the value of dm_active, the debugger must poll dmcontrol
+		 * until dm_active has taken the requested value */
+		bool dm_active = false;
+		uint32_t dm_control = 0;
+		while (!dm_active) {
+			if (!riscv_dmi_read(dmi, base_addr + RV_DM_CONTROL, &dm_control)) {
+				DEBUG_ERROR("error turning on DM!\n");
+				return;
+			}
+			dm_active = dm_control & 1;
+			uint8_t counter = 0;
+			if (++counter >= 100) {
+				DEBUG_ERROR("Timeout while trying to turn on DM\n");
+				return;
+			}
+		}
+
 		if (!riscv_dmi_read(dmi, base_addr + RV_DM_STATUS, &dm_status)) {
 			/* If we fail to read the status register, abort */
 			break;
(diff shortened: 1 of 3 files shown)


src/target/riscv_debug.c
```diff
@@ -304,6 +304,30 @@ void riscv_dmi_init(riscv_dmi_s *const dmi)
 	do {
 		/* Read out the DM's status register */
 		uint32_t dm_status = 0;
```
dragonmux, line 306: This one still needs moving down to line 330 please 🙂

src/target/riscv_debug.c
```diff
@@ -304,6 +304,30 @@ void riscv_dmi_init(riscv_dmi_s *const dmi)
...
+		/* After changing the value of dm_active, the debugger must poll dmcontrol
+		 * until dm_active has taken the requested value */
+		bool dm_active = false;
+		uint32_t dm_control = 0;
+		while (!dm_active) {
+			if (!riscv_dmi_read(dmi, base_addr + RV_DM_CONTROL, &dm_control)) {
+				DEBUG_ERROR("error turning on DM!\n");
+				return;
+			}
+			dm_active = dm_control & 1;
```
dragonmux, line 323: Please suffix this constant with `U` - this does a signed-unsigned conversion that can trigger UB otherwise during application of the mask operation (changing the sign bit on a signed number is not permitted for bitwise operations).

src/target/riscv_debug.c
```diff
@@ -304,6 +304,30 @@ void riscv_dmi_init(riscv_dmi_s *const dmi)
...
+			uint8_t counter = 0;
```
dragonmux, line 324: We would expect to see this defined just outside the `while` loop on line 317 - otherwise it won't have the intended effect.

Please suffix the initialiser and the comparison value with `U` to make them both unsigned.

dragonmux (commented):
This is looking much better 🙂. We've left a review reply on #2063 for one of the notes but have otherwise done a fresh review here. We are satisfied with the items you've addressed and mostly the items here are for subtle bugs we spotted.

We look forward to this being sorted so we can schedule its merge into v2.1 when the merge window opens for that.

src/target/riscv_debug.c
orsolic: I left this out because it wasn't part of the original change for ESP32-P4. Moved to line 330 and added 'U' prefix which, I assume, is needed.

dragonmux: Aye, indeed - thank you!

PR #2065, opened 2025-01-28
Feature: change platform GPIO slew rate when high SWCLK frequency is requested

ALTracer:
## Detailed description

* This is a new feature.
* The existing problem is poor signal integrity (ringing/reflections, ground bounce) of `stlinkv3` and `blackpill-f4` platforms resulting in failure of initial scan on setups with poor ground connections. 
* This PR solves it by runtime-matching STM32 GPIO OSPEED to the toggle rates indicated in datasheets.

Prior changes: 
`stlinkv3`: #1945, v1.10.0-1289-g 27cbec7e
`blackpill-f4`: #1717, v1.10.0-500-g b9969de3; and #2044, v2.0.0-rc1-3-g a3bb02e9

Because STLINK/V3E onboard Nucleo-H743ZI2 outright didn't work at highest reachable frequency (above 12 MHz) at 2 MHz slew rate, I gave 25 MHz a try and it worked. That platform is able to checksum target data at about 400 KiB/s.
MiniSTM32F4x1Cx is slower, but still can hit 8.72 MHz (96/11) without significant firmware changes, and this is uncomfortably outside DS10314 ranges (C_L = 10 pF, Vdd > 2.7, F_max(IO)out = 8 MHz).
But to reduce noise at lower frequencies, I suggest a platform callback which reverts to OSPEED_2MHz when crossing some frequency boundary. There are no series resistors, no buffers, and we can't expect users to always use two good ground wires, especially when other adapters work fine with just one. Shields and carriers can help this.
I don't think STM32F1-based platforms are affected, because a) fastest SWCLK primitives (no_delay) only reach 3.6-4.0 MHz, b) datasheet indicates 2 and 10 MHz f_max(IO)out for C_L = 50 pF.

## Closing issues

### DIFF
--- a/src/include/platform_support.h
+++ b/src/include/platform_support.h
@@ -70,6 +70,7 @@ void platform_max_frequency_set(uint32_t frequency);
 uint32_t platform_max_frequency_get(void);
 
 void platform_target_clk_output_enable(bool enable);
+void platform_ospeed_update(uint32_t frequency);
 
 #if CONFIG_BMDA == 0
 bool platform_spi_init(spi_bus_e bus);
--- a/src/platforms/common/blackpill-f4/blackpill-f4.c
+++ b/src/platforms/common/blackpill-f4/blackpill-f4.c
@@ -197,6 +197,15 @@ uint32_t platform_target_voltage_sense(void)
 }
 #endif
 
+void platform_ospeed_update(const uint32_t frequency)
+{
+	const uint8_t ospeed = frequency > 2000000U ? GPIO_OSPEED_25MHZ : GPIO_OSPEED_2MHZ;
+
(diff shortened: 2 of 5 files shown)


birkenfeld:
Works for me, however I have to set SWD frequency to 2 MHz to get the low speed because the default is 3 MHz.

Could you either change the default to 2 MHz or change the cutoff to 3 MHz? People who want really high speed will have to set the frequency anyway, while people like me who expect things to "just work" will be initially just as surprised with the cutoff in this patch.

Thanks!

dragonmux:
We'll add that: despite your self-classification as a feature, because of what this changes and that it fixes a reliability issue, we are treating this as a fix for a bug.

PR #1280, opened 2022-11-06
blackpill, bluepill: led, NRST, STM32F401CCU6, STM32F401CEU6

koendv:
## Detailed description

Various small patches.
This PR is submitted so people who search for these features find it. 

- blackpill:
  - LED blinks if bmp is active (like on stlinkv2)
  - support for NRST pin in case target needs to connect under reset
  - same binary for STM32F401CCU6, STM32F401CEU6 and STM32F411CEU6
  - compiles with: ``make PROBE_HOST=blackpillv2 ENABLE_RTT=1``
  
- blue pill:
  - LED blinks if bmp is active (like on stlinkv2)
  - compiles with: ``make PROBE_HOST=stlink BLUEPILL=1 ENABLE_RTT=1``

HTH

## Your checklist for this pull request

* [x] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md)
* [x] I've read the [guidelines for contributing](https://github.com/blackmagic-debug/blackmagic/blob/main/CONTRIBUTING.md) to this repository
* [x It builds for hardware native (`make PROBE_HOST=native`)
* [x] It builds as BMDA (`make PROBE_HOST=hosted`)
* [x] I've tested it to the best of my ability
* [x] My commit messages provide a useful short description of what the commits do

PR #2068, opened 2025-02-02
blackpill-f4: Enable internal weak pull-up on nRST, simplify set_val

ALTracer:
## Detailed description

* This is a bug fix for an half-implemented feature.
* The existing problem is floating nRST pin of `blackpill-f4`, which is, at least, inconsistent with other in-tree platforms.
* The PR solves it by enabling internal weak pull-up resistor on it.

The original implementation was contributed like this in #1280 v1.8.0-702-g 02986d67 and such code snippet is correct for F1 GPIO, where resistors are part of input stage, but superflous for F4 GPIO, which retains PU/PD config across all digital modes (Input, Output, AF). I don't even see a reason to change modes or output-settings away from Output + Open-Drain + Pull-up + Lowest OSPEED. `gpio_clear` activates the low FET, `gpio_set` deactivates it, so `GPIO_ACTIVE_LOW` in Linux device-tree gpio bindings speak, or inverse polarity (`!assert`). See `stlinkv3`, except I omit the busy-delay, more on that later.
For platform_init, I guess neither TRST nor NRST were set up into output modes. For F4, `gpio_set()` during Digital Input mode does not enable internal weak pull-up, contrary to F1. The `gpio_clear()` during Output open-drain worked, but only when the wire was connected, and otherwise the level read back was whatever/floating. This interferes with BMD ADIv5 and Cortex-M layers logic. I can't tell for sure where exactly. Another adjacent problem is inconsistent nrst control implementation in platforms, I believe any delays should be hoisted up from platforms (volatile busy-delays) to callsites, but this is almost out of scope here.

Tested on `blackpill-f411ce` with no shields/carriers in default pinout. Stabilizes the scans where previously bogus romtable entries were logged. Removes "SWD scan failed!" occasions when wiring and power are otherwise correct (and target's SWJ-DP is mapped and functional etc.)

* [x] It builds for hardware native (see [Building the firmware](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-firmware)) *but only affects blackpill-f4*

## Closing issues

Fixes #2062.

### DIFF
--- a/src/platforms/common/blackpill-f4/blackpill-f4.c
+++ b/src/platforms/common/blackpill-f4/blackpill-f4.c
@@ -112,6 +112,10 @@ void platform_init(void)
 	gpio_set(TRST_PORT, TRST_PIN);
 	gpio_mode_setup(TRST_PORT, GPIO_MODE_OUTPUT, GPIO_PUPD_PULLUP, TRST_PIN);
 	gpio_set_output_options(TRST_PORT, GPIO_OTYPE_OD, GPIO_OSPEED_2MHZ, TRST_PIN);
+	/* Pull up nRST pin */
+	gpio_set(NRST_PORT, NRST_PIN);
+	gpio_mode_setup(NRST_PORT, GPIO_MODE_OUTPUT, GPIO_PUPD_PULLUP, NRST_PIN);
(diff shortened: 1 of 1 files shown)


dragonmux (approved):
This LGTM - please rebase on `main` and we'll get this merged. Thank you for the contribution and tidy bugfix!

PR #372, opened 2018-07-16
[efm32] Add support for EFM32 devices with different DI and MSC layouts

richardeoin:
* DI layout is identified by attempting to read OUI from both layouts
* MSC address is passed to flashstub in r3

Retested EZR32LG230 (EZR Leopard Gecko M3)
Tested EFR32BG13P532F512GM32 (EFR Blue Gecko)

Achieves aims of PR #264 (I think) Thanks to @dholth and @ryankurte for inspiration
Fixes Issue #226

PR #2069, opened 2025-02-03
[efm32_probe] Fix detection of DI v2

mmmspatz:
## Detailed description

#402 introduced a bug with silabs devices (efm32/efr32) related to the parsing of the DI (Device Info) page. An example is [efr32xg13](https://www.silabs.com/documents/public/reference-manuals/efr32xg13-rm.pdf), which I believe would currently be broken. I do not have any of the relevant devices and cannot test this, but I noticed it while playing with adding support for the newer "series 2" parts (eg. EFRMG24).

The driver currently knows about two versions (it calls them "version 1" and "version 2") of the DI page with completely different layouts. It decides which version to use by:

1. Reading the EUI64 from the location specified by DI layout version 1
2. checking if the 24 bit OUI therein is equal to `EFM32_V1_DI_EUI_SILABS`, in which case layout version 1 is used
3. If not, checking if that same OUI is instead equal to `EFM32_V2_DI_EUI_ENERGYMICRO`, in which case version 2 is used
4. Defaulting to version 1 if neither OUI is matched

The problem with this if `EFM32_V2_DI_EUI_ENERGYMICRO` were to appear in the DI page, it woud be in the location specified by DI layout version 2, and so we need to perform another read from there. There's no way for the current code to ever detect v2.

So, that is the change this PR makes: Also try reading the OUI per layout version 2 (from the EUI48, because there is no EUI64 in v2). This is the behavior originally introduced in #372.

As #402 discusses, this strategy of relying on the OUI to differentiate parts versions is bad in the first place because there are over 100 OUIs assigned to Silicon Labs, and it's not actually known which devices use which OUI. A better idea might be to use the PARTNO value from the IDCODE register. It is mentioned in [AN1303](https://www.silabs.com/documents/public/application-notes/an1303-efr32-dci-swd-programming.pdf) that all series 2 parts (so, not what this driver supports) have `IDCODE == 0x6BA02477`, perhaps there is similar simple rule for the older parts supported by this driver? 

### DIFF
--- a/src/target/efm32.c
+++ b/src/target/efm32.c
@@ -377,6 +377,12 @@ static uint64_t efm32_v1_read_eui64(target_s *t)
 	return ((uint64_t)target_mem32_read32(t, EFM32_V1_DI_EUI64_1) << 32U) | target_mem32_read32(t, EFM32_V1_DI_EUI64_0);
 }
 
+/* Reads the EFM32 Extended Unique Identifier EUI48 (V2) */
+static uint64_t efm32_v2_read_eui48(target_s *t)
+{
+	return ((uint64_t)target_mem32_read32(t, EFM32_V2_DI_EUI48H) << 32U) | target_mem32_read32(t, EFM32_V2_DI_EUI48L);
+}
+
 /* Reads the Unique Number (DI V2 only) */
 static uint64_t efm32_v2_read_unique(target_s *t, uint8_t di_version)
 {
@@ -538,12 +544,13 @@ bool efm32_probe(target_s *t)
 {
 	/* Check if the OUI in the EUI is silabs or energymicro.
 	 * Use this to identify the Device Identification (DI) version */
-	uint8_t di_version = 1;
-	uint64_t oui24 = ((efm32_v1_read_eui64(t) >> 40U) & 0xffffffU);
-	if (oui24 == EFM32_V1_DI_EUI_SILABS) {
+	uint8_t di_version;
+	const uint32_t oui24_v1 = (efm32_v1_read_eui64(t) >> 40U) & 0xffffffU;
+	const uint32_t oui24_v2 = (efm32_v2_read_eui48(t) >> 24U) & 0xffffffU;
+	if (oui24_v1 == EFM32_V1_DI_EUI_SILABS) {
 		/* Device Identification (DI) version 1 */
 		di_version = 1;
-	} else if (oui24 == EFM32_V2_DI_EUI_ENERGYMICRO) {
+	} else if (oui24_v2 == EFM32_V2_DI_EUI_ENERGYMICRO) {
 		/* Device Identification (DI) version 2 */
 		di_version = 2;
 	} else {


src/target/efm32.c
```diff
@@ -538,12 +544,12 @@ bool efm32_probe(target_s *t)
...
-	uint8_t di_version = 1;
-	uint64_t oui24 = ((efm32_v1_read_eui64(t) >> 40U) & 0xffffffU);
-	if (oui24 == EFM32_V1_DI_EUI_SILABS) {
+	uint8_t di_version;
+	uint64_t oui24;
+	if (oui24 = ((efm32_v1_read_eui64(t) >> 40U) & 0xffffffU), oui24 == EFM32_V1_DI_EUI_SILABS) {
 		/* Device Identification (DI) version 1 */
 		di_version = 1;
-	} else if (oui24 == EFM32_V2_DI_EUI_ENERGYMICRO) {
+	} else if (oui24 = ((efm32_v2_read_eui48(t) >> 24U) & 0xffffffU), oui24 == EFM32_V2_DI_EUI_ENERGYMICRO) {
```
dragonmux, line 552: Per the lint pass, do not do assignments in `if` statements - as this is necessary, split these out into two separate assignments and checks please.

dragonmux (changes requested):

src/target/efm32.c
```diff
@@ -538,12 +544,13 @@ bool efm32_probe(target_s *t)
 {
 	/* Check if the OUI in the EUI is silabs or energymicro.
 	 * Use this to identify the Device Identification (DI) version */
-	uint8_t di_version = 1;
-	uint64_t oui24 = ((efm32_v1_read_eui64(t) >> 40U) & 0xffffffU);
-	if (oui24 == EFM32_V1_DI_EUI_SILABS) {
+	uint8_t di_version;
+	uint64_t oui24_v1 = ((efm32_v1_read_eui64(t) >> 40U) & 0xffffffU);
+	uint64_t oui24_v2 = ((efm32_v2_read_eui48(t) >> 24U) & 0xffffffU);
```
dragonmux, line 549: These can be made `const`, so let's do so. They also look to have masks that reduce them to 32 bits or less, so let's have them as `uint32_t`'s - this will improve codegen a bit. The outer parens are unnecessary and can be dropped.

dragonmux (changes requested):
We can see just one issue remaining which is to do with types/`const`, but otherwise this looks great. Please note your commit message needs to conform with the contribution guidelines for the repo. `efm32: ` as the prefix not that square bracket expression.

With that taken care of, we'll be happy to approve this and merge as a fix for v2.0 as this looks like a regression as you noted.

dragonmux:
> A better idea might be to use the PARTNO value from the IDCODE register. It is mentioned in [AN1303](https://www.silabs.com/documents/public/application-notes/an1303-efr32-dci-swd-programming.pdf) that all series 2 parts (so, not what this driver supports) have IDCODE == 0x6BA02477, perhaps there is similar simple rule for the older parts supported by this driver?

This is an ARM part number, so is not unique or identifying to these devices in any way. This only tells us that we're talking with a ADIv5 DPv2 part. Every ADIv5 DPv2 part shares this ID code (modulo the first nibble (6) which is a version number). `477` -> `43b` when decoded, which is ARM's JEP-106. `ba02` is the partno for a DP w/ the last nibble (2) indicating version. ba00 -> DPv0, ba01 -> DPv1, ba02 -> DPv2, and ba03 -> DPv3 (ADIv6).

dragonmux (approved):

PR #2073, opened 2025-02-11
blackpill-f4: change slew speed frequency cutoff to 3 MHz

birkenfeld:
## Detailed description

In #1717, the output slew rate for the JTAG/SWD pins on bluepill-f4 was changed to 25MHz. This leads to failing target scans even with low bus speeds when the target isn't connected by e.g. two ground wires. In #2065, this was mitigated by switching output slew rate depending on requested speed. However, the cutoff was set at 2MHz, while the default bus speed is 3MHz. This keeps the bad out-of-the-box experience for users that don't know or are not able to set their own bus frequency.

The 2MHz slew rate is fine for a signal frequency of 3MHz, see datasheet DS10314.

cc @ALTracer 

### DIFF
--- a/src/platforms/common/blackpill-f4/blackpill-f4.c
+++ b/src/platforms/common/blackpill-f4/blackpill-f4.c
@@ -143,7 +143,7 @@ void platform_init(void)
 	OTG_FS_GCCFG &= ~(OTG_GCCFG_VBUSBSEN | OTG_GCCFG_VBUSASEN);
 
 	/* By default, do not drive the SWD bus too fast. */
-	platform_max_frequency_set(3000000);
+	platform_max_frequency_set(2000000);
 }
 
 void platform_nrst_set_val(bool assert)


dragonmux:
Because the datasheet only gives us a guarantee that this slew rate will work to 2MHz (this is why the constant is "2MHz", not "4MHz" or another frequency; even if your specific silicon can do better, at your given voltage and temperature conditions; this does not mean all of the silicon used on Black Pills can at all voltages and temperatures), we would prefer the default frequency be dropped for this platform rather than bumping the slew cut-over point.

No need to close this PR to do that though, just switch the default `target_clk_divider` for this platform to one that achieves approximately 2MHz. Given how that particular global gets its default value, we would suggest making this change in the platform's `platform_init()` function toward the end of its execution.

ALTracer:
> we would prefer the default frequency be dropped for this platform

I agree with that approach, @birkenfeld edit this commit to reduce 3000000 to 2000000 argument in last call in platform_init() while keeping PR open, then I'll also agree with you. The fact that it initially worked for me on my wiring was one of strongest reasons this was contibuted (and lack of wide testing). Apparently we can't rely on 25°C and 3.3 Vdd.

birkenfeld:
Done!

dragonmux (approved):

PR #2074, opened 2025-02-11
command: pass optional baudrate argument to swo_init in Manchester mode

Misaka0x2730:
## Detailed description

I'm working on a custom platform, and because my Manchester decoding implementation is based on the RP2040 PIO, I need to pass the incoming signal's baud rate to the device in Manchester mode as well.  
I believe this change will not affect other platforms, since the baudrate argument in the swo command is optional, and in the general swo_init implementation, the baudrate is only used in UART mode.

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -666,7 +666,7 @@ static bool cmd_swo_enable(int argc, const char **argv)
 #endif
 #if SWO_ENCODING == 2 || SWO_ENCODING == 3
 	/* Handle the optional baud rate argument if present */
-	if (capture_mode == swo_nrz_uart && argc > decode_arg && argv[decode_arg][0] >= '0' && argv[decode_arg][0] <= '9') {
+	if (argc > decode_arg && argv[decode_arg][0] >= '0' && argv[decode_arg][0] <= '9') {
 		baudrate = strtoul(argv[decode_arg], NULL, 0);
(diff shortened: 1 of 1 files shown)


dragonmux:
Rather than using merge commits, please use rebase to update the branch instead.

Been giving the suggested change some careful thought, and well.. we have some reservations about accepting it, and some notes if this were to become something we did merge:

1) Why is it that your code needs the baud rate in Manchester mode? This mode needs to use CDR to extract the clock encoded in the bitstream back out and then recover the data, and because of how that works it inherently doesn't need a bit rate known in advance so we strongly question the path being taken if you're writing code that does.
2) If we actually accept this, the documentation for this should be updated to make the change discoverable.
3) The baud rate value must remain async-mode-only for all in-tree platforms as because of point 1 it doesn't make any sense for any of the in-tree stuff which is why the condition in the first place.

Unsure how you're planning to use the PIO to decode Manchester mode, but if you can find a way that allows it to "learn" the baud rate at runtime by looking at rising-to-falling/falling-to-rising edge timings, that would be significantly better. More robust, less error-prone, and feature parity with the in-tree platforms, noting that it is completely valid and okay for a target to change clocking and thus baud rate while keeping SWO enabled in Manchester mode, and BMD will automatically track this for the user and keep decoding.

PR #2075, opened 2025-02-12
Fix: Declare correct RAM sizes for STM32F7

### LINKED ISSUE #2006: [STM32F72x] Incorrect memory map
I tried debugging an STM32F722 with blackmagic probe, and it worked... until my code grew bigger.
I looked into the code, the memory region for both flash and RAM are incorrect.

# How to reproduce
- Place a breakpoint in a code located after 0x210000 (flash on ITCM interface)
- `set mem inaccessible-by-default off`
- Run code: the code does not stop at the breakpoint

# Solution

## RAM
It had no effect in my situation, but I replaced these two lines:
https://github.com/blackmagic-debug/blackmagic/blob/729b7730cf64b7dd46c8e472c0bc622051a48730/src/target/stm32f4.c#L425-L426
with
```
target_add_ram32(target, 0x20000000, 0x10000); /* 64kiB DTCM RAM */
target_add_ram32(target, 0x20010000, 0x30000); /* 192kiB RAM */
```

## Flash
And I made it work by replacing these 2 lines:
https://github.com/blackmagic-debug/blackmagic/blob/729b7730cf64b7dd46c8e472c0bc622051a48730/src/target/stm32f4.c#L477-L478
with
```
if (is_f7) {
    stm32f4_add_flash(target, ITCM_BASE, 0x10000, 0x4000, 0, split);
    stm32f4_add_flash(target, ITCM_BASE+0x10000, 0x10000, 0x10000, 4, split);
    stm32f4_add_flash(target, ITCM_BASE+0x20000, 0x60000, 0x20000, 5, split);
}
```

Which is exactly the code a few lines below, except the STM32F72xxx is not dual bank.
https://github.com/blackmagic-debug/blackmagic/blob/729b7730cf64b7dd46c8e472c0bc622051a48730/src/target/stm32f4.c#L486-L492

# Next steps
I am not confident in doing a pull request because of the intricate code paths.
If I had to do it, I would probably try to replace all the `if... else...` by an array of structures defining the memory map for each devices. I am not sure that this is the path you want to take, and maybe there are issues with this method I don't foresee, so I am leaving it to you to decide what to do.

ALTracer:
## Detailed description

* No new features.
* The existing problem is RTT Auto Search triggering lots of SWD Fault past end of physical SRAM on STM32F722.
* This PR solves it by shrinking declared memory map and dispatching datasheet values per-family.

As I've tried to inspect/debug a project on a board involving STM32F722, the RTT Auto Search feature (which is supposed to be plug&play without configuration) started scrubbing available RAM but fell off the 192 KiB cliff trying to scan all 384 KiB of AHB SRAM which should be present on STM32F777. Confirmed with `blackpill-f411ce` and BMDA. I know about `mon rtt mem` ranges but this sounds more like a workaround rather than fix, BMD should provide knowledge of DUT flash and RAM maps and sizes.

Note this was not rigorously tested with a real project yet. I don't like the seven function calls, this can be further reduced to three via stack locals, should be cheaper than a full LUT. Update: reduced, was +80, now is +48 bytes size-diff.
Tested on STM32F722RE to no longer generate "SWD access resulted in fault" messages during reads outside RAM bounds, and to create a smaller correct RAM map.

## Closing issues

Fixes #2006.

### DIFF
--- a/src/target/stm32f4.c
+++ b/src/target/stm32f4.c
@@ -32,6 +32,17 @@
  *   https://www.st.com/resource/en/reference_manual/rm0401-stm32f410-advanced-armbased-32bit-mcus-stmicroelectronics.pdf
  * GD32F4xx Arm® Cortex®-M4 32-bit MCU User Manual, Rev. 3.0
  *   https://www.gigadevice.com.cn/Public/Uploads/uploadfile/files/20240407/GD32F4xx_User_Manual_Rev3.0.pdf
+ * References for STM32F7xx:
+ * RM0385 - STM32F75xxx and STM32F74xxx advanced Arm®-based 32-bit MCUs, Rev. 9
+ *   https://www.st.com/resource/en/reference_manual/rm0385-stm32f75xxx-and-stm32f74xxx-advanced-armbased-32bit-mcus-stmicroelectronics.pdf
+ * RM0410 - STM32F76xxx and STM32F77xxx advanced Arm®-based 32-bit MCUs, Rev. 5
+ *   https://www.st.com/resource/en/reference_manual/rm0410-stm32f76xxx-and-stm32f77xxx-advanced-armbased-32bit-mcus-stmicroelectronics.pdf
+ * RM0431 - STM32F72xxx and STM32F73xxx advanced Arm®-based 32-bit MCUs, Rev. 4
+ *   https://www.st.com/resource/en/reference_manual/rm0431-stm32f72xxx-and-stm32f73xxx-advanced-armbased-32bit-mcus-stmicroelectronics.pdf
+ * AN4667 - STM32F7 Series system architecture and performance, Rev. 4
+ *   https://www.st.com/resource/en/application_note/an4667-stm32f7-series-system-architecture-and-performance-stmicroelectronics.pdf
+ * AN4826 - STM32F7 Series Flash memory dual bank mode, Rev. 2
+ *   https://www.st.com/resource/en/application_note/an4826-stm32f7-series-flash-memory-dual-bank-mode-stmicroelectronics.pdf
  */
 
 #include "general.h"
--- a/src/target/stm32f4.c
+++ b/src/target/stm32f4.c
@@ -423,17 +423,21 @@ static bool stm32f4_attach(target_s *const target)
 	/* And rebuild the RAM map */
 	bool use_dual_bank = !is_f7 && dual_bank;
 	if (is_f7) {
-		target_add_ram32(target, 0x00000000, 0x4000); /* 16kiB ITCM RAM */
-		if (target->part_id == ID_STM32F72X) {
-			target_add_ram32(target, 0x20000000, 0x10000); /* 64kiB DTCM RAM */
-			target_add_ram32(target, 0x20010000, 0x30000); /* 192kiB SRAM1/2 (176+16) */
-		} else if (target->part_id == ID_STM32F74X) {
-			target_add_ram32(target, 0x20000000, 0x10000); /* 64kiB DTCM RAM */
-			target_add_ram32(target, 0x20010000, 0x40000); /* 256kiB SRAM1/2 (240+16) */
-		} else if (target->part_id == ID_STM32F76X) {
-			target_add_ram32(target, 0x20000000, 0x20000); /* 128kiB DTCM RAM */
-			target_add_ram32(target, 0x20020000, 0x60000); /* 384kiB SRAM1/2 (368+16) */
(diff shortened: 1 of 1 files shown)


ALTracer:
Seven function calls costed more flash than two calls, so the #1710 approach is not optimal either.
Clarify whether I should also add the 8 datasheets and 3 reference manuals (RM0385, RM0410, RM0431) for STM32F7 to top of file.
I also have a minor fix for ITCM inconsistency for a follow-up PR, but it worsens `blank_check` behaviour (will check twice). Finally, I also will have a medium fix for STM32F4 RAM being pinned to `320kiB` which is too blanket, in another PR.

dragonmux:
> Clarify whether I should also add the 8 datasheets and 3 reference manuals (RM0385, RM0410, RM0431) for STM32F7 to top of file.

It would be best if these are specified at the top of the files, yes please - means it's easier to find them again to look them up and get the references as needed for contributors.

src/target/stm32f4.c
```diff
@@ -423,9 +423,21 @@ static bool stm32f4_attach(target_s *const target)
...
 	bool use_dual_bank = !is_f7 && dual_bank;
 	if (is_f7) {
-		target_add_ram32(target, 0x00000000, 0x4000);  /* 16kiB ITCM RAM */
-		target_add_ram32(target, 0x20000000, 0x20000); /* 128kiB DTCM RAM */
-		target_add_ram32(target, 0x20020000, 0x60000); /* 384kiB RAM */
+		uint32_t dtcm_size = 64U * 1024U; /* 64kiB DTCM RAM */
+		uint32_t ahbsram_size = 16U * 1024U;
+		if (target->part_id == ID_STM32F72X) {
+			ahbsram_size = (176U + 16U) * 1024U; /* 192kiB SRAM1/2 */
+		} else if (target->part_id == ID_STM32F74X) {
```
dragonmux, line 430: This is a single statement `if` statement - drop the braces please. Same for the following `else if`.

src/target/stm32f4.c
```diff
@@ -423,9 +423,21 @@ static bool stm32f4_attach(target_s *const target)
...
+			ahbsram_size = (240U + 16U) * 1024U; /* 256kiB SRAM1/2 */
+		} else if (target->part_id == ID_STM32F76X) {
+			dtcm_size = 128U * 1024U;            /* 128kiB DTCM RAM */
+			ahbsram_size = (368U + 16U) * 1024U; /* 384kiB SRAM1/2 */
+		}
+		target_add_ram32(target, 0x00000000, 0x4000); /* 16kiB ITCM RAM */
+		/* On STM32F7, DTCM and AHB SRAM are contiguous */
+		target_add_ram32(target, 0x20000000, dtcm_size);
+		target_add_ram32(target, 0x20000000 + dtcm_size, ahbsram_size);
```
dragonmux, line 439: As you are here working on these, please `U` suffix the addresses and size information and turn them into `#define`'s at the top of file as appropriate.

dragonmux (commented):
This is looking good - there are just a couple of formatting things. With them taking care of, we'll happily get this merged for v2.0.0.

src/target/stm32f4.c
ALTracer: Suffixed base addresses in new/edited code. Hoisted `STM32F7_DTCM_BASE` up top. Ignored 0x0U ITCM RAM base for now (`ITCM_BASE` is for the Flash alias, not RAM), and I'm not sure someone needs an alias for address zero.

dragonmux: We do as it has specific meaning and is a valid address in the map - given it's the `ITCM_BASE`, please name it as `STM32F7_ITCM_BASE`.

ALTracer: Named it `STM32F7_ITCM_RAM_BASE`, extending the other newly added names as well.

src/target/stm32f4.c
```diff
@@ -99,6 +110,11 @@
 #define AXIM_BASE 0x8000000U
 #define ITCM_BASE 0x0200000U
 
+#define STM32F7_ITCM_RAM_BASE 0x0U
```
dragonmux, line 113: Do note this should really have all 8 digits present for consistency please - we know they're all 0's but this at least makes the constants all follow the same form.

dragonmux (commented):
This seems reasonable. We've got just the one review note and then we think this should be good to mark as ready for merge.

src/target/stm32f4.c
ALTracer: Amended and fast-forwarded.

dragonmux (approved):

PR #2078, opened 2025-02-16
misc: Add bootloader build instructions to README.md

9names:
## Detailed description

Add basic instructions on how to build the BMP bootloader to README.md.
The existing documentation tells you how to build probe firmware but assumes you've already got a working bootloader - took me a while to work out how to build this, so I figure it is worth documenting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,6 +227,15 @@ You should now see the resulting binaries in `build`, in this case:
 
 These are the binary files you will use to flash to your probe.
 
+#### Building the bootloader
+
+If your probe hardware does not already have a bootloader on it, after building firmware as described above
+you can build the appropriate bootloader with the following command:
+
+```sh
+meson compile -C build boot-bin
+```
+
 ##### region `rom' overflowed
 
 It may happen, while working with non default configurations or the project's latest version from Git,


dragonmux (approved):
We do actually document this on the website in the [firmware hacking guide](https://black-magic.org/hacking/hacking.html), but having it more readily discoverable here and in such a cleanly described way seems reasonable to us, so LGTM - merging.

9names:
I think I skimmed through the hacking guide but missed that section.
If the bootloader commands were in a code block like every other console command on that page, and used meson to invoke the command rather than `cd`'ing and then invoking `ninja` directly, maybe it would stand out more?

I can't find a repo for the firmware hacking guide to make a PR for that too, so here's what I suggest for the updated hacking guide text:

---

To start, we need to create a build directory from a configuration with Meson and then ask Meson to run the build:

```console
meson setup build --cross-file=cross-file/native.ini
meson compile -C build
```

This will result in the following binary files in the build directory:

    blackmagic_native_firmware.elf - ELF binary of the Black Magic debug probe.

    blackmagic_native_firmware.bin - Flat binary of the Black Magic debug probe, load at 0x8002000.

If you need the bootloader as well, after running the above build run

```console
meson compile -C build boot-bin
```
which will compile the bootloader and generate a .bin file for it for use with provisioning tools. This results in two files:

    blackmagic_native_bootloader.elf - ELF binary of the Black Magic DFU bootloader.

    blackmagic_native_bootloader.bin - Flat binary of the DFU bootloader, load at 0x8000000.

```

dragonmux:
So, our only comments on the website side of things are: we did not know `meson compile` can be used to invoke specific sub-targets, but that's mostly irrelevant as it is just `ninja` under the hood anyway and we specifically want users to think before they go building the bootloader as it is *only* required for provisioning and never general updates and the like.

As for where the repo is - it's in the GH organisation for this project - [blackmagic-debug/black-magic-org](https://github.com/blackmagic-debug/black-magic-org). All open source.

PR #2083, opened 2025-02-22
[efm32_probe] Fix detection of DI v2 (backport to v1.10 branch)

mmmspatz:
See: https://github.com/blackmagic-debug/blackmagic/pull/2069

This is a clean cherry pick of commit 21127488158718ecd1cbf9a281e6db35f951d5a8


dragonmux:
Given this isn't a core "debug is broken, the firmware crashes", type fix, and that the v1.x branches are going into maintenance mode quite shortly in preference for pushing people onto the v2.x firmware series that is releasing imminently, we will not be backporting this at this time. Our advice for people using the EFM32 parts is to use the new firmware as it fixes a multitude of other bugs in the ADIv5 layer and remote protocol too.

PR #2090, opened 2025-03-12
Add basic TMS570 support

xobs:
## Detailed description

Add basic support for TI TMS570. This supports enumeration of the target, including configuring the ICEPICK-TYPE-C device present on the chip to enable enumeration.

This does not yet support flashing. Support for writing flash will need to come once the flash process is reverse engineered, or a flash stub will need to be created using official firmware.


src/include/buffer_utils.h
```diff
@@ -84,6 +108,15 @@ static inline uint64_t read_be8(const uint8_t *const buffer, const size_t offset
...
 
+static inline uint64_t read_le8(const uint8_t *const buffer, const size_t offset)
+{
+	return ((uint64_t)buffer[offset + 7] << 56U) | ((uint64_t)buffer[offset + 6] << 48U) |
+		((uint64_t)buffer[offset + 5] << 40U) | ((uint64_t)buffer[offset + 4] << 32U) |
+		((uint64_t)buffer[offset + 3] << 24U) | ((uint64_t)buffer[offset + 2] << 16U) |
+		((uint64_t)buffer[offset + 1] << 8U) | buffer[offset + 0];
+}
+
+
```
dragonmux, line 119: Extraneous new line.

src/target/cortexar.c
```diff
@@ -886,6 +888,17 @@ static target_s *cortexar_probe(
...
+	/* Read the CPSR to figure out if this target is big endian. This clobbers r0, so save
+	 * r0 beforehand and restore it after the read is complete.
+	 */
+	uint32_t r0 = cortexar_core_reg_read(target, 0U);
+	cortexar_run_insn(target, ARM_MRS_R0_CPSR_INSN);
+	priv->core_regs.cpsr = cortexar_core_reg_read(target, 0U);
+	cortexar_core_reg_write(target, 0U, r0);
+	if (priv->core_regs.cpsr & CORTEXAR_CPSR_BE) {
+		target->target_options |= TOPT_FLAVOUR_BE;
+	}
```
dragonmux, line 900: The braces here are superfluous and can be dropped.

src/target/cortexar.c
```diff
@@ -1171,7 +1223,13 @@ static inline bool cortexar_mem_write_fast(target_s *const target, const uint32_
 
 	/* Write each of the uint32_t's checking for failure */
 	for (size_t offset = 0; offset < count; ++offset) {
-		if (!cortexar_run_write_insn(target, ARM_STC_DTRRX_R0_POSTINC4_INSN, src[offset]))
+		uint32_t value;
+		if (target->target_options & TOPT_FLAVOUR_BE) {
+			value = read_be4((const void *const)src, offset * 4);
+		} else {
+			value = read_le4((const void *const)src, offset * 4);
+		}
```
dragonmux, line 1231: The braces here are superfluous and can be dropped. We would be surprised if `src` needs casing at all, it should automatically cast to `void *` and get `const` added as needed (removal of const requires casts, not addition).

Please suffix `4` with `U` to make the multiplication unsigned.

src/target/cortexar.c
```diff
@@ -1334,7 +1428,26 @@ static size_t cortexar_reg_read(target_s *const target, const uint32_t reg, void
...
+			write_le4(data, 0, value);
+			break;
+		}
+		case 8: {
+			uint64_t value;
+			if (target->target_options & TOPT_FLAVOUR_BE)
+				value = read_be8(reg_ptr, 0);
+			else
+				value = read_le8(reg_ptr, 0);
+			write_le4(data, 0, value);
```
dragonmux, line 1447: Should this not be `write_le8()`?

dragonmux (commented):
This looks to be progressing nicely, so we are giving you an interim review. Think our plan for merging this will be to split the AM335x support branch into the ICEPick part and then the AM335x part, and PR the ICEPick part for v2.1, that way this PR has a cleaner base to work from. If that sounds good to you, then we'll get the branches sorted out.

src/target/cortexar.c
xobs: The cast is required because the `read_*` functions take a `const uint8_t *const` and not `void *`:

```
../src/target/cortexar.c:1229:21: warning: incompatible pointer types passing 'const uint32_t *const' (aka 'const unsigned int *const') to parameter of type 'const uint8_t *' (aka 'const unsigned char *') [-Wincompatible-pointer-types]
 1229 |                         value = read_be4(src, offset * 4);
      |                                          ^~~
../src/include/buffer_utils.h:97:54: note: passing argument to parameter 'buffer' here
   97 | static inline uint32_t read_be4(const uint8_t *const buffer, const size_t offset)
      |                                                      ^
```

src/target/cortexar.c
xobs: It should be.

src/target/cortexar.c
dragonmux: In that case, you can simplify the cast to just `(const void *)` - the compiler will add the final `const` for us.

Issue #2094, opened 2025-03-18
Arm-Trustzone: Potential buffer/stack  overflow


mean00:
Hi,
I've been running into issues with cortex M33 recently (RP2350 and others)

One of the problem may be a buffer/stack  overflow. I spotted at least one.

If the chip has trustzone + FPU, the function cortexm_fault_unwind will only allocate registers to store GP + FPU registers, but the call to target_reg_reads will read all of them, including the trustzone ones.
That leads to a potential stack overflow.  In my case it was erasing "target" so it did not end well.

(NB, I was adding the msplim_[n]s/psplim_[n]s registers so it was triggered more easily)

Not sure what is the best way to solve this, it looks like a generic problem that could happen elsewhere.

Maybe   a CORTEXM_MAX_REG_COUNT  define and using that everywhere to allocate registers storage  ?
That way, there is a single spot to update with newer extension.

Thank you

dragonmux:
👋🏼 hi mean00,

Well figured out in regards that buffer bounds error - you are correct that w/ `CORTEXM_TOPT_TRUSTZONE` set, that would indeed lead to writing past the end of the regs buffer. If we are to define something like `CORTEXM_MAX_REG_COUNT` we would suggest it goes in the top of cortexm.c and is not exposed as things outside of this architecture implementation should not need or be relying on such a definition. It should only be necessary in the generic architecture implementation. We agree, however, that it would be the most straightforward way to prevent a part that has a FPU + TZ from causing BMD to smash stack.

If you would be willing to please PR such a fix, we will get it merged for v2.0's full release as we would say this is a bugfix at the very least.

PR #2095, opened 2025-03-18
cortemx: use a global define to make sure we allocate enough room for registers

mean00:
## Detailed description

This is a partial workaround for  #2094  
The core problem is depending on the option enabled ( FPU, Trustzone,..) the size needed for target_read/write_regs varies. Part of the code was wrongly computing the needed size and causing stack overflow/corruption and FPU+Trustzone was present.

The partial workaround is to use a define correctly settings the needed size and using it in cortemx.c
This is only partial because :
1- It is local only. There are similar wrong constructs in lpc55 code at least.
2- It does not really solve the core issue of ensuring the proper size has been allocated by design.

On option for the later would be to define a structure cortexm_registers / riscv_registers.. 
with the proper size. 

I tested it on my test case which adds more trustzone registers (msplim/psplim) and was crashing easily.
I dont have that crash any longer with that change injected. 
(The debugged MCU was a RP2350 with both trustzone and FPU)

* [ ] It builds for hardware native (see [Building the firmware](https://github.com/blackmagic-debug/blackmagic?tab
=readme-ov-file#building-black-magic-debug-firmware))

Does not link due to flash overflow

### DIFF
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -47,6 +47,9 @@
 
 #include <assert.h>
 
+/* This is the size (in 32 bits integers) you must allocate when reading cortem registers */
+#define CORTEXM_MAX_REG_COUNT (CORTEXM_GENERAL_REG_COUNT + CORTEX_FLOAT_REG_COUNT + CORTEXM_TRUSTZONE_REG_COUNT)
+
 static bool cortexm_vector_catch(target_s *target, int argc, const char **argv);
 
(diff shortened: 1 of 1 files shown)


dragonmux (approved):

Note that, on LPC55 devices we know what the core is and know how many registers those devices have - this is why that's actually safe on those parts. NXP would need to release new parts and, given their track record, would start a new part series if they were to introduce TZ in the LPC family. This seems sufficient for now.

PR #2096, opened 2025-03-20
Fix: minor deduplication in Puya driver

ALTracer:
## Detailed description

* No new features.
* The existing problem is code in `puya_flash_prepare()` is notably WET (or five-times), affecting BMF size in some configurations.
* This PR tries to dry it a bit by picking the low-hanging fruit: `epparaN` is replaced by an array and a loop.

Not tested on any py32 devices which I don't have.
Discovered via staring at `strings -n16 build-f411ce/blackmagic_blackpill_f411ce_firmware.bin | less` output.

First commit doesn't affect BMF size, or saves 136 bytes when DEBUG_TARGET is enabled (in Farpatch or else). Consider this as preparation to enabling `DEBUG_TARGET` firmware-wide. Which is unlikely to happen in v2.0, so this can be tracked to v2.1.

Second commit brings minor savings of 32 bytes in release, or 64 with logging, probably by eliminating 4 extra calls to `target_mem32_read32()` and `printf()`. It does lose `const`-ness of `eppara[5]` but the scope is fairly small. I guess I could even optimize this to a 20-byte block read `target_mem32_read()` but it'd change behaviour, and I don't know how Puya flash controller reacts to it.

CC #1817 by @ArcaneNibble 

### DIFF
--- a/src/target/puya.c
+++ b/src/target/puya.c
@@ -162,11 +162,11 @@ static bool puya_flash_prepare(target_flash_s *flash)
 	const uint32_t eppara2 = target_mem32_read32(flash->t, PUYA_FLASH_TIMING_CAL_BASE + hsi_fs * 20 + 8);
 	const uint32_t eppara3 = target_mem32_read32(flash->t, PUYA_FLASH_TIMING_CAL_BASE + hsi_fs * 20 + 12);
 	const uint32_t eppara4 = target_mem32_read32(flash->t, PUYA_FLASH_TIMING_CAL_BASE + hsi_fs * 20 + 16);
-	DEBUG_TARGET("PY32 flash timing cal 0: %08" PRIx32 "\n", eppara0);
-	DEBUG_TARGET("PY32 flash timing cal 1: %08" PRIx32 "\n", eppara1);
-	DEBUG_TARGET("PY32 flash timing cal 2: %08" PRIx32 "\n", eppara2);
-	DEBUG_TARGET("PY32 flash timing cal 3: %08" PRIx32 "\n", eppara3);
-	DEBUG_TARGET("PY32 flash timing cal 4: %08" PRIx32 "\n", eppara4);
+	DEBUG_TARGET("PY32 flash timing cal %d: %08" PRIx32 "\n", 0, eppara0);
+	DEBUG_TARGET("PY32 flash timing cal %d: %08" PRIx32 "\n", 1, eppara1);
+	DEBUG_TARGET("PY32 flash timing cal %d: %08" PRIx32 "\n", 2, eppara2);
+	DEBUG_TARGET("PY32 flash timing cal %d: %08" PRIx32 "\n", 3, eppara3);
+	DEBUG_TARGET("PY32 flash timing cal %d: %08" PRIx32 "\n", 4, eppara4);
 
 	target_mem32_write32(flash->t, PUYA_FLASH_TS0, eppara0 & 0xffU);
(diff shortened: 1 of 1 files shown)


src/target/puya.c
```diff
@@ -157,26 +157,20 @@ static bool puya_flash_prepare(target_flash_s *flash)
...
-	target_mem32_write32(flash->t, PUYA_FLASH_TS1, (eppara0 >> 16U) & 0x1ffU);
-	target_mem32_write32(flash->t, PUYA_FLASH_TS3, (eppara0 >> 8U) & 0xffU);
-	target_mem32_write32(flash->t, PUYA_FLASH_TS2P, eppara1 & 0xffU);
-	target_mem32_write32(flash->t, PUYA_FLASH_TPS3, (eppara1 >> 16U) & 0x7ffU);
-	target_mem32_write32(flash->t, PUYA_FLASH_PERTPE, eppara2 & 0x1ffffU);
-	target_mem32_write32(flash->t, PUYA_FLASH_SMERTPE, eppara3 & 0x1ffffU);
-	target_mem32_write32(flash->t, PUYA_FLASH_PRGTPE, eppara4 & 0xffffU);
-	target_mem32_write32(flash->t, PUYA_FLASH_PRETPE, (eppara4 >> 16U) & 0x3fffU);
+	uint32_t eppara[5] = {0};
+	for (unsigned int i = 0; i < 5; i++) {
```
dragonmux, line 161: This should really use `size_t` or `uint16_t` or so rather than a raw type.

src/target/puya.c
```diff
@@ -157,26 +157,20 @@ static bool puya_flash_prepare(target_flash_s *flash)
...
+		eppara[i] = target_mem32_read32(flash->t, PUYA_FLASH_TIMING_CAL_BASE + hsi_fs * 20 + i * 4U);
+		DEBUG_TARGET("PY32 flash timing cal %d: %08" PRIx32 "\n", i, eppara[i]);
```
dragonmux, line 163: `%d` is strictly for signed numbers, and `%u` for unsigned. If you use `uint16_t` for `i` then you can use `%u`, otherwise use `PRIu32` and cast the value to `(uint32_t)`.

ALTracer: Changed to %u and amended. Not changing first commit because 0-4 can be signed integer constants.

src/target/puya.c
ALTracer: Changed to uint16_t despite possibly worse codegen because that's not a "unsigned long" on any platforms and won't need %PRIu32/%lu breaking the format-string and reducing readability.

dragonmux: Using a narrower type should not result in any materially worse codegen - at least on any of the platforms we care about here - nor any penalty to runtime. That said, if we wanted to guarantee the fastest codegen.. `uint_fast8_t` or `uint_fast16_t` are available if that became a worry.

dragonmux (approved):
This LGTM, we'll get this merged when the v2.1 merge window opens. Thank you for the contribution!

dragonmux:
Please rebase this on `main` and we'll get it merged now the v2.1 merge window has opened.

ALTracer:
> Am I still needed to test something here?

Yes, I would prefer you took a test run of this 2096 and 2080 on any PY32F0 board you still have (and I still don't have any) -- the usual erase/flash/run, maybe reflash at five HSI scales to test voltage/frequency support.

dragonmux:
Given we're right at the start of the cycle, and as this code looks reasonable, we're going to merge this. If it turns out to be subtly wrong then a follow-up PR can be opened - we've got no problems with that.

PR #1863, opened 2024-07-11
Fix: Unify help messages for Unique ID with STM32

ALTracer:
## Detailed description

* Not a new feature, but rather a minor fix for size optimization.
* The existing problem is four distinct variants of the same idea of a message in UID commands or equivalents in target drivers for EFM32, nRF51, Renesas RA.
* This PR replaces the help messages to match the STM32 one, shrinking all builds by ~60 bytes thanks to linker relaxation.

I can't exactly test it live, due to no devices, but `strings -n12 blackmagic.bin | grep "unique"` no longer matches four entries, and the binary is smaller.

## Your checklist for this pull request

* [x] I've read the [Code of Conduct](https://github.com/blackmagic-debug/blackmagic/blob/main/CODE_OF_CONDUCT.md)
* [x] I've read the [guidelines for contributing](https://github.com/blackmagic-debug/blackmagic/blob/main/CONTRIBUTING.md) to this repository
* [x] It builds for hardware native (see [Building the firmware](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-firmware))
* [x] It builds as BMDA (see [Building the BMDA](https://github.com/blackmagic-debug/blackmagic?tab=readme-ov-file#building-black-magic-debug-app))
* [x] I've tested it to the best of my ability
* [x] My commit messages provide a useful short description of what the commits do

## Closing issues

PR #2097, opened 2025-03-20
Fix: log string deduplication

ALTracer:
## Detailed description

* No significant new features.
* The existing problem is suboptimal/repeating message strings for DEBUG_ERROR/WARN/TARGET().
* This PR solves it by factoring out unique substrings and enabling linker relaxation across newly-identical strings.

Discovered by staring at `strings -n16 build-f411ce/blackmagic_blackpill_f411ce_firmware.bin | less` output.
1. First commit saves up to 256 bytes when all targets are enabled and `DEBUG_TARGET()` is enabled.
2. In `stm32h7.c`, I drop the logline I added previously for H7B3 compat FLASH_CR patcher helper, because the driver is tested to work and that logline spams output during flashing. -116 bytes.
3. Renesas RA driver has three really long warnings with identical prefix/suffix and missing comma. -136 bytes.
4. LPC17xx & LPC40xx have similar mass-erase functions in which strings differ and hence don't get merged. -104 bytes.
5. LPC55xx reports IAP errors for five unique commands, and I decided to extract these instead. -52 bytes.

I would need some autobuilder script to measure size savings from each commit, but they're on the order of 50-100 bytes, for a total of about -292 bytes up to DEBUG_WARN, and -372 bytes more up to DEBUG_TARGET (-664).
Can't test on most of targets modified in PR scope. Idea similar to #1527 and #1863. Should have no effect on release firmware with disabled logging. When `-Ddebug_output=true`, default `native` and `swlink`/`stlink`/`bluepill` and `f072` builds still overflow by 17-23 KiB. The other probe platforms are not 128 KiB but 256 KiB or even larger.

## Closing issues

### DIFF
--- a/src/target/stm32h7.c
+++ b/src/target/stm32h7.c
@@ -500,7 +500,6 @@ static uint32_t stm32h7_flash_cr(uint32_t sector_size, const uint32_t ctrl, cons
 	/* H74x, H72x IP: 128 KiB and has PSIZE */
 	if (sector_size == FLASH_SECTOR_SIZE) {
 		command |= sector_number << STM32H7_FLASH_CTRL_SECTOR_NUM_SHIFT;
-		DEBUG_TARGET("%s: patching FLASH_CR from 0x%08" PRIx32 " to 0x%08" PRIx32 "\n", __func__, ctrl, command);
 		return command;
 	}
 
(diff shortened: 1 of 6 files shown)


src/target/stm32h7.c
dragonmux, line 1: How come the removal of the two DEBUG_TARGET's in this file? They are particularly useful when trying to debug Flash controller control register issues, which is why we left them in.

If there's a good reason though, then we'll be happy to accept this change.

dragonmux (dismissed):
LGTM, we'll get this merged when the v2.1 merge window opens. Thank you for the contribution!

ALTracer:
Rebased. Added `static` and macro guards to lpc55xx.c string reflection array, to keep release binary size-diff in check at +0 rom bytes. Log-enabled diff locally is -312, or -668 with DEBUG_TARGET unguarded (multiple loglevels in PR scope).

dragonmux (approved):
Thank you for fixing that small oops that was identified in Discord, and rebasing this! Merging.

PR #2103, opened 2025-04-03
Fix: CMSIS-DAP HID I/O

dragonmux:
## Detailed description

In this PR we address several issues encountered with certain CMSIS-DAP adaptors, notably ones from Atmel such as the JTAGICE3 which use HID (CMSIS-DAP v1.0.0, the unpublished spec) and massive packets (512 bytes) over HS.

We aim with this to fix the transfer behaviour for these adaptors (the packets are one byte too short before this, causing problems with the communication cycle), and provide various fallbacks and such for commands not implemented in the older versions of the spec.

The over-all result should be a more stable CMSIS-DAP experience and saner behaviour out of HID based adaptors.


esden (approved):
LGTM

PR #2104, opened 2025-04-03
Fix: Buffer utilities codegen

dragonmux:
## Detailed description

In this PR we address a codegen/optimisation issue that has been identified in the buffer utilities functions. These functions include the `read_{l,b}e{2,4,8}()` routines which do an endian-aware read of a buffer to (re)construct an integer.

The way these functions were expressed is somewhat naïve and results in extremely poor codegen (a sled of loads, shifts and adds) due to the need to deal with an offset into the buffer. It was quickly identified, however, that moving the data to a function-internal buffer with memcpy() and then doing the endian-dependant read produced the correct code output - one load, and depending on host endian, one swap.

This greatly improves the performance of these functions and their call footprint as they get inlined where they're called.

### DIFF
--- a/src/include/buffer_utils.h
+++ b/src/include/buffer_utils.h
@@ -61,27 +61,32 @@ static inline void write_be4(uint8_t *const buffer, const size_t offset, const u
 
 static inline uint16_t read_le2(const uint8_t *const buffer, const size_t offset)
 {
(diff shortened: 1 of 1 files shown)


esden (approved):
LGTM

PR #2108, opened 2025-04-17
DRAFT: samx5x: make `lock_bootprot` configurable

UtkarshVerma:
## Detailed description
For `samx5x` devices, blackmagic provides a `lock_bootprot` command to conveniently write-protect the bootloader section in the RAM. As stated on page 598 of the [datasheet](https://ww1.microchip.com/downloads/en/DeviceDoc/SAM_D5xE5x_Family_Data_Sheet_DS60001507F.pdf), the bootloader region can be marked using `BOOTPROT` value in the user page using $(15 - BOOTPROT) * 8 KiB$.

Currently, `lock_bootprot` hardcodes `BOOTPROT` to 0, which means that 120 KiB of the flash is write-protected. For bootloaders like [Adafruit's Arduino bootloader for SAMD51](https://github.com/adafruit/uf2-samdx1), this does not work since Arduino generates binaries that start with a 16 KiB offset, which is the size of the bootloader.

This PR allows the user to specify exactly how much space has to be reserved for the bootloader, like how it is handled for SAMD21 devices in blackmagic. 

Things I need review on:
- `lock_bootprot` accepts an argument which is the BOOTPROT value from the datasheet. It would be more developer-friendly if the command just accepted the number of 16 KiB pages to reserve and explained this in the `monitor help`  accordingly.
- `parse_unsigned` is currently copied over to samx5x source. How do I share it between samd and samx5x devices? 

### DIFF
--- a/src/target/samx5x.c
+++ b/src/target/samx5x.c
@@ -719,6 +719,19 @@ static int samx5x_set_flashlock(target_s *t, uint32_t value)
 	return 0;
 }
 
+static bool parse_unsigned(const char *str, uint32_t *val)
+{
+	char *end = NULL;
+	unsigned long num;
+
+	num = strtoul(str, &end, 0);
+	if (end == NULL || end == str)
+		return false;
+
+	*val = (uint32_t)num;
+	return true;
+}
+
 static bool samx5x_cmd_lock_flash(target_s *t, int argc, const char **argv)
 {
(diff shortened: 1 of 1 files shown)


UtkarshVerma:
This is just a quick draft I made while hacking around. I will tidy up the commits after I get the review.

PR #2125, opened 2025-06-04
Fix: vCont single stepping

### LINKED ISSUE #2124: Single step exiting an interrupt causes it to behave as if 'continue' was typed
This same issue is present in many other hardware debuggers - including ST's gdb-server + debug probe firmware. (But not in probe-rs, which has its own issues)

On a STM32G0B1
using a STLink V2.1 
Using Hubris OS
With memory protection and privileged/thread mode
On other systems using ST's debugger the same thing happened with normal C programs - I think it's probably intrinsic to disabling interrupts while single instruction stepping

Image

When using the `stepi` instruction to exit an ISR causes it as if 'continue' was typed ("zooms back around to the entry breakpoint")

dragonmux:
## Detailed description

In this PR we address the issue of single stepping having broken as reported in #2124.

As it turns out in newer GDBs, if we don't report that we support the signaled version of single stepping in response to `vCont?`, then single stepping will be entirely turned off, `vCont;s` will never be generated. With this fixed, it will and stepping with `stepi` goes back to working how one expects it to.

## Closing issues

Fixes #2124

### DIFF
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -133,7 +133,7 @@ int32_t gdb_main_loop(target_controller_s *const tc, const gdb_packet_s *const p
 			/**
 			 * Register data is unavailable
 			 * See: https://sourceware.org/gdb/current/onlinedocs/gdb.html/Packets.html#read-registers-packet
-			 * 
+			 *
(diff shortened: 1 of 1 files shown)


esden (approved):
LGTM

PR #2128, opened 2025-06-26
Feature: Supported targets remote reporting

dragonmux:
## Detailed description

In this PR we introduce a couple of new remote protocol commands to allow the probe to report what target architectures and target families have been compiled in and are supported by the firmware. This is designed to allow utilities like `bmputil` to discover target support from the probe to determine the firmware variant running (for automatic upgrades and `bmputil-cli probe info --list-targets`).

A follow-up PR is intended to introduce support for asking the probe for a complete targets support string - this string will be contained in the OTG Flash on BMP v2.3+ units and is purely for display purposes on the host. To this end we reserve the packets `HT` for a bitmask of supported targets in the future, and `Ht` for this string version. Such as string will list specific supported targets by family - eg `stm32f1,stm32f4,stm32u5,stm32h5,stm32h7` etc w/ comma as the delimiter.

While this is a feature PR, this is designed to solve the issue where this information was not otherwise discoverable, which prevents tooling from being built that can handle the variants situation and match firmware variants up to upgrade probes and such. Thus this is a bugfix in a manner of speaking.

### DIFF
--- a/src/platforms/hosted/remote/protocol_v4.c
+++ b/src/platforms/hosted/remote/protocol_v4.c
@@ -133,3 +133,37 @@ bool remote_v4_riscv_jtag_init(riscv_dmi_s *const dmi)
 	dmi->write = remote_v4_riscv_jtag_dmi_write;
 	return true;
 }
+
+uint64_t remote_v4_supported_architectures()
+{
+	/* Ask the probe what target architectures it supports */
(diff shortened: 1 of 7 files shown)


esden (approved):
LGTM, as discussed out of band this is being merged despite us being in RC mode right now, as we need to address these issues before the full release.

Issue #2180, opened 2026-01-28
Inconsistent `jtagtap_cycle()` implementations


ALTracer:
A method to generate N cycles of TCK with fixed values of TMS and TDI, `jtagtap_cycles()`, was added somewhat later than the other base methods to `jtag_proc` structure of function pointers. I have identified four places where this is used.
1. `jtagtap_init()` and equivalents, for emitting SWD LINERESET (50 cycles with TMS high), sometimes coded as a literal for-loop around `jtagtap_next()` or a prepared byte buffer.
https://github.com/blackmagic-debug/blackmagic/blob/54ddc3505c5ac3baaa0253ebc08a75444eabf976/src/platforms/common/jtagtap.c#L54-L59
2. jtag_scan.c:`jtag_sanity_check()`, called as early as `jtag_scan()` => `jtag_discover()` => `jtag_sanity_check()`. I have touched this recently in #2178.https://github.com/blackmagic-debug/blackmagic/blob/54ddc3505c5ac3baaa0253ebc08a75444eabf976/src/target/jtag_scan.c#L306-L307
3. adiv5_jtag.c:`adiv5_jtag_raw_access()`, for 8 idle cycles but only for ADI v6 JTAG-DP targets. Never encountered with ADI v5.https://github.com/blackmagic-debug/blackmagic/blob/54ddc3505c5ac3baaa0253ebc08a75444eabf976/src/target/adiv5_jtag.c#L166-L170
4. icepick.c:`icepick_router_handler()`, for 10 idle cycles after reconfiguring type-D controller. Very target-specific.
https://github.com/blackmagic-debug/blackmagic/blob/54ddc3505c5ac3baaa0253ebc08a75444eabf976/src/target/icepick.c#L135-L140

None of these places are particularly hot code (intensively called during debug). The problem is that BMDA backends for FTDI MPSSE, J-Link, CMSIS-DAP never set the pointer, so it stays NULL (in a `calloc()`'ed struct), allowing one to inadvertently segfault BMDA mid-scan.

The obvious solutions are a) null-check the pointer at all callsites and fall back to `jtagtap_next()`; b) code the wrappers which invoke corresponding methods bound to `jtagtap_next()`. Which can then evolve into proper functions optimized for bit-packing and buffering characteristic for the protocols of these low-level adapters.

[...]

PR #2182, opened 2026-01-29
Feature: implementation of `jtagtap_cycle` for JLink backend

ALTracer:
## Detailed description

* This is a minor new feature.
* The problem is rudimentary/poor implementation of `jtagtap_cycle()` in jlink.c backend used by BMDA. (and it didn't even exist until very recently)
* This PR solves it by providing one moderately optimized variant.

Tracking: #2180.
Note that `jlink_jtag_init()` used to rely on a direct `jlink_transfer()` of a baked byte buffer with 56 idle clocks and SWD-to-JTAG switch sequence, I replaced that with a `jlink_jtag_cycle()` call for 56-5=51 clocks and a shorter, 16-bit sequence. The caveat is that new function may raise an exception, and at scan time it results in immediate SIGABRT because it's unhandled (by any TRY-CATCH up the callsite). It also happens to be the first 0xCF opcode transfer. If such behaviour is unwanted, discard commit 1.

Tested on J-Link V9 against a RISC-V MCU.

## Closing issues

### DIFF
--- a/src/platforms/hosted/jlink_jtag.c
+++ b/src/platforms/hosted/jlink_jtag.c
@@ -123,15 +123,19 @@ static bool jlink_jtag_next(bool tms, bool tdi)
 
 static void jlink_jtag_cycle(const bool tms, const bool tdi, const size_t clock_cycles)
 {
-	uint8_t tms_buf[8] = {0};
-	uint8_t tdi_buf[8] = {0};
-	if (clock_cycles > 64U)
+	uint8_t tms_buf[512] = {0};
+	uint8_t tdi_buf[512] = {0};
+	if (clock_cycles > 4096U)
 		return;
-	const uint64_t all_ones = clock_cycles == 64U ? UINT64_MAX : (UINT64_C(1) << clock_cycles) - 1U;
-	const uint64_t tms_pattern = tms ? all_ones : 0U;
-	const uint64_t tdi_pattern = tdi ? all_ones : 0U;
-	memcpy(tms_buf, &tms_pattern, 8);
-	memcpy(tdi_buf, &tdi_pattern, 8);
+	const size_t clock_bytes = clock_cycles >> 3U;
+	memset(tms_buf, tms ? 0xffU : 0U, clock_bytes);
+	memset(tdi_buf, tdi ? 0xffU : 0U, clock_bytes);
+	const size_t clock_bits = clock_cycles & 7U;
+	if (clock_bits) {
+		const uint8_t ones = (1U << clock_bits) - 1U;
+		tms_buf[clock_bytes] = tms ? ones : 0U;
+		tdi_buf[clock_bytes] = tdi ? ones : 0U;
+	}
 	DEBUG_PROBE("jtagtap_cycle tms=%u tdi=%u, clock cycles: %zu\n", tms, tdi, clock_cycles);
 	const bool result = jlink_transfer(clock_cycles, tms_buf, tdi_buf, NULL);
 	if (!result)
--- a/src/platforms/hosted/jlink_jtag.c
+++ b/src/platforms/hosted/jlink_jtag.c
@@ -42,8 +42,6 @@ static void jlink_jtag_tdi_seq(bool final_tms, const uint8_t *data_in, size_t cl
 static bool jlink_jtag_next(bool tms, bool tdi);
 static void jlink_jtag_cycle(bool tms, bool tdi, size_t clock_cycles);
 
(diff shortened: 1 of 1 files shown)


src/platforms/hosted/jlink_jtag.c
```diff
@@ -56,7 +54,10 @@ bool jlink_jtag_init(void)
 
 	/* Ensure we're in JTAG mode */
 	DEBUG_PROBE("%s: Switch to JTAG\n", __func__);
-	if (!jlink_transfer(sizeof(jlink_switch_to_jtag_seq) * 8U, jlink_switch_to_jtag_seq, NULL, NULL)) {
+	jlink_jtag_cycle(true, false, 56U - 5U);
+	uint8_t tms_swd2jtag_seq[2] = {0x3cU, 0xe7U};
```
dragonmux: Please don't use the number 2 in the name here - `tms_swd_to_jtag_seq` reads easier and more naturally.

src/platforms/hosted/jlink_jtag.c
```diff
@@ -122,6 +123,17 @@ static bool jlink_jtag_next(bool tms, bool tdi)
 
 static void jlink_jtag_cycle(const bool tms, const bool tdi, const size_t clock_cycles)
 {
-	for (size_t i = 0; i < clock_cycles; i++)
-		jlink_jtag_next(tms, tdi);
+	if (clock_cycles > 63U)
+		return;
+	const uint64_t all_ones = (UINT64_C(1) << clock_cycles) - 1U;
+	const uint64_t tms_pattern = tms ? all_ones : 0;
+	const uint64_t tdi_pattern = tdi ? all_ones : 0;
```
dragonmux: These could do with `U` suffixing and you might as well use `UINT64_MAX` here unless the J-Link doesn't like the trailing set bits in the buffers - but that raises the max clock sequence length to 64 from 63.

src/platforms/hosted/jlink_jtag.c
```diff
@@ -122,6 +123,17 @@ static bool jlink_jtag_next(bool tms, bool tdi)
...
+	uint8_t tms_buf[8] = {0};
+	uint8_t tdi_buf[8] = {0};
+	memcpy(tms_buf, &tms_pattern, 8);
```
dragonmux: Please use `write_le8()` here not `memcpy()` as this makes invalid assumptions about byte layout within a `uint64_t` that will get violated on the wrong endian system.

dragonmux (changes requested):
Couple of things we notice reviewing this, but with them fixed we'll be happy to merge this.

src/platforms/hosted/jlink_jtag.c
```diff
@@ -122,6 +123,37 @@ static bool jlink_jtag_next(bool tms, bool tdi)
...
+#else
+	uint8_t tms_buf[512] = {0};
+	uint8_t tdi_buf[512] = {0};
+	if (clock_cycles > 4096U)
+		return;
+	const size_t clock_bytes = clock_cycles >> 3U;
+	for (size_t i = 0; i < clock_bytes; i++) {
+		tms_buf[i] = tms ? 0xffU : 0U;
+		tdi_buf[i] = tdi ? 0xffU : 0U;
+	}
```
dragonmux: Perhaps for this call `memset()`? (`memset(tms_buf, tms ? 0xffU : 0U, clock_bytes);`, etc)

src/platforms/hosted/jlink_jtag.c
```diff
@@ -122,6 +123,37 @@ static bool jlink_jtag_next(bool tms, bool tdi)
...
+	const size_t clock_bits = clock_cycles & 7U;
+	if (clock_bits) {
+		const uint8_t ones = 1U << clock_bits;
```
dragonmux: This looks to be missing its ` - 1U` to convert the value into a mask?

dragonmux (commented):
The suggested other way to express the cycle logic we think fits quite well - doesn't need byte ordering shenanigans, or wide shift and subtract - with the first block of the `#if`/`#else` removed and the two small notes below addressed, we'll be happy to merge this.

src/platforms/hosted/jlink_jtag.c
ALTracer: Yes, completely missed memset, I was looking at git history of jtag-related jlink functions, and they all seem to fill byte buffers. This way there's no 64-bit limit, you could generate up to 2 KiB of data per command (J-Link fw capacity).

src/platforms/hosted/jlink_jtag.c
ALTracer: The function in question does not exist, abandoning in favor of rewrite.

dragonmux (approved):
LGTM, merging. Thank you for the contribution and this improvement to the J-Link backend!

PR #2184, opened 2026-02-01
Fix: reuse GDB packet buffer for m/M-packets instead of VLA

### LINKED ISSUE #2042: Use of alloca in gdb m-packet handling can be optimized out
Looking at  #2041 I noticed there is a `alloca` use later in the affected code ([Line 153](https://github.com/ALTracer/blackmagic/blob/9c60b339dad849e8d554579a0c3141f9ed9c1ca0/src/gdb_main.c#L153)), and it can be optimized out by reusing the packet buffer.

Since we check previously that the requested memory will fit in the packet buffer after encoding we can use the packet buffer itself for the read safely. Care must be taken because later building the packet could possibly corrupt the read data (because each hex encoded byte will take 2 bytes in the buffer), but we can work around this by only using the upper half of the buffer, which should allow the encoding in place to be done safely.

ALTracer:
## Detailed description

* This is not a new feature, but still a potentially dangerous change. A sort of continuation of PR2014.
* The existing problem is gdb_main FSM performing VLA for m/M-packets.
* This PR solves it by reusing half of gdb packet buffer for (un)hexify operations.

Tested on BMDA (env CC=clang, -Db_sanitize=address) and on blackpill-f411ce against GD32VF103. Dump and compare-sections seem to match, GDB receives 512 byte (1024 hex-encoded) packets.
g/G-packets still depend on VLA because BMP supports multiple architectures with different amounts of general-purpose registers. `{armv7-m 20*4, armv8-m.main 24*4} + fpv4-sp-d16 33*4, armv7-a/r 17*4 + VFPv3 33*4, armv8-a 31*8, rv32i 33*4, rv32e 17*4, rv64i 33*8.` IMHO a static regcache of 512 bytes could be enough (with build-time static assert that no arch needs more), but it's not visibly broken now, so not touching.

## Closing issues

Closes #2042. @perigoso

### DIFF
--- a/src/gdb_main.c
+++ b/src/gdb_main.c
@@ -150,7 +150,7 @@ int32_t gdb_main_loop(target_controller_s *const tc, const gdb_packet_s *const p
 				break;
 			}
 			DEBUG_GDB("m packet: addr = %" PRIx32 ", len = %" PRIx32 "\n", addr, len);
-			uint8_t *mem = alloca(len);
+			char *mem = gdb_packet_buffer() + GDB_PACKET_BUFFER_SIZE / 2U;
 			if (target_mem32_read(cur_target, mem, addr, len))
 				gdb_put_packet_error(1U);
 			else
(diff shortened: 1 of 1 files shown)


dragonmux (approved):
LGTM, please rebase on `main` and we'll get this merged. Thank you for the contribtion! We quite like this way of working round the VLA problem - it's really a nice solution.

PR #2187, opened 2026-02-01
armv8m: make the debugger handle better faults through ICSR

mean00:
## Detailed description
The ArmV8m does not handle faults the same way. The two main differences are :
- The fault will trigger BEFORE executing the fault rather than after
- The info is split between between DFSR and the new ICSR
- 
As a result the current code does not handle fault very well.
Put a breakpoint on a fault (for example __asm__("udf #0") and try to "si" on it.

Gdb  will appear to get stuck there because the debugger will see the state  BEFORE the fault is actually executed and as a result not detect the fault.

The MR adds a separate path that , for ArmV8M, it checks if a fault is pending on ICSR 
The "si" command on a fault now works and you proceed on to the handler.

I tested it on a RP2350 and on another CortexM33 successfully. 
I'm not convinced all cases are covered as some parts are configurable, but at least it's working better.

Not sure either if the new two new registers should be renamed with _V8M_ in the name or something.

## Closing issues

N/A

### DIFF
--- a/src/target/cortexm.c
+++ b/src/target/cortexm.c
@@ -864,7 +864,21 @@ static target_halt_reason_e cortexm_halt_poll(target_s *target, target_addr64_t 
 	priv->dcache_enabled = ccr & CORTEXM_CCR_DCACHE_ENABLE;
 	priv->icache_enabled = ccr & CORTEXM_CCR_ICACHE_ENABLE;
 
-	if ((dfsr & CORTEXM_DFSR_VCATCH) && cortexm_fault_unwind(target))
+	bool fault_state = false;
+	// the V8 may stop before actually executing the instruction
+	// so reading dfsr might not work.
+	// Instead, we check if there are pending faults on ICSR
+	// meaning we stopped while trying to execute a fault
+	// but maybe did not execute it
+	if ((target->target_options & CORTEXM_TOPT_FLAVOUR_V8M)) {
+		const uint32_t icsr = target_mem32_read32(target, CORTEXM_ICSR);
+		const uint32_t pending = CORTEXM_ICSR_VEC_PENDING(icsr);
+		//  catch all pending faults
+		if (pending > 0U && pending < 8U)
+			fault_state = true;
+	} else
+		fault_state = (dfsr & CORTEXM_DFSR_VCATCH) != 0U;
+	if (fault_state && cortexm_fault_unwind(target))
 		return TARGET_HALT_FAULT;
 
 	/* Remember if we stopped on a breakpoint */
--- a/src/target/cortexm.h
+++ b/src/target/cortexm.h
@@ -69,6 +69,9 @@ extern unsigned cortexm_wait_timeout;
 #define CORTEXM_DWT_MASK(i) (CORTEXM_DWT_BASE + 0x024U + (0x10U * (i)))
 #define CORTEXM_DWT_FUNC(i) (CORTEXM_DWT_BASE + 0x028U + (0x10U * (i)))
 
+/* ARMv8 External Debug Fault Status Register */
+#define CORTEXM_EDFSR (CORTEXM_SCS_BASE + 0xf98U)
+#define CORTEXM_ICSR  (CORTEXM_SCS_BASE + 0xd04U)
 /* Application Interrupt and Reset Control Register (AIRCR) */
 #define CORTEXM_AIRCR_VECTKEY (0x05faU << 16U)
 /* Bits 31:16 - Read as VECTKETSTAT, 0xfa05 */
@@ -187,6 +190,10 @@ extern unsigned cortexm_wait_timeout;
 
 #define CORTEXM_XPSR_THUMB          (1U << 24U)
 #define CORTEXM_XPSR_EXCEPTION_MASK 0x0000001fU
+
+/* ICSR for ARMv8-M, the exception are the same as IPSR */
+#define CORTEXM_ICSR_VEC_PENDING(x) (((x) >> 12U) & 0x1ffU)
+#define CORTEXM_ICSR_VEC_ACTIVE(x)  (((x) >> 0U) & 0x1ffU)
 
 bool cortexm_attach(target_s *target);
 void cortexm_detach(target_s *target);


src/target/cortexm.h
```diff
@@ -188,6 +191,10 @@ extern unsigned cortexm_wait_timeout;
 #define CORTEXM_XPSR_THUMB          (1U << 24U)
 #define CORTEXM_XPSR_EXCEPTION_MASK 0x0000001fU
 
+/* ICSR for ArmV8m, the exception are the same as IPSR */
+#define CORTEXM_ICSR_VEC_PENDING(x) (((x) >> 12) & 0x1ff)
+#define CORTEXM_ICSR_VEC_ACTIVE(x)  (((x) >> 0) & 0x1ff)
```
dragonmux: Please suffix the numbers here with `U` so the bit shifting and masking is done unsigned.

src/target/cortexm.h
```diff
@@ -69,6 +69,9 @@ extern unsigned cortexm_wait_timeout;
 #define CORTEXM_DWT_MASK(i) (CORTEXM_DWT_BASE + 0x024U + (0x10U * (i)))
 #define CORTEXM_DWT_FUNC(i) (CORTEXM_DWT_BASE + 0x028U + (0x10U * (i)))
 
+/* Arm V8 External Debug Fault Status Register */
```
dragonmux: `ARMv8`

src/target/cortexm.h
```diff
@@ -188,6 +191,10 @@ extern unsigned cortexm_wait_timeout;
 #define CORTEXM_XPSR_THUMB          (1U << 24U)
 #define CORTEXM_XPSR_EXCEPTION_MASK 0x0000001fU
 
+/* ICSR for ArmV8m, the exception are the same as IPSR */
```
dragonmux: `ARMv8-M`

src/target/cortexm.c
```diff
@@ -864,7 +864,23 @@ static target_halt_reason_e cortexm_halt_poll(target_s *target, target_addr64_t
...
+	bool fault_state = false;
+	// the V8 may stop before actually executing the instruction
+	// so reading dfsr might not work.
+	// Instead, we check if there are pending faults on ICSR
+	// meaning we stopped while trying to execute a fault
+	// but maybe did not execut it
+	if ((target->target_options & CORTEXM_TOPT_FLAVOUR_V8M)) {
+		const uint32_t icsr = target_mem32_read32(target, CORTEXM_ICSR);
+		const uint32_t pending = CORTEXM_ICSR_VEC_PENDING(icsr);
+		if (pending != 0 && pending < 8) //  catch all faults
```
dragonmux: Given `pending` is unsigned, perhaps `(pending > 0U && pending < 8U)`? - also please let `clang-format` run here, the `{` should be on this line (or, well.. more the point.. aren't necessary with this `if` block at all).

Perhaps re-express this as `fault_state = pending > 0U && pending < 8U;`?

src/target/cortexm.c
```diff
@@ -864,7 +864,23 @@ static target_halt_reason_e cortexm_halt_poll(target_s *target, target_addr64_t
...
+		{
+			fault_state = true;
+		}
+	} else {
```
dragonmux: Please drop the else block braces, they're not necessary and not part of the code base code style.

src/target/cortexm.c
```diff
@@ -864,7 +864,23 @@ static target_halt_reason_e cortexm_halt_poll(target_s *target, target_addr64_t
...
+		fault_state = !!(dfsr & CORTEXM_DFSR_VCATCH);
```
dragonmux: Rather than using `!!` which is inobvious, please use `!= 0U` on the end of the expression for the bool conversion - it won't change codegen, but it does improve code clarity by better expression your intent, which isn't to double negate the value, but rather to check the value is not 0.

dragonmux (changes requested):
Got a few notes on going through this PR - with them fixed though we'll be happy to merge this.

mean00:
I'll update the MR this weekend
Thank you

src/target/cortexm.h
mean00: done

src/target/cortexm.h
mean00: done

src/target/cortexm.h
mean00: done

src/target/cortexm.c
mean00: done

src/target/cortexm.c
mean00: done

src/target/cortexm.c
mean00: done

mean00:
I think i got all the remarks as a separate commit for easy review
If that's ok, i'll squash them, it takes 10 sec to do so
Thank you

src/target/cortexm.c
```diff
@@ -864,7 +864,20 @@ static target_halt_reason_e cortexm_halt_poll(target_s *target, target_addr64_t
...
+	// the V8 may stop before actually executing the instruction
+	// so reading dfsr might not work.
+	// Instead, we check if there are pending faults on ICSR
+	// meaning we stopped while trying to execute a fault
+	// but maybe did not execute it
+	if ((target->target_options & CORTEXM_TOPT_FLAVOUR_V8M)) {
+		const uint32_t icsr = target_mem32_read32(target, CORTEXM_ICSR);
+		const uint32_t pending = CORTEXM_ICSR_VEC_PENDING(icsr);
+		if (pending > 0 && pending < 8) //  catch all faults
+			fault_state = true;
```
dragonmux: These are still missing their `U` suffixes to make the comparisons unsigned.

We think `fault_state = pending > 0U && pending < 8U;` with the comment on the line before would be more concise and should also be a `clang-tidy` lint.

dragonmux (commented):
This is looking better - there is one small note we have and once that's been addressed, we'll approve and merge this.

mean00:
This time i think it's correct. I've squashed it as the last change was really small & easy to check.
(slightly altered the commit comment to be in line with the naming for ARMv8-M)
Thank you.

dragonmux:
It would technically be `cortexm: ` for the commit message suffix as it's not based on the concept touched, but which file is touched.. however, we're going to approve and merge this as it stands as it looks good.

dragonmux (approved):
LGTM. Thank you for the contribution!

We'll do some cleanup after in a follow-up PR we have to do anyway (comment style, `if` with assign to `true` into straight truthy assignment) but this looks all functional which is the important part.

PR #2198, opened 2026-02-27
Feature: Add target for STM32WL3x

hallard:
For the STM32WL3x, stm32wb0.c is the best place because:
- Both are single-core Cortex-M0+ wireless MCUs
- Both share the same flash controller at 0x40001000 with identical register layout and command codes (0x11/0x22/0x33/0xBB)
- Same flash size formula: ((SIZE[15:0] + 1) << 2)
- Same erase granularity (2KB sectors)
- Both have the same write-then-poll pattern
- avoid creating specific target and increase flash size 

## Detailed description

This PR add ability to use STM32WL3x targets

Here below output of testing running demo

```
charles@office:debug$ arm-zephyr-eabi-gdb-py -ex 'set confirm off' -ex 'target extended-remote /dev/cu.usbmodem81D759AC1'
GNU gdb (Zephyr SDK 0.17.4) 12.1
Copyright (C) 2022 Free Software Foundation, Inc.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "--host=aarch64-build_apple-darwin23.6.0 --target=arm-zephyr-eabi".
<https://github.com/zephyrproject-rtos/sdk-ng/issues>.

For help, type "help".
Type "apropos word" to search for commands related to "word".
Remote debugging using /dev/cu.usbmodem81D759AC1
(gdb) monitor swdp_scan
Target voltage: 3.3V
Available Targets:
No. Att Driver
 1      STM32WL3 M0+
(gdb) attach 1
Attaching to Remote target
warning: No executable has been specified and target does not support
determining executable automatically.  Try using the "file" command.
0x10043d84 in ?? ()
(gdb) monitor erase_mass
Erasing device Flash: done
(gdb) file LCSC_BasicDemo.elf
Reading symbols from LCSC_BasicDemo.elf...
(gdb) load
Loading section .intvec, size 0xc0 lma 0x10040000
Loading section .text, size 0x6e90 lma 0x100400c0
Loading section .init, size 0x4 lma 0x10046f50
Loading section .fini, size 0x4 lma 0x10046f54
Loading section .ARM.exidx, size 0x8 lma 0x10046f58
Loading section .data, size 0x88 lma 0x10046f60
Start address 0x1004259c, load size 28648
Transfer rate: 3 KB/sec, 818 bytes/write.
(gdb) run
Starting program: LCSC_BasicDemo.elf 

```

### DIFF
--- a/src/target/stm32wb0.c
+++ b/src/target/stm32wb0.c
@@ -32,12 +32,17 @@
  */
 
 /*
- * This file implements support for STM32WB0x series devices, providing
- * memory maps and Flash programming routines.
+ * This file implements support for STM32WB0x and STM32WL3x series devices,
+ * providing memory maps and Flash programming routines.
+ *
+ * Both families share the same flash controller register layout and command
+ * set; the main differences are the flash bank base address, the debug
+ * infrastructure registers, and the write granularity.
  *
  * References:
  * RM0530 - STM32WB07xC and STM32WB06xC ultra-low power wireless 32-bit MCUs Arm®-based Cortex®-M0+ with Bluetooth® Low Energy and 2.4 GHz radio solution, Rev. 1
  * - https://www.st.com/resource/en/reference_manual/rm0530--stm32wb07xc-and-stm32wb06xc-ultralow-power-wireless-32bit-mcus-armbased-cortexm0-with-bluetooth-low-energy-and-24-ghz-radio-solution-stmicroelectronics.pdf
+ * RM0505 - STM32WL33 ultra-low power sub-GHz wireless 32-bit MCU Arm®-based Cortex®-M0+, Rev. 1
  */
 
 #include "general.h"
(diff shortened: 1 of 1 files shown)


src/target/stm32wb0.c
```diff
@@ -191,8 +240,8 @@ static bool stm32wb0_flash_write(
 {
 	target_s *const target = flash->t;
 
-	/* Start by telling the controller the first address we want to program */
-	target_mem32_write32(target, STM32WB0_FLASH_ADDRESS, (dest - STM32WB0_FLASH_BANK_BASE) >> 2U);
+	/* Start by telling the controller the first word address to program */
```
dragonmux: "word" here doesn't mean anything useful, please don't use the term. Is that 16-bit, 32-bit, etc? it's not even clear from ST's own literature as it depends on the part and the Flash configuration.. that is exactly why we stay away from that terminology. (STM32U5 runs a 128-bit Flash "word" size for example)

Please prefer `u32` or whatever suitable type, if not just saying something like "32-bit block".

src/target/stm32wb0.c
```diff
@@ -122,52 +148,75 @@ bool stm32wb0_probe(target_s *const target)
...
+		stm32wb0_add_flash(target, STM32WB0_FLASH_BANK_BASE, flash_size, STM32WB0_FLASH_PAGE_SIZE);
+	}
 	return true;
 }
 
 static bool stm32wb0_flash_wait_complete(target_s *const target, platform_timeout_s *const timeout)
 {
+	/* WL3x polls the raw interrupt status; WB0x polls the masked status register */
+	const uint32_t status_reg =
+		(target->part_id == ID_STM32WL3) ? STM32WL3_FLASH_IRQRAW : STM32WB0_FLASH_STATUS;
```
dragonmux: The parens here are not required - `==` binds tighter than `?` binds tighter than `=`, thus the order of operations is already guaranteed by C by writing `const uint32_t status_reg = target->part_id == ID_STM32WL3 ? STM32WL3_FLASH_IRQRAW : STM32WB0_FLASH_STATUS;`

